PR #2, opened 2023-04-29
fix test

vchiley:
found a few import errors and fixed a bunch of lint issues

### DIFF
--- a/llmfoundry/common/convert_dataset.py
+++ b/llmfoundry/common/convert_dataset.py
@@ -270,15 +270,17 @@ class ConcatTokensDataset(IterableDataset):
                 }
 
 
-def build_hf_dataset(dataset_name: str,
-                     split: str,
-                     mode: ConcatMode,
-                     max_length: int = None,
-                     bos_text: str = '',
-                     eos_text: str = '',
-                     no_wrap: bool = False,
-                     tokenizer: PreTrainedTokenizerBase = None,
-                     data_subset: Union[str, None] = None) -> IterableDataset:
+def build_hf_dataset(
+    dataset_name: str,
+    split: str,
+    mode: ConcatMode,
+    max_length: int = 0,
+    bos_text: str = '',
+    eos_text: str = '',
+    no_wrap: bool = False,
+    tokenizer: PreTrainedTokenizerBase = None,
+    data_subset: Union[str, None] = None,
+) -> IterableDataset:
     """Build an IterableDataset over the HF C4 or pile source data.
 
     Args:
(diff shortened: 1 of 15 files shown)


mvpatel2000 (approved):
remove the type ignore?  is it still necessary

dakinggg (approved):

llmfoundry/common/convert_dataset.py
```diff
@@ -300,6 +302,11 @@ def build_hf_dataset(dataset_name: str,
                                   data_subset=data_subset,
                                   split=split)
     else:
+        if not isinstance(tokenizer, PreTrainedTokenizerBase):
+            raise ValueError(
+                f'{tokenizer=} must be of type PreTrainedTokenizerBase')
+        if max_length == 0:
+            raise ValueError(f'max_length must be set.')
```
mvpatel2000, line 309: In this case, can we just not have it be a default arg of 0 and make it a required arg

mvpatel2000 (approved):
Many good catches. Thanks!

llmfoundry/common/fdiff.py
```diff
@@ -25,39 +25,38 @@ def __init__(self, diff_train_metrics=False, diff_eval_metrics=True):
 
     def batch_end(self, state: State, logger: Logger):
         if self.diff_train_metrics:
+            loss = state.loss.item()  # type: ignore
```
hanlint, line 28: I'm guessing the type error here is because `state.loss` can either be a tensor or a list. In this case, the better approach is to throw an error here that multiple losses are not supported by this callback. e.g. see: https://github.com/mosaicml/composer/blob/dev/composer/algorithms/mixup/mixup.py#L203

llmfoundry/common/fdiff.py
```diff
@@ -25,39 +25,38 @@ def __init__(self, diff_train_metrics=False, diff_eval_metrics=True):
...
-            self.train_prev_loss = state.loss.item()
+                logger.log_metrics(
+                    {'loss/train/total_fdiff': loss - self.train_prev_loss})
+            self.train_prev_loss = loss
 
             for k in self.train_prev_metric.keys():
                 logger.log_metrics({
                     f'metrics/train/{k}_fdiff':
-                        state.train_metric_values[k].item() -
+                        state.train_metric_values[k].item() -  # type: ignore
```
hanlint, line 37: The issue here is that `state.train_metric_values` is typed as a `Dict[str, float]` (https://github.com/mosaicml/composer/blob/dev/composer/core/state.py#L504), but this code assumes a tensor was put in. So is the `.item()` needed here? Is this masking an API misuse upstream?

llmfoundry/common/convert_dataset.py
vchiley: if we take the `if mode == ConcatMode.NO_CONCAT:` path, then max_length is not required.

hanlint: small nit: in this case, its common to do `max_length: Optional[int] = None` , and then check for `None`. Avoids using a magic number (0), and makes the usage clear in the type annotation. Also noticed that `max_length` is missing from the docstrings.

llmfoundry/common/fdiff.py
vchiley: This callback isn't used super often
I think you just found a bug
:thankyou:

PR #3, opened 2023-04-30
Fix READMEs

hanlint:
* moves `assets/` folder out of the python package folder
* fixes stale code links in the READMEs
* removes most references to specific line numbers in the README (those grow stale fast)

Open question: should `yamls` and `mcloud` folder still be inside the python package folder? If we want to include them in the distribution, then `setup.py` needs to be updated to include them as `package_data` (e.g. see Composer: https://github.com/mosaicml/composer/blob/dev/setup.py#L242)


hanlint:
Closing -- merged in #4

PR #5, opened 2023-04-30
Fix `mcli-1b-max-seq-len-8k.yaml`

hanlint:
Currently this YAML fails with an omegaconf interpolation error on `data_remote`. This PR comments out `data_remote`. If any omegaconf powerusers (maybe cc: @codestar12 ) can suggest an alternate fix, that would be ideal.


vchiley:
`null` is an alt option

hanlint:
Unfortunately `null` also did not work... maybe I'm missing something simple here.

abhi-mosaic:
I think this turned out to be an MCLI issue (removing keys with empty values). Gonna close for now and reopen later if we still need it.

PR #6, opened 2023-04-30
Fix `ComposerHFCausalLM` initialization

hanlint:
This PR changes the signature for the `ComposerHFCausalLM` class to match the centralization of tokenization creation introduced in a previous PR (https://github.com/mosaicml/examples/pull/314). Noticed when testing all the `mcloud` yamls in this repository.

Tested to work by manually launching training runs with `opt.yaml`, `gpt2-small.yaml` and `gpt-neo-125m.yaml`. 

I added a JIRA for a future improvement to add yaml loading tests to avoid future regressions.

Note: the same issue exists with `ComposerHFT5`, and `ComposerHFPrefixLM`. However, there are no yamls for those models in the repository, so I am unable to test. If someone can share a config, I am can add to this PR.


vchiley (approved):
nice catch

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -40,21 +41,11 @@ class ComposerHFCausalLM(HuggingFaceModelWithZLoss):
                 to validation metrics. Default: ``False``.
     """
 
-    def __init__(self,
-                 om_model_config: DictConfig,
-                 om_tokenizer_config: Optional[DictConfig] = None):
+    def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
```
vchiley, line 44: we went from being "optional" to not being optional

I think the tokenizer is only needed to get vocab_size (ie len(tokenizer)); this is probably part of the model (emb layer dim)
Otherwise, tokenizer is passed to `HuggingFaceModelWithZLoss` which is a subclass of HuggingFaceModel from Composer, where the tokenizer input is optional ([see here](https://github.com/mosaicml/composer/blob/052446000033d9026adabf8087f8e1e5d0e47abd/composer/models/huggingface.py#L67))

hanlint: Good point, let me fix this. I can just read the `vocab_size` from the `hf_config` correct?

hanlint: Actually, I'm not sure if the `Optional` from previous code was actually optional. If `om_tokenizer_config = None`, then `resolved_om_tokenizer_config['name']` would fail. I think the `type: ignore` is masking a bug there.

OK to merge?

vchiley: yes merge.

PR #18, opened 2023-05-03
Rename MosaicGPT to MPT

abhi-mosaic:
This PR does the following:
* global rename of MosaicGPT to MPT
* refactor `modeling_mpt.py` to match the class design of HFCausalLM models like [modeling_llama.py](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py)
  * `MPTModel` produces embeddings, most of the code is here
  * `MPTForCausalLM` has `self.model = MPTModel(...)` and on its forward, takes the embeddings and projects through `self.model.wte` and applies optional logit scaling
  * This opens the door to future classes like `MPTForSequenceClassification` and stuff like that

I've branched this off of the `abhi/restruct` branch for now to show the diffs in isolation


abhi-mosaic:
I've run through the end-to-end quickstart and it works! Next test would be to try and run a pretrained model (with weight names remapped) through HF generate and confirm it works...

mcli/mcli-1b-eval.yaml
```diff
@@ -29,7 +29,7 @@ parameters:
       model_max_length: ${max_seq_len}
 
   model:
-    name: mosaic_gpt
+    name: mpt
```
vchiley, line 32: Should the model name be `'mptforcausallm'`

vchiley (commented):
My question in general is should we mostly default to saying `mpt` or `mtpforcausallm` in the majority of places.

mcli/mcli-1b-eval.yaml
abhi-mosaic: Hmm I guess `mpt_for_causal_lm` is more correct. WDYT @dakinggg

dakinggg: I think to match our current registry setup it would be `mpt_causal_lm`? since we haven't quite finished the merging to just use the same `HFCausalLM` class. https://github.com/mosaicml/llm-foundry/blob/1c8704b6c59a7fc8905560fdfa2671009253fb10/llmfoundry/model_registry.py#L8-L13

dakinggg (approved):
This looks right as far as I can tell. We'll just have to go through bug bash after you two big prs are though and checkpoints are converted.

mcli/mcli-1b-eval.yaml
abhi-mosaic: ok I will adjust

PR #20, opened 2023-05-03
Update requirements

mvpatel2000:
Updates to latest composer release. Also cleans up a lot of the requirements to use composer targets. Also renames package

I'll handle pypi in a separate PR


setup.py
```diff
@@ -5,7 +5,7 @@
 
 from setuptools import setup
 
-_PACKAGE_NAME = 'mosaicml-llm'
+_PACKAGE_NAME = 'llm-foundry'
```
abhi-mosaic, line 8: Should we be dropping the `mosaicml-` part? This is the name that will show up in `pip list` and I think our other packages are like `mosaicml-examples`, and `mosaicml-cli`, etc.

@hanlint if you have a sec curious what you prefer

abhi-mosaic: Checked with @hanlint and I think `llm-foundry` is fine.

dakinggg (approved):

PR #23, opened 2023-05-03
convert examples v004 state dict to llm-foundry state dict

vchiley:
Abhi tried it on the final 7B model.


alextrott16 (commented):
This looks good to me. I'm not as intimately familiar with the recent naming conventions, so I'd prefer someone else to give the stamp of approval... But if no one else bites at the PR in the next 12 hours or so, then I'm happy to say that I trust your work and click approve.

As a separate question/comment, are we planning on putting more stuff in `scripts/migrate_mosaicexamples/`? It seems like `scripts/` is pretty focused around the standard phases of LLM forging (?) so it'd be nice to not let it get too cluttered. Maybe a `scripts/misc/` directory or something would be a good place to collect various QOL scripts like this.

abhi-mosaic:
Could we merge this somehow with #26? It seems like backwards compat could just be part of the `convert_composer_to_hf.py` script?

dakinggg:
I think we probably want a standalone convert script so that we can have converted composer checkpoints. I just added the config stuff to my script so I could test my script last night

dakinggg (approved):

PR #31, opened 2023-05-04
Speed up tests and add release yaml

mvpatel2000:
Speeds up cpu tests

deletes mosaic_gpt yamls

adds release workflow

image
Tests arae now 4 min + overhead. I expect docker part to speed up w caching, on composer it only takes 30sec.


scripts/data_prep/convert_dataset.py
```diff
@@ -114,6 +114,16 @@ def __init__(self,
         super().__init__(hf_split, folder_split, raw_samples, truncated_samples)
 
 
+class ValSmallConstants(DataSplitConstants):
```
dakinggg, line 117: this class looks unused?

dakinggg (commented):
why'd you delete a bunch of yamls? am i missing something?

abhi-mosaic:
> why'd you delete a bunch of yamls? am i missing something?

The `mosaic_gpt` folder is gone now, for some reason it was still around.

scripts/data_prep/convert_dataset.py
mvpatel2000: @codestar12 ? Abhi said u might be familiar w this file, what does it do

dakinggg: did you mean for this to say `ValXSmallConstants`? (still unused, but at least would match everything else then

PR #33, opened 2023-05-04
HF, JSON, and CSV dataset prep

codestar12:
Splits the concat and tokenization into their own classes that can be installed and seperates the HF specific logic so scripts are stand alone


vchiley:
btw since you're changing script names, there will be README's which need to be updated accordingly

dakinggg (approved):
🚢

PR #41, opened 2023-05-04
fix inference scripts and readme

dakinggg:


abhi-mosaic (approved):

scripts/inference/convert_hf_to_onnx.py
```diff
@@ -39,6 +40,28 @@
 from llmfoundry import MPTConfig, MPTForCausalLM
```
abhi-mosaic, line 40:
```suggestion
```

abhi-mosaic: I think this one can go, right?

PR #46, opened 2023-05-04
update readme and updt __init__

vchiley:
read through the llmfoundry and scripts/train README's
updated llmfoundry's `__int__.py` to import optim and utils

### DIFF
--- a/scripts/train/README.md
+++ b/scripts/train/README.md
@@ -32,7 +32,7 @@ To make yourself a copy of C4, use `convert_dataset.py` like so:
 ```bash
 # Download the 'train_small' and 'val_small' splits and convert to StreamingDataset format
 # This will take 20-60 seconds depending on your Internet bandwidth
-# You should see two folders: `./my-copy-c4/train_small` and `./my-copy-c4/val` that are each ~0.5GB
+# You should see two folders: `./my-copy-c4/train_small` and `./my-copy-c4/val_small` that are ~1.0GB total
 # Note: We are using the `--concat_tokens` option to pre tokenize our samples to be of the max sequence length without padding
 python ../data_prep/convert_dataset.py --dataset c4 --data_subset en --out_root ./my-copy-c4 --splits train_small val_small --concat_tokens 2048 --tokenizer EleutherAI/gpt-neox-20b --eos_text '<|endoftext|>'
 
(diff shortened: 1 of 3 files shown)


scripts/train/README.md
```diff
@@ -117,19 +117,19 @@ composer --world_size 16 --node_rank 1 --master_addr 0.0.0.0 --master_port 7501
...
 # export MASTER_ADDR=0.0.0.0
 # export MASTER_PORT=7501
-composer train.py yamls/mosaic_gpt/125m.yaml
+composer train.py yamls/mpt/125m.yaml
 ```
 
 You should see logs being printed to your terminal like so.
 You can also easily enable other experiment trackers like Weights and Biases or CometML,
-by using [Composer's logging integrations](https://docs.mosaicml.com/en/stable/trainer/logging.html).
+by using [Composer's logging integrations](hhttps://docs.mosaicml.com/projects/composer/en/latest/trainer/logging.html).
```
dakinggg, line 132: typo

scripts/train/README.md
```diff
@@ -54,11 +54,11 @@ To verify that the dataloader works, run a quick test on your `val_small` split
 # This will construct a `StreamingTextDataset` dataset from your `val` split,
 # pass it into a PyTorch Dataloader, and iterate over it and print samples.
 # Since we only provide a local path, no streaming/copying takes place.
-python ../data_prep/text_data.py --local_path ./my-copy-c4 --split val_small
+python ../../llmfoundry/data/text_data.py --local_path ./my-copy-c4 --split val_small
```
alextrott16, line 57: Can you double check that the extra `../` is necessary?

scripts/train/README.md
```diff
@@ -346,4 +346,4 @@ In previous blogs ([1](https://www.mosaicml.com/blog/farewell-oom), [2](https://
 we also demonstrated Auto Grad Accum, which takes things a step further by letting Composer determine the `device_train_microbatch_size` on its own.
 This makes our configs not only device-count-agnostic, but hardware-agnostic too!
 You can try out this feature by setting `device_train_microbatch_size: auto`, but bear in mind that FSDP support is still in alpha mode
-and may not always work with Auto Grad Accum (but we are working on it!).
+and may not always work with Auto Grad Accum (we are working on it!).
```
dakinggg, line 349: auto microbatchin not auto grad accum

dakinggg (approved):

alextrott16 (approved):
The selected components mentioned in `llmfoundry/README.md` seem like a bit of an odd collection. I think it's fine for now, but as a future TODO we may want to flesh this out a bit and/or organize it a bit more.

scripts/train/README.md
vchiley: yeah if you're in `llm-foundry/scripts/train/` and want to access `llm-foundry/llmfoundry/` you need to go up twice.

PR #47, opened 2023-05-05
Adding loss to MPTForCausalLM fwd if labels arg is not None

vchiley:
test:
1. follow [QuickStart](https://github.com/mosaicml/llm-foundry#quickstart)
2. keep training model with model def:
```
# Model
model:
  name: hf_causal_lm
  pretrained_model_name_or_path: mpt-1b-hf
  config_overrides:
    # WARNING: if setting `pretrained: true`, `max_position_embeddings` must match the
    # `max_position_embeddings` used during pre-training
    max_position_embeddings: ${max_seq_len}
  trust_remote_code: true  # not sure this is needed...
  pretrained: true  # false: only use the architecture; true: initialize with pretrained weights

```


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -636,8 +650,9 @@ def forward(self, batch):
 
     def loss(self, outputs, batch):
         targets = self.get_targets(batch)
-        return self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
-                            targets.view(-1))
+        model_output = self.loss_fn(
```
dakinggg, line 653: this isn't really the model output right? its the loss?

vchiley: yeah that edit wasn't needed...

dakinggg (approved):

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -446,16 +446,18 @@ def get_decoder(self):
...
+        past_key_values: Optional[List[Tuple[torch.FloatTensor]]] = None,
+        attention_mask: Optional[torch.ByteTensor] = None,
+        prefix_mask: Optional[torch.ByteTensor] = None,
+        sequence_id: Optional[torch.LongTensor] = None,
+        labels: Optional[torch.LongTensor] = None,
+        return_dict: Optional[bool] = None,
+        output_attentions: Optional[bool] = None,
+        output_hidden_states: Optional[bool] = None,
+        use_cache: Optional[bool] = None,
+    ):
```
dakinggg, line 460: we still need to bool cast the attention mask and prefix mask here

abhi-mosaic:
Ok i've tried pushing to this PR a couple different ways but have not figured out how to do it properly. Here's the diff required (1 file)

Screenshot 2023-05-04 at 7 41 49 PM

abhi-mosaic:
@vchiley you'll need to make this diff in order to fix the finetuning, and the steps you listed at the top of the PR will work

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -446,19 +446,26 @@ def get_decoder(self):
...
         return_dict = return_dict if return_dict is not None else self.config.return_dict
         use_cache = use_cache if use_cache is not None else self.config.use_cache
 
+        if attention_mask is not None:
+            attention_mask = attention_mask.bool()
+        if prefix_mask is not None:
+            prefix_mask = prefix_mask.bool()
```
abhi-mosaic, line 467: I think this has to go inside the `MPTModel`, not the `MPTForCausalLM`, to be most safe

vchiley: good point

PR #55, opened 2023-05-06
Update inference benchmarking script

abhi-mosaic:
This PR:
* standardizes some config names
* simplifies the logic for profiling and warmup
* adds an option for `use_cache`
* corrects the calculation for `ms_per_seq_output_token`, this is the latency per sequence-level token, not the latency per batch token


abhi-mosaic:
Tested by rerunning `1b.yaml` and `7b.yaml` and they match earlier results.

dakinggg (approved):

PR #84, opened 2023-05-09
Bump composer version

vchiley:
Was asked to run try [this streaming branch out](https://github.com/mosaicml/streaming/pull/219), so I tested updating composer as well. No issues; seems slightly faster on 125m model (might just be runtime variance, but I'm not sure)
Screenshot 2023-05-09 at 12 57 20 PM
Screenshot 2023-05-09 at 12 57 53 PM


dakinggg (approved):

mvpatel2000 (approved):
Raising min should not change anything (it will always use 14.1). but we should raise pin anyways

PR #85, opened 2023-05-09
Updates to prefixlm and t5

alextrott16:
Some of our past updates didn't reach the prefixlm or t5 model classes. This resolves that and updates the tests that missed this.

Adds a t5-small example YAML for dolly IFT, which runs locally on CPU w/:
`composer train.py yamls/hf_t5/t5-small_dolly_sft.yaml global_train_batch_size=2 device_eval_batch_size=2 device_train_microbatch_size=2 precision=fp32 fsdp_config=null`

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -22,6 +22,7 @@ from llmfoundry import (COMPOSER_MODEL_REGISTRY, ComposerHFCausalLM,
                         ComposerHFPrefixLM)
 from llmfoundry.models.layers import NORM_CLASS_REGISTRY, build_alibi_bias
 from llmfoundry.models.mpt import MPTConfig, MPTForCausalLM
+from llmfoundry.utils import build_tokenizer
 
 
 def get_config(conf_path='scripts/train/yamls/mpt/testing.yaml') -> DictConfig:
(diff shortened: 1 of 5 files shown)


tests/test_model.py
```diff
@@ -205,8 +206,10 @@ def test_full_forward_and_backward_gpt2_small(prefixlm, batch_size=2):
     else:
         neo_cfg.model.name = 'hf_causal_lm'
 
-    model = COMPOSER_MODEL_REGISTRY[neo_cfg.model.name](
-        neo_cfg.model, neo_cfg.tokenizer).to(device)
+    tokenizer = build_tokenizer(neo_cfg.tokenizer)
+
+    model = COMPOSER_MODEL_REGISTRY[neo_cfg.model.name](neo_cfg.model,
+                                                        tokenizer).to(device)
```
dakinggg, line 212: to avoid the silent issues from before, can you add an assert here that `model.tokenizer` is the actually a tokenizer? (same for the other test)

dakinggg (approved):
LGTM, thanks for unifying!

Taytay:
No need to hold this up from my question, but I saw there used to be an exact_match option for validation. Does that still exist somewhere? (It would come in handy for me)

alextrott16:
@Taytay We could add it back, I think. Could you file a feature request? (Just open a new issue and use the "enhancement" label -- you can assign it to me if it lets you)

Issue #89, opened 2023-05-10
TypeError: Object of type DictConfig is not JSON serializable for HFCausalLM


zanussbaum:
When using `train.py` in `scripts/train` with a config like 

```
model:
  name: hf_causal_lm
  device: cpu
  pretrained: true
  pretrained_model_name_or_path: mosaicml/mpt-7b
  config_overrides:
    attn_config:
      attn_impl: triton
```
we get the error 

```
Traceback (most recent call last):
  File "/home/paperspace/llm-foundry/scripts/train/train.py", line 376, in <module>
    main(cfg)
  File "/home/paperspace/llm-foundry/scripts/train/train.py", line 365, in main
    trainer.fit()
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/trainer/trainer.py", line 1766, in fit
    self._train_loop()
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/trainer/trainer.py", line 1996, in _train_loop
    self.engine.run_event(Event.BATCH_CHECKPOINT)
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/core/engine.py", line 293, in run_event
    self._run_nonlogger_callbacks(event)
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/core/engine.py", line 475, in _run_nonlogger_callbacks
    self._run_callbacks(event, callbacks)
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/core/engine.py", line 467, in _run_callbacks
    cb.run_event(event, self.state, self.logger)
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/core/callback.py", line 96, in run_event
    return event_cb(state, logger)
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/callbacks/checkpoint_saver.py", line 346, in batch_checkpoint
    self._save_checkpoint(
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/callbacks/checkpoint_saver.py", line 384, in _save_checkpoint
    saved_path = checkpoint.save_checkpoint(
  File "/home/paperspace/llm-foundry/env/lib/python3.9/site-packages/composer/utils/checkpoint.py", line 518, in save_checkpoint
    'state': state.state_dict(),
[...]

PR #90, opened 2023-05-10
hf dict cfg overrides

vchiley:
fixes issue: https://github.com/mosaicml/llm-foundry/issues/89

### DIFF
--- a/llmfoundry/models/hf/hf_causal_lm.py
+++ b/llmfoundry/models/hf/hf_causal_lm.py
@@ -46,11 +46,22 @@ class ComposerHFCausalLM(HuggingFaceModelWithZLoss):
     """
 
     def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
+        config_overrides = om_model_config.get('config_overrides', {})
+        attn_config_overrides = config_overrides.pop('attn_config', {})
+        init_config_overrides = config_overrides.pop('init_config', {})
+
         config = AutoConfig.from_pretrained(
             om_model_config.pretrained_model_name_or_path,
             trust_remote_code=om_model_config.get('trust_remote_code', True),
             use_auth_token=om_model_config.get('use_auth_token', False),
-            **om_model_config.get('config_overrides', {}))
+            **config_overrides)
+
+        if attn_config_overrides:
+            print(f'{attn_config_overrides=}')
+            config.attn_config.update(attn_config_overrides)
+        if init_config_overrides:
+            print(f'{init_config_overrides=}')
+            config.init_config.update(init_config_overrides)
 
         train_metrics = [
             LanguageCrossEntropy(len(tokenizer)),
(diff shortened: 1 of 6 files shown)


dakinggg (approved):
Maybe add a unit test?

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -47,7 +47,19 @@ def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
...
+        )
+
+        # set config overrides
+        for k, v in om_model_config.get('config_overrides', {}).items():
+            if not hasattr(config, k):
+                raise ValueError(
+                    f'config does not have attribute "{k}" to override ({k}: {v}).'
+                )
+
+            if isinstance(getattr(config, k), Dict):
```
dakinggg, line 59:
```suggestion
            if isinstance(getattr(config, k), Mapping):
```

might as well be permissive

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -47,7 +47,19 @@ def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
...
+                getattr(config, k).update(v)
```
dakinggg, line 60: do you want to check that any keys in `v` are present in `k`? that would prevent typos in the nested config from getting silently ignored.

vchiley: I was going to, but then realized that init_config might args which are not part of the default setting so i decided to let this checking slide and hope for the best... (pulling some HF ish)

dakinggg: is it worth defaulting all args in the init defaults?

alextrott16:
Should these same changes be incorporated into `hf_prefix_lm.py` and `hf_t5.py`?

llmfoundry/models/hf/hf_causal_lm.py
vchiley: sure

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -47,7 +47,25 @@ def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
...
+                    f'config does not have attribute "{k}" to override ({k}: {v}).'
+                )
+
+            attr = getattr(config, k)
+            if isinstance(attr, Mapping):
+                extra_keys = [_k for _k in v.keys() if _k not in attr.keys()]
+                if extra_keys:
+                    raise ValueError(f'Config dict override got unknown keys. '
+                                     f'Extra keys: {extra_keys}. '
+                                     f'Expected keys: {list(attr.keys())}.')
```
dakinggg, line 65: "Expected (a subset of) keys: ...`

dakinggg (approved):
LGTM

PR #101, opened 2023-05-10
Make mpt7b finetuning more obvious

samhavens:
Resolves RESEARCH-710

Also change the folder structure of the YAMLs to make it more clear which MPTs are pretrained from the HF Hub and which are an architecture for pretraining.

WARNING: This will not work until #90 is merged!


scripts/train/yamls/hf_causal_lm/mpt-7b_dolly_sft.yaml
```diff
@@ -0,0 +1,126 @@
...
+# Run Name
+run_name: # If left blank, will be read from env var $COMPOSER_RUN_NAME
+
+model:
+  name: hf_causal_lm
+  device: cpu
+  pretrained: true
+  pretrained_model_name_or_path: mosaicml/mpt-7b
+  config_overrides:
+    attn_impl: triton
```
dakinggg, line 13: this is no longer the correct way to do this, and actually is not possible until https://github.com/mosaicml/llm-foundry/pull/90 goes in

samhavens: Noted in the PR description. I think I am fine just waiting for #90 so that we can default to finetuning with Triton

vchiley:
Note: #90 is merged

scripts/train/yamls/finetune/1b_local_data_sft.yaml
```diff
@@ -1,3 +1,9 @@
+############################################################################################
+# NOTE: This uses the MPT _architecture_, but not the HuggingFace pretrained model!        #
+# Use this YAML to train directly from a Composer checkpoint.                              #
+# This is not the right YAML if you are trying to finetune a HuggingFace pretrained model. #
+############################################################################################
```
alextrott16, line 5: ❤️

scripts/train/yamls/finetune/mpt-7b_dolly_sft.yaml
```diff
@@ -0,0 +1,125 @@
+max_seq_len: 2048
+global_seed: 17
+
+# Run Name
+run_name: # If left blank, will be read from env var $COMPOSER_RUN_NAME
+
+model:
+  name: hf_causal_lm
+  device: cpu
```
alextrott16, line 9:
```suggestion
```
`device` isn't actually a config option, despite what all our YAMLs say. OOPS!
It's supposed to be `init_device`, but we can just remove the option because it needs to be `cpu` (which it defaults to) when you have `pretrained: true`, which all of these obviously will.

scripts/train/yamls/finetune/mpt-7b_dolly_sft.yaml
```diff
@@ -0,0 +1,125 @@
...
+  pretrained: true
+  pretrained_model_name_or_path: mosaicml/mpt-7b
+  config_overrides:
+    attn_impl: triton
```
alextrott16, line 13:
```suggestion
    attn_config:
      attn_impl: triton
```

alextrott16 (approved):
A couple minor suggestions to correct some issues with model configs but otherwise looks great!

BTW `t5-small_dolly_sft.yaml` also has the `device` instead of `init_device` issue, but I couldn't suggest a change on it because it was just a rename. Please apply the same fix there.

scripts/train/yamls/finetune/opt.yaml
alextrott16, line 1: I think this wound up in the wrong folder? Looks to be a pretrain file.

scripts/train/yamls/finetune/mpt-7b_dolly_sft.yaml
```diff
@@ -0,0 +1,125 @@
...
+  - 0.999
+  eps: 1.0e-8
+  weight_decay: 0
+
+algorithms:
+  gradient_clipping:
+    clipping_type: norm
+    clipping_threshold: 1.0
+
+  max_duration: 2ep  # 2-3 epochs seems like the sweet spot
```
dakinggg, line 83: I think this block is accidentally indented (i.e. its under `algorithms` right now

dakinggg (approved):

scripts/train/yamls/finetune/mpt-7b_dolly_sft.yaml
zhranj, line 1: Since this one does not have the header about not using the HuggingFace pretrained model as a baseline, does it mean it is indeed using it?

I noticed you did not publish models smaller than 7b to HF, do you have plans to do that in the future? Thanks!

alextrott16: That's correct. That YAML lets you finetune the MPT-7b model we put on the HF hub. @samhavens perhaps we should consider adding a header to the HF YAMLs as well?

TBD on other model releases :)

Issue #106, opened 2023-05-11
the problem about mosaicml-streaming


sysusicily:
when I follow pip install -e ".[gpu]"，I find the error about mosaicml-streaming 
#-------------------------------------------------------------------------------------------
root@7730f5bd29fa:/home/mosaicml/llm-foundry# pip list|grep stream
mosaicml-streaming     0.2.1
root@7730f5bd29fa:/home/mosaicml/llm-foundry#
root@7730f5bd29fa:/home/mosaicml/llm-foundry# python scripts/data_prep/convert_dataset_hf.py --dataset c4 --data_subset en --out_root ./my-copy-c4 --splits train_small val_small --concat_tokens 2048 --tokenizer EleutherAI/gpt-neox-20b --eos_text '<|endoftext|>'
Traceback (most recent call last):
  File "/home/mosaicml/llm-foundry/llmfoundry/__init__.py", line 8, in <module>
    from llmfoundry.data import (ConcatTokensDataset,
  File "/home/mosaicml/llm-foundry/llmfoundry/data/__init__.py", line 5, in <module>
    from llmfoundry.data.denoising import (MixtureOfDenoisersCollator,
  File "/home/mosaicml/llm-foundry/llmfoundry/data/denoising.py", line 19, in <module>
    from llmfoundry.data.text_data import StreamingTextDataset
  File "/home/mosaicml/llm-foundry/llmfoundry/data/text_data.py", line 15, in <module>
    from streaming import Stream, StreamingDataset
ImportError: cannot import name 'Stream' from 'streaming' (/usr/lib/python3/dist-packages/streaming/__init__.py)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/mosaicml/llm-foundry/scripts/data_prep/convert_dataset_hf.py", line 19, in <module>
    from llmfoundry.data.datasets import ConcatTokensDataset, NoConcatDataset
  File "/home/mosaicml/llm-foundry/llmfoundry/__init__.py", line 32, in <module>
    raise ImportError(
ImportError: Please make sure to pip install . to get the requirements for the LLM example.
root@7730f5bd29fa:/home/mosaicml/llm-foundry#

hanlint:
Hello @sysusicily , the `Stream` functionality was introduced in `v0.4.0` ([Release Notes](https://github.com/mosaicml/streaming/releases/tag/v0.4.0)), but you are on `v0.2.1`. Can you try:
```
[...]

PR #110, opened 2023-05-11
Add minimum `mosaicml-streaming` version

hanlint:
Our current requirements are `mosaicml[streaming,nlp,wandb]>0.14`, but the composer library itself has no minimum streaming requirement (`mosaicml-streaming<1`). This PR adds a `mosaicml-streaming>=0.4.1` minimum requirement since we use the `Streams` functionality introduced in 0.4.0. 

Fixes: #106

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ classifiers = [
 ]
 
 install_requires = [
-    'composer[nlp,streaming,wandb]>=0.14.1,<0.15',
+    'composer[nlp,wandb]>=0.14.1,<0.15',
     'mosaicml-streaming>=0.4.1,<0.5',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -47,7 +47,8 @@
 ]
 
 install_requires = [
-    'mosaicml[nlp,streaming,wandb]>=0.14.1,<0.15',
+    'composer[nlp,streaming,wandb]>=0.14.1,<0.15',
```
abhi-mosaic, line 50: can we remove the `streaming` part from here? Would prefer to just have one line installing streaming

abhi-mosaic (approved):
LGTM pending the one comment

PR #116, opened 2023-05-12
Add features to hf_generate

alextrott16:
Please ignore the branch name. FSDP was planned for this, along with other features, but this PR will just add the other features:
- Extra generate control knobs
- Sweep over multiple combinations of generate param values
- Prompts from text files
- Setting maximum batch size

### DIFF
--- a/scripts/inference/hf_generate.py
+++ b/scripts/inference/hf_generate.py
@@ -1,6 +1,8 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
+import itertools
+import os
 import random
 import time
 import warnings
@@ -8,7 +10,12 @@ from argparse import ArgumentParser, ArgumentTypeError, Namespace
 from contextlib import nullcontext
 
 import torch
-from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
+from composer.core import Precision
+from composer.models import HuggingFaceModel
+from composer.trainer.dist_strategy import prepare_fsdp_module
+from composer.utils import dist, get_device
+from transformers import (AutoConfig, AutoModelForCausalLM, AutoTokenizer,
+                          PreTrainedModel)
 
 
 def get_dtype(dtype):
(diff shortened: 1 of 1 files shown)


scripts/inference/hf_generate.py
```diff
@@ -58,12 +60,25 @@ def parse_args() -> Namespace:
...
+             'prompt contained in a txt file.'
+        )
     parser.add_argument('--max_seq_len', type=int, default=None)
-    parser.add_argument('--temperature', type=float, default=1.0)
-    parser.add_argument('--top_k', type=int, default=50)
-    parser.add_argument('--top_p', type=float, default=1.0)
+    parser.add_argument('--max_new_tokens', type=int, default=100)
+    parser.add_argument('--max_batch_size', type=int, default=None)
+    parser.add_argument('--temperature', type=float, nargs='+', default=[1.0])
+    parser.add_argument('--top_k', type=int, nargs='+', default=[50])
```
abhi-mosaic, line 71: Do you know if `--top_k 50` is a good default / the same one that HF generate uses? I put it there because I read it somewhere in a guide,  but then someone said that it was a weird settting. Maybe we should consider reverting to `None`?

Similarly for `top_p` and `temperature`, but I think those are set to behave as no-op right now

scripts/inference/hf_generate.py
```diff
@@ -20,7 +22,7 @@ def get_dtype(dtype):
         return torch.bfloat16
     else:
         raise NotImplementedError(
-            f'dtype {dtype} is not supported. '
+            f'dtype {dtype} is not supported. ' +\
```
abhi-mosaic, line 25: Is this a new syntax that pyright and lint are enforcing? Just curious, I haven't used it before

scripts/inference/hf_generate.py
```diff
@@ -204,67 +231,119 @@ def main(args: Namespace) -> None:
...
+        batches = []
+        if args.max_batch_size:
+            batch = []
+            for prompt in prompt_strings:
+                batch.append(prompt)
+                if len(batch) == args.max_batch_size:
+                    batches.append(batch)
+                    batch = []
+            if len(batch) > 0:
+                batches.append(batch)
```
abhi-mosaic, line 281:
```suggestion
        if args.max_batch_size:
            bs = args.max_batch_size
            batches = [prompts[i:i+bs] for i in range(0,len(prompts),bs)]

```

abhi-mosaic (approved):
Made a few minor comments but otherwise looks good! Thanks for adding optional batching support too.

scripts/inference/hf_generate.py
alextrott16: If we wanted to use the defaults set by HF (see here https://huggingface.co/docs/transformers/main/en/main_classes/text_generation#transformers.GenerationConfig), we'd set `--top_k 50` (which is surprising, kinda) and `--top_p 1.0`.

I'm in support of setting all of these generation knobs to the HF defaults, so that if the user doesn't set them they get the default behavior, but also we'd be exposing those defaults through the arg parser. I think we'd only need to change `top_p`.

scripts/inference/hf_generate.py
alextrott16: It's not being enforced, but my IDE hates implicit string concatenation and it has convinced me of the same.

I always see some annoying warning reminding me why the `+\` explicit string concatenation is safer/better, and it honestly just makes the code so much less ugly and distracting when the warning highlights go away.

scripts/inference/hf_generate.py
alextrott16: Scratch that, they already match the HF defaults.

PR #118, opened 2023-05-12
Set pad_token_id to tokenizer.pad_token_id if not set on command line

patrickhwood:
The hf_chat.py program emits this warning message before each chat response:

The attention mask and the pad token id were not set. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
Setting `pad_token_id` to `eos_token_id`:0 for open-end generation.

Fixed by setting pad_token_id to tokenizer.eos_token_id if not set on the command line.


alextrott16 (commented):
Thanks @patrickhwood!

@samhavens I'll defer approval to you because you know the script better, but the change here is what we do in `hf_generate.py` and it doesn't seem to cause any trouble.

samhavens (approved):
Thank you!

PR #121, opened 2023-05-13
Fix licensing typo & context window typo

brianjking:
• Storywriter is not commercially viable
• 65k not 64k context window for storywriter


jfrankle:
Closing with "won't fix."

The license on the model is Apache 2.0.

65k vs. 64k depends on whether you count in base 2 or base 10.

PR #136, opened 2023-05-15
Fix HF conversion script to upload to S3 after editing the files to be HF compatible

dakinggg:
There is probably some more cleanup that can be done to this script, but this change should at least fix the bug. The bug was that an incomplete folder would be what gets uploaded to cloud.


abhi-mosaic (approved):
Approving to fix the bug. Not 100% sure I understand how moving the object store creation fixes the issue. But as long as you've tested it, it LGTM.

PR #148, opened 2023-05-16
Adds precision to eval

mvpatel2000:
Adds precision to eval. Sets MPT to bf16. For some reason, BF16 + FSDP requires mixed_precision: FULL. It works fine without FSDP. FP16 also works fine and gives basically the same numbers with FSDP on any setting.


scripts/eval/yamls/mpt_eval.yaml
```diff
@@ -1,6 +1,7 @@
 max_seq_len: 2048
 tokenizer_name: EleutherAI/gpt-neox-20b
 seed: 1
+precision: amp_bf16
```
dakinggg, line 4: should we just use fp32 for all evals? cc @abhi-mosaic

mvpatel2000: fp32 does not work with FSDP

dakinggg (approved):
Please wait for @abhi-mosaic approval as well

abhi-mosaic (approved):
Looks good for now, and we'll debug `amp_bf16` + FSDP so we can reduce memory+compute requirements in the future.

PR #150, opened 2023-05-17
Changed the keep_zip default to False to comply with StreamingDataset

karan6181:
## Description
- Changed the keep_zip default to `False` to comply with StreamingDataset.


samhavens:
This is what Streaming requires, so I'm a little surprised that this hasn't caused problems already.

karan6181:
> This is what Streaming requires, so I'm a little surprised that this hasn't caused problems already.

The finetuning script (class `StreamingFinetuningDataset`) was having an issue, not the end-to-end training (class `StreamingTextDataset`)! So that's the reason why this is working for certain use-case.

samhavens (approved):

PR #157, opened 2023-05-17
Update StreamingDataset defaults

abhi-mosaic:
* Change default shuffle_algo to `py1b` rather than `py1s`
* Add support for `shuffle_block_size` defaulted to `1 << 18 ~= 256k`


alextrott16 (approved):
LGTM. Thanks!

Issue #154, opened 2023-05-17
Use the nvidia-supplied nvidia-ml-py instead of pynvml


mattip:
This project [depends on pynvml](https://github.com/mosaicml/llm-foundry/blob/93e3290fe2c228a50a67546e6191b71db7171ccb/setup.py#L57), where the nvidia-blessed bindings to nvml are in the [nvidia-ml-py](https://pypi.org/project/nvidia-ml-py/) package instead. This is causing problems for the [gpustat project](https://github.com/wookayin/gpustat/issues/153#issuecomment-1551036105), which needs some internals from nvidia-ml-py to properly support older nvml version. Would it be possible to switch?

hanlint:
Thanks @mattip , the `pynvml` dependency may be extraneous, we are confirming!

mattip:
Thanks!

PR #165, opened 2023-05-18
Remove `pynvml`

hanlint:
The repository's `setup.py` requires `pynvml`, however I quick search showed that nowhere in the code is `pynvml` used. Can this be safely removed?

Fixes: #154


abhi-mosaic:
@hanlint yes I believe this can be safely removed, it was used with the old HealthChecker callback but I think that's been upstreamed to Composer so the dep lives there now.

So I think we should re-raise the issue there?

PR #181, opened 2023-05-20
Revert "Torch2 (#177) (#178)"

dakinggg:
This reverts commit bb7f8bb1497ee77cd02d1203b2e6c10dfb1aa691.

We are temporarily reverting the change to support torch2, because it needs to be synced with a change to the models on the huggingface hub, otherwise `mosaicml/mpt-7b` cannot be used after installing foundry from main. Additionally, a change to (and probably a test for) the `convert_composer_to_hf.py` script is needed, as the new relative import inside a function is not getting parsed properly.


vchiley (approved):
😢

germanjke:
i have some problems from you datasets folder and datasets library, i just renamed your folder, it's was circular import error

germanjke:
@vchiley what's about use of mpt from your torch2.0 branch? it will be installed right?

PR #192, opened 2023-05-22
Tensor Parallel MLP with torch2.0

dskhudia:
This PR adds torch 2.0 based tensor parallel support for the ffn block. It's ported over from https://github.com/mosaicml/examples/pull/255 

Currently the trained weights don't match between parallel/no-parallel versions even in a simple example. Created a PyTorch issue: https://github.com/pytorch/pytorch/issues/102280

Test: I'll add a test (the one submitted in PyTorch issue) once it works fine.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,7 @@
 default_language_version:
   python: python3
 exclude: llmfoundry/models/layers/flash_attn_triton.py
+exclude: llmfoundry/models/utils/tensor_parallel.py
 repos:
 - repo: https://github.com/google/yapf
   rev: v0.32.0
--- a/llmfoundry/models/mpt/configuration_mpt.py
+++ b/llmfoundry/models/mpt/configuration_mpt.py
@@ -7,6 +7,8 @@ from typing import Dict, Optional, Union
 
 from transformers import PretrainedConfig
 
+from llmfoundry.models.utils.misc import is_torch_2_or_higher
+
 attn_config_defaults: Dict = {
     'attn_type': 'multihead_attention',
     'attn_pdrop': 0.0,
@@ -54,6 +56,7 @@ class MPTConfig(PretrainedConfig):
         embedding_fraction: float = 1.0,
         norm_type: str = 'low_precision_layernorm',
         use_cache: bool = False,
+        tensor_parallel_mlp: bool = False,
         init_config: Dict = init_config_defaults,
         **kwargs,
     ):
@@ -95,6 +98,8 @@ class MPTConfig(PretrainedConfig):
             norm_type (str): choose type of norm to use
             multiquery_attention (bool): Whether to use multiquery attention implementation.
             use_cache (bool): Whether or not the model should return the last key/values attentions
+            tensor_parallel_mlp (bool): Whether or not to tensor parallelize the MLPs. Only available with torch version
+                2.0 or higher.
             init_config (Dict): A dictionary used to configure the model initialization:
                 init_config.name: The parameter initialization scheme to use. Options: 'default_', 'baseline_',
                     'kaiming_uniform_', 'kaiming_normal_', 'neox_init_', 'small_init_', 'xavier_uniform_', or
@@ -128,6 +133,7 @@ class MPTConfig(PretrainedConfig):
         self.embedding_fraction = embedding_fraction
         self.norm_type = norm_type
         self.use_cache = use_cache
+        self.tensor_parallel_mlp = tensor_parallel_mlp
         self.init_config = init_config
         if 'name' in kwargs:
             del kwargs['name']
@@ -188,6 +194,10 @@ class MPTConfig(PretrainedConfig):
                       str) and self.logit_scale != 'inv_sqrt_d_model':
             raise ValueError(
                 f"{self.logit_scale=} is not recognized as an option; use numeric value or 'inv_sqrt_d_model'."
+            )
+        if not is_torch_2_or_higher() and self.tensor_parallel_mlp:
+            raise NotImplementedError(
+                'Tensor Parallel with torch version < 2.0.0 is not implemented. Please set tensor_parallel_mlp to false.'
             )
         if self.init_config.get('name', None) is None:
             raise ValueError(f"{self.init_config=} 'name' needs to be set.")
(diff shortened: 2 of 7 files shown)


dskhudia:
@vchiley : Are these test failures due to a recent change? These don't look related to this PR.

vchiley:
Screenshot 2023-05-25 at 11 43 04 AM

All tests pass from last PR merge

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -7,6 +7,8 @@
 
 from transformers import PretrainedConfig
 
+from llmfoundry.models.utils import is_torch_2_or_higher
```
dakinggg, line 10: im actually not entirely sure why, but i think changing this to `from llmfoundry.models.utils.misc import is_torch_2_or_higher` will fix your test failure. the way huggingface does imports is a bit tricky.

alextrott16 (commented):
TBH I don't feel like I am familiar enough with the pytorch TP tools to evaluate the correctness of this implementation. I don't notice anything obviously wrong with the MPT code (the config stuff looks good and the TP wrapping of the MPT blocks seems to make sense at a high level). But I don't think I can provide deeper feedback than that.

That said, if any specific questions come up that you think I'll have the answer to, don't hesitate to ask.

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -7,6 +7,8 @@
 
 from transformers import PretrainedConfig
 
+from llmfoundry.models.utils.misc import is_torch_2_or_higher
```
vchiley, line 10: [Composer has this util](https://github.com/mosaicml/composer/blob/dev/composer/utils/misc.py#L87). Should we just import it from there?

llmfoundry/models/utils/misc.py
vchiley, line 1: https://github.com/mosaicml/llm-foundry/pull/192/files#r1211983698

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -92,6 +116,33 @@ def __init__(self, config: MPTConfig):
...
+                    # Move them back to cpu so that FSDP wrapping sees all params on cpu.
+                    # Othewise FSDP wrapping fails as it sees some params on cpu and others on gpu.
+                    if config.init_device == 'cpu':
+                        block = block.to('cpu')
+                    new_blocks.append(block)
+                self.blocks = new_blocks
+            else:
+                raise ValueError('Enabling 2D parallelism with FSDP failed (call to enable_2d_with_fsdp). ' + \
+                    'Please make sure your torch installation supports 2D parallelism with FSDP')
+
```
vchiley, line 145: Could this be packaged away into a util fn so that we can keep the modeling_mpt.py file clean?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -92,6 +116,33 @@ def __init__(self, config: MPTConfig):
...
+                    block = parallelize_module(
+                        module=block,
+                        device_mesh=twod_mesh,
+                        parallelize_plan={'ffn': PairwiseSequenceParallel()},
+                        tp_mesh_dim=1)
+                    # Call to parallelize_module moves params to gpu if they are cpu params.
+                    # Move them back to cpu so that FSDP wrapping sees all params on cpu.
+                    # Othewise FSDP wrapping fails as it sees some params on cpu and others on gpu.
+                    if config.init_device == 'cpu':
+                        block = block.to('cpu')
```
vchiley, line 139: does this break meta init?

.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 default_language_version:
   python: python3
 exclude: llmfoundry/models/layers/flash_attn_triton.py
+exclude: llmfoundry/models/utils/tensor_parallel.py
```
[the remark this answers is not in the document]
vchiley: see [here](https://github.com/mosaicml/llm-foundry/pull/240/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R25)

llmfoundry/models/mpt/configuration_mpt.py
dakinggg: im ok with keeping this in foundry. minimizing composer dependencies has benefits for end users and huggingface compat.

llmfoundry/models/utils/__init__.py
```diff
@@ -6,12 +6,17 @@
...
 __all__ = [
     'AutoTokenizerForMOD', 'adapt_tokenizer_for_denoising',
     'convert_hf_causal_lm_to_prefix_lm', 'init_empty_weights',
     'add_bidirectional_mask_if_missing', 'generic_param_init_fn_',
-    'MODEL_INIT_REGISTRY'
+    'MODEL_INIT_REGISTRY', 'is_torch_2_or_higher'
 ]
+
+if is_torch_2_or_higher():
+    from llmfoundry.models.utils.tensor_parallel import PairwiseSequenceParallel
```
dakinggg, line 21: I guess this one doesn't need to be in a try/except for some reason? thats cool lol

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -92,6 +116,33 @@ def __init__(self, config: MPTConfig):
                 **config.to_dict(),
             ) for _ in range(config.n_layers)
         ])
+        if config.tensor_parallel_mlp and is_torch_2_or_higher():
+            if enable_2d_with_fsdp():
+                device_type = 'cuda' if get_device(
+                    None).name == 'gpu' else 'cpu'
+                world_size = dist.get_world_size()
+                node_count = world_size // dist.get_local_world_size()
```
dakinggg, line 124: could you add a comment here that this assumes a constant local world size across nodes (or only one node is smaller)? We have a TODO to at some point make sure everything (composer, foundry, streaming) work with non-constant local world size.

cliangyu:
Is self-attention parallelizable with some code modification?

dskhudia:
> Is self-attention parallelizable with some code modification?

It is but with code modifications. See https://pytorch.org/docs/stable/_modules/torch/distributed/tensor/parallel/multihead_attention_tp.html#TensorParallelMultiheadAttention

dakinggg:
@dskhudia should we close this?

dskhudia:
@dakinggg yes.

PR #193, opened 2023-05-23
torch2 updt with hf fixes

vchiley:
[Revert "Revert "Torch2"](https://github.com/mosaicml/llm-foundry/pull/181)
Add fixes so HF is happy

re-torch2+
cc @sashaDoubov @dskhudia


.github/workflows/pr-gpu.yaml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
     - main
     - release/*
-  pull_request_target:
+  pull_request:
```
dakinggg, line 7: don't merge

tests/test_hf_conversion_script.py
```diff
@@ -0,0 +1,122 @@
...
+
+def get_config(
+        conf_path='scripts/train/yamls/pretrain/testing.yaml') -> DictConfig:
+    os.environ['TOKENIZERS_PARALLELISM'] = 'false'
+    with open(conf_path) as f:
+        test_cfg = om.load(f)
+    return cast(DictConfig, test_cfg)
+
+
+def test_convert_and_generate_torch(tmp_path):
```
sashaDoubov, line 47: should this have a pytest decorator?

vchiley: I think this is fine

I'll let @dakinggg answer (he wrote this part)

tests/test_dataloader.py
```diff
@@ -177,7 +177,7 @@ def test_denoising_dataloader(decoder_only_format, pretokenize, packing_ratio):
                 'sequence_mask_ratios': 0.25,
             },
             'drop_last': False,
-            'num_workers': 0,
+            'num_workers': 4,
```
dakinggg, line 180: any reason you chose four?

tests/test_hf_conversion_script.py
dakinggg: yeah, this is ok. This is a CPU only test, the one below tests triton + gpu

tests/test_model.py
```diff
@@ -383,7 +383,7 @@ def test_loss_fn():
                                  eps=test_cfg.optimizer.eps,
                                  weight_decay=test_cfg.optimizer.weight_decay)
 
-    for i in range(25):
+    for i in range(15):
```
dakinggg, line 386: should I be worried about this?

tests/test_hf_mpt_gen.py
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    # build tokenizer
+    tokenizer = build_tokenizer(test_cfg.tokenizer)
+
+    # build model
+    model = COMPOSER_MODEL_REGISTRY[test_cfg.model.name](test_cfg.model,
+                                                         tokenizer)
+    test_cfg.n_params = sum(p.numel() for p in model.parameters())
+    print(f'{test_cfg.n_params=:.2e}')
```
dakinggg, line 52: remove?

tests/test_hf_mpt_gen.py
```diff
@@ -0,0 +1,63 @@
...
+import pytest
+from composer.core.precision import get_precision_context
+from composer.utils import dist, get_device, reproducibility
+from omegaconf import OmegaConf as om
+
+from llmfoundry import COMPOSER_MODEL_REGISTRY
+from llmfoundry.utils import build_tokenizer
+
+
+@pytest.mark.gpu
```
dakinggg, line 13: we want this to run in the cpu image, so i'd remove this mark and then just skip if device == 'gpu' and cuda not available.

tests/test_model.py
vchiley: I dont think so

I think this test is super aggressive anyway
(I personally think it should be like 1 or 2 steps and we should add a check for checking param.grad being allclose, but this is fine, I guess)

25 steps of SGD with a ton of numerical diffs... Why is this how we test the loss fn???
+ the if the torch2 crossentropy is implemented diff, it'll have diff numerical diff properties

tests/test_dataloader.py
vchiley: torch2 only allows prefetching if you have multiple workers (which makes sense)
I added workers instead of trying to strip out prefetching; could technically have gone either way.

@alextrott16 said it was ok (he wrote the test)

tests/test_hf_mpt_gen.py
vchiley: doesn't matter since pytest will hide it, but sure why not.

tests/test_hf_mpt_gen.py
vchiley: added
```
def test_init_hfhub_mpt_cpu():
    test_init_hfhub_mpt(device='cpu', attn_impl='torch')
```

vchiley: Screenshot 2023-05-23 at 2 33 52 PM

I think the bigger issue is that none of our tests run on the cpu img

I just emulated what was already there...

vchiley:
All tests pass
Screenshot 2023-05-24 at 4 02 17 PM

.github/workflows/pr-gpu.yaml
vchiley: undid

dakinggg (approved):

PR #204, opened 2023-05-24
Update inference README

abhi-mosaic:
I renamed the script to be a bit more descriptive and added a Table of Contents to the inference README.md.

The main reason for the rename is -- I want users to know that they are converting a **HuggingFace** checkpoint folder to FT rather than a **Composer** checkpoint to FT.


samhavens (approved):
I think this makes it more clear that this document is not a step-by-step instruction list, but a collection of documents that might help

dskhudia (approved):

PR #209, opened 2023-05-24
Use $RUN_NAME rather than $COMPOSER_RUN_NAME

abhi-mosaic:
I believe we should be using $RUN_NAME with MosaicML platform rather than $COMPOSER_RUN_NAME.

@siriuslee can you confirm?

### DIFF
--- a/mcli/mcli-1b-eval.yaml
+++ b/mcli/mcli-1b-eval.yaml
@@ -12,7 +12,7 @@ command: |
 
 image: mosaicml/pytorch:1.13.1_cu117-python3.10-ubuntu20.04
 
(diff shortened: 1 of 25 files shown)


siriuslee (commented):
So, we set both. I like `$RUN_NAME` personally, but I _think_ composer uses `$COMPOSER_RUN_NAME` so I'm not sure some of these comments are accurate

abhi-mosaic:
> but I think composer uses $COMPOSER_RUN_NAME

Ok understood. I think in LLM Foundry we explicitly set the `run_name` everywhere we use `Trainer` so we are not relying on Composer's defaults.

dakinggg (approved):

PR #210, opened 2023-05-25
Kv cache speed

vchiley:
There are rearranges (reshape and transpose) that happen [here](https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/layers/attention.py#L47)

you can kv cache before or after the transposes
if you do it before, then the entire cache needs to be transposed instead of just a seq len 1 tensor.
we do it before 🤦‍♂️ 

this PR makes it happen after
\+ some random style things.

Note: this only affects `attn_impl: torch`; triton and flash expect input shape: `[b, s, h, d_head]`

This PR also
- makes some code style changes
- fixes bug with `output_hidden_states` where the last state wasn't output (cc @dakinggg)
  - in [hf docs](https://huggingface.co/docs/transformers/main_classes/output#transformers.modeling_outputs.BaseModelOutputWithPast) its noted that last hidden state should be part of hidden states as well
- enable `output_attentions` if using `attn_impl: torch` (cc @samhavens)

# Results
Using the `llm-foundry/scripts/inference/benchmarking/benchmark.py` script with `yamls/1b.yaml` updating
```
model:
  attn_config:
    attn_impl: torch

model_dtype: bf16
autocast_dtype: bf16

batch_sizes: [1]
input_lengths: [3072]
output_lengths: [1, 11, 101]
```

we get:

```
name, latency (s), throughput (tokens/s), latency_per_sequence_output_token (ms)

============================== MAIN =============================================

MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_128_1, 0.015, 68.862, 14.522
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_128_11, 0.156, 70.695, 14.145
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_128_101, 1.443, 69.989, 14.288

MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_3072_1, 0.182, 5.489, 182.167
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_3072_11, 0.328, 33.494, 29.857
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_3072_101, 1.629, 61.994, 16.131

============================== This Branch =============================================

MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_128_1, 0.014, 69.881, 14.310
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_128_11, 0.136, 80.677, 12.395
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_128_101, 1.235, 81.767, 12.230

MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_3072_1, 0.182, 5.498, 181.882
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_3072_11, 0.304, 36.217, 27.611
MosaicDeepspeed_1xA100-80GB-bf16-GPT-1B_1_3072_101, 1.393, 72.526, 13.788
```

We see that this new code path helps with torch inference speed when using kv-cache.

### DIFF
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -19,7 +19,7 @@ from llmfoundry.models.layers.norm import LPLayerNorm
 def _reset_is_causal(num_query_tokens: int, num_key_tokens: int,
                      original_is_causal: bool):
     # disable causal when it is not needed
-    # absolutely necissary for flash, but useful for torch attn
+    # necessary for flash & triton for generation with kv_cache
     if original_is_causal and num_query_tokens != num_key_tokens:
         if num_query_tokens != 1:
             raise NotImplementedError(
@@ -51,6 +51,12 @@ def scaled_multihead_dot_product_attention(
     v = rearrange(value, 'b s (h d) -> b h s d', h=kv_n_heads)
 
     if past_key_value is not None:
+        # attn_impl: flash & triton use kernels which expect input shape [b, s, h, d_head].
+        # kv_cache is therefore stored using that shape.
+        # attn_impl: torch stores the kv_cache in the ordering which is most advantagious
+        # for its attn computation ie
+        # keys are stored as tensors with shape [b, h, d_head, s] and
+        # values are stored as tensors with shape [b, h, s, d_head]
         if len(past_key_value) != 0:
             k = torch.cat([past_key_value[0], k], dim=3)
             v = torch.cat([past_key_value[1], v], dim=2)
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -53,7 +53,7 @@ def scaled_multihead_dot_product_attention(
     if past_key_value is not None:
         # attn_impl: flash & triton use kernels which expect input shape [b, s, h, d_head].
         # kv_cache is therefore stored using that shape.
-        # attn_impl: torch stores the kv_cache in the ordering which is most advantagious
+        # attn_impl: torch stores the kv_cache in the ordering which is most advantageous
         # for its attn computation ie
         # keys are stored as tensors with shape [b, h, d_head, s] and
         # values are stored as tensors with shape [b, h, s, d_head]
@@ -72,7 +72,11 @@ def scaled_multihead_dot_product_attention(
     attn_weight = q.matmul(k) * softmax_scale
 
     if attn_bias is not None:
-        attn_bias = attn_bias[:, :, -s_q:, -s_k:]
+        # clamp to 0 necessary for torch 2.0 compile()
+        _s_q = max(0, attn_bias.size(2) - s_q)
+        _s_k = max(0, attn_bias.size(3) - s_k)
+        attn_bias = attn_bias[:, :, _s_q:, _s_k:]
+
         if (attn_bias.size(-1) != 1 and
                 attn_bias.size(-1) != s_k) or (attn_bias.size(-2) != 1 and
                                                attn_bias.size(-2) != s_q):
@@ -159,7 +163,10 @@ def flash_attn_fn(
         past_key_value = (key, value)
 
     if attn_bias is not None:
-        attn_bias = attn_bias[:, :, -query.size(1):, -key.size(1):]
+        # clamp to 0 necessary for torch 2.0 compile()
+        _s_q = max(0, attn_bias.size(2) - query.size(1))
+        _s_k = max(0, attn_bias.size(3) - key.size(1))
+        attn_bias = attn_bias[:, :, _s_q:, _s_k:]
 
     if attn_bias is not None:
         raise NotImplementedError(f'attn_bias not implemented for flash attn.')
@@ -268,7 +275,10 @@ def triton_flash_attn_fn(
         past_key_value = (key, value)
 
     if attn_bias is not None:
-        attn_bias = attn_bias[:, :, -query.size(1):, -key.size(1):]
+        # clamp to 0 necessary for torch 2.0 compile()
+        _s_q = max(0, attn_bias.size(2) - query.size(1))
+        _s_k = max(0, attn_bias.size(3) - key.size(1))
+        attn_bias = attn_bias[:, :, _s_q:, _s_k:]
 
     if dropout_p:
         raise NotImplementedError(
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -89,8 +89,7 @@ def scaled_multihead_dot_product_attention(
         attn_weight = attn_weight.masked_fill(
             ~key_padding_mask.view((b, 1, 1, s_k)), min_val)
 
-    reset_is_causal = _reset_is_causal(query.size(1), key.size(1), is_causal)
-    if reset_is_causal:
+    if is_causal and (not q.size(2) == 1):
         s = max(s_q, s_k)
         causal_mask = attn_weight.new_ones(s, s, dtype=torch.float16)
         causal_mask = causal_mask.tril()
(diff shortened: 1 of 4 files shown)


llmfoundry/models/layers/attention.py
```diff
@@ -18,6 +18,8 @@
 
 def _reset_is_causal(num_query_tokens: int, num_key_tokens: int,
                      original_is_causal: bool):
+    # disable causal when it is not needed
+    # absolutely necissary for flash, but useful for torch attn
```
samhavens, line 22: `necessary`

llmfoundry/models/layers/attention.py
```diff
@@ -59,6 +66,7 @@ def scaled_multihead_dot_product_attention(
     attn_weight = q.matmul(k) * softmax_scale
 
     if attn_bias is not None:
+        attn_bias = attn_bias[:, :, -s_q:, -s_k:]
```
dakinggg, line 69: @sashaDoubov I have some vague memory that torch compile blew up at negative indexing like this. Do you remember the details?

llmfoundry/models/layers/attention.py
```diff
@@ -232,6 +254,16 @@ def triton_flash_attn_fn(
...
 
+    if past_key_value is not None:
+        if len(past_key_value) != 0:
+            key = torch.cat([past_key_value[0], key], dim=1)
+            value = torch.cat([past_key_value[1], value], dim=1)
+
+        past_key_value = (key, value)
+
+    if attn_bias is not None:
+        attn_bias = attn_bias[:, :, -query.size(1):, -key.size(1):]
```
dakinggg, line 265: same comment about negative indexing

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -332,6 +337,8 @@ def forward(
                 # get the key tensor whose spec should be (batch, seq, dim), and
                 # collect the `seq`, so that the position embedding is shifted
                 past_position = past_key_values[0][0].size(1)
+                if self.config.attn_config['attn_impl'] == 'torch':
```
dakinggg, line 340:
```suggestion
                if self.attn_impl == 'torch':
```

llmfoundry/models/layers/attention.py
```diff
@@ -42,13 +45,17 @@ def scaled_multihead_dot_product_attention(
...
     q = rearrange(query, 'b s (h d) -> b h s d', h=n_heads)
-    k = rearrange(key, 'b s (h d) -> b h d s',
-                  h=1 if multiquery else n_heads)  # includes key.t()
-    v = rearrange(value, 'b s (h d) -> b h s d', h=1 if multiquery else n_heads)
+    kv_n_heads = 1 if multiquery else n_heads
+    k = rearrange(key, 'b s (h d) -> b h d s', h=kv_n_heads)
+    v = rearrange(value, 'b s (h d) -> b h s d', h=kv_n_heads)
 
-    min_val = torch.finfo(q.dtype).min
+    if past_key_value is not None:
```
dakinggg, line 53: can you document the difference in shape for torch kv cache vs others?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -332,6 +337,8 @@ def forward(
...
+                    past_position = past_key_values[0][0].size(3)
```
dakinggg, line 341: can you adjust the comment to reflect why this is different for torch?

llmfoundry/models/layers/attention.py
sashaDoubov: Yes I got a scary error of the form:
```[41](https://mosaicml.slack.com/archives/D056APM0WB0/p1683927713531659)
    return (as_strided(buf0, (4, 2048, 2048), (12582912, 6144, 1)), as_strided(buf0, (4, 2048, 2048), (12582912, 6144, 1), 2048), as_strided(buf0, (4, 2048, 2048), (12582912, 6144, 1), 4096), as_strided(primals_2, (1, 16, 2048, 2048), (32768, 2048, 2048, 1), -4192256), buf0, as_strided(primals_1, (8192, 2048), (2048, 1)), as_strided(primals_3, (6144, 2048), (2048, 1)), )
RuntimeError: Tensor: invalid storage offset -4192256
```
This seemed to arise when dims 2 and 3 were smaller than s_q and s_k, and was trying to access elements in the attn_bias tensor with large negative indices.
The fix was clamping, something like:
```
  # necessary change for torch 2.0, otherwise there's a RuntimeError with negatives
  clamped_query_dim = max(0, attn_bias.size(2) - query.size(1))
  clamped_key_dim = max(0, attn_bias.size(3) - key.size(1))

  attn_bias = attn_bias[:, :, clamped_query_dim:, clamped_key_dim:]
```

llmfoundry/models/layers/attention.py
vchiley: 😅

llmfoundry/models/layers/attention.py
vchiley: i'm not sure this is an issue
(i think u had a bug)

if `-s_q` or `-s_k` was `-4192256`, when `max(query.size(1)) = max_seq_len` and `max(key.size(1)) = max_seq_len`...
it seems like there was a bug somewhere...

or am I trippin?

@sashaDoubov is there a min repro of how i could hit this error?

llmfoundry/models/layers/attention.py
```diff
@@ -42,13 +45,23 @@ def scaled_multihead_dot_product_attention(
...
-
-    min_val = torch.finfo(q.dtype).min
+    kv_n_heads = 1 if multiquery else n_heads
+    k = rearrange(key, 'b s (h d) -> b h d s', h=kv_n_heads)
+    v = rearrange(value, 'b s (h d) -> b h s d', h=kv_n_heads)
+
+    if past_key_value is not None:
+        # attn_impl: flash & triton use kernels which expect input shape [b, s, h, d_head].
+        # kv_cache is therefore stored using that shape.
+        # attn_impl: torch stores the kv_cache in the ordering which is most advantagious
```
dakinggg, line 56:
```suggestion
        # attn_impl: torch stores the kv_cache in the ordering which is most advantageous
```

llmfoundry/models/layers/attention.py
dakinggg: python handles the negative indexing, but compile doesn't like it.

dakinggg: i dont recall why we were in a situation where one of those was highly negative, but i think it wasn't a bug?

dakinggg (commented):
lgtm, but want to resolve the negative indexing thing before merging

llmfoundry/models/layers/attention.py
sashaDoubov: Have discussed this with both @dakinggg and @vchiley, don't think this is an issue for our code, I was able to narrow down the issue to some manual manipulation that I was performing in a different `if` branch of the code. This was reproduced in this file here, but _shouldn't_ be an issue here. https://github.com/sashaDoubov/llm-foundry/blob/repro_tensor_issue/llmfoundry/models/layers/min_repro_2_failing.py

dakinggg: doesn't have to be part of this PR, but at some point we'll add torch.compile tests

vchiley: ok sounds like it should be fixed

dakinggg (approved):

PR #214, opened 2023-05-25
G-Eval

samhavens:
# G-Eval

This is an implementation of G-Eval, which uses GPT4 as a way to judge model outputs without a ground truth (some groups have started using GPT3.5 as a judge model... this feels pennywise and pound foolish, as only GPT4 is high enough quality to correlate with human ratings... maybe). It has two modes, Score and Compare. When scoring, it gives a score (by default 1-6). When comparing, it picks between two models and decides whether one is significantly better or if it is a tie. It has implementations of LIMA's Score and Compare prompts, as well as Vicuna's compare prompt. It also contains the Vicuna and Koala open evals.

The model(s) being evaluated can be local (assuming HF pipeline support), an HTTP endpoint (currently I think only OpenAI works), or a file of records of prompts and responses.

## Issues that likely need to be addressed

- [x] You likely want to cache your model responses and score them in one step, so I should have it so when you call the model locally or over HTTP it just saves it to a cache locally so you can re-run it without calling the model again.
- [x] I need assistance with deploying a model I want to test and updating this code to use our inference server
- [ ] Not sure if my way of generating from the local model is landing on GPU
- [x] Should the user be able to pass generation settings like top_p? How? More CLI args?
- [ ] Need comments and usage instructions.
- [ ] Possibly want to split the main script into multiple files, e.g. score.py and compare.py, to keep the arguments more manageable.
- [ ] `rubric` should not be part of the `GEval` class, but a property of a `Judge` — I was confused when I initially wrote it, but a Judge will always be coupled with a rubric.
- [ ] We want to randomize the location of the models when comparing, or always send 2 requests to the judge. This is because of evidence [like in this paper](https://arxiv.org/abs/2305.17926) that GPT4 biases the first model


scripts/eval/geval/README.md
```diff
@@ -0,0 +1,7 @@
+# G-Eval
+
+This is a starting point for implementing [G-Eval: NLG Evaluation using GPT-4 with Better Human Alignment](https://arxiv.org/abs/2303.16634), which is becoming a common evaluation technique in evaluating chat/instruction-tuned models.
+
+In `open_evals` we collect some common open-ended questions which are commonly used as prompts to compare domain-general models.
+
+The most repeatable way to run this is to run an open-ended eval through a model and store the prompt and responses in a JSONL file with the keys `prompt` and `response`. Then call this script with that file as the `model1` argument.
```
bmosaicml, line 7: Can we provide a script/command that does this for our MPT hf model?

samhavens:
@bmosaicml I am closing this

PR #219, opened 2023-05-25
Fixing benchmark mcli example with proper path and image

sashaDoubov:
The LLM-foundry benchmark script is out-of-date. Tested by running `mcli run -f mcli-benchmark-mpt.yaml` successfully


abhi-mosaic (approved):
Thanks @sashaDoubov !

PR #225, opened 2023-05-26
Add support for `device_map`

abhi-mosaic:
Based on community requests and PRs, adding these few lines should enable support for Accelerate's device map for multi-GPU inference (via naive pipelining)

After this is tested and merged, we should upgrade the `modeling_mpt.py` files on HF Hub for our 3 models.

### DIFF
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -483,7 +483,9 @@ class MPTForCausalLM(MPTPreTrainedModel):
                                    output_hidden_states=output_hidden_states,
                                    use_cache=use_cache)
 
-        logits = F.linear(outputs.last_hidden_state,
+        # move outputs to same device as weights for token embedding
+        # needed to support HF `device_map`
+        logits = F.linear(outputs.last_hidden_state.to(self.transformer.wte.weight.device),
                           self.transformer.wte.weight)
 
         if self.logit_scale is not None:
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -53,6 +53,7 @@ Tokenizer = Union[PreTrainedTokenizer, PreTrainedTokenizerFast]
 class MPTPreTrainedModel(PreTrainedModel):
     config_class = MPTConfig
     base_model_prefix = 'model'
+    _no_split_modules = ['MPTBlock']
 
 
 class MPTModel(MPTPreTrainedModel):
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -513,8 +513,9 @@ class MPTForCausalLM(MPTPreTrainedModel):
 
         # move outputs to same device as weights for token embedding
         # needed to support HF `device_map`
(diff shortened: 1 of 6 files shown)


hanlint:
is there a simple test we could add here?

samhavens:
@hanlint We should be able to test that it can be instantiated with the `device_map="auto"` flag

abhi-mosaic:
Yep, will add a test before merging. Just need to look into whether we need to package `accelerate` as a `tests` dependency

abhi-mosaic:
I added a test and manually tested on a 2-GPU machine, with a `device_map` that forces multi-gpu placement (Note using `device_map=auto` does work but it is insufficinet to use for a multigpu test because the model would have to be bigger than GPU #0 to be auto split)

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -510,8 +511,11 @@ def forward(
             use_cache=use_cache,
         )
 
-        logits = F.linear(outputs.last_hidden_state,
-                          self.transformer.wte.weight)
+        # move outputs to same device as weights for token embedding
+        # needed to support HF `device_map`
+        logits = F.linear(
+            outputs.last_hidden_state.to(self.transformer.wte.weight.device),
```
vchiley, line 517: Just want to verify that this does not break training with FSDP.
`self.transformer.wte.weight` should be sharded, so `self.transformer.wte.weight.device` might not make sense
or is it that durring training, within the FSDP block, `self.transformer.wte.weight` is set to the current device and therefore `device` is correct?

abhi-mosaic: @vchiley I will do some test runs with FSDP on single node 8-GPUs. Would that satisfy your concern?

FWIW I think the `transformer.wte.weight.device` will return the expected value and the storage location (sharded across the GPUs) is hidden from the end user. Since we are within the `transformer` FSDP block, after the forward all-gather, things look "normal"

vchiley: Why not just try 2 node
run for 100 itrs and if its fine, I'm fine.

2 node tests all the things

abhi-mosaic: Confirmed that a 2-node job with MPT-1B and FSDP works with this branch.

tests/test_model.py
```diff
@@ -751,6 +753,60 @@ def test_generate(attention_impl, device, alibi):
...
+        torch_dtype=torch.bfloat16,
+        trust_remote_code=True,
+        device_map=device_map,
+    )
+    out = pipe(
+        'The quick fox jumped over',
+        max_length=10,
+        do_sample=True,
+    )
+    print(out)
```
vchiley, line 807: nuke please

tests/test_model.py
```diff
@@ -751,6 +753,60 @@ def test_generate(attention_impl, device, alibi):
             generation_with_left_padding[:, 6:])
 
 
+@pytest.mark.gpu
+@pytest.mark.parametrize('world_size', [1, 2])
+def test_generate_with_device_map(tmp_path, world_size):
```
vchiley, line 758: Should we parameterize this with
```
@pytest.mark.parametrize('use_cache', [False, True])
```

(I think it would be passed to `MPTConfig`)

vchiley (approved):
lgtm

(left a few comments)

PR #227, opened 2023-05-26
Feat: Script to generate addition eval data

ksreenivasan:
Just a simple script to generate eval data in the right format for addition samples


jacobfulano:
Are you planning on adding a jsonl file with a generated dataset, as well as a yaml to make it part of our ICL suite?

ksreenivasan:
> Are you planning on adding a jsonl file with a generated dataset, as well as a yaml to make it part of our ICL suite?

@jacobfulano Yep! I didn't do it already because I wanted to do all 3-digit additions at first which is 1m rows. But I was planning to include a random subset. I'm currently talking with @bmosaicml to see if this is even necessary because potentially there's already a [standard bigbench dataset](https://huggingface.co/datasets/tasksource/bigbench/viewer/simple_arithmetic_json_subtasks/train) which does this.

ksreenivasan:
Closing PR for now since @bmosaicml is directly including the data.

Issue #67, opened 2023-05-08
FasterTransformer


xgal:
Hi, I saw in mpt model card that the models could run with FasterTransformer
I didn't find any details about that anywhere
can you guys share the conversion scripts or help there? 

Thanks

xgal:
Thanks @vchiley 
AFAIK it is using AliBi and some other things that aren't native in GPT FT version
maybe Bloom is more similar? and from playing with mpt around in HF version and try base as GPT / Bloom (by renaming and loading state dicts) I get nonsense so I wonder if there is some other impl. that prevents it straight forward such as other ordering in QKV layers or etc. 

wdyt ?

Thanks !!

vchiley:
> Transformer Engine

We've played around with [TE and H100 FP8 support](https://www.mosaicml.com/blog/coreweave-nvidia-h100-part-1)
It works and we'll include everything when we have more seat time with H100s so we can test everything more thoroughly.

SinanAkkoyun:
@vchiley Thank you so much! 🚀😍

meitalbensinai:
I also need the FT conversion script; it would be super helpful for me 🥰

SinanAkkoyun:
@vchiley You are also talking about uzing TransformerEngine in inference right?

dskhudia:
We will soon add conversion and run scripts

therealadityashankar:
I'd greatly appreciate that so so much!

dskhudia:
@xgal , @SinanAkkoyun , @meitalbensinai , @therealadityashankar : Conversion script was added in #169 . Please give it a try and let us know if you run into any issues.

therealadityashankar:
@dskhudia this is nice, I would love a script to run this with tokenizers, or a minimal example, so I can check this out

dskhudia:
@xgal , @SinanAkkoyun , @meitalbensinai , @therealadityashankar : The script to run the converted model has also been added. Let us know if you face any issues.

SinanAkkoyun:
Hi! What TPS is expected for the FT implementation?

dskhudia:
It depends on batching vs no batching, input/output lengths etc. but I see the following numbers on A100-40G:
<img width="408" alt="Screenshot 2023-05-30 at 3 55 09 PM" src="https://github.com/mosaicml/llm-foundry/assets/37562707/a749bb1a-7d16-4b91-8e56-daad073f858a">

SinanAkkoyun:
[...]

PR #229, opened 2023-05-27
Add a script to run mpt with FasterTransformer

dskhudia:
Adds a script to run a MPT checkpoint in FT format. 

Test to be added (Basically convert a 125m param model to FT and run it and compare outputs) in a future PR. Model run passes vibe check on some sample inputs. 

Fixes https://github.com/mosaicml/llm-foundry/issues/250 and https://github.com/mosaicml/llm-foundry/issues/67

### DIFF
--- a/scripts/inference/convert_hf_mpt_to_ft.py
+++ b/scripts/inference/convert_hf_mpt_to_ft.py
@@ -125,7 +125,7 @@ def convert_weight_to_ft_each(save_dir: str, infer_gpu_num: int,
 
         split_vals = np.split(data, infer_gpu_num, axis=-1)
         for j in range(infer_gpu_num):
-            save_path = os.path.join(save_dir, f'/model.{tensor_name}.{j}.bin')
+            save_path = os.path.join(save_dir, f'model.{tensor_name}.{j}.bin')
             split_vals[j].tofile(save_path)
             if config['no_bias']:
                 write_zero_bias(tensor_name, save_path, split_vals[j].shape[-1])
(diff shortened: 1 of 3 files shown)


scripts/inference/convert_hf_mpt_to_ft.py
```diff
@@ -200,37 +205,46 @@ def convert_mpt_to_ft(model_name_or_path: str, output_dir: str,
...
+                'qk_ln is enabled for this MPT model. This may not work as expected in FT. Use --force to force a conversion.'
+            )
+
+        with open(os.path.join(save_dir, 'config.ini'), 'w') as configfile:
             config.write(configfile)
     except:
         print(f'Failed to save the config in config.ini.')
         raise
 
+    sys.exit(1)
```
abhi-mosaic, line 247: Is this intended to be here?

scripts/inference/run_mpt_with_ft.py
```diff
@@ -0,0 +1,476 @@
+# Copyright 2022 MosaicML LLM Foundry authors
```
abhi-mosaic, line 1: Naming question -- does this script support any FT model or only MPT? If the former I might suggest renaming this script to something like `ft_generate.py` to signal that this script is analogous to `hf_generate.py`

abhi-mosaic (commented):
Have not tested myself yet so will refrain from approving just yet. Left small comments

scripts/inference/run_mpt_with_ft.py
dskhudia: This only supports MPT model.

scripts/inference/convert_hf_mpt_to_ft.py
dskhudia: Good catch. No. It was for some local testing I was doing. Removing it.

abhi-mosaic (approved):

TriLoo:
I have checked this script, can get same outputs from FasterTransformer as pytorch.

dskhudia:
@TriLoo : Wonderful. Thanks for checking. Do you mind sharing your setup/script for comparing?

TriLoo:
> @TriLoo : Wonderful. Thanks for checking. Do you mind sharing your setup/script for comparing?

sorry, I cannot share the script, but the official bloom example from FasterTransformer can be a good reference.

PR #234, opened 2023-05-27
Refactor logging

hanlint:
This PR refactors the logging to:
* centralize verbosity controls into the python logging levels, instead of passing `verbose` arguments to many helper methods. Because we default `python_log_level` to `DEBUG`, however, the logs start becoming very verbose. 
* downgrades most warnings to info level. `warnings` should only be used for errors or unexpected behavior that do not require raising an exceptions, not informative messages such as reporting expected behavior.
* removes the warnings around usage of torch vs. triton vs. flash. These have led to users mistakenly believing this is the source of other errors.

Note: This PR is API-breaking. Should I keep the `verbose` arguments to the methods, and add a deprecation warnings instead that this argument is being ignored?

TODO:

- [x] add python log level control in scripts
- [x] remove remaining `verbose` and `warnings`


llmfoundry/models/utils/param_init_fns.py
```diff
@@ -88,8 +86,7 @@ def generic_param_init_fn_(
         )
 
     if init_div_is_residual is not False:
-        if verbose > 1:
-            warnings.warn(
+        log.info(
                 f'Initializing _is_residual layers then dividing them by {div_is_residual:.3f}. ' +\
                 f'Set `init_div_is_residual: false` in init config to disable this.'
             )
```
vchiley, line 92:
```suggestion
        log.info(
            f'Initializing _is_residual layers then dividing them by {div_is_residual:.3f}. ' +\
            f'Set `init_div_is_residual: false` in init config to disable this.'
        )
```

there are a few instances where you leave the extra indent

llmfoundry/data/finetuning/tasks.py
```diff
@@ -226,33 +227,27 @@ def get_preprocessing_fn_from_str(self,
...
+            )
             return self._task_preprocessing_registry[preprocessor]
 
         try:
             import_path, function_name = preprocessor.split(':', maxsplit=1)
-            if verbose:
-                print(
-                    f'Importing preprocessing function via: `from {import_path} import {function_name}`'
-                )
+            log.info(
```
dakinggg, line 248: this log can go i think?

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -135,12 +136,20 @@ def __init__(
...
         self.norm_type = norm_type
         self.use_cache = use_cache
         self.init_config = init_config
         self.fc_type = fc_type
+        if verbose is not None:
+            warnings.warn(
+                DeprecationWarning(
+                    'verbose argument for MPTConfig will be deprecated in a future release. Use python_log_level instead.'
+                ))
+            logging.getLogger('llmfoundry').setLevel(
```
dakinggg, line 149: what if we just do nothing if verbose is passed in? deprecationwarning can say that you should use proper pythong logging now. it won't break anyone's code if verbose just starts doing nothing.

scripts/train/train.py
```diff
@@ -509,6 +510,8 @@ def main(cfg: DictConfig) -> Trainer:
         eval_loader.metric_names = eval_metric_names
         evaluators.insert(0, eval_loader)  # Put the base eval_loader first
 
+    logging.getLogger('llmfoundry').setLevel(python_log_level.upper())
```
dakinggg, line 513: could you add to `eval.py` as well?

dakinggg: python_log_level can be `None`, so lets just copy the way that composer does it (including the basicconfig setup)?
https://github.com/mosaicml/composer/blob/8d59da2293499f1faabff6276eee98344f334c64/composer/trainer/trainer.py#L902-L911

llmfoundry/models/utils/param_init_fns.py
dakinggg, line 1: I think we can just drop all the logs from this file.

dakinggg (approved):
I think its ok to have the API breaking change in all the lower level classes.

llmfoundry/models/layers/attention.py
```diff
@@ -16,6 +17,8 @@
 from llmfoundry.models.layers.fc import FC_CLASS_REGISTRY
 from llmfoundry.models.layers.norm import NORM_CLASS_REGISTRY
 
+log = logging.getLogger(__name__)
+
```
vchiley, line 21:
```suggestion
```

If you're getting rid of all prints from this file, you're not logging anything / don't need it (also don't need to import it above.)

llmfoundry/models/utils/param_init_fns.py
vchiley: I think I agree.
The logs in this file are for debugging more than anything.
Not sure they need to exist long term.

hanlint: Done

hanlint:
> what if we just do nothing if verbose is passed in? deprecationwarning can say that you should use proper pythong logging now. it won't break anyone's code if verbose just starts doing nothing. 

I'm fine with that too -- it's not strictly a deprecation then, but a loss of function. But I think that's OK.

llmfoundry/models/mpt/configuration_mpt.py
hanlint: Done

hanlint:
This PR is ready to merge, pending tests. I manually ran a few training runs, and the output was not verbose, even at debug, with normal usage.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -142,19 +146,12 @@ def __init__(self, config: MPTConfig):
...
 
-        # Print verbose info
-        if config.verbose and config.verbose > 2:
-            print(self)
-        if 'verbose' not in self.config.init_config:
-            self.config.init_config['verbose'] = self.config.verbose
-        if self.config.init_config['verbose'] > 1:
-            init_fn_name = self.config.init_config['name']
-            warnings.warn(f'Using {init_fn_name} initialization.')
+        logging.debug(self)
```
dakinggg, line 152:
```suggestion
        log.debug(self)
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -142,19 +146,12 @@ def __init__(self, config: MPTConfig):
...
+        logging.debug(
```
dakinggg, line 153:
```suggestion
        log.debug(
```

dakinggg (approved):
LGTM, thanks Hanlin!

PR #235, opened 2023-05-27
Fix a race condition in ICL eval

dakinggg:
There was a race condition where rank N could delete the ICL file before other ranks managed to load it. This fixes that by only allowing local rank 0 to delete ICL files.


abhi-mosaic (approved):
LGTM, thanks!

dakinggg:
A bit hard to "prove" this works, but before this change ~3/5 runs were failing (crash before starting eval) on r1z1, and now 5/5 succeed (make it to the start of eval).

PR #240, opened 2023-05-28
Fix Typing (part 1)

hanlint:
This is the first in a series of PRs that brings this library into compliance with `pyright`. No functional changes to the code should occur with these fixes.

Before:
```
68 errors, 610 warnings, 0 informations 
```

After:
```
28 errors, 443 warnings, 0 informations 
```


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -447,7 +459,7 @@ def __init__(self, config: MPTConfig):
             raise ValueError(
                 'MPTForCausalLM only supports tied word embeddings')
 
-        self.transformer = MPTModel(config)
+        self.transformer: MPTModel = MPTModel(config)
```
vchiley, line 462: Why do this here?
Does this set precedent that ALL variables should be typed?
Or is this to help alleviate a bunch of typing issues?

tests/test_model.py
```diff
@@ -352,13 +353,15 @@ def test_loss_fn():
...
-    torch.backends.cudnn.allow_tf32 = False
+    torch.backends.cuda.matmul.allow_tf32 = False  # type: ignore (third-party)
+    torch.backends.cudnn.allow_tf32 = False  # type: ignore (third-party)
 
     conf_path = 'scripts/train/yamls/pretrain/testing.yaml'
     with open(conf_path) as f:
         test_cfg = om.load(f)
 
+    assert isinstance(test_cfg, DictConfig)
+
```
vchiley, line 364:
```suggestion
    with open(conf_path) as f:
        test_cfg: DictConfig = om.load(f)

```

Would that accomplish the same thing?

llmfoundry/callbacks/generate_callback.py
```diff
@@ -84,9 +84,11 @@ def generate(self, state: State, logger: Logger):
 
         if dist.get_global_rank() == 0:
             if self.wandb_logger is not None:
-                artifact = wandb.Artifact(
-                    'generate_samples_' + str(wandb.run.id),  # type: ignore
-                    type='predictions')
+                assert wandb.run is not None, 'wandb should have started run'
```
vchiley, line 87: we generally prefer to raising errors instead of using asserts.

llmfoundry/callbacks/fdiff_callback.py
```diff
@@ -47,10 +47,12 @@ def batch_end(self, state: State, logger: Logger):
     def eval_end(self, state: State, logger: Logger):
         if self.diff_eval_metrics:
             evaluator = state.dataloader_label
-            metrics = list(state.eval_metrics[evaluator].keys())  # type: ignore
+            assert evaluator is not None, 'dataloader should have been set'
```
vchiley, line 50: raise error instead of assert

llmfoundry/__init__.py
```diff
@@ -24,7 +24,8 @@
 
 except ImportError as e:
     try:
-        is_cuda_available = torch.cuda.is_available()  # type: ignore
+        import torch
```
vchiley, line 27: Should we import torch before the large try block above?

dakinggg: I think the intention here was to raise a nice error anytime someone did not have the necessary reqs (including torch). We've now ended up in a situation where you have to install torch _before_ installing the repo, so this try/except for the torch import and follow on message may not be the most helpful way of doing it. we can probably just move the `import torch` out of the try/except as you said.

llmfoundry/callbacks/fdiff_callback.py
dakinggg: these asserts are ok because they are just for pyright, logically the `state.dataloader_label` should never be `None` at this point.

scripts/inference/convert_composer_to_hf.py
```diff
@@ -268,7 +268,7 @@ def convert_to_relative_import(
     return '.' + parts[-1]
 
 
-def find_module_file(module_name: str) -> str:
+def find_module_file(module_name: str) -> Optional[str]:
```
dakinggg, line 271: It looks like I use this function in a way that assumes the output is not `None`, so maybe this should raise an error if `module_file` is `None`.

llmfoundry/callbacks/generate_callback.py
dakinggg: same, assert should be ok here

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -31,19 +31,30 @@
...
+    init_empty_weights  # type: ignore (see note)
+from llmfoundry.models.utils.param_init_fns import (
+    generic_param_init_fn_,  # type: ignore (see note)
+    MODEL_INIT_REGISTRY,
+)
+# isort: on
 
 try:
-    from llmfoundry.models.layers.flash_attn_triton import flash_attn_func
+    from llmfoundry.models.layers.flash_attn_triton import \
```
dakinggg, line 56: this try/except needs to remain on a single line because of huggingface import issues. I believe you can achieve this with `isort: off` and `isort: on` around it.

llmfoundry/models/mpt/modeling_mpt.py
hanlint: This alleviates a bunch of typing issues down the road -- since `transformer` is now an `MPTModel`, static type checkers know that accesses to its attributes are valid and can retrieve the typing, as opposed to resolving to `Unknown`

tests/test_model.py
hanlint: No, that would fail the checker as `om.load` can also return a `ListConfig`. 

The assert here is not just to make the typechecker happy -- it also plays a valid input checking role in making sure that user configs are actually a dict and not a list at the top level.

llmfoundry/callbacks/generate_callback.py
hanlint: `assert` is the right choice here as these are system invariants that should never logically happen. See: https://stefan.sofa-rockers.org/2018/04/16/assertions-and-exceptions/ a good explanation.

vchiley (approved):

PR #255, opened 2023-05-31
Change mcli eval YAMLs to use `mixed_precision: FULL`

abhi-mosaic:
We have been seeing some inconsistency with MPT-7B eval with FSDP `mixed_precision: PURE` but everything is fine with `mixed_precision: FULL`. While we debug this, we should set the default values in our example YAMLs to `mixed_precision: FULL`.

### DIFF
--- a/mcli/mcli-1b-eval.yaml
+++ b/mcli/mcli-1b-eval.yaml
@@ -44,6 +44,7 @@ parameters:
   load_path: # Add your (optional) Composer checkpoint path here!
 
   device_eval_batch_size: 16
+  precision: fp32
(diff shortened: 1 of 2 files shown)


bmosaicml (approved):
should this apply to the opt eval file as well? and the yamls in the scripts/eval folder?

abhi-mosaic:
> should this apply to the opt eval file as well? and the yamls in the scripts/eval folder?

I believe all the eval YAMLs should be good now with `precision: fp32` and `fsdp_config.mixed_precision: FULL`.

PR #269, opened 2023-06-02
Fix model init when using 1 GPU

abhi-mosaic:


dblalock (approved):
LGTM. Just verified that this lets the quickstart training command run if you comment out the FSDP block. Doesn't make FSDP work, but that's fine.

abhi-mosaic:
Updated the PR to remove FSDP when running on single GPU, which should avoid the issues that arise with `torch2` and single GPU FSDP.

musabgultekin (approved):
This made 1xH100 working

PR #272, opened 2023-06-02
Fastertransformer: replace config['mpt'] with config['gpt']

dwyatte:
Perhaps this works with PyTorch-based fastertransformer because `run_mpt_with_ft.py` manually instantiates a `ParallelGPT` model, but the [Triton fastertransformer backend](https://github.com/triton-inference-server/fastertransformer_backend) requires the `gpt` key in the `config.ini`

This change makes it so that `convert_hf_mpt_to_ft.py` will produce a checkpoint that works with the Triton fastertransformer backend. I've only tested `convert_hf_mpt_to_ft.py` on my branch, but I think `run_mpt_with_ft.py` should work

### DIFF
--- a/scripts/inference/convert_hf_mpt_to_ft.py
+++ b/scripts/inference/convert_hf_mpt_to_ft.py
@@ -204,30 +204,30 @@ def convert_mpt_to_ft(model_name_or_path: str,
     hf_config = vars(model.config)
 
     config = configparser.ConfigParser()
-    config['mpt'] = {}
+    config['gpt'] = {}
     try:
-        config['mpt']['model_name'] = 'mpt' if hf_config[
+        config['gpt']['model_name'] = 'mpt' if hf_config[
             '_name_or_path'] == '' else hf_config['_name_or_path']
-        config['mpt']['head_num'] = str(hf_config['n_heads'])
+        config['gpt']['head_num'] = str(hf_config['n_heads'])
         n_embd = hf_config['d_model']
-        config['mpt']['size_per_head'] = str(n_embd // hf_config['n_heads'])
-        config['mpt']['inter_size'] = str(n_embd * hf_config['expansion_ratio'])
-        config['mpt']['max_pos_seq_len'] = str(hf_config['max_seq_len'])
-        config['mpt']['num_layer'] = str(hf_config['n_layers'])
-        config['mpt']['vocab_size'] = str(hf_config['vocab_size'])
-        config['mpt']['start_id'] = str(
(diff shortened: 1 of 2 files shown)


dskhudia:
@dwyatte : I haven't looked into triton backend yet (@nik-mosaic is looking into it). Is it possible for you to create a triton config for mpt so that it looks for mpt instead?

dskhudia:
@nik-mosaic any comments on this?

dwyatte:
> Is it possible for you to create a triton config for mpt so that it looks for mpt key instead?

@dskhudia, from looking at the fastertransformer code, I think loading based on the presence of the `"gpt"` key is controlled in the Fastertransformer Triton backend [here](https://github.com/NVIDIA/FasterTransformer/blob/652e61e/src/fastertransformer/triton_backend/multi_gpu_gpt/ParallelGptTritonModel.cc)

I'm not sure if they would be open to a change since it would be backward incompatible with previous config files

If you were planning to contribute an MPT-specific Fastertransformer implementation to support its specific divergences from GPT, I think probably an `mpt` key would be appropriate there. But given that it relies on the ParallelGPT model, it would be better to comply with the config file assumptions IMO

If you are looking to simply record that this is an MPT model and not GPT, I think you can insert extra keys in the config that remain unused

dskhudia:
@dwyatte : Alright. I agree with you. I was mainly looking for has_positional_encoding and use_attention_linear_bias and it seems both of these will be handled [here](https://github.com/NVIDIA/FasterTransformer/blob/652e61e/src/fastertransformer/triton_backend/multi_gpu_gpt/ParallelGptTritonModel.cc#L233-L236). Also with this change, we won't have to copy and modify the triton file and it will work as is.

dskhudia (approved):

dskhudia:
@dwyatte Thank you so much for this change and for testing the triton path.

dwyatte:
@nik-mosaic

Since `run_mpt_with_ft.py` is using `ParallelGPT`, [this](https://github.com/triton-inference-server/fastertransformer_backend/blob/a1d8ab2/all_models/gpt/fastertransformer/config.pbtxt) should be the relevant config.pbtxt and is what I am using (I actually parse this and adjust various parameters like parallelism and checkpoint path before writing it out, but otherwise it's the same basic file)

The main important part is 

```
parameters {
  key: "model_type"
  value: {
    string_value: "GPT"
  }
}
```

to tell the backend what type of model to try to load the weights into, but some of the input/output names/datatypes end up being model-specific too

PR #275, opened 2023-06-02
Remote JSONL IFT data

samhavens:
Support remote jsonl files for finetuning.

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -183,10 +183,17 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
                     cfg.dataset['hf_kwargs'] = kwargs
                     print(cfg.dataset)
                     dataset = dataset_constructor.build_from_hf(
-                        cfg.dataset, tokenizer=tokenizer)
+                        cfg.dataset,
+                        max_seq_len=cfg.dataset.max_seq_len,
+                        tokenizer=tokenizer,
+                    )
                     break
         else:
-            dataset = dataset_constructor.build_from_hf(cfg.dataset, tokenizer=tokenizer)
+            dataset = dataset_constructor.build_from_hf(
+                cfg.dataset,
+                max_seq_len=cfg.dataset.max_seq_len,
+                tokenizer=tokenizer,
+            )
 
         collate_fn, dataloader_batch_size = _build_collate_fn(
             cfg.dataset, tokenizer, device_batch_size)
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -2,10 +2,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import logging
+import tempfile
 from typing import Union
 
 import torch
-from composer.utils import dist
+from composer.utils import dist, get_file
 from omegaconf import DictConfig
 from torch.utils.data import DataLoader
 from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast
@@ -38,7 +39,9 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
             ---
             *** HuggingFace dataset config fields ***
             cfg.dataset.hf_name (str, optional): The name of the HuggingFace dataset
-                to use.
+                to use. Can also be a remote JSONL file in the format
+                (prompt, response) in a supported object store or http(s) URL,
+                in which case the builder will create a streaming dataset.
             cfg.dataset.hf_kwargs (DictConfig, optional): Additional kwargs to
                 pass to `datasets.load_dataset`, which can be used to load
                 a dataset from local files.
@@ -145,6 +148,14 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
         )
 
     else:
+        if '://' in cfg.dataset.hf_name:
+            with tempfile.TemporaryDirectory() as tmp_dir:
+                name = f'{cfg.dataset.hf_name}/{cfg.dataset.split}.jsonl'
+                location = f'{tmp_dir}/{name}'
+                get_file(cfg.dataset.hf_name, location)
+                cfg.dataset.hf_name = 'json'
+                cfg.dataset.hf_kwargs['data_dir'] = tmp_dir
+                cfg.dataset.hf_kwargs['split'] = cfg.dataset.split
         dataset = dataset_constructor.build_from_hf(cfg.dataset, tokenizer)
 
         collate_fn, dataloader_batch_size = _build_collate_fn(
@@ -192,6 +203,12 @@ def _validate_config(dataset_cfg: DictConfig):
                 'Those keys are used when building from a streaming dataset, but ' +\
                 'setting `hf_name` instructs the dataset to build from a HuggingFace dataset.'
             )
+        if '://' in dataset_cfg.hf_name.get('hf_name'):
+            if dataset_cfg.get('split') is None:
+                raise ValueError(
+                    'When using a HuggingFace dataset from a URL, you must set the ' +\
+                    '`split` key in the dataset config.'
+                )
     elif dataset_cfg.get('remote') is not None:
         # Using the streaming dataset codepath
         illegal_keys = ['hf_name', 'hf_kwargs', 'preprocessing_fn']
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -159,13 +159,12 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
                 supported_extensions = ['jsonl', 'csv', 'parquet']
                 for extension in supported_extensions:
                     name = f'{cfg.dataset.hf_name}/{cfg.dataset.split}.{extension}'
-                    destination = f'{tmp_dir}/{name}'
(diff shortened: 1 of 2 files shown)


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -192,6 +203,12 @@ def _validate_config(dataset_cfg: DictConfig):
                 'Those keys are used when building from a streaming dataset, but ' +\
                 'setting `hf_name` instructs the dataset to build from a HuggingFace dataset.'
             )
+        if '://' in dataset_cfg.hf_name.get('hf_name'):
+            if dataset_cfg.get('split') is None:
```
alextrott16, line 214: Recommend moving this error check into the start of the first `if '://' in cfg.dataset.hf_name:` block.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -145,6 +148,14 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
         )
 
     else:
+        if '://' in cfg.dataset.hf_name:
```
alextrott16, line 151: Worth getting a check from @eracah on this pattern. I think Composer already has utilities for checking whether a path points to a remote object store supported by `get_file`.

dakinggg, line 151: recommend using `parse_uri` (also from `from composer.utils import parse_uri`)

samhavens: but this could also be valid with a value like `sam-mosaic/dolly-hhrlhf`

llmfoundry/data/finetuning/dataloader.py
alextrott16: Oops, ignore this! Github's presentation of the code had me totally backwards on where this was getting added. This looks correct to me!

llmfoundry/data/finetuning/dataloader.py
dakinggg: i don't follow, what could also be valid? It looks like this code path is only supposed to execute when the `cfg.dataset.hf_name` is a `{s3|gcp|oci}://` uri path.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -145,6 +148,21 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+                    name = f'{cfg.dataset.hf_name}/{cfg.dataset.split}.{extension}'
+                    location = f'{tmp_dir}/{name}'
+                    try:
+                        get_file(cfg.dataset.hf_name, location)
+                    except FileNotFoundError as e:
+                        if extension == supported_extensions[-1]:
+                            raise e
+                        continue
+                    cfg.dataset.hf_name = extension
+                    cfg.dataset.hf_kwargs['data_dir'] = tmp_dir
```
dakinggg, line 169: this `tmp-dir` will be destroyed outside of the context manager

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -145,6 +148,21 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+                    cfg.dataset.hf_kwargs['split'] = cfg.dataset.split
```
alextrott16, line 170: When we call `dataset = dataset_constructor.build_from_hf(cfg.dataset, tokenizer)` the `build_from_hf` method will set the `split` kwarg based on `cfg.dataset.split`, so I think you do not need this line.

llmfoundry/data/finetuning/dataloader.py
samhavens: Are you saying something like:

```python
if '://' in cfg.dataset.hf_name:
  backend, bucket_name, path = parse_uri(cfg.dataset.hf_name)
```

or are you saying that there is a way to use `parse_uri` to not do the `'://' in` check?

samhavens: oh, maybe

```python
backend, bucket_name, path = parse_uri(cfg.dataset.hf_name)
if backend != '':
```
?

llmfoundry/data/finetuning/dataloader.py
samhavens: Is it then better to make a non-temporary directory, or have the loading stay happen in the context manager and duplicate that line in an `else` block?

alextrott16: The second option should work fine. I think the only reason you would not do it is that it is *mildly* harder to read. But perfectly acceptable IMHO.

llmfoundry/data/finetuning/dataloader.py
dakinggg: yeah, the second thing you said. except double check if its backend `None` or empty string

samhavens:
When I run this I am seeing the s3 download fail at 29% with

```sh
Downloading ift/jsonl_test:   0%|          | 0.00/906k [00:00<?, ?iB/s]
Downloading ift/jsonl_test:  29%|██▉       | 262k/906k [00:00<00:00, 3.01MiB/s]

Downloading ift/jsonl_test:   0%|          | 0.00/906k [00:00<?, ?iB/s]Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/composer/utils/object_store/s3_object_store.py", line 120, in get_object_size
    obj = self.client.get_object(Bucket=self.bucket, Key=self.get_key(object_name))
  File "/usr/lib/python3/dist-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/lib/python3/dist-packages/botocore/client.py", line 964, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.

The above exception was the direct cause of the following exception:

  File "/usr/lib/python3/dist-packages/composer/utils/file_helpers.py", line 504, in get_file
    return get_file(
  File "/usr/lib/python3/dist-packages/composer/utils/file_helpers.py", line 471, in get_file
    _get_file(
  File "/usr/lib/python3/dist-packages/composer/utils/file_helpers.py", line 526, in _get_file
    total_size_in_bytes = object_store.get_object_size(path)
  File "/usr/lib/python3/dist-packages/composer/utils/object_store/s3_object_store.py", line 122, in get_object_size
    _ensure_not_found_errors_are_wrapped(self.get_uri(object_name), e)
  File "/usr/lib/python3/dist-packages/composer/utils/object_store/s3_object_store.py", line 26, in _ensure_not_found_errors_are_wrapped
    raise FileNotFoundError(f'Object {uri} not found') from e
FileNotFoundError: Object s3://mosaicml-internal-checkpoints-shared/ift/jsonl_test.symlink not found
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -145,7 +148,31 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+                    'When using a HuggingFace dataset from a URL, you must set the ' +\
+                    '`split` key in the dataset config.'
+                )
+            with tempfile.TemporaryDirectory() as tmp_dir:
+                supported_extensions = ['jsonl', 'csv', 'parquet']
+                for extension in supported_extensions:
+                    name = f'{cfg.dataset.hf_name}/{cfg.dataset.split}.{extension}'
+                    location = f'{tmp_dir}/{name}'
+                    try:
+                        get_file(cfg.dataset.hf_name, location)
```
dakinggg, line 164: you probably want to only do the downloading on rank 0 (and make everyone else wait for the download to be finished

samhavens: gee, if only there were a nice utility for that

samhavens: @dakinggg sorry, fixed

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -145,7 +148,34 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+                    name = f'{cfg.dataset.hf_name}/{cfg.dataset.split}.{extension}'
+                    destination = f'{tmp_dir}/{name}'
+                    try:
+                        with dist.local_rank_zero_download_and_wait(
+                                destination):
+                            if dist.get_local_rank() == 0:
+                                get_file(name, destination, overwrite=True)
+                    except FileNotFoundError as e:
+                        if extension == supported_extensions[-1]:
+                            raise e
```
dakinggg, line 170: can we raise a nicer error here? otherwise the last filenotfound will be about a parquet path, which might be confusing to a user if they were (e.g.) trying to use a jsonl.

dakinggg (approved):
LGTM

alextrott16 (approved):
LGTM. Seems fine to merge once you're confident it works as expected when testing.

abhi-mosaic:
I think all that's left here @vchiley is to do a quick test with a remote JSON file (e.g. upload the sample data to S3 and try pointing at it). And then we add a sentence in the finetuning instructions clarifying that "Yes, you can point `cfg.dataset.hf_name` at a remote URI"

vchiley:
https://wandb.ai/mosaic-ml/domain-adapt-remote-jsonl-ift/runs/jox5h54l shows training with this functionality
Screenshot 2023-06-21 at 10 13 03 PM
seems legit

alextrott16 (approved):
I like.

PR #276, opened 2023-06-02
Add `device_map` support for `hf_generate.py` and `hf_chat.py`

abhi-mosaic:
This PR enables `--device_map auto` which enables using these scripts with very large models that don't fit on a single GPU.

This also removes the need for FSDP support @alextrott16 !

### DIFF
--- a/scripts/inference/README.md
+++ b/scripts/inference/README.md
@@ -101,7 +101,10 @@ output_tok_per_sec=292.03tok/sec
 
 The argument for `--name_or_path` can be either the name of a model that exists on the HF Hub, such as `gpt2`, `facebook/opt-350m`, etc. or the path to a HF checkpoint folder, such as `my_hf_model/` like we exported above.
 
-For MPT models specifically, you can pass args like `--init_device cuda:0`, `--attn_impl triton`, and `--max_seq_len 4096` to speed up generation time or alter the max generation length at inference time (thanks to ALiBi).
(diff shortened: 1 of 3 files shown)


dskhudia:
How is it parallelizing the model?

abhi-mosaic:
> How is it parallelizing the model?

@dskhudia the HF `device_map` feature is very simple, it just stores the model weights on different devices (and it respects a model attribute called `_no_split_modules` that the model author writes) and then it move activations from one GPU to the next during forward. Basically pipelining with one microbatch. So there is no speedup, but you have more GPU memory now to host the model weights.

abhi-mosaic:
I tried refactoring the defaults so that it defaults to `device_map=auto` (which leads to much faster CPU init than leaving it out). I think this is better UX but this requires that we have `accelerate` installed. 

To satisfy this without users running into an error and having to `pip install acclerate` every time they use our scripts, I moved `accelerate` to be a LLM Foundry dependency, which I don't like... the alternative would be to separate the `inference/` folder deps into something like `pip install .[inference]` which would include `accelerate`, `onnx`, `onnxruntime`, etc. 

WDYT about the two options @alextrott16 ?

alextrott16 (approved):
This looks great! Thanks for adding this feature! This is much preferable compared to using FSDP, since the latter is the wrong tool for the job.

Approving because the overall structure looks good and I don't spot any obvious bugs. I trust that you will confirm that it works as expected.

samhavens:
Just chiming in that I am already using this branch for testing the chat model!

PR #271, opened 2023-06-02
adding te Linear for fp8 support

vchiley:
ran `composer train/train.py train/yamls/pretrain/mpt-3b.yaml` also with `model.fc_type=te` and `precision=amp_fp8`
Result:
```
torch: throughput/device/tokens_per_sec: 23.7k
te: throughput/device/tokens_per_sec: 23.7k
te with fp8: throughput/device/tokens_per_sec: 29.4k
```

Note there does seem to be this error when activation ckpt is enabled when `activation_checkpointing_reentrant: false`. If we set `activation_checkpointing_reentrant: true`, then act ckpt works fine without `amp_fp8`; with `amp_fp8` the issue still persists.
(previously, circa summer 2022, `activation_checkpointing_reentrant: true` resulted in some difficulties which is why we set it to `false`; not sure if this is necessary still...)
ActCkpt error will be added to the comments.
(the error with `amp_fp8` might be a composer impl of fp8 issue)

PR #279, opened 2023-06-04
Add 8-bit LION optimizer

dblalock:
Adds an 8-bit version of the LION optimizer. Some non-obvious aspects of this include:

- CUDA kernels for int8 quantizing and dequantizing floats. Kernels use numba since I got stonewalled by Triton bugs.
- A fused CUDA kernel for the LION update
- We only quantize tensors with 1024 elements or more for simplicity (and since small tensors don't take much space or time anyway)
- We also quantize the quantization scales. So the quantized repr of a length N tensor is:
    - N int8 values
    - N/16 int8 scales
    - N/1024 fp32 scale scales
- We use a scaling algorithm I haven't seen before where we store the maximum for each row and column. I won't explain it here but basically it means you need 2 outliers instead of one to ruin the scaling for other values. 
- We preprocess everything via signed square root before quantizing. I also haven't this seen before but it makes it super hard to get screwed by overflow and underflow, and reduces quantization error in my offline experiments.

Code changes:
- Adds `numba` to the GPU dependencies in setup.py
- Adds `lion8b.py` and `_quantize_kernels.py` to `llm-foundry/optim`
- Adds `Lion8bit` to `llm-foundry/optim/__init__.py`
- Adds `lion8b` as an option in `llm-foundry/optim/builders.py`
- Adds `test_lion8b.py` to the tests. I'd like to test the kernels directly as well, but this is effectively an integration test for all that logic.
- Changes the pre-commit config to allow use of `dict()` with kwargs; not sure why this was disallowed

### DIFF
--- a/.github/mcp/mcp_pytest.py
+++ b/.github/mcp/mcp_pytest.py
@@ -1,9 +1,6 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
-# Copyright 2022 MosaicML Composer authors
-# SPDX-License-Identifier: Apache-2.0
-
 """Run pytest using MCP."""
 
 import argparse
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,54 @@
+name: Docker
+on:
+  push:
+    branches:
+    - main
+  workflow_dispatch: {}
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'mosaicml'
+    strategy:
+      matrix:
+        include:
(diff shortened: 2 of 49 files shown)


vchiley:
to clarify, this cannot be used on CPUs (no that anyone wants to train on CPUs, but just want to verify)

dblalock:
It will never actually quantize on CPUs. It should still run on CPUs with non-quantized states.

vchiley:
Added lion8b to [this](https://github.com/mosaicml/llm-foundry/pull/271#issuecomment-1595337924)

Screenshot 2023-06-16 at 4 46 54 PM
Screenshot 2023-06-16 at 4 47 11 PM

dblalock:
Replaced by #514

PR #292, opened 2023-06-06
Upgrade to `mosaicml-streaming==0.5.x`

abhi-mosaic:
- updates streaming version
- enables arg checking via passing in args as **kwargs


llmfoundry/data/text_data.py
```diff
@@ -238,7 +235,7 @@ def build_text_dataloader(
                     cfg.dataset.get('split', None),
                     proportion=stream.get('proportion', None),
                     repeat=stream.get('repeat', None),
-                    samples=stream.get('samples', None),
+                    choose=stream.get('samples', None),
```
vchiley, line 238:
```suggestion
                    choose=stream.get('choose', None),
```

vchiley (approved):
Would it make sense to run 125M with v0.4 and v0.5 to verify training looks the same?

Approving to unblock. Generally lgtm

abhi-mosaic:
Yup will run some convergence tests before merging

llmfoundry/data/text_data.py
```diff
@@ -238,7 +235,7 @@ def build_text_dataloader(
                     cfg.dataset.get('split', None),
                     proportion=stream.get('proportion', None),
                     repeat=stream.get('repeat', None),
-                    samples=stream.get('samples', None),
```
dakinggg, line 241: given the way we are accessing these args with `.get`, can we add some simple validation based on streaming version and the `stream` cfg dict? i know we have a full validation jira planned, but i think this one is likely to bite people.

abhi-mosaic: Yeah I can add some kind of validation, these `.get`s worry me too

abhi-mosaic:
Would recommend waiting on this merge until we get a hotfix release 0.5.1

llmfoundry/data/text_data.py
vchiley: Would [**kwargs handling kwarg checking](https://github.com/mosaicml/llm-foundry/pull/292/commits/bd57a87a5ade3bc4e83fac02316ecf506040e540) do the trick?

vchiley:
With baseline usage, main and this branch have similar performance
Screenshot 2023-06-20 at 4 06 34 PM
MFU is also effectively exactly the same

the only diff is that the update uses slightly less sys mem
Screenshot 2023-06-20 at 4 08 28 PM

PR #294, opened 2023-06-06
Update README.md

abhi-mosaic:
Update the numbers in the "How many GPUs do I need..." paragraph. These old numbers were for AdamW, but the writing and math references LionW.


growlix:
One table that would be nice to add: minimum # GPUs needed to run a given model architecture/seq len/precision format/other relevant hparams, for common GPUs (A/H100 40gb/80gb)

abhi-mosaic:
> One table that would be nice to add: minimum # GPUs needed to run a given model architecture/seq len/precision format/other relevant hparams, for common GPUs (A/H100 40gb/80gb)

I'll add this to our JIRA todo for both train and inference

PR #296, opened 2023-06-07
Sam/chat v2

samhavens:
This fixes a bug in `hf_chat.py` where the custom system prompt and user and assistant format strings were ignored

It also cleans up the implementation and does streaming generation

### DIFF
--- a/scripts/inference/hf_chat.py
+++ b/scripts/inference/hf_chat.py
@@ -5,7 +5,7 @@ import time
 import warnings
 from argparse import ArgumentParser, ArgumentTypeError, Namespace
 from contextlib import nullcontext
-from typing import Any, Dict, Union
+from typing import Any, Dict, List, Union
 
 import torch
 from transformers import (AutoConfig, AutoModelForCausalLM, AutoTokenizer,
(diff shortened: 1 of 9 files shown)


scripts/inference/hf_chat.py
```diff
@@ -5,16 +5,126 @@
...
+                          PreTrainedTokenizer, PreTrainedTokenizerFast,
+                          StoppingCriteria, StoppingCriteriaList, TextStreamer)
 
 Tokenizer = Union[PreTrainedTokenizer, PreTrainedTokenizerFast]
 
 
-def get_dtype(dtype):
+class ChatFormatter:
+
+    def __init__(self, system: str, user: str, assistant: str) -> None:
```
dakinggg, line 20: could you add a simple docstring here?

scripts/inference/hf_chat.py
```diff
@@ -5,16 +5,126 @@
...
+        self.system = system if system else '<|im_start|>system\nA conversation between a user and an LLM-based AI assistant. The assistant gives helpful and honest answers.<|im_end|>\n'
+        self.user = user if user else '<|im_start|>user\n{}<|im_end|>\n'
+        self.assistant = assistant if assistant else '<|im_start|>assistant\n{}<|im_end|>\n'
+        self.response_prefix = '<|im_start|>assistant\n'
+
+
+class Conversation:
+
+    def __init__(self, model, tokenizer: Tokenizer, chat_format: ChatFormatter,
```
dakinggg, line 29: same, simple docstring would be great

scripts/inference/hf_chat.py
```diff
@@ -5,16 +5,126 @@
...
+                 generate_kwargs: Dict[str, Any]) -> None:
+        self.model = model
+        self.tokenizer = tokenizer
+        self.chat_format = chat_format
+
+        stop_token_ids = self.tokenizer.convert_tokens_to_ids(
```
dakinggg, line 35: can you add a check here that those tokens convert to a single token id? i think the rest of this code assumes that (although please correct me if im wrong)

dakinggg (approved):
Didn't review super carefully. I assume it works

PR #297, opened 2023-06-07
Set eval shuffle to False

eldarkurtic:
I think eval dataloaders should not shuffle data for consistent evaluation results


abhi-mosaic (approved):
Good catch, thank you!

I did a global ctrl-F and I think this is the only place we have this bug.

PR #298, opened 2023-06-07
Adding Custom Embedding, Enabling us to initialize on Heterogeneous Devices

bcui19:
The below code adds a SharedEmbedding class that let's us get rid of a `F.linear` call. This is necessary with certain wrapping structures (our HF ones), otherwise FSDP emits a strange error.

### DIFF
--- a/llmfoundry/models/hf/hf_fsdp.py
+++ b/llmfoundry/models/hf/hf_fsdp.py
@@ -152,16 +152,6 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel) -> None:
             child._fsdp_wrap = True
     
     block_type = type(model_block[0])  # type: ignore
-    # When using the HF LM models,
-    # the weights of the self.lm_head and self.transformer.wte are tied.
-    # This tying occurs inside the `self.post_init()` function.
-    # This is a hurdle for FSDP because they need to be in the same FSDP block
-    # These lines ensures that both modules stay together in the top-most block when
-    # the model has this tying enabled (almost all do; this property defaults to True)
-    if model.config.tie_word_embeddings:
-        causal_base_model._fsdp_wrap = False  # type: ignore
-        tied_embeddings._fsdp_wrap = False  # type: ignore
-        lm_head._fsdp_wrap = False  # type: ignore
 
     # FSDP Wrap and Activation Checkpoint every model block
     model.fsdp_wrap_fn = lambda module: isinstance(module, block_type)
--- a/llmfoundry/models/hf/hf_fsdp.py
+++ b/llmfoundry/models/hf/hf_fsdp.py
@@ -136,6 +136,21 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel) -> None:
             raise ValueError(
                 f'Unable to FSDP-wrap this model! `{mod_name}` does not ' +
                 'follow common layer/weight naming conventions.')
+    
+    for child in model.children():
+        if isinstance(child, type(causal_base_model)):
+            continue
+        if isinstance(child, torch.nn.Module):
+            child._fsdp_wrap = True
+    
+    # For FSDP with models with different intiailizations, we need to tag all child modules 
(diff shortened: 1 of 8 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -79,13 +80,13 @@ def __init__(self, config: MPTConfig):
...
 
-        self.wte = nn.Embedding(config.vocab_size,
+        self.wte = SharedEmbedding(config.vocab_size,
                                 config.d_model,
                                 device=config.init_device)
         if not self.alibi:
-            self.wpe = nn.Embedding(config.max_seq_len,
-                                    config.d_model,
-                                    device=config.init_device)
+            self.wpe = SharedEmbedding(config.max_seq_len,
```
alextrott16, line 87: I don't think there's any need for `wpe` to be a `SharedEmbedding`. It shouldn't really matter but I think it helps clarify the role of `SharedEmbedding` to limit it to the embedding that is actually shared.

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -135,17 +136,22 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel) -> None:
...
-    # When using the HF LM models,
-    # the weights of the self.lm_head and self.transformer.wte are tied.
-    # This tying occurs inside the `self.post_init()` function.
-    # This is a hurdle for FSDP because they need to be in the same FSDP block
-    # These lines ensures that both modules stay together in the top-most block when
-    # the model has this tying enabled (almost all do; this property defaults to True)
-    if model.config.tie_word_embeddings:
-        causal_base_model._fsdp_wrap = False  # type: ignore
-        tied_embeddings._fsdp_wrap = False  # type: ignore
-        lm_head._fsdp_wrap = False  # type: ignore
```
alextrott16, line 148: I think getting rid of this entirely might be a bit heavy handed, especially if it breaks tied embeddings throughout the stack.

Instead, I recommend that we keep this logic for this PR and create a separate PR where we add in logic for mixed CPU/meta initialization. I feel like that separate PR would be the place where we would want to change around this `if` block to throw an informative error in the case that `tie_word_embeddings==True` and `tied_embeddings is not lm_head` and the user is attempting to do the mixed initialization.

alextrott16 (commented):
A couple comments...

1. I think we need to make sure we feel solid about the modifications to `hf_fsdp.py` in this PR. See my comment there. Also, do we need to change the code for encoder-decoder in that file?

2. I _think_ that changing `nn.Embedding` to `SharedEmbedding` will not require us to change the `convert_examples_ckpt.py` conversion script (because the parameter names in the checkpoint will be preserved, I think). Does that sound right to you? If no, we'll need to update the conversion script too.

llmfoundry/models/hf/hf_fsdp.py
bcui19: Yeah sounds good. I'll make a separate PR. I think we'd need the user to specify the mixed initialization flag, since otherwise we have no other way to know how the other models are initialized on other processes... Wdyt

bcui19:
@alextrott16 1. Yeah let me create a separate PR for the HF changes. We'd need to propagate it to also enc-dec. Also I think we need to pass in a flag for the mixed initializations? see here: https://github.com/mosaicml/llm-foundry/pull/298#discussion_r1223384528

2. It should be fine because the module name doesn't change e.g. it's still `self.wte`

llmfoundry/models/hf/hf_fsdp.py
alextrott16: Yes, totally agree. I think the simplest thing would be to allow the HF model wrappers to use `cfg.init_device = "mixed"` (or whatever we want to call it). In a perfect world, this would be the default "cpu" initialization works when using FSDP and a compatible model, but detecting the latter condition might not be trivial.
Happy to discuss whether this is possible offline.

alextrott16 (approved):
My comments were addressed, and I'm on board with these MPT changes. Approving to unblock.

PR #303, opened 2023-06-08
Huggingface Mixed Initialization

bcui19:
Adding mixed initialization to support HF models. This allows rank 0 to load on `cpu` and all other ranks to load onto `meta`.

Tested here:

https://wandb.ai/mosaic-ml/hf-init-devices/runs/bqothkss?workspace=user-bcui

Also tested the difference when initializing w/ `cpu` and with `mixed` here (pythia-410m):

https://wandb.ai/mosaic-ml/hf-init-devices?workspace=user-bcui

### DIFF
--- a/llmfoundry/models/hf/hf_causal_lm.py
+++ b/llmfoundry/models/hf/hf_causal_lm.py
@@ -15,6 +15,7 @@ from omegaconf import DictConfig
 from transformers import (AutoConfig, AutoModelForCausalLM, PreTrainedTokenizer,
                           PreTrainedTokenizerFast)
 
+from llmfoundry.models.hf.hf_fsdp import hf_get_init_device
 from llmfoundry.models.hf.model_wrapper import HuggingFaceModelWithZLoss
 from llmfoundry.models.utils import init_empty_weights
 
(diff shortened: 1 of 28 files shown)


llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -136,16 +148,39 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel) -> None:
...
+                continue
+            if isinstance(child, torch.nn.Module):
+                child._fsdp_wrap = True
+
+        if model.config.tie_word_embeddings:
+            warnings.warn((
+                'The passed in HuggingFaceModel has tied word embeddings '
+                'and the passed in initialization device is `mixed.` '
+                'In order to support this initializaiton schema, we are going to '
+                'break weight tying.'))
```
alextrott16, line 173: I strongly think this should be an error. Our warnings are way too easy to miss and IMO someone should have to explicitly set the config overrides to have `tie_word_embeddings: False`.
If it is made to be an error, we might also want to expand the check a bit because MPT has `tie_word_embeddings: True` but the tie won't break because `tied_embeddings == lm_head`. I _think_ that as long as those point to the same module they will, by definition, be on the same FSDP block and the weight tying won't be (perhaps even couldn't be?) broken.

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -97,7 +102,7 @@ def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
                     config,
                     trust_remote_code=trust_remote_code,
                 )
-        elif init_device == 'meta':
+        elif resolved_init_device == 'meta':
             if om_model_config.pretrained:
```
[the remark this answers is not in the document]
bcui19: I added code to hot swap out pretrained when the `resolved_init_device = 'meta'` to always be False for that rank.

llmfoundry/models/hf/hf_fsdp.py
bcui19: I agree, changed to an error...

PR #309, opened 2023-06-09
Make convert_mpt_to_ft importable when this repo is installed

dskhudia:
same as title.


dakinggg (commented):
Can you make it so the script itself still lives in `scripts`, but all that script does is import the helper function and call it?

dskhudia:
@dakinggg : Actually abandoning this PR for now. Installing llm-foundry will overwrite our custom PyTorch build in inference images.

PR #310, opened 2023-06-10
Fix bug with saving optimizer states with MonolithicCheckpointSaver Callback

eracah:
Currently there is a bug where only the model state_dict_type is 'full' and not the optimizer state state dict type.
This PR fixes that using the `fsdp_get_optim_state_dict` helper function

### DIFF
--- a/llmfoundry/callbacks/monolithic_ckpt_callback.py
+++ b/llmfoundry/callbacks/monolithic_ckpt_callback.py
@@ -8,7 +8,8 @@ from pathlib import Path
 
 import torch
 from composer.core import Callback, State
-from composer.core.state import fsdp_state_dict_type_context, fsdp_get_optim_state_dict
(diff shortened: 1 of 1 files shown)


dakinggg (approved):
LGTM assuming you've tested

eracah:
I ran a manual test:
I saved with mono checkpoint saver in job mono-ckpt-test-8-NZ2JJ0 to create `s3://mosaicml-internal-checkpoints-test/evan-test/mono-ckpt-test/mono-ckpt-test-8-NZ2JJ0/mono_ba2.pt` and then
I loaded `s3://mosaicml-internal-checkpoints-test/evan-test/mono-ckpt-test/mono-ckpt-test-8-NZ2JJ0/mono_ba2.pt` in
job `mono-ckpt-test-8-Bdi3np`

Seemed to work well. Resumed seemlessly starting from batch 3

PR #320, opened 2023-06-13
Better error messages for ckpt conversion script

dskhudia:
This script expects a full composer checkpoint. Make it clear via error messages. Also change main function name as I people are using it as an api in their flow.


abhi-mosaic (approved):
LGTM!

PR #322, opened 2023-06-14
Fix generate callback to work with precision context

dakinggg:
still works: https://wandb.ai/mosaic-ml/daniel-debug/runs/ycsym27d?workspace=user-danielking


abhi-mosaic (approved):
LGTM

PR #323, opened 2023-06-14
Allow MPT past the tied word embeddings error

dakinggg:
```
In [3]: cfg = transformers.AutoConfig.from_pretrained('mosaicml/mpt-7b', trust_remote_code=True)
Explicitly passing a `revision` is encouraged when loading a configuration with custom code to ensure no malicious code has been contributed in a newer revision.

In [4]: cfg.model_type
Out[4]: 'mpt'
```


bcui19 (approved):
LGTM!

PR #324, opened 2023-06-14
Remove demo column

samhavens:
While we sort out exactly which models will have demos and how they will be hosted, lets remove the link.


vchiley (approved):
seems legit

samhavens:
Talked to Abhi... we will get the 7b demo up and just keep this

PR #329, opened 2023-06-14
hard set bias(alibi) precision

vchiley:
Hard set bias(alibi) precision so that `FSDP: mixed_precision: PURE | FULL` produce same eval scores.

We can see that when running eval from main with `PURE`, eval perf degredes.
If we run it with this fix, we improve and are within noise of eval with `FULL`
Screenshot 2023-06-14 at 10 38 34 AM


dakinggg (approved):
Please make the change to the hub models too :D

PR #334, opened 2023-06-15
Load on rank 0 only flag

mvpatel2000:
Adds new config flag. Needs new Composer release


eracah (approved):
sure

mvpatel2000:
Holding until Composer release

mvpatel2000:
Actually, this doesn't require a pin, so I can merge. It'll just no-op in config

PR #338, opened 2023-06-16
Add script to update hub code from foundry

dakinggg:
Adds a script to open PRs against huggingface hub models to keep in sync with foundry code.

TODO:
- [x] add a "test uploaded model" arg
- [x] run and open the prs

Opened some prs with this script:
https://huggingface.co/mosaicml/mpt-7b/discussions/63
https://huggingface.co/mosaicml/mpt-7b-instruct/discussions/51
https://huggingface.co/mosaicml/mpt-7b-chat/discussions/29
https://huggingface.co/mosaicml/mpt-7b-storywriter/discussions/47

### DIFF
--- a/scripts/misc/update_hub_code.py
+++ b/scripts/misc/update_hub_code.py
@@ -1,0 +1,54 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import os
+import tempfile
+from typing import List
+
+import transformers
+
+from llmfoundry import MPTConfig, MPTForCausalLM
+from llmfoundry.utils.huggingface_hub_utils import \
+    edit_files_for_hf_compatibility
+
+
+def main(hf_repos_for_upload: List[str]):
+    from huggingface_hub import HfApi
+    api = HfApi()
+
+    # register config auto class
+    MPTConfig.register_for_auto_class()
+
+    # register model auto class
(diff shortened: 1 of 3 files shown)


scripts/misc/update_hub_code.py
```diff
@@ -0,0 +1,75 @@
...
+            _ = pr_tokenizer.batch_decode(generation)
+
+            print(f'Opening PR against {repo}')
+            result = api.upload_folder(
+                folder_path=original_save_dir,
+                repo_id=repo,
+                use_auth_token=True,
+                repo_type='model',
+                allow_patterns=['*.py'],
+                commit_message='LLM-foundry update',
```
abhi-mosaic, line 56: Could we add a datetime here?

abhi-mosaic (approved):
This LGTM, I am kinda confused by the usage of `torch.nn.Embedding` rather than `nn.Embedding` in the HF Hub PR https://huggingface.co/mosaicml/mpt-7b/discussions/61/files but it's not breaking anything. If there's an easy way to fix that in this PR and push it up too, that would be cool. But not gating.

dakinggg:
our main code uses `torch.nn.Embedding`, at some point someone updated the hub code to be `nn.Embedding`, this is just bringing them back in sync. doesn't really matter which one so ill just leave it as `torch.nn.Embedding`

scripts/misc/update_hub_code.py
dakinggg: added

PR #346, opened 2023-06-20
Feature/peft compatible models

danbider:
Edits needed to support a combo of composer with hf/peft.

Pipeline is: 
1. load a hf model e.g., mpt-7b
2. use hf/peft to add lora modules or adapter modules.
3. convert that peft model (that is loaded into python) into a composer model (use my new function for this)
4. train in composer (required adding the `inputs_embeds` args to `model.forward()`.


llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -148,3 +152,36 @@ def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
...
+
+
+class ComposerHFCausalLMFromPython(HuggingFaceModelWithZLoss):
+    """Configures a :class:`.HuggingFaceModel` around a Causal LM that is loaded in memory.
+    Args:
+        model (peft.peft_model.PeftModel or transformers.PreTrainedModel): The HF model loaded into python memory.
+        tokenizer (PreTrainedTokenizer): The tokenizer that the model will use.
+    """
+
+    def __init__(self, model: Union[peft.peft_model.PeftModel,transformers.PreTrainedModel], tokenizer: Tokenizer):
```
dakinggg: since the only change here is how the base model is constructed, should we just change the arg to `ComposerHFCausalLM` to accept `Union[DictConfig, peft.peft_model.PeftModel, transformers.PreTrainedModel`?

danbider: @dakinggg the reason I decided against it is that I didn't want to contaminate the main func with flags that immediately returns a composer model, but happy to change.

dakinggg: My opinion is that's it's better to unify them. I think it will just be a simple `if instance(model, DictConfig): do all the extra stuff else: pass through the model`. I think we should refactor (not in this pr) to have a dedicated builder function that takes the config anyway, instead of this class accepting a config as a constructor arg.

danbider: ok, will make the modification now and test on my end.

danbider:
refactored the hf convertor to a single function as suggested by @dakinggg. tested it on my end and ran pre-commit successfully. I want to move forward and push the code updates to the hub.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -316,16 +320,21 @@ def forward(
                 'prefix_mask is a required argument when MPT is configured with prefix_lm=True.'
             )
 
+        # rase a not implemented error if input_embeds is not None (this ia an arg in huggingface transformers, we want to support it for peft)
```
samhavens, line 323:
```suggestion
# Raise a not implemented error if input_embeds is not None (this is an arg in huggingface transformers and we need to support it for PEFT)
```

samhavens:
Tests are failing with
```
___________________ ERROR collecting tests/test_training.py ____________________
ImportError while importing test module '/llm-foundry/tests/test_training.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
llmfoundry/__init__.py:8: in <module>
    from llmfoundry.data import (ConcatTokensDataset,
llmfoundry/data/__init__.py:5: in <module>
    from llmfoundry.data.denoising import (MixtureOfDenoisersCollator,
llmfoundry/data/denoising.py:20: in <module>
    from llmfoundry.models import utils
llmfoundry/models/__init__.py:4: in <module>
    from llmfoundry.models.hf import (ComposerHFCausalLM, ComposerHFPrefixLM,
llmfoundry/models/hf/__init__.py:4: in <module>
    from llmfoundry.models.hf.hf_causal_lm import (ComposerHFCausalLM,
llmfoundry/models/hf/hf_causal_lm.py:10: in <module>
    import peft
E   ModuleNotFoundError: No module named 'peft'
```

c9o:
Hello @danbider , could you share your yamls for MPT peft/lora training? Thanks.

llmfoundry/__init__.py
```diff
@@ -46,6 +46,7 @@
     'MPTForCausalLM',
     'ComposerMPTCausalLM',
     'ComposerHFCausalLM',
+    'ComposerHFCausalLMFromPython',
```
mvpatel2000, line 49: is this actually added anywhere?

mvpatel2000 (commented):
Mostly LGTM! Left a few comments. Will approve on next pass / w review from NLP person

setup.py
```diff
@@ -62,6 +62,10 @@
     'cmake>=3.25.0,<=3.26.3',  # required for triton-pre-mlir below
     # PyPI does not support direct dependencies, so we remove this line before uploading from PyPI
     'triton-pre-mlir@git+https://github.com/vchiley/triton.git@triton_pre_mlir_sm90#subdirectory=python',
+    'loralib',  # lora
+    'peft',  # lora
+    'bitsandbytes',  # 8bit
+    'scipy'
```
samhavens, line 68: what is `scipy` for? also, let's pin down the versions to a range or value

scripts/train/train.py
```diff
@@ -160,8 +162,28 @@ def main(cfg):
 
     # Build Model
     print('Initializing model...')
-    with init_context:
-        model = build_composer_model(cfg.model, tokenizer)
+    # if cfg.peft exists and is lora
```
samhavens, line 165: remove this comment, I think it's outdated?

scripts/train/train.py
```diff
@@ -160,8 +162,28 @@ def main(cfg):
...
+    if cfg.get('lora', None) is not None:
```
samhavens, line 166: I think this block should be in `build_composer_model`

I'm wondering if we can clean this block and the big `if` block in the init of `ComposerHFCausalLM` by having `build_composer_model` call a new `build_lora_model` function under the appropriate conditions?

llmfoundry/__init__.py
danbider: good catch; should be removed. this is an older implementation i had

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -87,64 +69,116 @@ def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
...
-            os.remove(signal_file_path)
+            z_loss = om_model_config.get('z_loss', 0.0)
+
+        # elif the model is either a PeftModel or a PreTrainedModel
+        elif isinstance(
+                om_model_config,
+            (peft.peft_model.PeftModel, transformers.PreTrainedModel)):
+            model = om_model_config
+            init_device = 'cpu'
+            z_loss = 0.0
```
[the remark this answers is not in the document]
danbider: i didn't think about this at all; I can try and see if `meta` works...

scripts/train/train.py
danbider: yeah sounds good!

setup.py
danbider: done; pinned down everything but peft which changes daily and we need their changes. 
`scipy` is a dep of `bitsandbytes` though it doesn't appear in their setup.py (raised an issue there).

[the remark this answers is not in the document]
danbider: sent to gpu

chris-aeviator:
@stoperro according to https://github.com/mosaicml/llm-foundry/pull/416 just use the ordinary peft code (huggingface has ready to go PEFT notebooks) or with llm-foundry add 
grafik

palash04:
Hey @chris-aeviator, I noticed that in the repository, LoRA currently only supports MPT models. Can we perform LoRA fine-tuning on other models such as LLAMA?

dakinggg:
@palash04 this is getting fixed in https://github.com/mosaicml/llm-foundry/pull/435

PR #347, opened 2023-06-21
updt composer to 0.15.0

vchiley:


dakinggg (approved):
LGTM, should we do a training run to verify no behavior change?

vchiley:
125M matches 125M main
3B matches 3B main

Screenshot 2023-06-21 at 9 23 51 AM

PR #348, opened 2023-06-21
Filter out sequences where prompt is longer than max length, rather than dropping them on the fly later

dakinggg:
For the HuggingFace code path for finetuning datasets, we prefilter out examples where the prompt is longer than max sequence length. This was previously causing the run to hang when using small dtms

before: `test-finetune-hang-before-9UXsdA` (hangs at batch 7)
after: `test-finetune-hang-after-o8hS44` (past batch 7, and emitted warning)


codestar12 (approved):
LGTM. I'm a little concerned there aren't tests for anything this is touching, but the changes are straight forward.

PR #354, opened 2023-06-22
Revert "Filter out sequences where prompt is longer than max length, rather than dropping them on the fly later"

codestar12:
Reverts mosaicml/llm-foundry#348


dakinggg (approved):
Approving to unblock

PR #359, opened 2023-06-22
Adds max seq len filter before finetuning ds

vchiley:
https://github.com/mosaicml/llm-foundry/pull/358


llmfoundry/data/finetuning/tasks.py
```diff
@@ -248,9 +253,20 @@ def dataset_mapper(example: Dict):
             dataset_mapper,
             batched=False,
             remove_columns=columns_to_remove,
+            num_proc=max(os.cpu_count() - 2, 1),
```
alextrott16, line 256:
```suggestion
```

llmfoundry/data/finetuning/tasks.py
```diff
@@ -248,9 +253,20 @@ def dataset_mapper(example: Dict):
...
         )
+        prompt_length_filtered_dataset = tokenized_dataset.filter(
+            lambda example: len(example['input_ids']) < max_seq_len,
+            num_proc=max(os.cpu_count() - 2, 1))
```
alextrott16, line 260:
```suggestion
        )
```

alextrott16 (approved):
This seems to be working for me now. I recommend removing the `num_proc` stuff from the dataset mapping calls because they produce the most horrifying logs and I irrationally distrust them.

Otherwise LGTM. (until of course it breaks mysteriously again)

Issue #371, opened 2023-06-26
Incorrect implementation of RMSNorm


vancoykendall:
There is a small bug the in the repo's RMSNorm implementation:
https://github.com/mosaicml/llm-foundry/blob/7abae0d079e8c99a8f1eb73861b65e6766e45018/llmfoundry/models/layers/norm.py#L54-L58

Since `torch.rsqrt` (reciprocal sqrt) is used instead of `torch.sqrt`, `x` should be _multiplied_ by the result instead of divided by it:

```python
output = x / torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + eps)
```
becomes
```python
output = x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + eps)
```

When training a model with the current, incorrect implementation the loss diverges. With the fix the loss appears normal.

PR #372, opened 2023-06-26
fixed rmsnorm bug. Changed division to multiply since using torch.rsqrt

vancoykendall:
RMSNorm was implemented incorrectly as noted in #371. This is the fix.


vchiley (approved):
Nice catch.
Good thing we haven't done much testing with RMSNorm 😅

PR #387, opened 2023-06-28
Fixing hf_generate bug to account for pre-tokenization

ksreenivasan:
Pre-tokenization causes the tokenizer to clean up spaces sometimes `lorem , ipsum` -> `lorem, ipsum`. This can causes incorrect trimming while printing the generated text. Fixing that here.


ksreenivasan:
Test script:
Run `python inference/hf_generate.py -n mosaicml/mpt-30b-instruct --model_dtype=bf16 --autocast_dtype=bf16 --device_map=auto --prompts "Here's a wonderful sentence. Here's a less wonderful sentence with a comma , after a space. 1+2="`

Output:
```
Using device=None and device_map='auto'
Using model_dtype=torch.bfloat16
Loading HF Config...
Explicitly passing a `revision` is encouraged when loading a configuration with custom code to ensure no malicious code has been contributed in a newer revision.

Loading HF tokenizer...
Loading HF model with dtype=torch.bfloat16...
Explicitly passing a `revision` is encouraged when loading a model with custom code to ensure no malicious code has been contributed in a newer revision.
Instantiating an MPTForCausalLM model from /root/.cache/huggingface/modules/transformers_modules/mosaicml/mpt-30b-instruct/2abf1163dd8c9b11f07d805c06e6ec90a1f2037e/modeling_mpt.py
You are using config.init_device='cpu', but you can also use config.init_device="meta" with Composer + FSDP for fast initialization.
Loading checkpoint shards: 100%|__________________________________________________________________________________________________________________________________________________________________________________________________________________| 7/7 [00:29<00:00,  4.15s/it]
n_params=29957200896
Using autocast with dtype=torch.bfloat16...

Generate seed:
42

Generate kwargs:
{'max_new_tokens': 100, 'temperature': 1.0, 'top_p': 1.0, 'top_k': 50, 'repetition_penalty': 1.0, 'no_repeat_ngram_size': 0, 'use_cache': True, 'do_sample': True, 'eos_token_id': 0, 'pad_token_id': 0}

Tokenizing prompts...
Warming up...
Generating responses...
####################################################################################################
Here's a wonderful sentence. Here's a less wonderful sentence with a comma , after a space. 1+2=3 but I don't always believe it. It's a shame that 1+2=3 is sometimes untrue. That untruth made me sad. I could have prevented that falsehood by adding ; instead of, in the latter sentence -- that would have helped me.

No, it wouldn't!
####################################################################################################
```

scripts/inference/hf_generate.py
```diff
@@ -329,6 +333,15 @@ def _generate(encoded_inp):
...
             total_output_tokens = output_tokens.sum()
+
+            if total_output_tokens == 0:
+                print('\n\nWarning: No non-special output tokens generated.')
+                print(
+                    'This can happen if the generation only contains padding/eos tokens.'
+                )
+                debug_decoded_gen = tokenizer.batch_decode(
+                    encoded_gen, skip_special_tokens=False)
+                print(f'Debug: Full generation: {debug_decoded_gen}\n\n')
```
alextrott16, line 344: A couple notes on this:
- As written, when this triggers it will re-print all the prompts which were already printed above.
- This won't trigger in the case where the batch has at least 1 prompt that led to a non-empty response.

So, I recommend folding this logic into the the block where it prints the prompt+continuation for each sequence in the batch. Ideally, it would see that `continuation` is empty and then add the printouts. I don't think we need to print the full generation with all the special tokens -- that could easily end up being just a flood of EOS in the batched setting.

alextrott16 (changes requested):
I think there are a few changes that need to be made to make this work better in the setting where the batch size is greater than 1. I've pointed those out in the comments.

In addition, can you please add a line so that `gen_tokens` is at least 1?
`gen_tokens = torch.maximum(gen_tokens, torch.ones_like(gen_tokens))`

This should help correct the divide-by-zero error that happens when the model immediately generates an EOS _and_ the EOS and pad tokens are the same (which is common).

scripts/inference/hf_generate.py
```diff
@@ -319,7 +319,11 @@ def _generate(encoded_inp):
             # Print generations
             delimiter = '#' * 100
             for prompt, gen in zip(batch, decoded_gen):
-                continuation = gen[len(prompt):]
+                # decode the encoded prompt to handle the case when the tokenizer
+                # trims extra spaces or does other pre-tokenization things
+                effective_prompt = tokenizer.batch_decode(
+                    encoded_inp['input_ids'], skip_special_tokens=True)[0]
```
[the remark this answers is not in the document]
ksreenivasan: Good catch! It shouldn't. I'll fix it.

scripts/inference/hf_generate.py
ksreenivasan: Definitely a more elegant solution. I just wanted to throw this in with the tokenization fix since it currently just throws the `divide_by_zero` warning. I'll include this but I also will open a separate PR with more changes to `hf_generate` soon to handle padding, delimiters etc.

alextrott16 (approved):
LGTM

PR #397, opened 2023-06-29
make peft installs a extra_dep

vchiley:
If I run anything I get this warning:
```
===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please run

python -m bitsandbytes

 and submit this information together with your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
```

this gates the parameter efficient fine tuning installs so that the above only happens if a user wants it to.

### DIFF
--- a/llmfoundry/models/hf/hf_causal_lm.py
+++ b/llmfoundry/models/hf/hf_causal_lm.py
@@ -55,7 +55,7 @@ class ComposerHFCausalLM(HuggingFaceModelWithZLoss):
     """
 
     def __init__(self,
-                 om_model_config: Union[DictConfig, **model_types],
+                 om_model_config: Union[DictConfig, model_types],
                  tokenizer: Tokenizer):
 
         # set up training and eval metrics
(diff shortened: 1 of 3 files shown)


llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -24,6 +23,12 @@
 from llmfoundry.models.hf.model_wrapper import HuggingFaceModelWithZLoss
 from llmfoundry.models.utils import init_empty_weights
 
+try:
+    from peft.peft_model import PeftModel
+    model_types = (PeftModel, transformers.PreTrainedModel)
+else:
```
samhavens, line 29:
```suggestion
except ImportError:
```

samhavens (commented):
I think `else -> except`, otherwise LGTM

llmfoundry/models/hf/hf_causal_lm.py
vchiley: wanted to push before a mtg.
that was hilarious 🤣

samhavens (approved):
LGTM

samhavens (approved):
LGTM2: Return of the LGTM

PR #407, opened 2023-06-30
add install for AMD beta support

vchiley:
Got a node and tested these instructions.
Only weird thing was that I needed to downgrade numpy
Also the amd flash attn build takes FOREVER!

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ pip3 install torch torchvision torchaudio --index-url https://download.pytorch.o
 ```
 **Lastly**, install the ROCm enabled flash attention (instructions [here](https://github.com/ROCmSoftwarePlatform/flash-attention/tree/flash_attention_for_rocm2#amd-gpurocm-support)).
 
-Notes: 
+Notes:
 1. `attn_impl: triton` does not work.
 1. We don't yet have a docker img where everything works perfectly. You might need to up/down grade some packages (in our case, we needed to downgrade to `numpy==1.23.5`) before everything works without issue.
 
--- a/README.md
+++ b/README.md
@@ -97,8 +97,7 @@ If you have success/failure using LLM Foundry on other systems, please let us kn
 | H100-80GB                 | 2.0.1            | 11.8         | :white_check_mark: Supported  |
 | A10-24GB                  | 1.13.1           | 11.7         | :construction: In Progress    |
 | A10-24GB                  | 2.0.1            | 11.7, 11.8   | :construction: In Progress    |
-
-We also have beta support for MI250 AMD GPUs.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,6 +98,8 @@ If you have success/failure using LLM Foundry on other systems, please let us kn
 | A10-24GB                  | 1.13.1           | 11.7         | :construction: In Progress    |
 | A10-24GB                  | 2.0.1            | 11.7, 11.8   | :construction: In Progress    |
```
abhi-mosaic, line 100: Can we add the AMD MI250 as a line here like | 2.0.1 | RoCm 5.4 |  :construction: In Progress

README.md
```diff
@@ -151,6 +153,30 @@ pip install -e ".[gpu]"  # or pip install -e . if no NVIDIA GPU
...
+```bash
+git clone https://github.com/mosaicml/llm-foundry.git
+cd llm-foundry
+
+# Creating and activate a virtual environment
+python3 -m venv llmfoundry-venv-amd
+source llmfoundry-venv-amd/bin/activate
+
+# installs
+pip install cmake packaging torch
```
abhi-mosaic, line 170: should torch be here?

vchiley: yes
its a bit screwy

torch is needed to install some deps
then out `pip install -e .` installs whatever specified version of torch it decides (usually a CPU version)
then we have to reinstall the amd torch version

vchiley: actually let me try without that torch
maybe we just need to install torch twice instead of 3 times...

vchiley: ok so it works IFF the system already has some version of torch installed.
in case ppl don't have that going for them, these instruction are complete.
worse case, pip says: requirement satisfied.

README.md
```diff
@@ -97,6 +97,7 @@ If you have success/failure using LLM Foundry on other systems, please let us kn
 | H100-80GB                 | 2.0.1            | 11.8         | :white_check_mark: Supported  |
 | A10-24GB                  | 1.13.1           | 11.7         | :construction: In Progress    |
 | A10-24GB                  | 2.0.1            | 11.7, 11.8   | :construction: In Progress    |
+| MI250                     | 2.0.1            | RoCm 5.4     | :construction: In Progress    |
```
abhi-mosaic, line 100:
```suggestion
| MI250                     | 2.0.1            | ROCm 5.4     | :construction: In Progress    |
```

abhi-mosaic (approved):
LGTM!

README.md
vchiley: 🤦‍♂️

PR #408, opened 2023-07-01
updt dtype of causal mask

vchiley:
We do it the previous way because onnx had a compat issue with bfloat

Now XLA has issues with float16


alextrott16 (approved):
:stamp:

samhavens (approved):
:shipit:

PR #420, opened 2023-07-03
rm peft from pypi package

vchiley:
PyPI does not support direct dependencies, so we remove PEFT before PyPI build


codestar12 (approved):
LGTM

PR #427, opened 2023-07-05
I think the README for json data is different

Taishi-N324:
The jsonl one has a bug that if the files are compressed by zstd, the learning starts and those files are erased.

`venv/lib/python3.10/site-packages/streaming/base/format/base/reader.py`
In
`def init_local_dir(self, filenames_present: Set[str], keep_zip: bool) -> bool:`
If we look inside self.file_pairs without the function
In the hug,
`[(FileInfo(basename='shard.00010.mds', bytes=67092637, hashes={}), None)]
`The jsonl one with zstd is.
`[(FileInfo(basename='shard.00000.mds', bytes=67092639, hashes={}), FileInfo(basename='shard.00000.mds.zstd', bytes=12403006, hashes ={}))] 
`The part of
`venv/lib/python3.10/site-packages/streaming/base/format/base/reader.py
`in the later part of the process, which is the part that deletes the file.


Taishi-N324:
You can unify each of the folders that converted the separate jsonl format data

codestar12:
> The jsonl one has a bug that if the files are compressed by zstd, the learning starts and those files are erased.
> 
> `venv/lib/python3.10/site-packages/streaming/base/format/base/reader.py` In `def init_local_dir(self, filenames_present: Set[str], keep_zip: bool) -> bool:` If we look inside self.file_pairs without the function In the hug, `[(FileInfo(basename='shard.00010.mds', bytes=67092637, hashes={}), None)] `The jsonl one with zstd is. `[(FileInfo(basename='shard.00000.mds', bytes=67092639, hashes={}), FileInfo(basename='shard.00000.mds.zstd', bytes=12403006, hashes ={}))] `The part of `venv/lib/python3.10/site-packages/streaming/base/format/base/reader.py `in the later part of the process, which is the part that deletes the file.

Hi, @Taishi-N324 

Could you go into more detail? Its not clear to me what the problem is this PR is solving. Is there an open issue it relates to?

Taishi-N324:
@codestar12 

JSON data
Using the convert_dataset_json.py script...

# Convert json dataset to StreamingDataset format
python convert_dataset_json.py \
  --path ./example_data/arxiv.jsonl \
  --out_root my-copy-arxiv --split train \
  --concat_tokens 2048 --tokenizer EleutherAI/gpt-neox-20b --eos_text '<|endoftext|>' \
  --compression zstd
Where --path can be a single json file, or a folder containing json files. --split denotes the intended split (hf defaults to train).


If you try to train with the dataset created in , the dataset placed in local will be deleted.
It seems that zstd has a bug that erases the dataset

Processing in hugginface is not exactly zstd

HuggingFace data
In this example, we use the convert_dataset_hf.py script to convert a HuggingFace c4 dataset into a StreamingDataset, using the EleutherAI/gpt-neox-20b tokenizer. The resulting directory is saved at ./llm-foundry/scripts/data_prep/my-copy-c4.

Currently supports c4 and The Pile.

# Convert C4 dataset to StreamingDataset format
python convert_dataset_hf.py \
  --dataset c4 --data_subset en \
  --out_root my-copy-c4 --splits train_small val_small \
  --concat_tokens 2048 --tokenizer EleutherAI/gpt-neox-20b --eos_text '<|endoftext|>' \
  --compression zstd

float-trip:
Tagging @karan6181, who was looking into this bug when I mentioned it on Slack a while back

dakinggg:
This issue should be fixed in the latest streaming release, please let us know if you are still encountering it.

PR #432, opened 2023-07-06
Add linear layer and ffn config to enable TransformerEngine layers (with FP8)

vchiley:
This PR adds a config for Linear Layers and FFN modules which allows the use TransformerEngine's te.Linear and te.LayerNormMLP modules (which have fp8 with amp.fp8).
\+ I did a little cleanup

This PR is built on top of https://github.com/mosaicml/llm-foundry/pull/271

In the future, this'll also allow us to add and prototype other linear layers and ffn blocks. Furthermore it enables us to configure TP/SP for the MLP block, in the `build_ffn` util fn.

AMP FP8 training gets results which are nearly identical to AMP BF16:
image
but has faster runtime.

Furthermore `ffn_config_defaults: ffn_type: te_ln_mlp` allows us to use [transformer engine's LayerNormMLP](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/pytorch.html#transformer_engine.pytorch.LayerNormMLP) layer which has SP and TP support if configured correctly.


llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -40,13 +46,13 @@ def __init__(
         d_model: int = 2048,
         n_heads: int = 16,
         n_layers: int = 24,
-        expansion_ratio: int = 4,
```
dakinggg, line 43: Lets not break backwards compat just for this

vancoykendall:
I'm excited that this PR is being worked on, mainly because I've been extending the `MPTMLP` myself in order to try using gated linear units (GEGLU, SwiGLU, etc) and also decided to add a `ffn_config` in my implementation. I was wondering though, why `expansion_ratio` is not a part of the `ffn_config`? From my understanding, `expansion_ratio` is only used when building the ffn block, so it makes more sense to me for it to be defined in the `ffn_config`. Also, what does `fc` stand for in `fc_type`? It isn't clear to me.

vchiley:
@vancoykendall 

`fc` stands for **f**ully **c**onnected layer

While `expansion_ratio` should really be part of the `ffn_config`, moving it into the `ffn_config` breaks backwards compatibility with models already uploaded to HuggingFace hub and all of our internal checkpoints so we are leaving that config variable alone.

scripts/train/train.py
```diff
@@ -59,6 +59,32 @@ def validate_config(cfg):
...
 
+    if (cfg.model.get('fc_type', 'torch') != 'te' and 'te' not in cfg.model.get(
+            'ffn_config', {}).get('ffn_type', 'mptmlp') and
+            'fp8' in cfg.precision):
+        warnings.warn(
+            "fp8 only supported for te.Linear layers. Either set `cfg.model.fc_typ='te'` or "
+            "`cfg.model.ffn_config.ffn_type='te_ln_mlp'` to enable layers using fp8 precision."
+        )
+
+    if ((cfg.model.get('fc_type', 'torch') == 'te' or 'te' not in cfg.model.get(
```
sashaDoubov, line 70: these `if` conditions are a little hard to parse quickly because of the defaults. Would it make sense to make to do several levels of `if` statements + a quick comment, rather than as a single `if` line?

vchiley: I split the conditional.

re comment
I feel like the warning is comment enough
```
            warnings.warn(
                '`te.Linear` layers do not support activation_checkpointing with '
                '`activation_checkpointing_reentrant = False`. '
                'Setting cfg.fsdp_config.activation_checkpointing_reentrant=True.'
            )
```
Is there anything else you'd add to it?

sashaDoubov: Yea that should be clear enough. I think the defaults make it harder to parse, but they are necessary for bwd compatibility/ability to not have to fully specify everything so it should be fine. Thanks!

sashaDoubov:
LGTM

PR #440, opened 2023-07-07
Add MaxSeqLen to ICL tasks

vchiley:
~~Add MaxSeqLen to ICL tasks in `tasks.yaml`~~

~~We have it in [`tasks_light.yaml`](https://github.com/mosaicml/llm-foundry/blob/main/scripts/eval/yamls/tasks_light.yaml) but not here~~

Seems like the more correct approach would be to enable a `icl_max_seq_len` cfg setting.

### DIFF
--- a/scripts/eval/yamls/tasks.yaml
+++ b/scripts/eval/yamls/tasks.yaml
@@ -1,3 +1,4 @@
+icl_seq_len: 2048
 icl_tasks:
 -
   label: jeopardy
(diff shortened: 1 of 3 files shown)


bmosaicml (commented):
Typically we have been setting the default for all tasks at once in eval cfg (e.g. in `eval/yamls/hf_eval.yaml`). I think we might be better off leaving this in that yaml because it's typically the yaml people will edit before running the code. At any rate I've been using 1024 as the seqlen for the model gauntlet.

PR #441, opened 2023-07-07
Add Programming to Foundry

rishab-partha:
Added programming datasets and added programming to the eval yamls/models.


danbider (commented):
looks good to me. @codestar12 it'll be useful for me if this got merged.

scripts/eval/yamls/model_gauntlet.yaml
```diff
@@ -112,3 +112,17 @@ model_gauntlet:
     - name: boolq
       num_fewshot: 10
       random_baseline: 0.5
+  - name: programming
+    benchmarks:
+    - name: human_eval
+      num_fewshot: 0
```
bmosaicml, line 118: I know it is not conventional to do few shot prompting, but this seems like a task where models would give more reliable responses with 5-10 in context examples. Could you launch a run with e.g. MPT30B/Falcon 40B with 0, 5, 10 shot of all these benchmarks and see whether perf improves?

scripts/eval/local_data/programming/processed_humaneval_c.jsonl
bmosaicml, line 1: Might not be worth including a benchmark with only 9 examples. Can this be combined with the C++ benchmark? If not might be best to drop it from the gauntlet entirely

rishab-partha: ah yea this is supposed to be a proof of concept, for like how we would create a c/c++ dataset

rishab-partha: we can check it in, but not add to gauntlet

rishab-partha: ^^ i think this might be ideal b/c other ppl might have to take on creating this kind of dataset in the future

bmosaicml (approved):
Looks great. Note to future self: per discussion with Rishab the official v0.1 of model gauntlet will remove human eval c and set num_fewshot to 3 for all programming tasks

PR #448, opened 2023-07-10
use mono checkpoints

samhavens:
Users are ending up with sharded checkpoints, unable to convert to HF format. I believe that removing this line will cause us to save mono checkpoints.


dakinggg (approved):
LGTM, just wanted to make sure there wasn't a reason it was originally written this way

samhavens:
Same

PR #467, opened 2023-07-18
Autoresume default on 

mvpatel2000:
As much as possible, we want to turn on autoresume when we specify all criteria. This avoids cases where users turn on watchdog but do not turn on autoresume, resulting in a very bad user experience (restart from scratch...).

This PR turns on autoresume by default when all criteria are met.

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -292,8 +292,7 @@ def main(cfg):
         for name, algorithm_cfg in (cfg.get('algorithms') or {}).items()
     ]
 
-    # Build the Trainer
-    print('Building trainer...')
+    # Set autoresume default on if possible
     save_folder = cfg.get('save_folder', None)
     save_latest_filename = cfg.get('save_latest_filename',
                                    'latest-rank{rank}.pt')
@@ -302,6 +301,9 @@ def main(cfg):
     autoresume_default = False
     if cfg.run_name is not None and save_folder is not None and save_latest_filename is not None and not save_overwrite and not save_weights_only:
         autoresume_default = True
+
+    # Build the Trainer
+    print('Building trainer...')
     trainer = Trainer(
         run_name=cfg.run_name,
         seed=cfg.seed,
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -300,6 +300,9 @@ def main(cfg):
     save_weights_only = cfg.get('save_weights_only', False)
     autoresume_default = False
     if cfg.run_name is not None and save_folder is not None and save_latest_filename is not None and not save_overwrite and not save_weights_only:
+        print(
+            'As run_name, save_folder, and save_latest_filename are set, changing autoresume default to True...'
+        )
         autoresume_default = True
 
     # Build the Trainer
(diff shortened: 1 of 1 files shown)


scripts/train/train.py
```diff
@@ -294,6 +294,14 @@ def main(cfg):
...
     # Build the Trainer
     print('Building trainer...')
+    save_folder = cfg.get('save_folder', None)
+    save_latest_filename = cfg.get('save_latest_filename',
+                                   'latest-rank{rank}.pt')
+    save_overwrite = cfg.get('save_overwrite', False)
+    save_weights_only = cfg.get('save_weights_only', False)
+    autoresume_default = False
+    if cfg.run_name is not None and save_folder is not None and save_latest_filename is not None and not save_overwrite and not save_weights_only:
+        autoresume_default = True
```
nancyhung, line 304: Nit: can you write a comment here about turning `autoresume` on if the requirements are configured correctly?

scripts/train/train.py
```diff
@@ -294,6 +294,14 @@ def main(cfg):
 
     # Build the Trainer
     print('Building trainer...')
+    save_folder = cfg.get('save_folder', None)
+    save_latest_filename = cfg.get('save_latest_filename',
+                                   'latest-rank{rank}.pt')
+    save_overwrite = cfg.get('save_overwrite', False)
+    save_weights_only = cfg.get('save_weights_only', False)
+    autoresume_default = False
```
nancyhung, line 302: Btw I don't think this actually configures `autoresume` by default on. If and only if all requirements are configured properly, then we turn on `autoresume`. Shouldn't we emit a warning that `autoresume` is not turned on by default if one of these requirements is not met?

mvpatel2000: Yea, I think this is the closest we can get. All of these are on by default except save folder, which users need to do autoresume / always provide.

I think it's ok to not emit a warning. These are rarely touched, and this is a second line of defense -- normally all of our configs have autoresume on anyways. This will trigger on >95% of user workloads, so I'm not really concerned about it

scripts/train/train.py
```diff
@@ -292,6 +292,16 @@ def main(cfg):
...
     ]
 
+    # Set autoresume default on if possible
+    save_folder = cfg.get('save_folder', None)
+    save_latest_filename = cfg.get('save_latest_filename',
+                                   'latest-rank{rank}.pt')
+    save_overwrite = cfg.get('save_overwrite', False)
+    save_weights_only = cfg.get('save_weights_only', False)
+    autoresume_default = False
+    if cfg.run_name is not None and save_folder is not None and save_latest_filename is not None and not save_overwrite and not save_weights_only:
```
vchiley, line 302: Lint allowed this?
What are the sweet whispers?

(ignorable comment)

vchiley (approved):
Given `autoresume=False` is the composer default, I'd throw a `warnings.warn` msg telling the user we are changing it to `True` if the correct conditions are met; else throw a `warnings.warn` that the conditions are not met (optional).

scripts/train/train.py
mvpatel2000: There's no paranthesis :)

PR #469, opened 2023-07-19
Support eval loader when finetuning from JSONL files in object stores

samhavens:
`datasets` caching was causing the error, saving to a non-temp dir, with different dirs per split, fixes it

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -157,7 +157,8 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
                     '`split` key in the dataset config.'
                 )
             supported_extensions = ['jsonl', 'csv', 'parquet']
-            finetune_dir = "/data/finetuning"
+            # using Tempdir causes issues with HF datasets caching
+            finetune_dir = f'/llm-foundry/data/finetuning/{cfg.dataset.split}'
             for extension in supported_extensions:
                 name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
                 destination = str(
@@ -178,11 +179,7 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
                             f'Could not find {name}, looking for another extension'
                         )
                     continue
-                # 'json' causes special behavior in the dataset constructor
-                cfg.dataset.hf_name = extension if extension != 'jsonl' else 'json'
-                kwargs = cfg.dataset.get('hf_kwargs', {})
-                kwargs['data_files'] = destination
-                cfg.dataset['hf_kwargs'] = kwargs
+                cfg.dataset.hf_name = finetune_dir
                 print(cfg.dataset)
                 dataset = dataset_constructor.build_from_hf(
                     cfg.dataset,
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -249,7 +249,7 @@ def _build_hf_dataset_from_remote(cfg: DictConfig, tokenizer: Tokenizer):
     Args:
         cfg (DictConfig): The configuration dictionary containing the necessary parameters to load the dataset.
             This includes:
-                - dataset.hf_name: The name of the HuggingFace dataset to download.
+                - dataset.hf_name: The path of the HuggingFace dataset to download.
                 - dataset.split: The dataset split to download (e.g., 'train', 'validation', 'test').
                 - dataset.max_seq_len: The maximum sequence length for tokenizing the dataset.
 
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -272,7 +272,7 @@ def _build_hf_dataset_from_remote(cfg: DictConfig, tokenizer: Tokenizer):
             os.path.abspath(f'{finetune_dir}/{cfg.dataset.split}.{extension}'))
         # Since we don't know exactly what the extension will be, since it is one of a list
         # use a signal file to wait for instead of the desired file
-        signal_file_path = os.path.join(finetune_dir, 'the_eagle_has_landed')
+        signal_file_path = os.path.join(finetune_dir, '.the_eagle_has_landed')
         if dist.get_local_rank() == 0:
             try:
                 get_file(name, destination, overwrite=True)
@@ -290,7 +290,7 @@ def _build_hf_dataset_from_remote(cfg: DictConfig, tokenizer: Tokenizer):
 
             os.makedirs(os.path.dirname(signal_file_path), exist_ok=True)
             with open(signal_file_path, 'wb') as f:
-                f.write(b'local_rank0_completed_autoresume')
+                f.write(b'local_rank0_completed_download')
 
         # Avoid the collective call until the local rank zero has finished trying to download the checkpoint
         # so that we don't timeout for large downloads. This syncs all processes on the node
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -157,7 +157,9 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
                 )
             supported_extensions = ['jsonl', 'csv', 'parquet']
             # using Tempdir causes issues with HF datasets caching
-            finetune_dir = f'/llm-foundry/data/finetuning/{cfg.dataset.split}'
+            finetune_dir = os.path.join(
+                os.path.dirname(os.path.realpath(__file__)),
+                f'downloaded_finetuning_data/{cfg.dataset.split}')
             for extension in supported_extensions:
                 name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
                 destination = str(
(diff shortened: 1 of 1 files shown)


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -157,39 +156,36 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
-                    cfg.dataset['hf_kwargs'] = kwargs
-                    print(cfg.dataset)
-                    dataset = dataset_constructor.build_from_hf(
-                        cfg.dataset,
-                        max_seq_len=cfg.dataset.max_seq_len,
-                        tokenizer=tokenizer,
-                    )
-                    break
+            # using Tempdir causes issues with HF datasets caching
+            finetune_dir = f'/llm-foundry/data/finetuning/{cfg.dataset.split}'
```
dakinggg, line 160: This doesn't work as a hardcoded dir, as this path may not exist and would be a bit surprising to the user for their data to end up there. I'd suggest something like `os.path.join(os.path.dirname(os.path.realpath(__file__)), "downloaded_finetuning_data/{cfg.dataset.split}`

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -157,39 +156,36 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+            for extension in supported_extensions:
+                name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
+                destination = str(
+                    os.path.abspath(
+                        f'{finetune_dir}/{cfg.dataset.split}.{extension}'))
+                try:
+                    with dist.run_local_rank_zero_first():
```
dakinggg, line 167: I'm probably just forgetting something, but I think we just want `get_file` to only run on local rank 0? When it was a tmp path, you were maybe running it on all ranks cause the the tmp_dir was different on each rank, but now that its a fixed path can we just download it once?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -157,39 +156,36 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
-                            ) from e
-                        else:
-                            print(
-                                f'Could not find {name}, looking for another extension'
-                            )
-                        continue
-                    # 'json' causes special behavior in the dataset constructor
-                    cfg.dataset.hf_name = extension if extension != 'jsonl' else 'json'
-                    kwargs = cfg.dataset.get('hf_kwargs', {})
-                    kwargs['data_files'] = destination
```
dakinggg, line 184: was deleting these intentional? are they unused?

dakinggg: (`data_files` and `hf_kwargs` that is)

samhavens: yes and yes. rather than data_files we now use path (the only non kw arg). lots of way to do the same thing

llmfoundry/data/finetuning/dataloader.py
samhavens: thanks, fixed.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -157,15 +156,23 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+            for extension in supported_extensions:
+                name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
+                destination = str(
+                    os.path.abspath(
+                        f'{finetune_dir}/{cfg.dataset.split}.{extension}'))
+                # Since we don't know exactly what the extension will be, since it is one of a list
+                # use a signal file to wait for instead of the desired file
+                signal_file_path = os.path.join(
+                    os.path.dirname(os.path.realpath(__file__)),
+                    'the_eagle_has_landed.txt')
```
dakinggg, line 172:
```suggestion
                    '.the_eagle_has_landed')
```

dakinggg: This should be defined outside of the for loop i think?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -157,15 +156,23 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
                 )
             supported_extensions = ['jsonl', 'csv', 'parquet']
-            with tempfile.TemporaryDirectory() as tmp_dir:
-                for extension in supported_extensions:
-                    name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
-                    destination = str(
-                        os.path.abspath(
-                            f'{tmp_dir}/{cfg.dataset.split}.{extension}'))
+            # using Tempdir causes issues with HF datasets caching
+            finetune_dir = os.path.join(
```
dakinggg, line 160: should probably up front call `os.makedirs` on the finetune_dir

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -178,18 +185,31 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
-                        cfg.dataset,
-                        max_seq_len=cfg.dataset.max_seq_len,
-                        tokenizer=tokenizer,
-                    )
-                    break
+
+                    os.makedirs(os.path.dirname(signal_file_path),
+                                exist_ok=True)
+                    with open(signal_file_path, 'wb') as f:
+                        f.write(b'local_rank0_completed_autoresume')
```
dakinggg, line 192:
```suggestion
                        f.write(b'local_rank0_download_successful')
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -178,18 +185,31 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+
+                # Avoid the collective call until the local rank zero has finished trying to download the checkpoint
```
dakinggg, line 194:
```suggestion
                # Avoid the collective call until the local rank zero has finished trying to download the file
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -178,18 +185,31 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+                # so that we don't timeout for large downloads. This syncs all processes on the node
+                with dist.local_rank_zero_download_and_wait(signal_file_path):
+                    # Then, wait to ensure every node has finished downloading the checkpoint
```
dakinggg, line 197:
```suggestion
                    # Then, wait to ensure every node has finished downloading the file
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -157,15 +156,23 @@ def build_finetuning_dataloader(cfg: DictConfig, tokenizer: Tokenizer,
...
+                f'downloaded_finetuning_data/{cfg.dataset.split}')
+            for extension in supported_extensions:
+                name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
+                destination = str(
+                    os.path.abspath(
+                        f'{finetune_dir}/{cfg.dataset.split}.{extension}'))
+                # Since we don't know exactly what the extension will be, since it is one of a list
+                # use a signal file to wait for instead of the desired file
+                signal_file_path = os.path.join(
+                    os.path.dirname(os.path.realpath(__file__)),
```
dakinggg, line 171: lets just write this into `finetune_dir`?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -269,6 +235,84 @@ def _validate_config(dataset_cfg: DictConfig):
...
+        os.path.dirname(os.path.realpath(__file__)),
+        f'downloaded_finetuning_data/{cfg.dataset.split}')
+    os.makedirs(finetune_dir, exist_ok=True)
+    for extension in supported_extensions:
+        name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
+        destination = str(
+            os.path.abspath(f'{finetune_dir}/{cfg.dataset.split}.{extension}'))
+        # Since we don't know exactly what the extension will be, since it is one of a list
+        # use a signal file to wait for instead of the desired file
+        signal_file_path = os.path.join(finetune_dir, 'the_eagle_has_landed')
```
dakinggg, line 275:
```suggestion
        signal_file_path = os.path.join(finetune_dir, '.the_eagle_has_landed')
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -269,6 +235,84 @@ def _validate_config(dataset_cfg: DictConfig):
...
+                        f'at {cfg.dataset.hf_name}/{cfg.dataset.split}'
+                    ) from e
+                else:
+                    print(
+                        f'Could not find {name}, looking for another extension')
+                continue
+
+            os.makedirs(os.path.dirname(signal_file_path), exist_ok=True)
+            with open(signal_file_path, 'wb') as f:
+                f.write(b'local_rank0_completed_autoresume')
```
dakinggg, line 293:
```suggestion
                f.write(b'local_rank0_completed_download')
```

dakinggg (approved):
LGTM with a couple last comments. Also if you could put evidence of testing this in the pr description that would be great.

PR #470, opened 2023-07-20
Fix missing import

bcui19:


dakinggg (approved):
Separately can you add a test that would've caught this?

dakinggg:
@bcui19 I manually tested that quickstart works again after this change, so I'm going to go ahead and merge, but please add a test that would've caught this. Thanks!

PR #475, opened 2023-07-20
Add MLFlow as a logger option

aspfohl:
Installs mflow by default and allows for mlflow logging!

You just need to specify the mlflow logger in the yaml, this will even work for hosted mlflow:
```
  loggers:
    mlflow:
      experiment_name: /Users/<email>/<experimentfolder>
```


eracah (commented):
Looks pretty good. Can you add `mlflow.log_params(params=hyperparameters)` to the [log_config](https://github.com/mosaicml/llm-foundry/blob/ca0bf0fc83b27d734d53483d2e87fdb60c462a68/llmfoundry/utils/config_utils.py#L91) function similar to how WandB does it

llmfoundry/utils/config_utils.py
```diff
@@ -97,3 +97,11 @@ def log_config(cfg: DictConfig):
             raise e
         if wandb.run:
             wandb.config.update(om.to_container(cfg, resolve=True))
+
+    if 'mlflow' in cfg.get('loggers', {}):
+        try:
+            import mlflow
+        except ImportError as e:
+            raise e
+        if mlflow.active_run():
```
aspfohl, line 106: I think [this](https://mlflow.org/docs/latest/tracking.html#logging-functions) is what I want?

eracah: actually, we tested it last week. we're good!

eracah (approved):
LGTM!

PR #477, opened 2023-07-20
Adding pyright to pre-commit

bcui19:
This PR adds pyright to precommit and gets it to pass so that we stop making things worse. Things that should still be done in the future include (1) getting rid of as many type ignores as possible and (2) adding return types to stuff (3) fix a todo with the model gauntlet callback which currently does not actually work as a callback.

I've also removed python 3.8 from the list of python versions. The `ast.unparse` function used in the editing of the files for huggingface compatibility does not exist on python 3.8, and I think it is ok to require >=3.9

- [x] Run through quickstart and make sure nothing is broken

### DIFF
--- a/llmfoundry/callbacks/generate_callback.py
+++ b/llmfoundry/callbacks/generate_callback.py
@@ -75,9 +75,11 @@ class Generate(Callback):
         dummy_input = device.tensor_to_device(dummy_input)
         with get_precision_context(state.precision):
             with torch.no_grad():
+                assert isinstance(model.model, torch.nn.Module)
                 _ = model.model(input_ids=dummy_input)
 
-            output_token_ids = model.model.generate(
+            assert hasattr(model.model, 'generate')
+            output_token_ids = model.model.generate(  # type: ignore
                 input_ids=tokenized_input['input_ids'],
                 attention_mask=tokenized_input['attention_mask'],
                 synced_gpus=True,
--- a/llmfoundry/callbacks/generate_callback.py
+++ b/llmfoundry/callbacks/generate_callback.py
@@ -58,6 +58,11 @@ class Generate(Callback):
         tokenizer = cast(Tokenizer, state.model.tokenizer)
         device = state.device
 
+        if not hasattr(model.model, 'generate'):
+            raise ValueError(
+                f'Cannot generate from model {model.model.__class__.__name__} because it does not have a `generate` method'
+            )
+
         # stash the original original value of padding_side because generation requires left padding
         original_padding_side = tokenizer.padding_side
         tokenizer.padding_side = 'left'
--- a/llmfoundry/callbacks/generate_callback.py
+++ b/llmfoundry/callbacks/generate_callback.py
@@ -83,7 +83,6 @@ class Generate(Callback):
                 assert isinstance(model.model, torch.nn.Module)
                 _ = model.model(input_ids=dummy_input)
 
-            assert hasattr(model.model, 'generate')
             output_token_ids = model.model.generate(  # type: ignore
                 input_ids=tokenized_input['input_ids'],
                 attention_mask=tokenized_input['attention_mask'],
--- a/llmfoundry/callbacks/generate_callback.py
+++ b/llmfoundry/callbacks/generate_callback.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """Periodically log generations to wandb from a set of prompts."""
-from typing import List, Union, cast
+from typing import Any, List, Union, cast
 
 import torch
 import wandb
@@ -16,7 +16,8 @@ Tokenizer = Union[PreTrainedTokenizer, PreTrainedTokenizerFast]
 
 class Generate(Callback):
 
-    def __init__(self, prompts: List[str], batch_log_interval: int, **kwargs):
+    def __init__(self, prompts: List[str], batch_log_interval: int,
+                 **kwargs: Any):
         """Periodically log generations to wandb from a set of prompts.
 
         In the main view for a run, there will be a table that will show the _last_ logged generations.
--- a/llmfoundry/callbacks/model_gauntlet_callback.py
+++ b/llmfoundry/callbacks/model_gauntlet_callback.py
@@ -35,7 +35,8 @@ class ModelGauntlet(Callback):
         weighting (Weighting): The weighting scheme used to balance different tasks within each category.
                                Either assign them all equal weight, assign them weight proportional
                                to the dataset size, or assign them weight proportional to the log2 of the dataset size.
-        substract_random_baseline (bool): Flag determining whether to subtract random baseline accuracy
+                               Options are 'EQUAL', 'SAMPLE_SZ', and 'LOG_SAMPLE_SZ'.
+        subtract_random_baseline (bool): Flag determining whether to subtract random baseline accuracy
                                           from the performance on each individual benchmark before aggregating.
         rescale_accuracy (bool): Flag determining whether to rescale the accuracy on each benchmark
                                  by (1-random_baseline_accuracy) before aggregating. Using this ensures that all benchmarks max out at 1.0.
@@ -45,7 +46,7 @@ class ModelGauntlet(Callback):
     def __init__(self,
                  logger_keys: dict,
                  categories: dict,
-                 weighting: Weighting = Weighting.EQUAL,
+                 weighting: str = 'EQUAL',
                  subtract_random_baseline: bool = True,
                  rescale_accuracy: bool = True,
                  benchmark_sizes: Optional[dict] = None):
@@ -69,9 +70,14 @@ class ModelGauntlet(Callback):
 
             for benchmark in category['benchmarks']:
                 bench_name = f"{benchmark['name']}/{benchmark['num_fewshot']}-shot"
-                cumulative_samples = max(
-                    sum(count for name, count in benchmark_sizes.items()
-                        if name.startswith(bench_name)), 1)
+
+                if self.weighting != Weighting.EQUAL:
+                    assert benchmark_sizes is not None
+                    cumulative_samples = max(
+                        sum(count for name, count in benchmark_sizes.items()
+                            if name.startswith(bench_name)), 1)
+                else:
+                    cumulative_samples = -1  # pyright
 
                 if self.weighting == Weighting.EQUAL:
                     weight = 1
@@ -79,14 +85,19 @@ class ModelGauntlet(Callback):
                     weight = cumulative_samples
                 elif self.weighting == Weighting.LOG_SAMPLE_SZ:
                     weight = max(math.log(cumulative_samples, 2), 1)
+                else:
+                    weight = None
+                    raise ValueError(
+                        f'Unexpected weighting scheme {self.weighting}')
 
                 benchmark['weighting'] = weight
 
-    def compute_averages(self, logger_data):
+    def compute_averages(self, logger_data: Logger):
 
         results = {}
         pat = re.compile(
-            'metrics/(.*?)/(\d+)-shot(/.*?)?/InContextLearning(.*)')
+            'metrics/(.*?)/(\d+)-shot(/.*?)?/InContextLearning(.*)'  # type: ignore
+        )
         for key in self.logger_keys:
             match = pat.match(key)
             val = logger_data.data[key][0][1].item()
--- a/llmfoundry/callbacks/model_gauntlet_callback.py
+++ b/llmfoundry/callbacks/model_gauntlet_callback.py
@@ -78,18 +78,16 @@ class ModelGauntlet(Callback):
                             if name.startswith(bench_name)), 1)
                 else:
                     cumulative_samples = -1  # pyright
-
+                
+                weight = None
                 if self.weighting == Weighting.EQUAL:
                     weight = 1
                 elif self.weighting == Weighting.SAMPLE_SZ:
                     weight = cumulative_samples
                 elif self.weighting == Weighting.LOG_SAMPLE_SZ:
                     weight = max(math.log(cumulative_samples, 2), 1)
-                else:
-                    weight = None
-                    raise ValueError(
-                        f'Unexpected weighting scheme {self.weighting}')
 
+                assert weight is not None
                 benchmark['weighting'] = weight
 
     def compute_averages(self, logger_data: Logger):
--- a/llmfoundry/callbacks/model_gauntlet_callback.py
+++ b/llmfoundry/callbacks/model_gauntlet_callback.py
@@ -78,7 +78,7 @@ class ModelGauntlet(Callback):
                             if name.startswith(bench_name)), 1)
                 else:
                     cumulative_samples = -1  # pyright
-                
+
                 weight = None
                 if self.weighting == Weighting.EQUAL:
                     weight = 1
--- a/llmfoundry/callbacks/model_gauntlet_callback.py
+++ b/llmfoundry/callbacks/model_gauntlet_callback.py
@@ -100,7 +100,10 @@ class ModelGauntlet(Callback):
         )
         for key in self.logger_keys:
             match = pat.match(key)
-            val = logger_data.data[key][0][1].item()
+
+            # TODO(bmosaicml) This needs to be factored for this callback to work as a normal callback
+            # and therefore for the typing to be fixed
+            val = logger_data.data[key][0][1].item()  # type: ignore
 
             if match:
                 eval_name = match.group(1)
--- a/llmfoundry/data/denoising.py
+++ b/llmfoundry/data/denoising.py
@@ -13,7 +13,7 @@ import torch
 from omegaconf import DictConfig
 from omegaconf import OmegaConf as om
 from torch.utils.data import DataLoader
-from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast
+from transformers import PreTrainedTokenizerBase
 
 from llmfoundry.data.packing import BinPackWrapper
 from llmfoundry.data.text_data import StreamingTextDataset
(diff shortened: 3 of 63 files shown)


llmfoundry/callbacks/generate_callback.py
```diff
@@ -74,9 +75,11 @@ def generate(self, state: State, logger: Logger):
         dummy_input = device.tensor_to_device(dummy_input)
         with get_precision_context(state.precision):
             with torch.no_grad():
+                assert isinstance(model.model, torch.nn.Module)
                 _ = model.model(input_ids=dummy_input)
 
-            output_token_ids = model.model.generate(
+            assert hasattr(model.model, 'generate')
```
hanlint, line 81: This might want to be a `ValueError`, since a user could conceivably try this callback with a model that doesn't have generate function?

llmfoundry/callbacks/generate_callback.py
```diff
@@ -74,9 +75,11 @@ def generate(self, state: State, logger: Logger):
...
+            output_token_ids = model.model.generate(  # type: ignore
```
hanlint, line 82: If you checked the attribute in the previous line, why should this be a type ignore?

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -45,7 +46,7 @@ class ModelGauntlet(Callback):
     def __init__(self,
                  logger_keys: dict,
                  categories: dict,
-                 weighting: Weighting = Weighting.EQUAL,
+                 weighting: str = 'EQUAL',
```
hanlint, line 49: This is a breaking API change -- can we avoid this?

llmfoundry/data/denoising.py
```diff
@@ -678,7 +677,8 @@ def _sample_span_lengths(total_tokens: int, num_spans: int) -> np.ndarray:
         """
         span_markers = np.less(np.arange(total_tokens - 1), num_spans -
                                1)[np.random.permutation(total_tokens - 1)]
-        span_start_indicator = np.concatenate([[0], span_markers])
+        span_start_indicator = np.concatenate([[0],
+                                               span_markers])  # type: ignore
```
hanlint, line 681: This seems to be pure numpy types... why do we need type ignore here? Same with all of the below.

llmfoundry/models/hf/hf_prefix_lm.py
```diff
@@ -145,7 +144,7 @@ def __init__(self, om_model_config: DictConfig, tokenizer: Tokenizer):
 
         return composer_model
 
-    def forward(self, batch):
+    def forward(self, batch: MutableMapping):
```
hanlint, line 147: I've never seen this type applied to a `batch` before, might be confusing to users. Is there a more narrow type possible here?

llmfoundry/models/layers/ffn.py
```diff
@@ -76,6 +74,7 @@ def build_ffn(
             device=device,
         )
     elif ffn_type == 'te_ln_mlp':
+        assert te is not None
```
hanlint, line 77: This should be an error return to the user right? e.g please install transformer engine..

hanlint (commented):
Seriously, thank you for doing this. I only reviewed the `llmfoundry` folder files (stopped at `pyproject.toml` in this first pass). A few general notes:
* I saw `MutableMapping` or `Mapping` used for `batch`. Is it possible to be consisent.. and is there a more narrow type, e.g. `Dict` that could be used?
* noted a few `assert` that maybe shoudl be informative user errors

llmfoundry/callbacks/generate_callback.py
dakinggg: That is a great question that I could not figure out 😅 it was only python 3.9 that it failed, not 3.10. so I conceded on this one.

scripts/eval/eval.py
```diff
@@ -41,7 +46,10 @@ def load_model(model_cfg, tokenizer, fsdp_config, num_retries):
                     )
 
 
-def evaluate_model(model_cfg, run_name, model_gauntlet_df):
+def evaluate_model(model_cfg: DictConfig, cfg: DictConfig, run_name: str,
+                   model_gauntlet_df: Optional[pd.DataFrame]):
+    print('_' * 30)
```
dakinggg, line 51: Remove

llmfoundry/callbacks/model_gauntlet_callback.py
dakinggg: So, this callback has a couple issues that mean I strongly suspect no one is using this as is (see the comment in `compute_averages` for the other major issue). One of the issue is that this default actually doesn't work. It would then try to access weighting via `Weighting[Weighting.EQUAL]` which is not a valid way to access an enum. Given that we mostly instantiate via yaml, I assume this was getting instantiated via a string from a yaml and the default has never been run.

llmfoundry/data/denoising.py
dakinggg: Yeah, I do not know. I accepted these type ignores in favor of attempting to actually get this pr in. There are a variety of type ignores throughout the repo that should be revisited, but I figured better to stop making things worse. Trade off of how much time I spend on this pr vs reward.

llmfoundry/models/hf/hf_prefix_lm.py
dakinggg: So, I _think_ this is the right type although open to a different suggestion. Sometimes the HF batch input was a `UserDict` (which is why this isn't just a `Dict` type). A `Union[Dict, UserDict]` probably also works, but `Mapping` does encompass the properties needed for a batch. As for the `MutableMapping` vs `Mapping`, they could all be `MutableMapping` without issue. Just set them based on whether the code in question tried to modify the batch or not.

llmfoundry/models/layers/ffn.py
dakinggg: The error is actually in the config validation (https://github.com/mosaicml/llm-foundry/blob/bd8127252c660e45ed01413645d29427f86c085a/llmfoundry/models/mpt/configuration_mpt.py#L218-L227), so at this point we can assume the import has already been checked and this assert is really just for pyright.

llmfoundry/callbacks/generate_callback.py
dakinggg: Yeah, thats fair. Will change.

llmfoundry/callbacks/model_gauntlet_callback.py
hanlint: OK -- can you file a JIRA ticket for @bmosaicml to fix?

dakinggg: Yeah, I think he already knows/is discussing with Evan how to implement this in a better way. @bmosaicml is there already a jira?

Skylion007 (approved):

scripts/data_prep/convert_dataset_hf.py
```diff
@@ -252,7 +252,8 @@ def _est_progress_denominator(total_samples: int, chars_per_sample: int,
         return total_samples * est_tokens_per_sample // max_length
 
 
-def build_dataloader(dataset, batch_size, num_workers) -> DataLoader:
+def build_dataloader(dataset: Dataset, batch_size: int,
+                     num_workers: Optional[int]) -> DataLoader:
```
vchiley, line 256:
```suggestion
                     num_workers: Optional[int] = None) -> DataLoader:
```

llmfoundry/optim/lion.py
```diff
@@ -103,14 +106,14 @@ def step(self, closure: Optional[Callable] = None):
...
-    def dist_reduce_metrics(self, optimizer_metrics):
+    def dist_reduce_metrics(self, optimizer_metrics: Dict[str, torch.Tensor]):
         for metric in optimizer_metrics:
             if metric.startswith('l2_norm'):
                 reduced = optimizer_metrics[metric]
                 if dist.get_world_size() > 1:
                     dist.all_reduce(reduced, reduce_operation='SUM')
 
-                optimizer_metrics[metric] = math.sqrt(reduced)
+                optimizer_metrics[metric] = torch.tensor(math.sqrt(reduced))
```
vchiley, line 116: is changing the type here ok?

llmfoundry/optim/adaptive_lion.py
```diff
@@ -385,14 +391,14 @@ def step(self, closure: Optional[Callable] = None):
...
-    def dist_reduce_metrics(self, optimizer_metrics):
+    def dist_reduce_metrics(self, optimizer_metrics: Dict[str, torch.Tensor]):
         for metric in optimizer_metrics:
             if metric.startswith('l2_norm'):
                 reduced = optimizer_metrics[metric]
                 if dist.get_world_size() > 1:
                     dist.all_reduce(reduced, reduce_operation='SUM')
 
-                optimizer_metrics[metric] = math.sqrt(reduced)
+                optimizer_metrics[metric] = torch.tensor(math.sqrt(reduced))
```
vchiley, line 401: type change ok here?

tests/test_model.py
```diff
@@ -703,35 +710,39 @@ def test_generate(attention_impl, device, alibi):
...
+    batched_input_ids = composer_device.tensor_to_device(batched_input_ids)
     batched_attention_mask = torch.tensor([[0, 0, 0, 1, 1, 1],
                                            [0, 0, 1, 1, 1, 1]]).bool()
-    batched_attention_mask = device.tensor_to_device(batched_attention_mask)
+    batched_attention_mask = composer_device.tensor_to_device(
+        batched_attention_mask)
 
-    with get_precision_context('amp_bf16' if device.name == 'gpu' else 'fp32'):
+    with get_precision_context('amp_bf16' if composer_device.name ==
+                               'gpu' else 'fp32'):
```
vchiley, line 745: since this isn't marked as a gpu test, `if device.name == 'gpu'` will always be False?

vchiley (approved):
bunch of ignorable comments
besides random nits, lgtm

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -92,7 +92,7 @@ def hf_get_hidden_layers(model: PreTrainedModel):
     return findattr(model, hidden_layers_attrs)
 
 
-def hf_get_init_device(init_device: str):
+def hf_get_init_device(init_device: Optional[str]):
```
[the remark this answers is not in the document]
dakinggg: Optional type actually just means the arg can be None, not that the argument is optional to specify.

llmfoundry/models/layers/attention.py
```diff
@@ -260,9 +266,13 @@ def triton_flash_attn_fn(
...
             raise RuntimeError(
                 'Requirements for `attn_impl: triton` not installed. Either (1) have a CUDA-compatible GPU '
+                +
                 'and `pip install .[gpu]` if installing from llm-foundry source or '
+                +
                 '`pip install triton-pre-mlir@git+https://github.com/vchiley/triton.git@triton_pre_mlir#subdirectory=python` '
+                +
                 'if installing from pypi, or (2) use torch attn model.attn_config.attn_impl=torch (torch attn_impl will be slow). '
+                +
                 'Note: (1) requires you have CMake and PyTorch already installed.'
```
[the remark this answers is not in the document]
dakinggg: Apparently

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -23,16 +23,18 @@
...
 from llmfoundry.models.layers.attention import attn_bias_shape, build_attn_bias
 from llmfoundry.models.layers.blocks import MPTBlock
 from llmfoundry.models.layers.custom_embedding import SharedEmbedding
-from llmfoundry.models.layers.fc import FC_CLASS_REGISTRY
-from llmfoundry.models.layers.ffn import FFN_CLASS_REGISTRY, MPTMLP, build_ffn
+from llmfoundry.models.layers.fc import FC_CLASS_REGISTRY as FC_CLASS_REGISTRY
+from llmfoundry.models.layers.ffn import \
+    FFN_CLASS_REGISTRY as FFN_CLASS_REGISTRY
+from llmfoundry.models.layers.ffn import MPTMLP as MPTMLP
+from llmfoundry.models.layers.ffn import build_ffn as build_ffn
```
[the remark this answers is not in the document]
dakinggg: Just so pyright doesn't think it's unused

tests/test_onnx.py
```diff
@@ -69,13 +71,14 @@ def test_onnx_export(tmp_path):
     with torch.no_grad():
         orig_out = mpt(**sample_input)
 
-    import onnx  # type: ignore
-    import onnx.checker  # type: ignore
-    import onnxruntime as ort  # type: ignore
+    import onnx
+    import onnx.checker
```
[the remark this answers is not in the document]
dakinggg: Yeah you would think so, but no. It complains about checker module not being exported if you don't do this.

llmfoundry/optim/lion.py
dakinggg: Yeah, they're all supposed to be tensors, this is the one exception. The code that uses it checks if it's a tensor and then calls .item if it is

setup.py
```diff
@@ -71,7 +71,7 @@
     'pytest>=7.2.1,<8',
     'pytest_codeblocks>=0.16.1,<0.17',
     'pytest-cov>=4,<5',
-    'pyright==1.1.296',
+    'pyright==1.1.256',
```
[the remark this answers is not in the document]
dakinggg: Someday a brave soul will upgrade pyright. Today is not that day.

llmfoundry/callbacks/generate_callback.py
```diff
@@ -16,7 +16,8 @@
 
 class Generate(Callback):
 
-    def __init__(self, prompts: List[str], batch_log_interval: int, **kwargs):
+    def __init__(self, prompts: List[str], batch_log_interval: int,
+                 **kwargs: Any):
```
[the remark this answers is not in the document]
dakinggg: counter: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -69,27 +70,40 @@ def __init__(self,
...
 
                 if self.weighting == Weighting.EQUAL:
                     weight = 1
                 elif self.weighting == Weighting.SAMPLE_SZ:
                     weight = cumulative_samples
                 elif self.weighting == Weighting.LOG_SAMPLE_SZ:
                     weight = max(math.log(cumulative_samples, 2), 1)
+                else:
+                    weight = None
+                    raise ValueError(
```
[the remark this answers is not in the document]
dakinggg: yeah, that works. this was just for pyright so ill just move it and use assert

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -119,21 +120,26 @@ def prepare_hf_model_for_fsdp(model: PreTrainedModel, init_device: str) -> None:
...
-                                        init_device: str) -> None:
+                                        init_device: Optional[str]) -> None:
     """FSDP wrap a HuggingFace decoder.
 
     Wrap any model for FSDP which follows one of the 3 existing conventions from
     HuggingFace for decoder-only LLMs.
     """
     causal_base_model = hf_get_causal_base_model(model)
+    if causal_base_model is None:
+        raise ValueError(
```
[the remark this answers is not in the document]
dakinggg: reasonable

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -141,6 +141,7 @@ def __init__(
         self.use_cache = use_cache
         self.init_config = init_config
         self.fc_type = fc_type
+        self.bias = None
```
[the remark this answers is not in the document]
dakinggg: oh yeah, something about something beiing undefined when it get set at the bottom of this file

tests/test_hf_v_mpt.py
```diff
@@ -40,7 +40,8 @@
     ('torch', 0.0, False, None, True),
     ('triton', 0.0, False, None, True),
 ])
-def test_compare_hf_v_mpt(attn_impl, dropout, alibi, mask_val, no_attn_mask):
+def test_compare_hf_v_mpt(attn_impl: str, dropout: float, alibi: bool,
+                          mask_val: Optional[int], no_attn_mask: bool):
```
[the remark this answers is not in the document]
dakinggg: I think its only along one code path that it has to not be none?

llmfoundry/callbacks/generate_callback.py
```diff
@@ -74,9 +80,11 @@ def generate(self, state: State, logger: Logger):
         dummy_input = device.tensor_to_device(dummy_input)
         with get_precision_context(state.precision):
             with torch.no_grad():
+                assert isinstance(model.model, torch.nn.Module)
                 _ = model.model(input_ids=dummy_input)
 
-            output_token_ids = model.model.generate(
+            assert hasattr(model.model, 'generate')
```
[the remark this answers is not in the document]
dakinggg: lets find out

tests/test_model.py
dakinggg: that seems right. isn't harming anything though so im just going to leave it

llmfoundry/callbacks/model_gauntlet_callback.py
dakinggg: https://mosaicml.atlassian.net/browse/RESEARCH-801

setup.py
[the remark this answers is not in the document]
dakinggg: (this change was just to match the version that is in precommit, and the version that composer uses)

PR #489, opened 2023-07-25
If Alibi is on, we should turn learned_pos_emb to False

bcui19:
If we use `alibi` we need to turn the `learned_pos_emb` to False. Otherwise we still go down the `learned_pos_emb` codepath.

### DIFF
--- a/llmfoundry/models/mpt/configuration_mpt.py
+++ b/llmfoundry/models/mpt/configuration_mpt.py
@@ -147,6 +147,8 @@ class MPTConfig(PretrainedConfig):
             del kwargs['loss_fn']
         if self.attn_config.get('alibi', False):
             self.learned_pos_emb = False
+            warnings.warn(
+                f'alibi is turned on, setting `learned_pos_emb` to `False.`')
         super().__init__(**kwargs)
 
         self._validate_config()
@@ -210,7 +212,7 @@ class MPTConfig(PretrainedConfig):
         if self.init_config.get('name', None) is None:
             raise ValueError(f"{self.init_config=} 'name' needs to be set.")
         if not self.learned_pos_emb and not self.attn_config['alibi']:
-            raise warnings.warn(
+            warnings.warn(
                 f'Positional information not being provided to the model using either learned_pos_emb or alibi.'
             )
         if self.fc_type == 'te' or self.ffn_config['ffn_type'] == 'te_ln_mlp':
--- a/llmfoundry/models/mpt/configuration_mpt.py
+++ b/llmfoundry/models/mpt/configuration_mpt.py
@@ -145,6 +145,8 @@ class MPTConfig(PretrainedConfig):
             del kwargs['name']
         if 'loss_fn' in kwargs:
             del kwargs['loss_fn']
+        if self.attn_config.get('alibi', False):
+            self.learned_pos_emb = False
         super().__init__(**kwargs)
 
         self._validate_config()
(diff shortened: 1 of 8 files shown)


vchiley:
Similarly the argument could be made that `if alibi == False` set `learned_pos_emb  = True` BUT I figured `alibi = False` and `learned_pos_emb  = False` is `NoPE` so I left the option open.
Similarly, I wanted to give ppl the ability to use both `alibi`  and `learned_pos_emb`

maybe add a warning or something???
maybe just ignore me... 😅

bcui19:
Hm... Okay, so one thing is if we train models w/ ALiBi (without `learned_pos_embed`) and load that model with `learned_pos_embed=True` it could really affect model performance (since now you have this random position embedding...)

I also am not 100% sure what's the right thing. The "safest" thing might be to add a warning that we're changing it?

vchiley:
if we load a trained model, should the model config be part of the ckpt?

samhavens:
would a good middle ground being changing the default value so that `learned_pos_emb: bool = False,`? Alternatively, we can default `learned_pos_emb` to `None`, and then later change the `None` to `learned_pos_emb = not alibi`

bcui19:
@vchiley It might be good to independently save it from composer, since also the composer checkpoint is wildly unwieldy.

@samhavens I agree, so I did sth like this:

```
        if self.learned_pos_emb is None:
            self.learned_pos_emb = not self.attn_config.get('alibi', False)
            warnings.warn(f'`learned_pos_emb` has not been set, setting it to {self.learned_pos_emb}')
```
And it leads to this super not well understood error... ): Let me know if I'm doing something obviously wrong (I think there's something weird where in transformers where the class gets initialized twice...:
https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L816

And we also can't set this up in `_validate_config` since we need it if we were to create a model using `from_pretrained`.

image

So I think unfortunately, I think I'm just going to add an additional warning, unless someone else has a way around this.

vchiley:
@bcui19 it sets it to false, then sets it to true?

bcui19:
I think... When you create MPTConfig the way we do it now... I think HF does something weird and it ends up creating two `MPTConfig`'s...

samhavens:
If HF is going to make us choose between extremely cryptic warnings like that or just "set pos_emb to false if alibi is true" then I guess we should revert to what you had before and just let the user know they specified alibi so we are setting pos emb to false.

Unless, does setting the default differently solve this? `learned_pos_emb: bool = False` ?

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -208,7 +212,7 @@ def _validate_config(self):
         if self.init_config.get('name', None) is None:
             raise ValueError(f"{self.init_config=} 'name' needs to be set.")
         if not self.learned_pos_emb and not self.attn_config['alibi']:
-            raise warnings.warn(
+            warnings.warn(
```
vchiley, line 215: Having neither be set is a valid config of NoPE

bcui19: Yeah, if you choose to `raise` it leads to an error (so you can't actually use NoPE):

image

vchiley: I'm an idiot

bcui19: haha naw all good

vchiley (approved):

bcui19:
Running w/ the current set of changes:

image

Going to merge. This is partly me assuming that very few people would want to mix positional embeddings w/ Alibi

PR #492, opened 2023-07-27
Grouped Query Attention + Refactor Attn

sashaDoubov:
Adding grouped query attention to LLM-foundry, and making the `GQA` class a superclass of MQA and MHA attention, as it is a generalization of those two variants of attention. 
Things to note:
- we currently use `repeat_interleave` to make the grouped query tensor the same dimensions as multi-head attention, which does allocate new memory, compared to using `expand`. This can be updated in the future, but is the safer bet for now, given that we previously saw edge-cases with using `expand` vs `repeat` for particular `head_dim` settings causing NaNs
- Part of this change is also changing how we initialize the QKV matrix, which used to initialize (n_heads * head_dim, d_model), but now is changed to initialize each n_head (head_dim, d_model) matrix separately

### DIFF
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -36,7 +36,7 @@ def scaled_multihead_dot_product_attention(
     key: torch.Tensor,
     value: torch.Tensor,
     n_heads: int,
-    kv_n_heads: int,
+    kv_n_heads: Optional[int] = None,
     past_key_value: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
     softmax_scale: Optional[float] = None,
     attn_bias: Optional[torch.Tensor] = None,
@@ -55,6 +55,12 @@ def scaled_multihead_dot_product_attention(
                 'The direct use of the multiquery arg is deprecated. Setting kv_n_heads=1 automatically. Please set kv_n_heads=1 explicitly to remove this warning.'
             ))
         kv_n_heads = 1
+    elif kv_n_heads is None:
+        warnings.warn(
+            DeprecationWarning(
+                'Not specifying a value for the kv_n_heads arg is deprecated. Setting kv_n_heads=n_heads automatically. Please set kv_n_heads=n_heads explicitly to remove this warning.'
+            ))
+        kv_n_heads = n_heads
 
     q = rearrange(query, 'b s (h d) -> b h s d', h=n_heads)
     k = rearrange(key, 'b s (h d) -> b h d s', h=kv_n_heads)
@@ -156,7 +162,7 @@ def flash_attn_fn(
     key: torch.Tensor,
     value: torch.Tensor,
     n_heads: int,
-    kv_n_heads: int,
+    kv_n_heads: Optional[int] = None,
     past_key_value: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
     softmax_scale: Optional[float] = None,
     attn_bias: Optional[torch.Tensor] = None,
@@ -181,6 +187,12 @@ def flash_attn_fn(
                 'The direct use of the multiquery arg is deprecated. Setting kv_n_heads=1 automatically. Please set kv_n_heads=1 explicitly to remove this warning.'
             ))
         kv_n_heads = 1
+    elif kv_n_heads is None:
+        warnings.warn(
+            DeprecationWarning(
+                'Not specifying a value for the kv_n_heads arg is deprecated. Setting kv_n_heads=n_heads automatically. Please set kv_n_heads=n_heads explicitly to remove this warning.'
+            ))
+        kv_n_heads = n_heads
 
     if past_key_value is not None:
         if len(past_key_value) != 0:
@@ -264,7 +276,7 @@ def triton_flash_attn_fn(
     key: torch.Tensor,
     value: torch.Tensor,
     n_heads: int,
-    kv_n_heads: int,
+    kv_n_heads: Optional[int] = None,
     past_key_value: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
     softmax_scale: Optional[float] = None,
     attn_bias: Optional[torch.Tensor] = None,
@@ -311,6 +323,12 @@ def triton_flash_attn_fn(
                 'The direct use of the multiquery arg is deprecated. Setting kv_n_heads=1 automatically. Please set kv_n_heads=1 explicitly to remove this warning.'
             ))
         kv_n_heads = 1
+    elif kv_n_heads is None:
+        warnings.warn(
+            DeprecationWarning(
+                'Not specifying a value for the kv_n_heads arg is deprecated. Setting kv_n_heads=n_heads automatically. Please set kv_n_heads=n_heads explicitly to remove this warning.'
+            ))
+        kv_n_heads = n_heads
 
     if past_key_value is not None:
         if len(past_key_value) != 0:
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -462,12 +462,12 @@ class GroupedQueryAttention(nn.Module):
             qkv = qkv.clamp(min=-self.clip_qkv, max=self.clip_qkv)
 
         query, key, value = qkv.split(
-        [
-            self.d_model, 
-            self.kv_n_heads * self.head_dim,
-            self.kv_n_heads * self.head_dim,
-        ],
-        dim=2,
+            [
+                self.d_model,
+                self.kv_n_heads * self.head_dim,
+                self.kv_n_heads * self.head_dim,
+            ],
+            dim=2,
         )
 
         key_padding_mask = attention_mask
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -68,8 +68,6 @@ def scaled_multihead_dot_product_attention(
 
     # grouped query case
     if kv_n_heads > 1 and kv_n_heads < n_heads:
-        # note that the interleave happens at dim 1,
-        # unlike the specialized kernels for flash attention and triton
         k = k.repeat_interleave(n_heads // kv_n_heads, dim=1)
         v = v.repeat_interleave(n_heads // kv_n_heads, dim=1)
 
@@ -211,8 +209,10 @@ def flash_attn_fn(
     elif kv_n_heads < n_heads:
         # Each query belong to a group of kv heads of group size n_heads // kv_n_heads
         # We repeat each kv head by the group size number to use use the underlying MHA kernels
-        key = key.repeat_interleave(n_heads // kv_n_heads, dim=2)
-        value = value.repeat_interleave(n_heads // kv_n_heads, dim=2)
+        # done along the head dimension = 1
+        key_unpad = key_unpad.repeat_interleave(n_heads // kv_n_heads, dim=1)
+        value_unpad = value_unpad.repeat_interleave(n_heads // kv_n_heads,
+                                                    dim=1)
 
     dropout_p = dropout_p if training else 0.0
 
@@ -329,6 +329,7 @@ def triton_flash_attn_fn(
     elif kv_n_heads < n_heads:
         # Each query belong to a group of kv heads of group size n_heads // kv_n_heads
         # We repeat each kv head by the group size number to use use the underlying MHA kernels
+        # done along dim = 2, unlike the implementation for flash and torch attn
         key = key.repeat_interleave(n_heads // kv_n_heads, dim=2)
         value = value.repeat_interleave(n_heads // kv_n_heads, dim=2)
 
@@ -375,6 +376,7 @@ class GeneralizedAttention(nn.Module):
 
         self.head_dim = d_model // n_heads
 
+        assert self.kv_n_heads > 0, 'kv_n_heads should be greater than zero'
         assert self.kv_n_heads <= self.n_heads, 'The number of KV heads should be less than or equal to Q heads'
         assert self.n_heads % self.kv_n_heads == 0, 'Each Q head should get the same number of KV heads, so n_heads must be divisible by kv_n_heads'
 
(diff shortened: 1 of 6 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -76,11 +77,21 @@ def __init__(self, config: MPTConfig):
...
 
-        self.attn_impl = config.attn_config['attn_impl']
-        self.prefix_lm = config.attn_config['prefix_lm']
-        self.attn_uses_sequence_id = config.attn_config['attn_uses_sequence_id']
-        self.alibi = config.attn_config['alibi']
-        self.alibi_bias_max = config.attn_config['alibi_bias_max']
+        # necessary to copy in order to be able to pop arguments so that we can pass in
+        # a subset of arguments into the Attention() init function as **kwargs
+        # and remove unnecessary args and still support checkpointing the original att_config
+        config.attn_config_mutable = copy.deepcopy(config.attn_config)
```
sashaDoubov, line 83: This was a workaround to make `**attn_config` work, because certain arguments like `attn_type, prefix_lm, alibi` are not used in `MultiHeadAttention()`, but we want to be able to use `**attn_config`. Other possible workarounds:

```
1. .pop() when using these arguments (but I wasn't sure if this messes up checkpoints?)
2. **kwargs in the init arguments (don't love this as you know)
3. adding some of these arguments into the __init__ function of Attention
4. removing these arguments manually before passing into the init function ie.  attn_config_subset = {key: value for key, value in att_config.items() if key not in [attn_type, prefix_lm, alibi]}
```

llmfoundry/models/layers/attention.py
```diff
@@ -377,17 +401,21 @@ def __init__(
...
-            3 * self.d_model,
+            self.d_model + 2 * self.kv_n_heads * self.head_dim,
             **fc_kwargs,
         )
         # for param init fn; enables shape based init of fused layers
-        fuse_splits = (d_model, 2 * d_model)
+        fuse_splits = [
+            i * self.head_dim
+            for i in range(1, self.n_heads + 2 * self.kv_n_heads)
+        ]
```
sashaDoubov, line 411: I've sanity checked this, but confirming this init

vchiley: lgtm

llmfoundry/models/layers/attention.py
```diff
@@ -432,7 +460,11 @@ def forward(
         if self.clip_qkv:
             qkv = qkv.clamp(min=-self.clip_qkv, max=self.clip_qkv)
 
-        query, key, value = qkv.chunk(3, dim=2)
+        query, key, value = qkv.split([
+            self.d_model, self.kv_n_heads * self.head_dim,
+            self.kv_n_heads * self.head_dim
+        ],
+                                      dim=2)
```
vchiley, line 467:
```suggestion
        query, key, value = qkv.split(
        [
            self.d_model, 
            self.kv_n_heads * self.head_dim,
            self.kv_n_heads * self.head_dim,
        ],
        dim=2,
        )
```

Does this version still pass lint?
(the version you have just looks weird)

llmfoundry/models/layers/attention.py
vchiley, line 1: love how cleanly this worked out

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -77,7 +77,7 @@ def __init__(
             emb_pdrop (float): The dropout probability for the embedding layer.
             learned_pos_emb (bool): Whether to use learned positional embeddings
             attn_config (Dict): A dictionary used to configure the model's attention module:
-                attn_type (str): type of attention to use. Options: multihead_attention, multiquery_attention
+                attn_type (str): type of attention to use. Options: multihead_attention, multiquery_attention, grouped_query_attention
```
vchiley, line 80: add something like: `kv_n_heads (int): number of heads in grouped_query_attention if using grouped_query_attention`

tests/test_model.py
```diff
@@ -373,7 +373,7 @@ def test_loss_fn():
     assert isinstance(test_cfg, DictConfig)
 
     test_cfg.device = 'cuda:0'
-    test_cfg.model.init_device = 'cuda:0'
+    test_cfg.model.init_device = 'cpu'
```
vchiley, line 376: is this to help the test be less flakey?

vchiley (commented):
`GeneralizedAttention` is a valid name, but should be call it `GroupedQueryAttention` to not confuse ppl?

or alternatively create
```
class GroupedQueryAttention(GeneralizedAttention):
    def __init__(..., groups=G, ...):
        assert n_heads / G == 0
        super().__init__(..., kv_n_heads=n_heads//G, ...)
```

or something like that

tests/test_model.py
sashaDoubov: yea, otherwise, with the new `init` change, it fails after fewer steps (ex. 3). I don't have a great sense of why that is,  but seems like the new version may introduce more/different non-determinism for this particular test.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -76,11 +77,21 @@ def __init__(self, config: MPTConfig):
...
+
+        # this is an argument into attention init, so should NOT be popped
+        self.attn_impl = config.attn_config_mutable['attn_impl']
```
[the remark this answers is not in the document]
sashaDoubov: yes, `self.attn_impl` is used directly a few times, notably in `_attn_bias`

llmfoundry/models/layers/attention.py
```diff
@@ -432,7 +461,14 @@ def forward(
         if self.clip_qkv:
             qkv = qkv.clamp(min=-self.clip_qkv, max=self.clip_qkv)
 
-        query, key, value = qkv.chunk(3, dim=2)
+        query, key, value = qkv.split(
+            [
```
sashaDoubov, line 465: @vchiley looks like this is how the pre-commit formatting makes it look

llmfoundry/models/layers/blocks.py
```diff
@@ -52,23 +52,18 @@ def __init__(
         super().__init__()
 
         norm_class = NORM_CLASS_REGISTRY[norm_type.lower()]
-        assert isinstance(attn_config['attn_type'], str)
-        attn_class = ATTN_CLASS_REGISTRY[attn_config['attn_type']]
+
+        attn_type = attn_config_mutable.pop('attn_type')
```
dakinggg, line 56: I don't love the mutable config thing. What do you think about the option where inside this function you just make a copy of the attn_config with the extraneous args removed? Then the caller of the block creation does not need to do any popping, and this function doesn't have side effects on the original attn config.

dakinggg (commented):
Could you please include a comparison run for a normal mha model before and after this pr? I'd like to make sure we don't have a perf regression (or at least know the quantity of it)

llmfoundry/models/layers/attention.py
```diff
@@ -44,11 +45,9 @@ def scaled_multihead_dot_product_attention(
     dropout_p: float = 0.0,
     training: bool = False,
     needs_weights: bool = False,
-    multiquery: bool = False,
```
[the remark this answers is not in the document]
sashaDoubov: I think the underlying assumption was that we can't break the api to the `MultiHeadAttention` class, but that for these helper functions it is okay to change the arguments. Thoughts?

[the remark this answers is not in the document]
dakinggg: yeah, i would be ok with it if there was a need, but I think its easy enough to not break that we should deprecate. technically these are public apis, so better to not break, and to generally get into that habit for this repo.

llmfoundry/models/layers/blocks.py
sashaDoubov: yea I think that might be cleaner. I'll change the implementation accordingly.

llmfoundry/models/layers/attention.py
```diff
@@ -367,6 +384,14 @@ def __init__(
 
         self.d_model = d_model
         self.n_heads = n_heads
+        self.kv_n_heads = kv_n_heads
+
+        self.head_dim = d_model // n_heads
+
+        assert self.kv_n_heads > 0, 'kv_n_heads should be greater than zero'
```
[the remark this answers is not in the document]
sashaDoubov: done

llmfoundry/models/layers/attention.py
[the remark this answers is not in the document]
sashaDoubov: okay, I have addressed this

llmfoundry/models/layers/blocks.py
```diff
@@ -29,8 +29,8 @@ def __init__(
         device: Optional[str] = None,
         **kwargs: Any,
     ):
-        if attn_config is None:
-            attn_config = {
+        if attn_config_mutable is None:
```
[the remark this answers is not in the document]
sashaDoubov: hopefully no longer relevant

llmfoundry/models/layers/attention.py
```diff
@@ -47,8 +48,15 @@ def scaled_multihead_dot_product_attention(
     multiquery: bool = False,
 ) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[Tuple[torch.Tensor,
                                                                 torch.Tensor]]]:
+
+    if multiquery:
+        warnings.warn(
+            DeprecationWarning(
+                'The direct use of the multiquery arg is deprecated. Set kv_n_heads=1 instead.'
```
dakinggg, line 55:
```suggestion
                'The direct use of the multiquery arg is deprecated. Setting kv_n_heads=1 automatically. Please set kv_n_heads=1 explicitly to remove this warning.'
```

llmfoundry/models/layers/attention.py
```diff
@@ -161,6 +175,13 @@ def flash_attn_fn(
 
     check_valid_inputs(query, key, value)
 
+    if multiquery:
+        warnings.warn(
+            DeprecationWarning(
+                'The direct use of the multiquery arg is deprecated. Set kv_n_heads=1 instead.'
```
dakinggg, line 181:
```suggestion
                'The direct use of the multiquery arg is deprecated. Setting kv_n_heads=1 automatically. Please set kv_n_heads=1 explicitly to remove this warning.'
```

llmfoundry/models/layers/attention.py
```diff
@@ -278,6 +305,13 @@ def triton_flash_attn_fn(
...
```
dakinggg, line 311:
```suggestion
                'The direct use of the multiquery arg is deprecated. Setting kv_n_heads=1 automatically. Please set kv_n_heads=1 explicitly to remove this warning.'
```

llmfoundry/models/layers/attention.py
```diff
@@ -367,6 +411,23 @@ def __init__(
 
         self.d_model = d_model
         self.n_heads = n_heads
+        self.kv_n_heads = kv_n_heads
+
+        self.head_dim = d_model // n_heads
+
+        if self.kv_n_heads <= 0:
+            raise ValueError('kv_n_heads should be greater than zero')
```
dakinggg, line 419:
```suggestion
            raise ValueError('kv_n_heads should be greater than zero.')
```

llmfoundry/models/layers/attention.py
```diff
@@ -367,6 +411,23 @@ def __init__(
...
+
+        if self.kv_n_heads > self.n_heads:
+            raise ValueError(
+                'The number of KV heads should be less than or equal to Q heads'
```
dakinggg, line 423:
```suggestion
                'The number of KV heads should be less than or equal to Q heads.'
```

llmfoundry/models/layers/attention.py
```diff
@@ -367,6 +411,23 @@ def __init__(
...
+            )
+
+        if self.n_heads % self.kv_n_heads != 0:
+            raise ValueError(
+                'Each Q head should get the same number of KV heads, so n_heads must be divisible by kv_n_heads'
```
dakinggg, line 428:
```suggestion
                'Each Q head should get the same number of KV heads, so n_heads must be divisible by kv_n_heads.'
```

llmfoundry/models/layers/blocks.py
```diff
@@ -51,24 +51,30 @@ def __init__(
         del kwargs  # unused, just to capture any extra args from the config
         super().__init__()
 
+        norm_class = NORM_CLASS_REGISTRY[norm_type.lower()]
```
dakinggg, line 54: accidentally added duplicate line?

llmfoundry/models/layers/blocks.py
```diff
@@ -51,24 +51,30 @@ def __init__(
...
+
         norm_class = NORM_CLASS_REGISTRY[norm_type.lower()]
         assert isinstance(attn_config['attn_type'], str)
         attn_class = ATTN_CLASS_REGISTRY[attn_config['attn_type']]
 
+        # necessary to avoid passing extraneous args into attn_class while allowing the use of **kwargs
+        args_to_exclude_in_attn_class = [
```
dakinggg, line 61: convert to set instead of list

dakinggg (commented):
few nits, but lgtm pending the performance comparison

sashaDoubov:
> Could you please include a comparison run for a normal mha model before and after this pr? I'd like to make sure we don't have a perf regression (or at least know the quantity of it)

Here are some perf numbers, which look good outside of the loss spikes. However, I did see loss spikes with both "before and after" the change:

image

image

image

llmfoundry/models/layers/blocks.py
sashaDoubov: done

llmfoundry/models/layers/blocks.py
sashaDoubov: good catch, done.

dakinggg (approved):

llmfoundry/models/mpt/modeling_mpt.py
sashaDoubov: addressed

PR #495, opened 2023-07-31
Fix Local World Size

rishab-partha:
Make LOCAL_WORLD_SIZE set in pytest to fix non-dist tests


dakinggg:
same comment as composer pr

rishab-partha:
fixed!

dakinggg (approved):
Will give a couple hours for @mvpatel2000 to potentially object, then will merge to unblock CI

PR #500, opened 2023-08-02
fix boolean for reentrant setting

dakinggg:
run with this fixed, used to hang: https://wandb.ai/mosaic-ml/daniel-debug/runs/b0spb0xb?workspace=user-danielking


vchiley (approved):
lgtm

PR #501, opened 2023-08-02
Enable gauntlet training

bmosaicml:
This PR enables us to run the gauntlet during training. It also enables you to set a subset of batches to run eval over in order to speed up ICL during training.

It also changes eval to no longer use the InMemoryLogger and instead pull the metrics from the State.

Eval test run:  `all-eval-F2oe1D` 
Training test run: `test-1b-C5QG5k` `test-1b-c7sqCh `

https://wandb.ai/mosaic-ml/gauntlet/runs/rrbn0qbb?workspace=user-jemdohmann

image

Eval results still good:

```
| model_name               |   average |   world_knowledge |   commonsense_reasoning |   language_understanding |   symbolic_problem_solving |   reading_comprehension |
|:-------------------------|----------:|------------------:|------------------------:|-------------------------:|---------------------------:|------------------------:|
| mosaicml/mpt-7b-instruct |  0.354255 |          0.398764 |                0.415097 |                 0.371509 |                   0.171216 |                0.414691 |
```


llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -120,8 +126,11 @@ def compute_averages(self, logger_data: Logger):
                     results[key] = [val]
         return {k: sum(v) / len(v) for k, v in results.items()}
 
-    def eval_end(self, state: State, logger: Logger):
-        new_metrics = self.compute_averages(logger)
+    def eval_after_all(self, state: Optional[State], logger: Logger):
+        inmemorylogger = get_in_memory_logger(logger)
```
abhi-databricks, line 130: @bmosaicml what's the reason we need to grep the logs from the InMemoryLogger rather than read the eval metrics out of the state directly?

scripts/eval/yamls/mpt_eval.yaml
```diff
@@ -24,14 +24,16 @@ models:
     attn_config:
       attn_impl: triton
 
-load_path: # Add your (optional) Composer checkpoint path here!
+  load_path: # Add your (optional) Composer checkpoint path here!
```
dakinggg, line 27: this indentation seems wrong?

scripts/eval/yamls/hf_eval.yaml
```diff
@@ -37,9 +37,11 @@ models:
 device_eval_batch_size: 4
 
 # FSDP config for model sharding
-fsdp_config:
-  sharding_strategy: FULL_SHARD
-  mixed_precision: FULL
+# fsdp_config:
```
dakinggg, line 40: should this be commented out?

scripts/eval/yamls/hf_eval.yaml
```diff
@@ -37,9 +37,11 @@ models:
...
+#   sharding_strategy: FULL_SHARD
+#   mixed_precision: FULL
+#   forward_prefetch: True
+#   limit_all_gathers: True
 
-icl_tasks: 'eval/yamls/tasks_light.yaml'
+icl_tasks: 'eval/yamls/tasks.yaml'
```
dakinggg, line 46: is this change intentional?

scripts/eval/eval.py
```diff
@@ -131,23 +135,21 @@ def main(cfg: DictConfig):
...
 
         benchmark_to_taxonomy = {}
         if model_gauntlet is not None:
             for t in model_gauntlet.categories:
                 for b in t.benchmarks:
                     benchmark_to_taxonomy[b.name] = t.name
 
-        model_results = calculate_markdown_results(logger_keys,
-                                                   in_memory_logger.data,
+        model_results = calculate_markdown_results(logger_keys, logger,
```
dakinggg, line 152: why are we using markdown results? This seems like an odd choice for outputting structured data

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -90,18 +98,16 @@ def __init__(self,
...
         pat = re.compile(
             'metrics/(.*?)/(\d+)-shot(/.*?)?/InContextLearning(.*)'  # type: ignore
         )
         for key in self.logger_keys:
             match = pat.match(key)
-
-            # TODO(bmosaicml) This needs to be factored for this callback to work as a normal callback
-            # and therefore for the typing to be fixed
-            val = logger_data.data[key][0][1].item()  # type: ignore
+            if key not in logger_destination.data:
```
dakinggg, line 108: when would this condition happen?

scripts/eval/eval.py
```diff
@@ -131,23 +135,21 @@ def main(cfg: DictConfig):
...
-        (in_memory_logger, logger_keys, model_gauntlet_callback, model_gauntlet,
+        (logger, logger_keys, model_gauntlet_callback, model_gauntlet,
          model_gauntlet_df) = evaluate_model(model_cfg, cfg, cfg.run_name,
                                              model_gauntlet_df)
 
         if model_gauntlet_callback is not None:
-            # TODO(bmosaicml) This needs to be refactored to fix the typing issue
-            composite_scores = model_gauntlet_callback.eval_end(
-                None, in_memory_logger)  # type: ignore
+            composite_scores = model_gauntlet_callback.eval_after_all(
```
dakinggg, line 143: shouldn't the `eval_after_all` just write some data to the `model_gauntlet_callback` (or to a file if you prefer)? Then you don't need to call the event explicitly, as it should just run on the event as intended after eval.

llmfoundry/utils/builders.py
```diff
@@ -174,7 +181,7 @@ def build_icl_evaluators(icl_tasks: Union[str, ListConfig],
     else:
         icl_tasks_list = icl_tasks
 
-    def _validate_cfg(icl_cfg: DictConfig):
+    def _validate_cfg(icl_cfg: Any):
```
dakinggg, line 184: this shouldn't be type `Any`, it should be a `DictConfig`

dakinggg: you probably just need to move the assert below above the call to this function

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -20,6 +21,13 @@ class Weighting(Enum):
     LOG_SAMPLE_SZ = 3
 
 
+def get_in_memory_logger(logger: Logger):
+    for lg in logger.destinations:
+        if isinstance(lg, InMemoryLogger):
+            return lg
+    raise Exception("Couldn't find InMemoryLogger in logger destinations!")
```
hanlint, line 28: `Exception` should almost never be used. In thise case, you should throw a `ValueError`.

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -90,18 +98,16 @@ def __init__(self,
...
         )
         for key in self.logger_keys:
             match = pat.match(key)
-
-            # TODO(bmosaicml) This needs to be factored for this callback to work as a normal callback
-            # and therefore for the typing to be fixed
-            val = logger_data.data[key][0][1].item()  # type: ignore
+            if key not in logger_destination.data:
+                continue
+            val = logger_destination.data[key][-1][1].item()
```
hanlint, line 110: was the index change from `[0]` -> `[-1]` intentional here?

llmfoundry/callbacks/model_gauntlet_callback.py
hanlint: +1 this is not very robust and strongly prefer a better way here.

mcli/mcli-hf-eval.yaml
```diff
@@ -1,7 +1,7 @@
 integrations:
 - integration_type: git_repo
   git_repo: mosaicml/llm-foundry
-  git_branch: v0.2.0
+  git_branch: # v0.2.0
```
hanlint, line 4: We need a tagged branch here, can't be tip of dev as that is too unstable.

mcli/mcli-1b-gauntlet.yaml
```diff
@@ -0,0 +1,189 @@
...
+  pip_install: -e .[gpu]
+  ssh_clone: false
+-
+  integration_type: wandb
+  entity: mosaic-ml
+  project: gauntlet
+
+compute:
+  gpus: 8 # <insert number of gpus>  # update for hero run
+  cluster: r1z1 # <insert cluster name> # update for hero run
```
hanlint, line 17: Please remove specific internal cluster names

mcli/mcli-1b-gauntlet.yaml
```diff
@@ -0,0 +1,189 @@
...
+-
+  integration_type: git_repo
+  git_branch: v0.2.0
+  git_repo: mosaicml/llm-foundry
+  pip_install: -e .[gpu]
+  ssh_clone: false
+-
+  integration_type: wandb
+  entity: mosaic-ml
+  project: gauntlet
```
hanlint, line 13: our yaml examples should not have specific mosaicml setups.

mcli/mcli-1b-gauntlet.yaml
```diff
@@ -0,0 +1,189 @@
...
+  console_log_interval: 100ba
+  progress_bar: false
+  python_log_level: DEBUG
+  loggers:
+    wandb: {}
+
+  # Checkpointing
+  # autoresume: true
+  # save_filename: ep{epoch}-ba{batch}/rank{rank}.pt
+  # save_folder: oci://mosaicml-internal-checkpoints/llm-hero-test/7-6-23/mpt-1b # update for hero run
```
hanlint, line 101: please remove hardcoded values

mcli/mcli-1b-gauntlet.yaml
```diff
@@ -0,0 +1,189 @@
...
+      eos_token_id: ${eos_token_id}
+      max_seq_len: ${max_seq_len}
+      num_canonical_nodes: ${num_canonical_nodes}
+      shuffle: true
+      shuffle_seed: ${global_seed}
+      streams:
+        c4:
+          local: /tmp/dataset/c4/
+          proportion: 0.8
+          remote: oci://mosaicml-internal-dataset-c4/sem-dedupe/23-04-05/dataset/pretokencat/EleutherAI-gpt-neox-20b/0pt8/
```
hanlint, line 183: remove

mcli/mcli-1b-gauntlet.yaml
```diff
@@ -0,0 +1,189 @@
...
+          split: train
+        markdown:
+          local: /tmp/dataset/markdown/
+          proportion: 0.2
+          remote: oci://mosaicml-internal-datasets/stack-split-neox/markdown/
```
hanlint, line 188: remove

mcli/mcli-hf-eval.yaml
```diff
@@ -41,250 +41,14 @@ parameters:
...
-  #     pretrained: true
-  #     use_auth_token: false
-
 
   # FSDP config for model sharding
   fsdp_config:
     sharding_strategy: FULL_SHARD
     mixed_precision: FULL
+    forward_prefetch: True
+    limit_all_gathers: True
```
hanlint, line 50: why are these needed?

scripts/eval/yamls/mpt_eval.yaml
```diff
@@ -1,4 +1,4 @@
-max_seq_len: 2048
+max_seq_len: 1024
```
hanlint, line 1: why this change?

scripts/eval/yamls/mpt_eval.yaml
```diff
@@ -24,14 +24,16 @@ models:
...
-load_path: # Add your (optional) Composer checkpoint path here!
+  load_path: # Add your (optional) Composer checkpoint path here!
 
 device_eval_batch_size: 16
 
 # FSDP config for model sharding
 fsdp_config:
   sharding_strategy: FULL_SHARD
   mixed_precision: FULL
+  forward_prefetch: True
```
hanlint, line 35: Why are these changed?

scripts/train/train.py
```diff
@@ -240,11 +242,27 @@ def main(cfg: DictConfig):
...
+            cfg.max_seq_len,
+            cfg.device_eval_batch_size,
+            icl_subset_num_batches=cfg.get('icl_subset_num_batches', None))
         evaluators.extend(icl_evaluators)
+        if 'model_gauntlet' in cfg:
+            if isinstance(cfg.model_gauntlet, str):
+                with open(cfg.model_gauntlet, 'r') as icl_f:
+                    model_gauntlet_cfg = om.load(icl_f)
+                model_gauntlet = model_gauntlet_cfg.model_gauntlet
+            else:
```
hanlint, line 259: Shouldn't there be a different type check here?

scripts/train/train.py
```diff
@@ -259,11 +277,16 @@ def main(cfg: DictConfig):
...
+    callbacks: List[Callback] = [
         build_callback(name, callback_cfg)
         for name, callback_cfg in (cfg.get('callbacks') or {}).items()
     ]
 
+    if model_gauntlet_callback is not None:
+        callbacks.append(model_gauntlet_callback)
+        if not any(isinstance(l, InMemoryLogger) for l in loggers):
+            loggers.append(build_logger('in_memory_logger', {}))
+
```
hanlint, line 289: we should never silently add an unwanted callback. suggest we find a different way to pass data, rather than rely on a logger destination.

hanlint (changes requested):
Thanks for putting this PR together, left a few comments:
* the new yaml files have unexplained changes (batch size, fsdp config), and also have several hardcoded internal OCI bucket names and cluster names. Please remove the hardcoded values prior to pushing PRs.
* can we avoid silently adding in an InMemoryLogger, purely for the purposes of passing around data? Seems like there's a deeper design issue if we have to resort to this approach.

llmfoundry/callbacks/model_gauntlet_callback.py
bmosaicml: yes, since logging events to the same key append to a list, in order to retrieve the most recent logging we need to use index -1

llmfoundry/callbacks/model_gauntlet_callback.py
eracah: I think you should use `state.eval_metric_values` here. It should contain a dictionary of all recently computed metrics, keyed by metric name. The one issue is it is not keyed by dataloader name, so if there are multiple dataloaders with the same metric there will be a collision, but we can patch in that change to enable this PR

llmfoundry/callbacks/model_gauntlet_callback.py
eracah: `state.eval_metric_values` will contain the most recent eval metrics

scripts/train/train.py
eracah: I thiiink `state.eval_metric_values` should work here

eracah (commented):
Might be worth using state to get the metrics instead of `inmemorylogger`

mcli/mcli-hf-eval.yaml
bmosaicml: Brandon found that it made FSDP + inference faster to use these flags

scripts/eval/yamls/mpt_eval.yaml
bmosaicml: this is the standard length used for our gauntlet evals

scripts/eval/yamls/hf_eval.yaml
bmosaicml: yeah I want all the configs to reflect the standard gauntlet usage

scripts/eval/eval.py
bmosaicml: I'm open to other suggestions. What did you have in mind?

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -27,7 +26,7 @@ class ModelGauntlet(Callback):
     specification provided in the constructor.
 
     Args:
-        logger_keys (dict): These are the exact keys that the individual benchmark metrics will be
```
hanlint, line 30: This is API breaking -- is this an absolutely neccessary change, and if so, can we have backwards compatibility with a deprecation warning?

tests/test_training.py
```diff
@@ -88,3 +88,27 @@ def test_train(device: str):
...
+            'icl_task_type':
+                'language_modeling'
+        })
+    ])
+    test_cfg.icl_subset_num_batches = 1  # -1 to evaluate on all batches
+    test_cfg.model_gauntlet = 'eval/yamls/model_gauntlet.yaml'
+    test_cfg.icl_seq_len = 128
+    test_cfg.max_duration = '1ba'
+    test_cfg.eval_interval = '1ba'
+    main(test_cfg)
```
hanlint, line 114: tests need some sort of assert or check

mcli/mcli-hf-eval.yaml
hanlint: I don't think this was resolved?

bmosaicml: Isn't v.0.2.0 stable? We can also just comment it out entirely and let users put whatever tag/branch they want

llmfoundry/callbacks/model_gauntlet_callback.py
bmosaicml: It's not strictly necessary, so I can change it back!

tests/test_training.py
bmosaicml: I think ideally we would check that the ModelGauntlet was constructed and its eval_after_all method was called but I am not if it's possible to Mock that if the object is constructed within Main. Fwiw, the other unit test in this method has no asserts and simply checks whether the code fails

dakinggg: Yeah, these tests are really "does it crash". For the gauntlet, there should separately be unit tests testing the basic functionality.

bmosaicml: I think it is important to test the full workflow of running train and seeing if it logs properly. 

I am trying to unit test this by using console logger but I think the definition of console logger's fit_end and eval_end methods preclude the gauntlet callback from working properly. This is because the gauntlet callback runs at eval_after_all. As it currently works, the console logger runs at eval_end and at fit end, but anything logged between eval_end and fit_end (e.g. during eval_after_all) will just sit in purgatory and not get flushed.

Lmk how you think I should proceed.

mcli/mcli-hf-eval.yaml
hanlint: Yes `v0.2.0` is stable, but this diff comments it out, which means that this yaml defaults to using the tip of `dev`. I think this is OK, but we should update this to a version upon the next release. Defer to @dakinggg here.

dakinggg: My preference is to have the MCLI yamls remain pinned to releases and then update them when we do a release. If we really _need_ to break this so we can give the yaml to a customer, it is ok. But the yaml should pin to a specific commit that works (e.g. https://github.com/mosaicml/llm-foundry/blob/667006889b5aab99fbecf1f927704fa7645e1799/mcli/mcli-llama2-finetune.yaml#L1-L6 for llama2)

tests/test_training.py
dakinggg: I guess `ConsoleLogger` needs to flush on `eval_after_all`? Can you make a JIRA for that and assign it to Evan please?

In the meantime...can you test using `InMemoryLogger`?

bmosaicml: I'm not sure that InMemoryLogger will work. How will I access the logger's data from inside the test?

mcli/mcli-hf-eval.yaml
bmosaicml: Should I add v.0.2.0 back in?

tests/test_training.py
dakinggg: I'm assuming in the test you are talking about you want to test calling `main`. Assuming that is true, you can return the `Trainer` from main, and then use the returned object to get whatever other objects you need.

bmosaicml: ok I will modify main to return trainer...good call

abhi-mosaic:
Sorry if this is a silly question, but is it meant to be `gauntlet` or `model_gauntlet`? I feel like the former is a bit more accurate? Or even `eval_gauntlet`, but `model_gauntlet` sounds a bit strange for a callback that averages eval metrics.

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -44,7 +43,7 @@ class ModelGauntlet(Callback):
     """
 
     def __init__(self,
-                 logger_keys: dict,
+                 logger_keys: list,
```
hanlint, line 46: Since this is an API change, can we add in a check below, e.g.
```
if isinstance(logger_keys, dict):
   raise ValueError('logger_keys now requires a list type as input, not a dict') 
```

The code path now seems to support both, but down the road, people might have silent failures if we start depending on the type being a list instead of a dict.

hanlint (approved):
LGTM, with one minor suggestion. But please get an approval from another reviewer before merging.

llmfoundry/callbacks/model_gauntlet_callback.py
bmosaicml: done!

bmosaicml:
> Sorry if this is a silly question, but is it meant to be `gauntlet` or `model_gauntlet`? I feel like the former is a bit more accurate? Or even `eval_gauntlet`, but `model_gauntlet` sounds a bit strange for a callback that averages eval metrics.

I've been calling it model gauntlet since we run models through it haha. We could call it just gauntlet or eval gauntlet if you prefer? just lmk

abhi-databricks (approved):
Excited to start using this in our training runs! Thanks @bmosaicml for getting this over the finish line.

As discussed offline, I think we should change the name from `model_gauntlet` to `gauntlet` or `eval_gauntlet`, up to you. Approving to unblock but will leave it to you to rename and merge.

llmfoundry/callbacks/model_gauntlet_callback.py
```diff
@@ -90,56 +92,47 @@ def __init__(self,
...
-                    subcat = subcat[1:]
-                    if f'metrics/{eval_name}/{num_shot}-shot/InContextLearning{metric}' not in results:
-                        results[f'metrics/{eval_name}/{num_shot}-shot/InContextLearning{metric}'] = []
-                    results[
-                        f'metrics/{eval_name}/{num_shot}-shot/InContextLearning{metric}'].append(
-                            val)
-                else:
-                    results[key] = [val]
+
+            # starting at index 1 skips the "metric" part of they key which is superfluous
```
dakinggg, line 100:
```suggestion
            # starting at index 1 skips the "metric" part of the key which is superfluous
```

scripts/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2022 MosaicML LLM Foundry authors
```
dakinggg, line 1: why do we need to make scripts a module? Would be better to not, and anything that needs to be importable move into the llmfoundry folder. If its just needed for tests, you can just append to the path to make things importable.

scripts/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import scripts.data_prep as data_prep
+import scripts.eval as eval
+import scripts.train as train
+
+__all__ = ['train', 'data_prep', 'eval']
+
+__version__ = '0.2.0'
```
dakinggg, line 10: I don't think the version info belongs here?

scripts/train/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2022 MosaicML LLM Foundry authors
```
dakinggg, line 1: same q

tests/test_eval.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Add repo root to path so we can import scripts and test it
+repo_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.append(repo_dir)
+
+from scripts.eval.eval import main  # noqa: E402
+
+
+@pytest.fixture(autouse=True)
+def set_correct_cwd():
```
dakinggg, line 19: why do we need this?

dakinggg (commented):
The tests that you have added are really slow, nearly doubling the time it takes CI to run

```
196.68s call     tests/test_training.py::test_train_gauntlet
161.70s call     tests/test_training.py::test_train[cpu]
120.18s call     tests/test_model_gauntlet.py::test_gauntlet_callback[True-False]
120.00s call     tests/test_model_gauntlet.py::test_gauntlet_callback[False-True]
119.20s call     tests/test_model_gauntlet.py::test_gauntlet_callback[True-True]
118.90s call     tests/test_model_gauntlet.py::test_gauntlet_callback[False-False]
```

Can you please reduce the intensity of the tests? Probably run less eval tasks or subset batches or something?

tests/test_eval.py
bmosaicml: tasks.yaml contains a bunch of dataset paths that are relative to the scripts directory

bmosaicml:
> The tests that you have added are really slow, nearly doubling the time it takes CI to run
> 
> ```
> 196.68s call     tests/test_training.py::test_train_gauntlet
> 161.70s call     tests/test_training.py::test_train[cpu]
> 120.18s call     tests/test_model_gauntlet.py::test_gauntlet_callback[True-False]
> 120.00s call     tests/test_model_gauntlet.py::test_gauntlet_callback[False-True]
> 119.20s call     tests/test_model_gauntlet.py::test_gauntlet_callback[True-True]
> 118.90s call     tests/test_model_gauntlet.py::test_gauntlet_callback[False-False]
> ```
> 
> Can you please reduce the intensity of the tests? Probably run less eval tasks or subset batches or something?

Not sure `test_train_gauntlet` can be sped up anymore as it's now about the same speed as test_train. Should I just cut it? Or make it GPU only?

mcli/mcli-1b-eval.yaml
```diff
@@ -21,46 +21,39 @@ compute:
...
-    n_heads: 16 # Modified 24->16 so that d_head == 128 to statisfy FlashAttention
-    n_layers: 24
-    expansion_ratio: 4
-    max_seq_len: ${max_seq_len}
-    vocab_size: 50368
-    attn_config:
-      attn_impl: triton
-
-  load_path: # Add your (optional) Composer checkpoint path here!
+  max_seq_len: 1024
```
dakinggg, line 25: does this yaml run using foundry release 0.2.0. I'm assuming not. We can't change yamls that are pinned to a foundry release out of sync with new releases.

mcli/mcli-hf-eval.yaml
```diff
@@ -38,253 +38,17 @@ parameters:
...
+
 
   icl_tasks: 'eval/yamls/tasks.yaml'
-  model_gauntlet: 'eval/yamls/model_gauntlet.yaml'
+  eval_gauntlet: 'eval/yamls/eval_gauntlet.yaml'
```
dakinggg, line 54: same version compatibility comment

scripts/eval/eval.py
```diff
@@ -276,60 +265,67 @@ def main(cfg: DictConfig):
...
-            if eval_name not in results[num_shot]:
-                results[num_shot][eval_name] = {}
-            if metric not in results[num_shot][eval_name]:
-                results[num_shot][eval_name][metric] = []
-
-            results[num_shot][eval_name][metric].append({
-                'val': val,
-                'subcat': subcat
-            })
+        # dl_name consists is either 2-tuple (benchmark_name, num_fewshot)
```
dakinggg, line 299:
```suggestion
        # dl_name is either 2-tuple (benchmark_name, num_fewshot)
```

scripts/eval/yamls/tasks_lite.yaml
dakinggg, line 1: Can we please not change the name of this file? It will break many yamls.

scripts/train/train.py
```diff
@@ -395,6 +409,30 @@ def main(cfg: DictConfig):
     # Build tokenizer
     tokenizer = build_tokenizer(tokenizer_config)
 
+    # Build Model
```
dakinggg, line 412: this doesn't seem right. why is this code changed?

scripts/train/train.py
```diff
@@ -244,11 +245,24 @@ def main(cfg: DictConfig):
...
                                                           must_exist=False,
                                                           default_value=None)
-    icl_tasks_config: Optional[ListConfig] = pop_config(cfg,
+    icl_tasks_config: Optional[Union[ListConfig,
+                                     str]] = pop_config(cfg,
                                                         'icl_tasks',
                                                         must_exist=False,
                                                         default_value=None)
-
+    eval_gauntlet_config: Optional[Union[DictConfig,
```
dakinggg, line 253: In order to not break existing yamls, please check for both `model_gauntlet` and `eval_gauntlet`, and emit a deprecation warning if `model_gauntlet` is used.

scripts/eval/eval.py
```diff
@@ -48,18 +46,18 @@ def load_peft_model(model_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
         try:
             trust_remote_code = model_cfg.get('trust_remote_code', True)
             use_auth_token = model_cfg.get('use_auth_token', False)
-            model = model_registry[model_cfg.name].from_pretrained(
+            underlying_model = model_registry[model_cfg.name].from_pretrained(
```
dakinggg, line 49: This change doesn't seem like it should be part of this pr?

scripts/eval/eval.py
```diff
@@ -170,15 +159,14 @@ def evaluate_model(model_cfg: DictConfig, dist_timeout: Union[float, int],
...
-            model_gauntlet, model_gauntlet_df)
+    return (trainer, logger_keys, eval_gauntlet_callback, eval_gauntlet_df)
 
 
 def main(cfg: DictConfig):
     om.resolve(cfg)
     model_configs: ListConfig = pop_config(cfg, 'models', must_exist=True)
-    model_gauntlet_config: Optional[Union[str, DictConfig]] = pop_config(
-        cfg, 'model_gauntlet', must_exist=False, default_value=None)
+    eval_gauntlet_config: Optional[Union[str, DictConfig]] = pop_config(
```
dakinggg, line 168: same comment as train script. need to not break existing yamls, so look for both old key and new key.

dakinggg (changes requested):
see remaining comments around backwards compatibility and a couple code changes that I think were a bad merge.

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -19,6 +19,21 @@ class Weighting(Enum):
     LOG_SAMPLE_SZ = 3
 
 
+class ModelGauntlet(Callback):
```
dakinggg, line 22: Ah I didn't notice initially that the file name changed too. So, this should actually be in the original file name `model_gauntlet_callback.py`

mcli/mcli-1b-eval.yaml
```diff
@@ -1,7 +1,7 @@
 integrations:
 - integration_type: git_repo
   git_repo: mosaicml/llm-foundry
-  git_branch: v0.2.0
+  # git_branch: # Specify your git branch
```
dakinggg, line 4: This still needs to be pinned, so if this yaml is updated to only work with main, we need to repin after this PR goes in.

scripts/train/train.py
```diff
@@ -474,7 +502,7 @@ def main(cfg: DictConfig):
         assert model.train_metrics is not None
         eval_metric_names = list(model.train_metrics.keys())
         eval_loader.metric_names = eval_metric_names
-        evaluators.insert(0, eval_loader)  # Put the base eval_loader first
+        evaluators.insert(0, eval_loader)
```
dakinggg, line 505: Can you put the comment back?

mcli/mcli-1b-eval.yaml
bmosaicml: what should I pin it to?

dakinggg (approved):

PR #506, opened 2023-08-07
Add runtime error in train.py if yaml config is improperly formatted with extraneous or missing values

j316chuck:
## Description
This PR enables us to sanity check our train yaml configuration files before we run the full training pipeline. This enables us to catch errors in the YAML config before a training run starts. If a yaml config is improperly formatted with extraneous or missing values, a `NameError` will be thrown. 

## Unit Test: 
Added unit tests to make sure we raise a `NameError` and/or warn the user if the yaml is incorrectly formatted. Warnings are used if the parameter has an optional default value. 

## Integration Test:
Before and after training runs show the same loss curves throughout training. 
Screenshot 2023-08-15 at 10 57 55 AM

**mpt-125m training branch master**: https://wandb.ai/mosaic-ml/chuck-runs/runs/2jq1uy86
**mpt-125m training branch chuck/add_yaml_sanity_check_train**: https://wandb.ai/mosaic-ml/chuck-runs/runs/n8wlqwec

All runs at: https://wandb.ai/mosaic-ml/chuck-runs?workspace=user-chuck-tang98

### DIFF
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -12,6 +12,7 @@ from omegaconf import OmegaConf as om
 
 from llmfoundry.models.utils import init_empty_weights
 
+
 def pop_config(cfg: DictConfig,
                key: str,
                must_exist: bool = True,
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -26,7 +26,7 @@ def pop_config(cfg: DictConfig,
     if value is not None:
         return value
     elif must_exist:
-        raise RuntimeError(
+        raise NameError(
             f'The {key} parameter is missing and must exist for execution. Please check your yaml.'
         )
     else:
@@ -112,8 +112,8 @@ def process_init_device(model_cfg: DictConfig, fsdp_config: Optional[Dict]):
 def log_config(cfg: DictConfig):
     """Logs the current config and updates the wandb and mlflow configs.
 
-    This function can be called multiple times to update the wandb config with
-    different variables.
+    This function can be called multiple times to update the wandb and MLflow
+    config with different variables.
     """
     print(om.to_yaml(cfg))
     if 'wandb' in cfg.get('loggers', {}):
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -90,6 +90,11 @@ def process_init_device(model_cfg: DictConfig, fsdp_config: Optional[Dict]):
 
 
 def log_config(cfg: DictConfig):
+    """Logs the current config and updates the wandb config if available.
+
+    This function can be called multiple times to update the wandb config with
+    different variables.
+    """
     print(om.to_yaml(cfg))
     if 'wandb' in cfg.get('loggers', {}):
         try:
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -4,13 +4,32 @@
 import contextlib
 import math
 import warnings
-from typing import Dict, Optional, Union
+from typing import Any, Dict, Optional, Union
 
 from composer.utils import dist
 from omegaconf import DictConfig
 from omegaconf import OmegaConf as om
 
 from llmfoundry.models.utils import init_empty_weights
+
+def pop_config(cfg: DictConfig,
+               key: str,
+               must_exist: bool = True,
+               default_value: Any = None) -> Any:
+    """Pop a value from the main config file and return it.
+
+    If the key does not exist, return the default_value or raise a RuntimeError
+    depending on the must_exist flag.
+    """
+    value = cfg.pop(key, None)
+    if value is not None:
+        return value
+    elif must_exist:
+        raise RuntimeError(
+            f'The {key} parameter is missing and must exist for execution. Please check your yaml.'
+        )
+    else:
+        return default_value
 
 
 def calculate_batch_size_info(global_batch_size: int,
@@ -90,7 +109,7 @@ def process_init_device(model_cfg: DictConfig, fsdp_config: Optional[Dict]):
 
 
 def log_config(cfg: DictConfig):
-    """Logs the current config and updates the wandb config if available.
+    """Logs the current config and updates the wandb and mlflow configs.
 
     This function can be called multiple times to update the wandb config with
     different variables.
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -184,11 +184,11 @@ def main(cfg: DictConfig):
     om.resolve(cfg)
 
     # Set seed first
-    seed: int = cfg.pop('seed')
+    seed: int = pop_config(cfg, 'seed', must_exist=True)
     reproducibility.seed_all(seed)
 
     # Initialize distributed setting with seed
-    dist_timeout: Union[int, float] = cfg.pop('dist_timeout', 600.0)
+    dist_timeout: Union[int, float] = pop_config(cfg, 'dist_timeout', must_exist=False, default_value=600.0)
     dist.initialize_dist(get_device(None), timeout=dist_timeout)
 
     # Get global and device batch size information from distributed/single node setting
@@ -279,12 +279,12 @@ def main(cfg: DictConfig):
     autoresume = pop_config(cfg,'autoresume', must_exist=False, default_value=autoresume_default)
 
     # Pop known unused parameters.
-    cfg.pop('data_local', None)
-    cfg.pop('data_remote', None)
-    cfg.pop('global_seed')
-    cfg.pop('global_train_batch_size')
-    cfg.pop('n_gpus')
-    cfg.pop('device_train_grad_accum')
+    pop_config(cfg, 'data_local', must_exist=False)
+    pop_config(cfg, 'data_remote', must_exist=False)
+    pop_config(cfg, 'global_seed', must_exist=False)
+    pop_config(cfg, 'global_train_batch_size', must_exist=False)
+    pop_config(cfg, 'n_gpus', must_exist=False)
+    pop_config(cfg, 'device_train_grad_accum', must_exist=False)
 
     # Warn users for unused parameters
     for key in cfg:
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -3,13 +3,13 @@
 import os
 import sys
 import warnings
-from typing import Any, List, Optional, Union
+from typing import List, Optional, Union, Dict
 
 import torch
 from composer import Trainer
 from composer.core import Evaluator
 from composer.utils import dist, get_device, reproducibility
-from omegaconf import DictConfig
+from omegaconf import DictConfig, ListConfig
 from omegaconf import OmegaConf as om
 from transformers import PreTrainedTokenizerBase
 
@@ -22,7 +22,7 @@ from llmfoundry.utils.builders import (build_algorithm, build_callback,
                                        build_optimizer, build_scheduler,
                                        build_tokenizer)
 from llmfoundry.utils.config_utils import (log_config, process_init_device,
-                                           update_batch_size_info)
+                                           update_batch_size_info, pop_config)
 
 
 def validate_config(cfg: DictConfig):
@@ -168,29 +168,6 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
         raise ValueError(f'Not sure how to build dataloader with config: {cfg}')
 
 
-def pop_config(cfg: DictConfig,
-               key: str,
-               must_exist: bool = True,
-               default_value: Any = None) -> Any:
-    """Pop a value from the main config file and return it.
-
-    If the key does not exist, return the default_value or raise a RuntimeError
-    depending on the must_exist flag.
-    """
-    value = cfg.pop(key, None)
-    if value:
-        return value
-    elif must_exist:
-        raise RuntimeError(
-            f'The {key} parameter is missing and must exist for execution. Please check your yaml.'
-        )
-    else:
-        warnings.warn(
-            f'No {key} parameters specified. {key} will not be used during execution.'
-        )
-        return default_value
-
-
 def main(cfg: DictConfig):
     # Filter deprecation warning from torch internal usage
     warnings.filterwarnings(
@@ -211,11 +188,10 @@ def main(cfg: DictConfig):
     reproducibility.seed_all(seed)
 
     # Initialize distributed setting with seed
-    dist_timeout: int = cfg.pop('dist_timeout', 600.0)
+    dist_timeout: Union[int, float] = cfg.pop('dist_timeout', 600.0)
     dist.initialize_dist(get_device(None), timeout=dist_timeout)
 
     # Get global and device batch size information from distributed/single node setting
-    # TODO remove this as we should treat configs as immutable objects
     cfg = update_batch_size_info(cfg)
 
     print('Logging config...')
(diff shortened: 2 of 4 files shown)


j316chuck:
Right now we are still seeing these values after popping most of the keys from the config
```
(Pdb) p cfg
{'data_local': 'my-copy-c4', 'data_remote': None, 'max_seq_len': 2048, 'global_seed': 100, 'global_train_batch_size': 256, 'seed': '${global_seed}', 'n_gpus': 1, 'device_train_grad_accum': 16, 'n_params': 125311488}
```

scripts/train/yamls/pretrain/mpt-125m.yaml
```diff
@@ -75,6 +75,7 @@ eval_interval: 500ba
 eval_first: false
 eval_subset_num_batches: -1
 global_train_batch_size: 256
+save_overwrite: true
```
dakinggg, line 78: why'd you add this?

scripts/train/train.py
```diff
@@ -180,156 +188,190 @@ def main(cfg: DictConfig):
...
     print('Logging config...')
     log_config(cfg)
 
-    if cfg.get('eval_first',
+    # Eval first if requested
+    if cfg.pop('eval_first',
                False) and trainer.state.timestamp.batch.value == 0:
         trainer.eval()
 
+    check_config_only_contains_special_params(cfg)
```
dakinggg, line 373: this should probably happen before the eval call

scripts/train/train.py
```diff
@@ -25,6 +25,14 @@
                                            update_batch_size_info)
 
 
+def check_config_only_contains_special_params(cfg: DictConfig): 
+    """ Checks that the config is empty except for special parameters after we loaded everything in trainer. """
+    SPECIAL_UNPOPPED_PARAMETERS = ['data_local', 'data_remote', 'max_seq_len', 'global_seed', 'global_train_batch_size', 'n_gpus', 'device_train_grad_accum']
```
dakinggg, line 30: I think we should figure out a way to make this not quite so brittly hardcoded. We can chat tomorrow

dakinggg (commented):
Looks like the right track! One high level comment is that this doesn't completely close the linked jira. Lets chat tomorrrow

scripts/train/train.py
```diff
@@ -181,153 +200,232 @@ def main(cfg: DictConfig):
...
+    autoresume_default = False
+    necessary_autoresume_condition_met = run_name and save_folder and save_latest_filename
+    skip_autoresume_conditions = save_overwrite or save_weights_only
+    if necessary_autoresume_condition_met and not skip_autoresume_conditions:
+        print(
+            'As run_name, save_folder, and save_latest_filename are set, changing autoresume default to True...'
+        )
+        autoresume_default = True
+    autoresume = cfg.pop('autoresume', autoresume_default)
+
```
j316chuck, line 306: should this be refactored into a function?

scripts/train/train.py
```diff
@@ -181,153 +200,232 @@ def main(cfg: DictConfig):
...
+                                                      'loggers',
+                                                      must_exist=False)
+    callback_configs: Optional[DictConfig] = pop_config(cfg,
+                                                        'callbacks',
+                                                        must_exist=False)
+    algorithm_configs: Optional[DictConfig] = pop_config(cfg,
+                                                         'algorithms',
+                                                         must_exist=False)
+
+    device_train_batch_size: int = cfg.pop('device_train_batch_size')
```
j316chuck, line 258: Should we:
- have type annotations? 
- use `pop_config` function in which case the warnings will be very verbose which may be good? 
- list all the DEFAULT_VALUES at the top of the file?

j316chuck (commented):
added a v2 of this yaml config restructuring. 

will add unit tests as well in this PR.

llmfoundry/utils/config_utils.py
```diff
@@ -90,6 +90,12 @@ def process_init_device(model_cfg: DictConfig, fsdp_config: Optional[Dict]):
 
 
 def log_config(cfg: DictConfig):
+    """Logs the current config to stdout and updates the wandb config if
```
dakinggg: looks like it support mlflow as well

scripts/train/train.py
```diff
@@ -164,170 +164,266 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
         )
-
     else:
         raise ValueError(f'Not sure how to build dataloader with config: {cfg}')
 
 
-def main(cfg: DictConfig):
-    # Check for incompatibilities between the model and data loaders
-    validate_config(cfg)
+def pop_config(cfg: DictConfig,
```
dakinggg, line 171: we can move this to `config_utils.py` and then reuse it across scripts

scripts/train/train.py
```diff
@@ -164,170 +164,266 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+               key: str,
+               must_exist: bool = True,
+               default_value: Any = None) -> Any:
+    """Pop a value from the main config file and return it.
+
+    If the key does not exist, return the default_value or raise a RuntimeError
+    depending on the must_exist flag.
+    """
+    value = cfg.pop(key, None)
+    if value:
```
dakinggg, line 181:
```suggestion
    if value is not None:
```
The current if statement will not return the number 0 :)

scripts/train/train.py
```diff
@@ -164,170 +164,266 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+        return value
+    elif must_exist:
+        raise RuntimeError(
+            f'The {key} parameter is missing and must exist for execution. Please check your yaml.'
+        )
+    else:
+        warnings.warn(
```
dakinggg, line 188: This should probably not be a warning, because there is no expected user action to correct it. It should probably be a log.debug, except that we haven't set up logging in foundry properly, so maybe just leave out any log/print.

scripts/train/train.py
```diff
@@ -164,170 +164,266 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
-        f'torch.distributed.*_base is a private function and will be deprecated.*'
+        'torch.distributed.*_base is a private function and will be deprecated.*'
     )
 
-    cfg.dist_timeout = cfg.get('dist_timeout', 600.0)
+    # Check for incompatibilities between the model and data loaders
+    validate_config(cfg)
+
+    # Resolve all interpolation variables as early as possible
+    om.resolve(cfg)
```
dakinggg, line 207: did you test this line? I thought previously when I did this I had to do `om.to_container(cfg, resolve=True)`, but maybe that was because I needed an actual dict out. Should we just convert to an actual dict here?

scripts/train/train.py
```diff
@@ -164,170 +164,266 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
 
-    reproducibility.seed_all(cfg.seed)
-    dist.initialize_dist(get_device(None), timeout=cfg.dist_timeout)
+    # Set seed first
+    seed: int = cfg.pop('seed')
```
dakinggg, line 210: should this user your helper?

scripts/train/train.py
```diff
@@ -164,170 +164,266 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+    reproducibility.seed_all(seed)
 
-    # Run Name
-    if cfg.get('run_name') is None:
-        cfg.run_name = os.environ.get('RUN_NAME', 'llm')
+    # Initialize distributed setting with seed
+    dist_timeout: int = cfg.pop('dist_timeout', 600.0)
```
dakinggg, line 214: I think this can be int or float? not sure, but the default is specified as a float

scripts/train/train.py
```diff
@@ -164,170 +164,264 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+                                                      must_exist=False)
+    callback_configs: Optional[DictConfig] = pop_config(cfg,
+                                                        'callbacks',
+                                                        must_exist=False)
+    algorithm_configs: Optional[DictConfig] = pop_config(cfg,
+                                                         'algorithms',
+                                                         must_exist=False)
+
+    # Mandatory parameters for trainer
+    device_train_batch_size: int = cfg.pop('device_train_batch_size')
```
dakinggg, line 259: everything here and below should use the helper function?

scripts/train/train.py
```diff
@@ -164,170 +164,264 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+    device_eval_batch_size: int = cfg.pop('device_eval_batch_size')
+    max_duration: Union[int, str] = cfg.pop('max_duration')
+    eval_interval: Union[int, str] = cfg.pop('eval_interval')
+    precision: str = cfg.pop('precision')
+    max_seq_len: int = cfg.pop('max_seq_len')
+
+    # Optional parameters for will be set to default values if not specified.
```
dakinggg, line 266: Note to self to double check all the defaults

scripts/train/train.py
```diff
@@ -164,170 +164,264 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+    # Warn if fsdp is enabled but user only has 1 GPU
     if dist.get_world_size() == 1 and fsdp_config is not None:
         warnings.warn(
             'FSDP is not applicable for single-GPU training. Reverting to DDP.')
-        cfg.pop('fsdp_config')
         fsdp_config = None
 
-    init_context = process_init_device(cfg.model, fsdp_config)
+    # Initialize context
+    init_context = process_init_device(model_config, fsdp_config)
```
dakinggg, line 316: This call modifies the config, so we should log the config after it

scripts/train/train.py
```diff
@@ -164,170 +164,264 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
         tokenizer,
-        cfg.device_train_batch_size,
+        device_train_batch_size,
     )
+
+    ## Evaluation
     print('Building eval loader...')
     evaluators = []
-    if 'eval_loader' in cfg:
+    if eval_loader_config:
```
dakinggg, line 346:
```suggestion
    if eval_loader_config is not None:
```

scripts/train/train.py
```diff
@@ -164,170 +164,264 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
-                                metric_names=list(model.train_metrics.keys()))
+                                dataloader=eval_dataloader,
+                                metric_names=eval_metric_names)
         evaluators.append(eval_loader)
 
-    if 'icl_tasks' in cfg:
-        icl_evaluators, _ = build_icl_evaluators(cfg.icl_tasks, tokenizer,
-                                                 cfg.max_seq_len,
-                                                 cfg.device_eval_batch_size)
+    if icl_tasks_config:
```
dakinggg, line 356:
```suggestion
    if icl_tasks_config is not None:
```

scripts/train/train.py
j316chuck: fixed to use helper

scripts/train/train.py
j316chuck: fixed

scripts/train/train.py
j316chuck: fixed

scripts/train/train.py
j316chuck: checked

scripts/train/train.py
j316chuck: logged

j316chuck (commented):
thx for the rev

llmfoundry/utils/config_utils.py
j316chuck: thx added

tests/test_yaml_inputs.py
```diff
@@ -0,0 +1,89 @@
...
+            'scheduler',
+            'max_duration',
+            'eval_interval',
+            'precision',
+            'max_seq_len',
+        ]
+        for param in mandatory_params:
+            orig_param = cfg.pop(param)
+            with pytest.raises(
+                (omegaconf.errors.ConfigAttributeError, RuntimeError)):
```
dakinggg, line 55: Why is this either a config attribute error or a runtime error?

tests/test_yaml_inputs.py
```diff
@@ -0,0 +1,89 @@
...
+        ]
+        old_cfg = copy.deepcopy(cfg)
+        for param in optional_params:
+            orig_value = cfg.pop(param, None)
+            updated_param = param + '-mispelling'
+            cfg[updated_param] = orig_value
+            with warnings.catch_warnings(record=True) as warning_list:
+                try:
+                    main(cfg)
+                except:
```
dakinggg, line 84: why is except pass here? I thought it would always emit a warning not crash

llmfoundry/utils/config_utils.py
```diff
@@ -90,6 +110,11 @@ def process_init_device(model_cfg: DictConfig, fsdp_config: Optional[Dict]):
 
 
 def log_config(cfg: DictConfig):
+    """Logs the current config and updates the wandb and mlflow configs.
+
+    This function can be called multiple times to update the wandb config with
```
dakinggg, line 115: Can it also be called multiple times with mlflow?

scripts/train/train.py
```diff
@@ -164,170 +165,315 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
 
-    # Run Name
-    if cfg.get('run_name') is None:
-        cfg.run_name = os.environ.get('RUN_NAME', 'llm')
+    # Set seed first
+    seed: int = pop_config(cfg, 'seed', must_exist=True)
+    reproducibility.seed_all(seed)
 
-    # Get batch size info
+    # Initialize distributed setting with seed
```
dakinggg, line 191: why does this comment. mention seed?

scripts/train/train.py
```diff
@@ -164,170 +165,315 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+    # Get global and device batch size information from distributed/single node setting
     cfg = update_batch_size_info(cfg)
 
-    # Read FSDP Config as a dict
-    fsdp_config = cfg.get('fsdp_config', None)
-    fsdp_config = om.to_container(fsdp_config,
-                                  resolve=True) if fsdp_config else None
-    assert isinstance(fsdp_config, Dict) or fsdp_config is None
+    print('Logging config...')
+    log_config(cfg)
```
dakinggg, line 202: The log config call actually needs to happen _after_ the trainer is created, because creating the trainer is what initializes wandb and mlflow.

scripts/train/train.py
```diff
@@ -164,170 +165,315 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+    load_weights_only: bool = pop_config(cfg,
+                                         'load_weights_only',
+                                         must_exist=False,
+                                         default_value=False)
+    load_ignore_keys: Optional[List[str]] = pop_config(cfg,
+                                                       'load_ignore_keys',
+                                                       must_exist=False,
+                                                       default_value=None)
+    # Enable autoresume from model checkpoints if possible
+    autoresume_default = not save_overwrite and not save_weights_only
```
dakinggg, line 339: I don't think this is the same default condition as before

scripts/train/train.py
```diff
@@ -164,170 +165,315 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+    autoresume = pop_config(cfg,
+                            'autoresume',
+                            must_exist=False,
+                            default_value=autoresume_default)
+
+    # Pop known unused parameters.
```
dakinggg, line 345: These are all used as interpolation variables? If so, can you update the comment to say that?

scripts/train/train.py
```diff
@@ -164,170 +165,315 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
     if dist.get_world_size() == 1 and fsdp_config is not None:
         warnings.warn(
             'FSDP is not applicable for single-GPU training. Reverting to DDP.')
-        cfg.pop('fsdp_config')
         fsdp_config = None
 
-    init_context = process_init_device(cfg.model, fsdp_config)
+    # Initialize context
+    init_context = process_init_device(model_config, fsdp_config)
+    if fsdp_config:
```
dakinggg, line 367: same, we can remove this and just log once after trainer init

scripts/train/train.py
```diff
@@ -164,170 +165,315 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
             print_trainable_parameters(model)  # should not be 100%
         else:  # standard model
-            model = build_composer_model(cfg.model, tokenizer)
-    cfg.n_params = sum(p.numel() for p in model.parameters())
-    print(f'{cfg.n_params=:.2e}')
+            model = build_composer_model(model_config, tokenizer)
+
+    # Log number of parameters
+    n_params = sum(p.numel() for p in model.parameters())
+    log_config(om.create({'n_params': n_params}))
```
dakinggg, line 385: same, this needs to happen after trainer init

scripts/train/train.py
```diff
@@ -164,170 +165,315 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
     # Scheduler
-    scheduler = build_scheduler(cfg.scheduler)
+    scheduler = build_scheduler(scheduler_config)
 
     # Loggers
     loggers = [
-        build_logger(name, logger_cfg)
-        for name, logger_cfg in (cfg.get('loggers') or {}).items()
-    ]
+        build_logger(str(name), logger_cfg)
```
dakinggg, line 422: why did this need a str cast?

dakinggg (commented):
Could you please include a weights and biases plot with a short run before this pr and after this pr so we can make sure everything is still working?

scripts/train/train.py
j316chuck: The `cfg` will be empty after the `trainer` has been initialized since we are popping out from it to check for improperly formatted variables. 

A slightly verbose solution is to create a `logged_cfg` variable that we pop all variables from `cfg` to logged_cfg to solely for logging purposes.

dakinggg: Ah, good point. Should we make a deepcopy at the start and log that at the end?

j316chuck: ya let's do that

scripts/train/train.py
j316chuck: ah good catch

scripts/train/train.py
j316chuck: updated comment

scripts/train/train.py
j316chuck: done

scripts/train/train.py
j316chuck: yep yep

scripts/train/train.py
j316chuck: It's of type `DictKeyType` whereas build_logger takes type string

tests/test_yaml_inputs.py
j316chuck: It's a config error if we use the variable in `validate_config`. It's a runtime error if we `pop_config` when the key is missing and must_exist=True

https://github.com/mosaicml/llm-foundry/blob/90132f035d00de9064d3246553ac770fbce20ce1/llmfoundry/utils/config_utils.py#L29-L31

tests/test_yaml_inputs.py
j316chuck: it will crash as the dataset does not exist; we have to import the dataset creation from the `test_training.py` test fixes

j316chuck: Another thing we can do is add a dry run option to trainer to only call up to trainer.init()?

alextrott16 (commented):
I'll withhold from giving a formal "Approve" because @dakinggg is providing the most comprehensive reviewing. That said, this looks great to me at a high level, so I soft approve!

One nit is that `must_exist` and `default_value` seem to be redundant in `pop_config`. That is, if a default is provided then `must_exist=False` and if not then `must_exist=True`. So, you could probably get rid of `must_exist` as an explicit argument and clean up the code.

Otherwise, I love what's going on here.

j316chuck:
Thanks for the rev @alextrott16. 

The caveat with default_value and must_exist is that there is an edge case where the `must_exist` can be False but `default_value` for the variable is still None which is why we need both variables.

j316chuck:
Test that diffs the two output logs from wandb also show that they are almost identical in config ✅ 
[output (1).log](https://github.com/mosaicml/llm-foundry/files/12340413/output.1.log)
[output.log](https://github.com/mosaicml/llm-foundry/files/12340414/output.log)

scripts/train/train.py
```diff
@@ -164,170 +166,323 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+                                         must_exist=False,
+                                         default_value=False)
+    load_ignore_keys: Optional[List[str]] = pop_config(cfg,
+                                                       'load_ignore_keys',
+                                                       must_exist=False,
+                                                       default_value=None)
+    # Enable autoresume from model checkpoints if possible
+    autoresume_default: bool = False
+    if logged_cfg.get('run_name', None) is not None \
+        and save_folder is not None \
```
dakinggg, line 340: I think this condition is still missing the check on `save_latest_filename`?

dakinggg (commented):
Looking good! Left one last comment, and if @vchiley could take another pass that would great

scripts/train/train.py
j316chuck: `save_latest_filename` is always set to something by default so no need to do checks against None condition.

vchiley:
as discussed offline
it would be nice to avoid adding 150+ lines of config checking to the `train.py` training entry point.
almost all config checking would probably find a good home in `utils/config_utils.py`.


Aside
Another nice check for args would be, in `[build_optimizer](https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/utils/builders.py#L91)`, you could `name = cfg.pop('name')` at the top of the fn, convert the cfg into a dict, then **kwargs all optimizer args.
This way all optimizer args MUST be valid.

dakinggg:
This pr adds very little net new code, and is mostly a reorganization with popping added, and the config parsing code isn't reusable anywhere else so I would probably leave it here, but not opposed to pulling it into a helper function. Will leave it up to Chuck what he thinks is cleaner.

The second comment is valid, but will be in a future pr. Since this one is a big change, would prefer not to tack on little bits and instead do them separately

j316chuck:
Agreed, this PR is mainly a refactoring of code and adding type checks for readability, type safety, and early error checking. Let's not refactor all the configs into a helper function just yet and  defer to future PR when we try to remove kwargs. 

As to the second comment, yes, we will do that in a future PR as well so that functions touch less config and more explicit function arguments.

tests/test_yaml_inputs.py
j316chuck: resolved offline; running a short test that crashes is best for this use case.

dakinggg (approved):

PR #509, opened 2023-08-08
Remove old optimizer logs

mvpatel2000:
The optimizer monitor logs very aggressively. This PR changes removes some logging which is not useful per @bmosaicml's recommendation


vchiley (approved):
if @bmosaicml says they're not super useful, then lgtm

PR #514, opened 2023-08-10
8-bit LION, take 2

dblalock:
Adds an 8-bit version of the LION optimizer. Also features 1 byte of (optional) auxiliary error correction state for each parameter to make pure bf16 training work.

Code changes:
- Adds `lion8b.py` to `llm-foundry/optim`
- Adds `DecoupledLionW_8bit` to `llm-foundry/optim/__init__.py`
- Adds `lion8b` as an option in `llm-foundry/optim/builders.py`
- Adds `test_lion8b.py` to the tests.
- Adds `mosaicml-turbo` to the GPU dependencies in `setup.py`. This is the repo that currently holds all the CUDA kernels. These are in a separate repo for now to avoid complicating LLM foundry {install, deps, source code}.
- Adds an optional `master_weight_dtype` field in `train.py`. If set to bf16 or fp16, the script does `model.to(dtype=<that dtype>)` before training. This works when we have error correction turned on.
- Tweaks `config_utils.py` to set FSDP's param_dtype to None if the master weights are already fp16/bf16.

Non-obvious design choices:
- Features the option to store a few extra bits per parameter to facilitate bf16 or f16 master weights, which should save us space and hopefully squeeze out a little more MFU.
- There's an undocumented `_fused` arg that's kind of needed for testing and maybe (?) should be part of the API in case someone wants to check whether it's causing issues. I kind of want to get rid of this though once we trust the kernel logic fully.
- This should work fine on CPUs—it just doesn't quantize. We could save some LoC if we instead said "sorry, you need a GPU to run this" but I'm worried about users wanting to prototype/debug locally.
- We have a "_MaybeQuantizedTensor" class to hold the quantized optimizer states. This is a pretty good solution but I don't love having a tensor-like object that doesn't implement the full Tensor interface.

There's enough test coverage here that I'm not super worried about these choices, but wanted to highlight them in case someone has strong opinions.

### [**WandB report**](https://wandb.ai/mosaic-ml/lion8b-v2/reports/LION8b-MPT-Results--Vmlldzo1MTMzMTE5)


llmfoundry/utils/builders.py
```diff
@@ -115,6 +115,10 @@ def build_optimizer(cfg: DictConfig, model: torch.nn.Module):
                                   timeout=cfg.timeout,
                                   lr_penalty=cfg.lr_penalty,
                                   min_scale=cfg.min_scale)
+    elif cfg.name.lower() == 'decoupled_lionw_8b':
+        # str() cast is just for pyright
+        kwargs = {str(k): v for k, v in cfg.items() if k != 'name'}
```
vchiley, line 120: if you want to make all optimizer inits use `**kwargs`, that would help people avoid wrong args
and you can probably `cfg.pop('name')` at the top of the builder fn

llmfoundry/utils/config_utils.py
```diff
@@ -86,6 +86,25 @@ def process_init_device(model_cfg: DictConfig, fsdp_config: Optional[Dict]):
...
+
+    # no mixed precision needed for weights when they're already 16 bits
+    master_dtype = model_cfg.get('master_weights_dtype')
+    small_dtypes = ('bf16', 'f16', 'float16', 'bfloat16', 'amp_fp16',
+                    'amp_bf16')
+    if fsdp_config and master_dtype in small_dtypes:
+        reduce_dtype = None
+        buffer_dtype = None
+        mixed_precision = fsdp_config.get('mixed_precision')
+        if isinstance(mixed_precision, Mapping):
```
vchiley, line 98: what happens when people just use `mixed_precision: PURE` or `DEFAULT`?

tests/test_lion8b.py
vchiley, line 1: Is there a way to test that elastic sharding works (Evan will prob be the best person to talk to about this).

tests/test_lion8b.py
```diff
@@ -0,0 +1,422 @@
...
+    with pytest.raises(AssertionError):  # opt step modified the weights
+        torch.testing.assert_close(W_orig, W)
+
+    # every momentum should be nonzero with infinite precision, but
+    # might be zero after quantization
+    param_state = opt.state[W]  # type:ignore using tensor as key
+    momentum = param_state['exp_avg'].materialize()
+    assert momentum.shape == (D, D)
+    momentum = momentum.ravel()
+    assert momentum is not None
```
vchiley, line 57: if it has shape (2 lines above) why assert is not None here?

tests/test_lion8b.py
```diff
@@ -0,0 +1,422 @@
...
+    opt = Lion8bit([W],
+                   _fused=fused,
+                   betas=(.5, .5),
+                   quantize=(device != 'cpu'),
+                   weight_decay=weight_decay,
+                   error_correction=use_errors)
+
+    zeros_grad = torch.zeros_like(W)
+    for _ in range(5):
+        W.grad = zeros_grad
```
vchiley, line 91: https://pytorch.org/docs/stable/generated/torch.Tensor.zero_.html#torch.Tensor.zero_

vchiley, line 91:
```suggestion
    for _ in range(5):
        W.grad.zeros_()
```

https://pytorch.org/docs/stable/generated/torch.Tensor.zero_.html#torch.Tensor.zero_

tests/test_lion8b.py
```diff
@@ -0,0 +1,422 @@
...
+    opt.step()
+    opt.zero_grad()
+
+    with pytest.raises(AssertionError):  # opt step modified the weights
+        torch.testing.assert_close(W_orig, W)
+
+    # every momentum should be nonzero with infinite precision, but
+    # might be zero after quantization
+    param_state = opt.state[W]  # type:ignore using tensor as key
+    momentum = param_state['exp_avg'].materialize()
```
vchiley, line 54: Is there a `torch.Tensor.materialize()` member method?
Can't find it in the docs... can this be replaced with something people can reference?

llmfoundry/utils/config_utils.py
dblalock: I had `mixed_precision: PURE` in my runs. All the mixed precision dtypes just default to `None` if the config isn't a mapping.

llmfoundry/utils/builders.py
dblalock: Can I do this in a separate PR? I admittedly wasn't super disciplined about this but it would be nice to not risk breaking something unrelated in this one.

llmfoundry/optim/lion8b.py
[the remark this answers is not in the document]
dblalock: These are great questions.

1. Yes, a private repo where I'm the owner. I don't think this is the long-term solution but I'm not sure what is. Maybe move this repo to our org? Maybe move the code to perf repo? I see this as analogous to your triton fork for now, but pip installable from pypi. Happy to move code wherever though if we reach some agreement.
2. There actually *is* a copy of the optimizer there (since that's where I debugged everything) but I copied everything but the CUDA + python wrappers to llmfoundry to improve the user experience. The main UX point is ensuring that all the errors get thrown from code that users can inspect. This again may not be the right decision long-term.
3. Yes, my training runs were on H100s.

tests/test_lion8b.py
dblalock: Negative--this is specific to our MaybeQuantizedTensor. We make sure that our exp_avg state is always an instance of this class.

I'll add some clarifying comments.

tests/test_lion8b.py
dblalock: So I started to change this, but then I then I realized it's like this because torch optimizers' `zero_grad` [defaults to setting the grad to None now](https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.zero_grad.html); this means we actually do have to assign to `.grad` since `W.grad == None`. Or we could `zero_grad(set_to_none=False)`, but I'd rather test the default behavior.

tests/test_lion8b.py
eracah: I sent this to Davis: https://github.com/mosaicml/composer/blob/dev/tests/trainer/test_fsdp_checkpoint.py#L498-L582

llmfoundry/utils/config_utils.py
vchiley: iirc
if your runs used `mixed_precision: PURE`, then mixed_precision was not a mapping, then in your runs `reduce_dtype` and `buffer_dtype` defaulted to `None`, which means in FSDP they default to fp32, therefore you were not using PURE

tests/test_lion8b.py
```diff
@@ -0,0 +1,422 @@
...
+        # since we're getting the same batch every time and have a small
+        # learning rate, our gradients should point in the same direction
+        # at each step. Consequently, our momentum should grow each step.
+        state_for_param = opt.state[W]  # type:ignore using tensor as key
+        momentum = state_for_param['exp_avg'].materialize()
+        assert momentum is not None and momentum.shape == W.shape
+        if prev_momentum is not None:
+            momentum_changes = momentum - prev_momentum
+            assert torch.all(momentum_changes >= 0)
+            assert momentum_changes.max() > 0
```
vchiley, line 151:
```
        # since we're getting the same batch every time and have a small
        # learning rate, our gradients should point in the same direction
        # at each step. Consequently, our momentum should grow each step.
```

that describes why its norm or abs should grow, not why it should always be positive.

tests/test_lion8b.py
```diff
@@ -0,0 +1,422 @@
...
+@pytest.mark.parametrize('D', [4, 12])  # vectorized and unvectorized impls
+@pytest.mark.parametrize('dtype', _FLOAT_DTYPES)
+def test_lion8b_fused_unfused_unquantized_same(w_init: str, grad_strategy: str,
+                                               D: int,
+                                               dtype: torch.dtype) -> None:
+    torch.manual_seed(123)
+    device = 'cuda'
+
+    # each optimizer gets a different weight matrix to optimize
+    if w_init == 'cyclic':
```
vchiley, line 177: is this a particularly bad case?
why test this init?

tests/test_lion8b.py
```diff
@@ -0,0 +1,422 @@
...
+        W0 = ((W0 // 2 % 3) - 1).to(dtype=dtype)
+    elif w_init == 'rand':
+        W0 = torch.rand(
+            size=(D, D), device=device, requires_grad=False,
+            dtype=dtype) * 2 - 1
+        W0 += .01 * torch.sign(W0)  # bound away from 0 to cap rel errors
+        W0 = W0.to(dtype=dtype)
+    else:  # here for pyright
+        raise ValueError('Unrecognized w_init: ', w_init)
+    W0.add_(W0.sign())  # bound away from zero so decay won't flip sign
```
vchiley, line 191: given we use really high weight decay, is this a realistic setting?

llmfoundry/utils/config_utils.py
dblalock: Do they default to fp32 specifically, or whatever dtype they are in the model? I have a separate `model.to(dtype=torch.bfloat16)` [in `train.py`](https://github.com/mosaicml/llm-foundry/blob/71e3f9c426e7490e1032a1e5a801bde07c295aeb/scripts/train/train.py#L222).

I *think* FSDP is using bf16 weights and grads when requested based on the ~6GB memory savings across 8 GPUs on a 7B model [in wandb](https://wandb.ai/mosaic-ml/lion8b-v2/reports/LION8b-MPT-Results--Vmlldzo1MTMzMTE5). (7e9 params * 6 bytes saved per param) / 8 devices = 5.25GB. And the marginal gain of ~3.7GB going from {4B grad, 4B weight, 1B opt state} -> {2B grad, 3B weight, 1B opt state} is consistent with saving 7e9/8 * 3bytes = 2.625GB.

It's actually a little weird to me that we're saving *more* memory than we'd expect, but maybe this is the pytorch allocator or FSDP not binpacking perfectly.

tests/test_lion8b.py
dblalock: mostly because it was good for debugging. I can rm it.

tests/test_lion8b.py
dblalock: ah, this is a relic from when I checked that the signs didn't get flipped in an earlier version of this test. I just removed this line.

llmfoundry/utils/config_utils.py
vchiley: I guess if its None, then param_dtype is used: https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp/api.py#L125

llmfoundry/optim/lion8b.py
```diff
@@ -0,0 +1,417 @@
...
+                 weights: torch.Tensor,
+                 momentums: _MaybeQuantizedTensor,
+                 lr: float,
+                 beta1: float,
+                 beta2: float,
+                 weight_decay: float = 0,
+                 errors: Optional[torch.Tensor] = None,
+                 fused: bool = True) -> None:
+
+    if momentums.is_quantized() and fused:
```
vchiley, line 396: seems like a value error is brewing... something like:
```
if fused and not momentums.is_quantized():
    ValueError('requested fused variant of lion8b, but momentum is not quantized.')
```

llmfoundry/optim/lion8b.py
```diff
@@ -0,0 +1,417 @@
...
+
+    def set_data(self, data: torch.Tensor) -> None:
+        if not (self._try_quantize and data.is_cuda):
+            self.data = data.to(dtype=torch.float32)
+            self.quantized = None
+            self.scales = None
+        else:
+            self.data = None
+            assert self._f_encode is not None  # pyright
+            self.quantized, self.scales = self._f_encode(data)
```
vchiley, line 259: flipping the if else makes more sense (the `not` can throw ppl off)

Should an error be thrown if `self._try_quantize and not data.is_cuda`?

llmfoundry/optim/lion8b.py
```diff
@@ -0,0 +1,417 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple
+
+import torch
+
+_KEY_MOMENTUM = 'exp_avg'
```
vchiley, line 8: just because its what pytorch does, I'd skip the keys and just use the names throughout (eg [adam](https://github.com/pytorch/pytorch/blob/main/torch/optim/adam.py))

llmfoundry/optim/lion8b.py
dblalock: Flipped the if and else and made this throw.

tests/test_lion8b.py
dblalock: I added [a test](https://github.com/mosaicml/llm-foundry/blob/abdc6a6d883e795a63c63c9e11a6e445c04a94fa/tests/test_lion8b.py#L401) for FSDP state dict saving and loading with 2 gpus and all three sharding strategies. @eracah correct me if I'm wrong, but I think testing elastic sharding is gonna be a much more involved process. 

Since this is already a 1k line PR, I vote for adding the elasticity test in a follow-up PR if we determine we need it.

llmfoundry/utils/config_utils.py
```diff
@@ -116,6 +116,25 @@ def process_init_device(model_cfg: DictConfig, fsdp_config: Optional[Dict]):
             # Set defaults for mixed initialization
             fsdp_config.setdefault('use_orig_params', False)
             fsdp_config.setdefault('load_monolith_rank0_only', True)
+
+    # no mixed precision needed for weights when they're already 16 bits
+    master_dtype = model_cfg.get('master_weights_dtype')
+    small_dtypes = ('bf16', 'f16', 'float16', 'bfloat16', 'amp_fp16',
```
vchiley, line 122: here and in train.py, you define magic strings (representing dtypes)
can these be thrown into a `llmfoundry/util/precision.py` files (or something like that)

I dont like that composer already has things like this
we can do better.

tests/test_lion8b.py
```diff
@@ -0,0 +1,549 @@
...
+    assert times[True] < times[False] + atol
+    assert times[True] < times['NA'] + atol
+    assert times['ecc'] < times['NA'] + atol
+
+    if False:  # change to True to check on thruput
+        print('')
+        print('time fused (ms):       ', times[True] * 1e3)
+        print('time fused+ecc (ms):   ', times['ecc'] * 1e3)
+        print('time unfused (ms):     ', times[False] * 1e3)
+        print('time unquantized (ms): ', times['NA'] * 1e3)
```
vchiley, line 549: iirc pytest doesn't print stuff out unless you add a  -s flag. ie just printing this probably wont hurt

vchiley:
**High level design point**
If `DecoupledLionW_8bit` is requested, it kind of seems like it should always be quantized.
If people are running on CPU or do not ask for the quantized version, they should be running `DecoupledLionW` not `DecoupledLionW_8bit`...

Should we make sure that the state dict of `DecoupledLionW_8bit` can be loaded into `DecoupledLionW` (and vice versa), then error out directing people to use `DecoupledLionW` if they are in a setting where `DecoupledLionW_8bit` isn't available.
This setup would mean that we can get rid of the _MaybeQuantizedTensor cls (as well as potentially some other code)

Yes this point should have been brought up a LONG time ago. I'm not saying we should do this, just opening the discussion.
The downside is obviously that it would be tougher to test quantized vs not quantized.
(The answer to this will probably be: we're going to enable the non-quantized version in the `DecoupledLionW_8bit` class as well 😄 )


Also can the state dict of `DecoupledLionW_8bit` can be loaded into `DecoupledLionW` (and vice versa)?

tests/test_lion8b.py
```diff
@@ -0,0 +1,549 @@
...
+
+# run just this test with:
+# python3 -m composer.cli.launcher -n 2 --master_port 26000 -m pytest -m gpu tests/test_lion8b.py::test_fsdp_save_load  # noqa
+@pytest.mark.gpu
+@pytest.mark.world_size(2)
+@pytest.mark.parametrize('dtype', _FLOAT_DTYPES)
+@pytest.mark.parametrize('use_errors', [False, True])
+@pytest.mark.parametrize('state_sharding',
+                         [_FULL_STATE, _SHARDED_STATE, _LOCAL_STATE])
+def test_fsdp_save_load(dtype: torch.dtype, use_errors: bool,
```
eracah, line 409: nice. really good coverage test! any chance you can add a test similar to this where you actually pass mod and opt to `Trainer` and train for a batch or two?

dblalock: How important do you estimate this is? We've used it for a bunch of training runs and, given the way the optimizer works, FSDP should really have no idea it's different from any other optimizer. Trying to scope this down because it's already a 1kloc PR that's been "done" for a week already and research people are using it from branches and stuff right now

eracah: ok if you've done manual tests then it's probably fine

dblalock:
@vchiley How would you feel about just making this *be* DecoupledLionW in a future PR? So we'd have one optimizer with a `quantize` flag instead of two classes?

vchiley:
@bmosaicml originally wrote lion implementation
@bmosaicml do you have an opinion on this?

llmfoundry/optim/lion8b.py
```diff
@@ -0,0 +1,429 @@
...
+
+def lion_step_unfused(grads: torch.Tensor,
+                      weights: torch.Tensor,
+                      momentums: torch.Tensor,
+                      lr: float,
+                      beta1: float,
+                      beta2: float,
+                      weight_decay: float = 0) -> torch.Tensor:
+    # f32 cast to match fused impl + for compatibility with f32 grads or weights
+    momentums = momentums.to(dtype=torch.float32)
```
bmosaicml, line 317: Just for my own comprehension - every parameter is upcast during the lerp and then downcast before moving onto the next parameter?

dblalock: Yes, basically. The parameters are kept in their original dtype, but we ensure the momentums are always f32 during the update (even if they were quantized). If they weren't quantized, this upcast is a no-op.

dblalock:
@bmosaicml so are you good with "yes, leave the quantization flag and make this the single LION implementation"?

llmfoundry/utils/config_utils.py
dblalock: turned this into [a JIRA issue](https://mosaicml.atlassian.net/browse/RESEARCH-804)

vchiley (approved):

tests/test_lion8b.py
dblalock: done

PR #515, opened 2023-08-10
Pr eval lora

danbider:
Modified from @SingL3 [PR](https://github.com/mosaicml/llm-foundry/pull/433) to pin down peft version. 

Previously saw conflicts between the `huggingface/peft/main` and the `accelerate` library that was pinned down in llm-foundry.

Tested this on an interactive session with:
* hf lora module as tested by @SingL3.
* a Composer checkpoint of a trained peft/lora model (trained on pile-of-law). 
eval works fine in both cases.


codestar12 (approved):
LGTM

PR #517, opened 2023-08-10
Fix training integration test

j316chuck:
This PR fixes the `test_training` integration test for both cpu and gpu training runs. 

Before, this test was skipped (x-fail) since the `my-copy-c4` dataset did not exist. This PR adds the logic to create a mocked `my-copy-c4` dataset, kicks of a simple training run, and checks that the codebase reproduces a golden loss score. 

Before: 
- https://github.com/mosaicml/llm-foundry/actions/runs/5800250754/job/15721956423?pr=510 
```
tests/test_training.py::test_train[None-cpu] XFAIL (c4 dataset not s...) [ 97%]
tests/test_training.py::test_train[None-cuda] SKIPPED (testing with ...) [ 97%]
tests/test_training.py::test_train[0.036-cpu] XFAIL (c4 dataset not ...) [ 98%]
tests/test_training.py::test_train[0.036-cuda] SKIPPED (testing with...) [ 98%]
tests/test_training.py::test_train[inv_sqrt_d_model-cpu] XFAIL (c4 d...) [ 99%]
tests/test_training.py::test_train[inv_sqrt_d_model-cuda] SKIPPED (t...) [100%]

```

After: 
- https://github.com/mosaicml/llm-foundry/actions/runs/5826923270/job/15801908817?pr=517
```
2023-08-10 23:12:52 [    INFO] Setting seed to 17 (reproducibility.py:159)
2023-08-10 23:12:52 [    INFO] Added key: store_based_barrier_key:1 to store for rank: 0 (distributed_c10d.py:319)
2023-08-10 23:12:52 [    INFO] Rank 0: Completed store-based barrier for key:store_based_barrier_key:1 with 1 nodes. (distributed_c10d.py:353)
2023-08-10 23:12:52 [    INFO] The number of tokens in the tokenizer is less than the number of tokens in the model. You may want to resize the model embeddings to 50277 from 50368 by calling `model.resize_token_embeddings(len(tokenizer))` before calling the `HuggingFaceModel` constructor. The vocab size is sometimes intentionally set to a multiple of 32 or 64 to improve performance. (huggingface.py:111)
2023-08-10 23:12:52 [    INFO] Setting seed to 17 (reproducibility.py:159)
2023-08-10 23:12:52 [    INFO] Run name: llm (trainer.py:1000)
2023-08-10 23:12:52 [    INFO] Stepping schedulers every batch. To step schedulers every epoch, set `step_schedulers_every_batch=False`. (trainer.py:99)
2023-08-10 23:12:52 [    INFO] Setting seed to 17 (trainer.py:1380)
2023-08-10 23:12:52 [    INFO] Setting seed to 17 (reproducibility.py:159)
2023-08-10 23:12:52 [    INFO] Using precision Precision.FP32 (trainer.py:1913)
PASSED                                                                   [ 99%]
```

### DIFF
--- a/llmfoundry/data/text_data.py
+++ b/llmfoundry/data/text_data.py
@@ -8,7 +8,6 @@ from itertools import islice
 from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Union
 
 import numpy as np
-import streaming.base.util
 import torch
 import transformers
 from omegaconf import DictConfig
@@ -240,9 +239,6 @@ def build_text_dataloader(
             # stream is the streams kwargs
             # fwd all kwargs with **stream allows streaming to check args
             streams.append(Stream(**stream))
-
-    # clean up stale shared memory
-    streaming.base.util.clean_stale_shared_memory()
 
     # build dataset potentially with streams
     dataset = StreamingTextDataset(
--- a/llmfoundry/data/text_data.py
+++ b/llmfoundry/data/text_data.py
@@ -8,6 +8,7 @@ from itertools import islice
 from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Union
 
 import numpy as np
+import streaming.base.util
 import torch
 import transformers
 from omegaconf import DictConfig
@@ -240,7 +241,7 @@ def build_text_dataloader(
             # fwd all kwargs with **stream allows streaming to check args
             streams.append(Stream(**stream))
 
-    import streaming.base.util
+    # clean up stale shared memory
     streaming.base.util.clean_stale_shared_memory()
 
     # build dataset potentially with streams
--- a/llmfoundry/data/text_data.py
+++ b/llmfoundry/data/text_data.py
@@ -240,6 +240,9 @@ def build_text_dataloader(
             # fwd all kwargs with **stream allows streaming to check args
             streams.append(Stream(**stream))
 
+    import streaming.base.util
+    streaming.base.util.clean_stale_shared_memory()
+
     # build dataset potentially with streams
     dataset = StreamingTextDataset(
         tokenizer=tokenizer,
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -170,6 +170,7 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
 
 
 def main(cfg: DictConfig) -> Trainer:
+    validate_config(cfg)
     # Filter deprecation warning from torch internal usage
     warnings.filterwarnings(
         action='ignore',
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -170,7 +170,9 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
 
 
 def main(cfg: DictConfig):
+    # Check for incompatibilities between the model and data loaders
     validate_config(cfg)
+
     # Filter deprecation warning from torch internal usage
     warnings.filterwarnings(
         action='ignore',
@@ -330,6 +332,7 @@ def main(cfg: DictConfig):
 
     print('Starting training...')
     trainer.fit()
+
     print('Done.')
 
 
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -169,10 +169,7 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
         raise ValueError(f'Not sure how to build dataloader with config: {cfg}')
 
 
-def main(cfg: DictConfig):
-    # Check for incompatibilities between the model and data loaders
-    validate_config(cfg)
-
+def main(cfg: DictConfig) -> Trainer:
     # Filter deprecation warning from torch internal usage
     warnings.filterwarnings(
         action='ignore',
@@ -332,8 +329,8 @@ def main(cfg: DictConfig):
 
     print('Starting training...')
     trainer.fit()
-
     print('Done.')
+    return trainer
 
 
 if __name__ == '__main__':
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -169,7 +169,7 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
         raise ValueError(f'Not sure how to build dataloader with config: {cfg}')
 
 
-def main(cfg: DictConfig) -> Trainer:
+def main(cfg: DictConfig):
     validate_config(cfg)
     # Filter deprecation warning from torch internal usage
     warnings.filterwarnings(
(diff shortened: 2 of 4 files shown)


llmfoundry/data/text_data.py
```diff
@@ -240,6 +240,9 @@ def build_text_dataloader(
             # fwd all kwargs with **stream allows streaming to check args
             streams.append(Stream(**stream))
 
+    import streaming.base.util
```
dakinggg, line 243: This import can be at the top of the file

tests/test_training.py
```diff
@@ -52,30 +86,40 @@ def gpt_tiny_cfg(conf_path: str = 'scripts/train/yamls/pretrain/mpt-125m.yaml'):
...
-        "Using the 'grad_clip_norm' field in Trainer is deprecated. Please usethe GradientClipping Algorithm in composer.algorithms.gradient_clipping."
-    )
-
-    test_cfg = gpt_tiny_cfg(
-        conf_path='scripts/train/yamls/pretrain/mpt-125m.yaml')
-    test_cfg.eval_subset_num_batches = 2
-    if logit_scale:
-        test_cfg.model.logit_scale = logit_scale
+def test_train(device: str):
+    create_c4_dataset_xsmall()
```
dakinggg, line 90: should probably `reproducibility.seed_all` here to ensure fixed randomness.

dakinggg (approved):

llmfoundry/data/text_data.py
```diff
@@ -240,6 +240,9 @@ def build_text_dataloader(
...
+    streaming.base.util.clean_stale_shared_memory()
```
vchiley, line 244: Is this safe?

What happens if I'm running 2 jobs on the same instance / gpu? Will the second one I launch kill the first since it'll clean up its dataloader?

llmfoundry/data/text_data.py
vchiley: ➕

scripts/train/train.py
```diff
@@ -169,10 +169,7 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
         raise ValueError(f'Not sure how to build dataloader with config: {cfg}')
 
 
-def main(cfg: DictConfig):
-    # Check for incompatibilities between the model and data loaders
-    validate_config(cfg)
```
vchiley, line 174: BLOCKING COMMENT:

`validate_config(cfg)` should still be called.

tests/test_training.py
vchiley: main seeds_all right away, but I still agree, its not a bad idea to do it here.

tests/test_training.py
```diff
@@ -52,30 +86,40 @@ def gpt_tiny_cfg(conf_path: str = 'scripts/train/yamls/pretrain/mpt-125m.yaml'):
...
+        test_cfg.fsdp_config = None
+        test_cfg.model.attn_config.attn_impl = 'torch'
+        test_cfg.model.loss_fn = 'torch_crossentropy'
         test_cfg.precision = 'fp32'
 
-    main(test_cfg)
+    trainer = main(test_cfg)
+    trainer_loss: float = trainer.state.loss.item()  # type: ignore
+
+    expected_golden_loss = 11.80
```
vchiley, line 123: non-blocking comment

The magic number test seems like a recipe for future failures...

vchiley (changes requested):

llmfoundry/data/text_data.py
j316chuck: fixed

llmfoundry/data/text_data.py
dakinggg: Actually, yeah I agree with vitaliy. Can you do this cleanup in the test only please?

scripts/train/train.py
dakinggg: ➕ thanks for catching

tests/test_training.py
dakinggg: Yeah let's see how it goes. Easy enough to fix if it is flaky

llmfoundry/data/text_data.py
j316chuck: @karan6181 can you chime in on if this command is safe for 2 jobs? 

For more context, I couldn't create both the streaming dataset and run the training pipeline without errors without this line (most likely because stale shared memory was left over).

tests/test_training.py
j316chuck: this is tested in main so probably don't want to add here imo as it would be overwritten anyways

llmfoundry/data/text_data.py
karan6181: If we are running more than one job, calling `streaming.base.util.clean_stale_shared_memory()` would be devastating, it will close existing shared memory which is an unintended behavior. Ideally, one should call this method only once during their entire run. 

>For more context, I couldn't create both the streaming dataset and run the training pipeline without errors without this line (most likely because stale shared memory was left over).

If the job finishes successfully, then I would expect perfect clean up and you can run the same job on the same machine again. What error do you see when you don't include `streaming.base.util.clean_stale_shared_memory()` ?

tests/test_training.py
j316chuck: agreed for flakiness; let's remove it, I'll just add an assert True at the end

scripts/train/train.py
j316chuck: thanks; fixed this

llmfoundry/data/text_data.py
j316chuck: Thanks @karan6181 resolved offline by creating separate datasets on local disk based on the test device

tests/test_training.py
```diff
@@ -52,30 +98,8 @@ def gpt_tiny_cfg(conf_path: str = 'scripts/train/yamls/pretrain/mpt-125m.yaml'):
...
-        )
-        test_cfg.model.init_device = 'cpu'
-        test_cfg.model.attn_impl = 'torch'
-        test_cfg.precision = 'fp32'
-
+def test_train(device: str):
+    dataset_name = create_c4_dataset_xsmall(device)
+    test_cfg = gpt_tiny_cfg(dataset_name, device)
     main(test_cfg)
+    assert True, 'training crashed somewhere'
```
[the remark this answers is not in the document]
j316chuck: removed

tests/test_training.py
```diff
@@ -52,30 +98,8 @@ def gpt_tiny_cfg(conf_path: str = 'scripts/train/yamls/pretrain/mpt-125m.yaml'):
...
-        pytest.xfail(
-            'FSDP in PyTorch 1.13 does not support precision `Precision.FP32` with sharding_strategy `FULL_SHARD.`'
-        )
-        test_cfg.model.init_device = 'cpu'
-        test_cfg.model.attn_impl = 'torch'
-        test_cfg.precision = 'fp32'
-
+def test_train(device: str):
+    dataset_name = create_c4_dataset_xsmall(device)
+    test_cfg = gpt_tiny_cfg(dataset_name, device)
```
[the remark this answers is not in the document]
j316chuck: fixed

vchiley (approved):

PR #518, opened 2023-08-10
Monkeypatch flash attention for llama2 models

dakinggg:
This PR adds a monkeypatch of triton flash attention for llama2 models. If we start doing this for more models we can try to generalize into an algorithm, but until that time I think this monkeypatch is good enough.

TODO:
- [ ] rebase after #492 is merged
- [ ] Copy profiling results for 7b
- [ ] Copy profiling results for 70b


dakinggg:
Closing in favor of #520

PR #523, opened 2023-08-12
Update max duration due to mcli api change

mvpatel2000:
Update max duration due to mcli api change in pytest. API now accepts hours instead of seconds


vchiley (approved):
could you cite the api change in the pr description

PR #526, opened 2023-08-16
Adapt composer -> HF conversion script to support all causal lms

dakinggg:
Adapts the composer -> hf conversion script to support all hf causal lms instead of just MPT. Also removes one of the helper functions that is now fully implemented in composer. The rest can move to composer after a couple more bugs are fixed upstream.

Evidence that the llama2 test passes:
Screenshot 2023-08-15 at 1 36 48 PM

Manual tested training, converting, and loading back in a llama2 model as well.


mvpatel2000 (approved):
> The rest can move to composer after a couple more bugs are fixed upstream.

what are the bugs?

dakinggg:
I have jira for them, but its with saving the tokenizer for llama2, can't use a temp dir, and need to not assume that all filenames are unique

PR #534, opened 2023-08-17
add formatting for llama2chat

samhavens:
This adds formatting classes for ChatML (MPT) and Llama2 chat. In `hf_chat.py` we automatically infer the formatting needed based on model name for these two formatting types.

Why? This makes it a lot easier to test MPT-chat and llama2 chat models with our `hf_chat.py` script. Also, these formatters can be useful for storing chat data in a syntaxless way (list of lists) and easily reformatting for ChatML or llama2 style chat finetuning. It also allows for easy translation from llama2 formatting to ChatML formatting.


samhavens:
`tokenizers` added functionality that covers this behavior

PR #547, opened 2023-08-23
Move the model creation to the last step before trainer creation

dakinggg:
The model download from huggingface is often quite slow, so this PR moves the model creation to the last step before the trainer creation. This way any other misconfigurations or dataset init validation will fail faster without having to wait for the model download from huggingface.

Before with bad data path: `fast-fail-data-before-1-5KeAKO` (fails after model download)
After with bad data path: `fast-fail-data-after-1-HMYrlQ` (fails before model download)
After with normal data path: `fast-fail-data-normal-3-5YV31Z` (trains and evals as normal)

Closes [GRT-2320](https://mosaicml.atlassian.net/browse/GRT-2320)

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -390,25 +390,6 @@ def main(cfg: DictConfig):
     # Build tokenizer
     tokenizer = build_tokenizer(tokenizer_config)
 
-    # Build Model
-    print('Initializing model...')
-    with init_context:
-        if lora_config is not None:  # frozen model + trainable lora modules
-            model: ComposerHFCausalLM = build_composer_peft_model(
-                model_config.pretrained_model_name_or_path, lora_config['args'],
-                tokenizer)
-            print_trainable_parameters(model)  # should not be 100%
-        else:  # standard model
-            model = build_composer_model(model_config, tokenizer)
-
-    # Log number of parameters
-    n_params = sum(p.numel() for p in model.parameters())
-    logged_cfg.update({'n_params': n_params})
-
-    # Optimizer
-    optimizer_name: str = optimizer_config.pop('name')
-    optimizer = build_optimizer(model, optimizer_name, optimizer_config)
-
     # Scheduler
     scheduler_name: str = scheduler_config.pop('name')
     scheduler = build_scheduler(scheduler_name, scheduler_config)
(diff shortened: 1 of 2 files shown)


scripts/train/train.py
```diff
@@ -442,22 +423,50 @@ def main(cfg: DictConfig):
...
+    optimizer = build_optimizer(model, optimizer_name, optimizer_config)
+
+    # Now add the eval metrics
+    if eval_loader_config is not None:
+        assert eval_loader is not None
+        assert model.train_metrics is not None
+        eval_metric_names = list(model.train_metrics.keys())
+        eval_loader.metric_names = eval_metric_names
+        evaluators = [eval_loader
+                     ] + evaluators  # Put the base eval_loader first
```
mvpatel2000, line 468:
```suggestion
        evaluators.insert(eval_loader, 0)  # Put the base eval_loader first
```
I think this is a bit easier to parse and fits on one line. Optional

mvpatel2000 (approved):
LGTM. Good idea

scripts/train/train.py
```diff
@@ -442,22 +423,50 @@ def main(cfg: DictConfig):
...
+    n_params = sum(p.numel() for p in model.parameters())
+    logged_cfg.update({'n_params': n_params})
+
+    # Optimizer
+    optimizer_name: str = optimizer_config.pop('name')
+    optimizer = build_optimizer(model, optimizer_name, optimizer_config)
+
+    # Now add the eval metrics
+    if eval_loader_config is not None:
+        assert eval_loader is not None
```
j316chuck, line 463: nit: should we raise a ValueError here?

tests/test_train_inputs.py
```diff
@@ -17,6 +18,38 @@
 from scripts.train.train import main  # noqa: E402
 
 
+def make_fake_index_file(path: str) -> None:
```
j316chuck, line 21: Thanks making this mock data function!

j316chuck (approved):
lgtm ✅

scripts/train/train.py
dakinggg: nah, this is just for pyright

PR #549, opened 2023-08-24
Fix propagation of `drop_last` and add error message when it would produce no batches

dakinggg:
`drop_last` was not correctly propagated to the HF finetuning dataloader, this fixes that. It also adds a nice error message if your settings would result in 0 batches being produced, as this is not a valid setting (and currently causes some cryptic errors in composer).


mvpatel2000 (approved):
Love friendly errors.

dakinggg:
Converted back to draft because there was a bunch of stuff wrong with this. Sorry about that.

tests/test_dataloader.py
```diff
@@ -282,3 +285,69 @@ def test_finetuning_dataloader(decoder_only_format: bool,
...
+
+    expected_keys = ['input_ids', 'attention_mask', 'labels']
+    expected_keys += ['bidirectional_mask']
+
+    error_context = contextlib.nullcontext()
+    if (dist.get_world_size() * device_batch_size > dataset_size) and drop_last:
+        error_context = pytest.raises(ValueError, match='Your dataset')
+
+    with error_context:
+        _ = build_finetuning_dataloader(cfg, tokenizer, device_batch_size)
```
mvpatel2000, line 350:
```suggestion
        build_finetuning_dataloader(cfg, tokenizer, device_batch_size)
```
not really necessary. maybe fine to be explicit ur ignoring it

mvpatel2000 (approved):

PR #550, opened 2023-08-24
Fix conversion of MPT to HF

eldarkurtic:
Converting MPT checkpoint to HF breaks with:
```shell
*** TypeError: 'MptAttentionConfig' object does not support item assignment
```
and this PR should fix it.


dakinggg:
Hey @eldarkurtic, this is actually a different between the MPT implementation in LLM foundry and the MPT implementation in the `transformers` library (maintained by Hugging Face). So your current fix would break conversions for models trained with the LLM foundry model implementation. To fix properly you would need to conditionally check which implementation it is.

dakinggg:
Closing in favor of #561

PR #560, opened 2023-08-25
Upgrade composer version

dakinggg:
new regression group looks good: https://wandb.ai/mosaic-ml/llm-foundry-regressions/groups/09-05-2023%3A%3A19%3A27%3A44%3A%3A2a791a1ceb7ac620ad3149cc1792b6f1fcc815e7/workspace?workspace=user-danielking


irenedea:
@dakinggg What should we generally look for with the regression tests you posted? I suppose decreasing loss for training?

dakinggg:
I'm just comparing to my previous runs of them. Each run is a group within that wandb project. Once Chuck's stuff is done they will be compared against a fixed "golden" run.

dakinggg:
So in this case I'm just seeing that the curves and number are identical to the ones I ran before.

dakinggg:
Although one of the runs actually failed...so looking into that now

dakinggg:
Ok one of them didn't actually ever run and I just missed it on my original PR. Gonna update in this PR.

dakinggg:
Ok, the failing regression test was user erorr. updated to mixed init. I'm comfortable merging given that validation done. We will set up proper regression tests once Chuck's stuff is ready.

PR #570, opened 2023-08-31
Add script for MDS conversion of bucket of text files

irenedea:
Similar to the other data conversion scripts in foundry, this one is aimed at a continued pretraining API, which will accept a remote bucket containing text files and pretokenconcat and convert to mds for continued pretraining. 

https://databricks.atlassian.net/browse/GRT-2273

Manual Test

* Trained mpt-125m on the first 100 text files in a remote s3 bucket for 1 epoch (i.e. 6 batches)
* Text files were processed with 4 processes, EleutherAI/gpt-neox-20b tokenizer
* Tested with both mds shards uploaded to s3 and local mds shards.

https://gist.github.com/irenedea/7256479e61519b73bf7f231c7597610c

total tokens found when writing to mds: 3028992 (This is equivalent to 5 full batches with a 6th batch that has remaining samples) 

Without data duplication (StreamingDatatset `num_canonical_nodes=1`):
`time/token` is 3028992 at the end of training

With data duplication:
`time/token` is 3145728 at the end of training

num of tokens for 6 full batches: 256 *6 * 2048 = 3145728


scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,430 @@
...
+def get_object_names(input_folder: str) -> List[str]:
+    object_store = maybe_create_object_store_from_uri(input_folder)
+    if object_store is not None:
+        _, _, folder_prefix = parse_uri(input_folder)
+        names = [
+            name for name in object_store.list_objects(folder_prefix)
+            if name.endswith('.txt')
+        ]
+    else:
+        # input_folder is a local folder
```
irenedea, line 154: Do we want to support local input folders?

setup.py
```diff
@@ -47,7 +47,7 @@
 ]
 
 install_requires = [
-    'mosaicml[libcloud,wandb,mlflow]>=0.15.0,<0.16',
+    'mosaicml[libcloud,wandb,mlflow]>=0.16,<0.17',
```
irenedea, line 50: @dakinggg Is it okay to upgrade the mosaicml version? We need at least 0.16 for ObjectStore.list_objects

scripts/data_prep/convert_text_to_mds.py
irenedea: yes we do, both local input and output folders

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,329 @@
...
+):
+    object_store = maybe_create_object_store_from_uri(input_folder)
+    _, _, folder_prefix = parse_uri(input_folder)
+
+    # Download file_names
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        downloading_iter = DownloadingIterable(file_names, folder_prefix,
+                                               tmp_dir, object_store)
+        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
+        tokenizer.model_max_length = 5000000000
```
irenedea, line 178: add comment to say this is a hack to get rid of a HF tokenizer warning

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,329 @@
...
+    if output_object_store is not None:
+        # Download and read the done file from the remote object store
+        _, _, output_folder_prefix = parse_uri(output_root)
+        try:
+            with tempfile.TemporaryDirectory() as tmp_dir:
+                done_file = os.path.join(tmp_dir, done_file_name)
+                output_object_store.download_object(
+                    os.path.join(output_folder_prefix, done_file_name),
+                    done_file)
+                done_file_contents = open(done_file).read().splitlines()
```
dakinggg, line 221: context managers on the opens in this function?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,329 @@
...
+         input_folder=args.input_folder,
+         concat_tokens=args.concat_tokens,
+         eos_text=args.eos_text,
+         bos_text=args.bos_text,
+         no_wrap=args.no_wrap,
+         max_workers=args.max_workers,
+         compression=args.compression,
+         processes=args.processes,
+         reprocess=args.reprocess,
+         args_str=str(args))
```
dakinggg, line 329: do we want to include all the args here? e.g. changing the max workers shouldn't affect whether we need to reprocess, right?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,329 @@
...
+        help=
+        'The text to append to each example to separate concatenated examples')
+    parser.add_argument(
+        '--no_wrap',
+        default=False,
+        action='store_true',
+        help=
+        'Whether to let text examples wrap across multiple training examples')
+    parser.add_argument(
+        '--processes',
```
dakinggg, line 77: maybe change the name of this to differentiate more clearly from the `max_workers` arg?

tests/test_convert_text_to_mds.py
```diff
@@ -0,0 +1,198 @@
...
+        max_workers=1,
+        compression='zstd',
+        processes=processes,
+        args_str='Namespace()',
+        reprocess=False,
+    )
+
+
+# Mock starmap with no multiprocessing
+def _mock_starmap(func: Callable, args: Iterable):
```
dakinggg, line 77: any reason this actually needs to be mocked?

tests/test_convert_text_to_mds.py
```diff
@@ -0,0 +1,198 @@
...
+    assert mock_object_store.download_object.call_count == n_text_files + 1  # text files + done file
+    assert mock_object_store.upload_object.call_count == processes + 2  # shard per process + done file + index.json
+
+    if processes > 1:
+        merge_shard_groups.assert_called_once()
+
+    total_object_names = 0
+    for call_args in download_and_convert.call_args_list:
+        object_names = call_args[0][0]
+        assert (len(object_names) == n_text_files / processes
```
dakinggg, line 138: can we test when this is ragged too? like object names doesn't evenly divide into processes groups

dakinggg (commented):
I think you're already in process of doing this, but would like a manual test that this script produces data we can train on, and expected dataset size looks right before merging.

scripts/data_prep/utils.py
```diff
@@ -0,0 +1,174 @@
...
+            info['raw_data']['basename'] = new_basename
+
+            if info['zip_data'] is not None:
+                old_basename = info['zip_data']['basename']
+                new_basename = with_id(old_basename, shard_id)
+                info['zip_data']['basename'] = new_basename
+
+            old_filename = os.path.join(subdir, old_basename)
+            new_filename = os.path.join(root, new_basename)
+            assert not os.rename(old_filename, new_filename)
```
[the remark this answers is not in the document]
irenedea: It checks that the rename has completed because the return is None. I took it from the mds example code, but I think we can remove it because os.rename should raise an exception if there are issues.

tests/test_convert_text_to_mds.py
irenedea: multi processing unfortunately messes with the function call counting with mocking

scripts/data_prep/convert_text_to_mds.py
irenedea: how about changing max_workers to max_mds_writer_workers?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,329 @@
...
+                       columns=columns,
+                       max_workers=max_workers,
+                       compression=compression) as out:
+            for sample in tqdm(samples):
+                out.write(sample)
+
+
+def is_remote_path(path: str) -> bool:
+    backend, bucket, _ = parse_uri(path)
+    return backend != '' or bucket != ''
```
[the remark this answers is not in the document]
irenedea: I arrived to that logic since a local path (according to the doc string of parse_uri) is where `backend == '' and bucket == ''`, so `is_remote_path => not (backend == '' and bucket == '') => backend != '' or bucket != ''`

scripts/data_prep/convert_text_to_mds.py
irenedea: Good point, I suppose we need to include all the args except max_workers, reprocess?

tests/test_convert_text_to_mds.py
irenedea: hmm, i think i might know how to get around that after taking a second look... let's see

irenedea: yeah, that's going to be tricky to get working with nice mocking/call counts because memory isn't shared between processes. So we could count the calls by writing things to files instead of using the nice Mock.call_count if it's really important for us to test with full multiprocessing. wdyt?

dakinggg: ah thats ok. we can rely on manual test for this.

scripts/data_prep/convert_text_to_mds.py
dakinggg: sgtm

scripts/data_prep/convert_text_to_mds.py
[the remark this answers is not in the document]
dakinggg: huh, fair enough. not sure if the upstream implementation is wrong or not...but we def need a backend and a bucket for our purposes here, so lets just check that.

scripts/data_prep/convert_text_to_mds.py
dakinggg: I think thats right.

scripts/data_prep/utils.py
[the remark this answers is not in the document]
dakinggg: ah ok. yeah, those don't seem necessary

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,329 @@
...
+    bos_text: str,
+    no_wrap: bool,
+    compression: str,
+    max_workers: int,
+) -> Iterable:
+    objs_per_group = math.ceil(len(object_names) / n_groups)
+    for group, i in enumerate(range(0, len(object_names), objs_per_group)):
+        output_subdir = os.path.join(output_root, str(group))
+        yield (
+            object_names[i:i + objs_per_group],
```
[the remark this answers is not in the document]
irenedea: do you mean min? for preventing to indexing out of bounds?

scripts/data_prep/utils.py
```diff
@@ -0,0 +1,174 @@
...
+    ):
+        """Iterable that downloads files from an object store before yielding.
+
+        If object_store is None, input_folder_prefix is treated as a local path.
+
+        text samples.
+
+        Args:
+            object_names (List[str]): Names of objects to to download
+            input_folder_prefix (str): Object store prefix to download from
```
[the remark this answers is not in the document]
irenedea: That's correct. In this case, it is used to not to determine the remote download path, but to determine the output file path, which is usually going to be within a temporary folder.

I think I can actually just remove the need for the prefix all together and just use the full paths from object names

scripts/data_prep/convert_text_to_mds.py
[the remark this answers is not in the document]
dakinggg: oops yeah. meant min

scripts/data_prep/utils.py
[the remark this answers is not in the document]
dakinggg: ah ok, makes sense.

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,355 @@
...
+    compression: str,
+    processes: int,
+    args_str: str,
+    reprocess: bool,
+):
+    done_file_name = get_done_file_name()
+    is_remote_output = is_remote_path(output_folder)
+
+    object_names = get_object_names(input_folder)
+    object_names = object_names[:10]
```
irenedea, line 281: remove

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,329 @@
...
+
+    if is_remote_output:
+        # Upload the local output to the remote location
+        output_object_store = cast(
+            ObjectStore, maybe_create_object_store_from_uri(output_folder))
+        _, _, output_folder_prefix = parse_uri(output_folder)
+        pattern = os.path.join(local_output_folder, '*')
+        files_to_upload = sorted(glob(pattern))
+
+        # TODO: Use multi threading to upload files?
```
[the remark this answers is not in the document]
irenedea: we could do this, but the functions require state and logger args (init and upload_file). The usage of those args is pretty minimal and close to none...should we just pass dummy objects? I don't really like the idea of relying on understanding the internal implementation to use the functions because it could break easily with a change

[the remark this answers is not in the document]
irenedea: i think a better but more involved solution would be to refactor the composer code to move the logic into a class that doesn't implement LoggerDestination and the LoggerDestination version can call that class... wdyt?

We could keep this for now and then refactor in a separate PR(s)

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,358 @@
...
+        with MDSWriter(out=output_folder,
+                       columns=columns,
+                       max_mds_writer_workers=max_mds_writer_workers,
+                       compression=compression) as out:
+            for sample in tqdm(dataset):
+                total_tokens_bytes += len(sample['tokens'])
+                out.write(sample)
+        total_tokens = total_tokens_bytes / 8
+        print('tokens', total_tokens_bytes, total_tokens)
+        return total_tokens
```
irenedea, line 209:
```suggestion
        print(f'Converting to MDS format...')
        with MDSWriter(out=output_folder,
                       columns=columns,
                       max_mds_writer_workers=max_mds_writer_workers,
                       compression=compression) as out:
            for sample in tqdm(dataset):
                out.write(sample)
```

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,353 @@
...
+
+    if processes > 1:
+        # Download and convert the text files in parallel
+        args = get_task_args(object_names, local_output_folder, input_folder,
+                             processes, tokenizer_name, concat_tokens, eos_text,
+                             bos_text, no_wrap, compression,
+                             max_mds_writer_workers)
+        with ProcessPoolExecutor(max_workers=processes) as executor:
+            print('all tokens',
+                  sum(executor.map(download_and_convert_starargs, list(args))))
```
irenedea, line 300:
```suggestion
            list(executor.map(download_and_convert_starargs, args))
```

llmfoundry/utils/data_prep_utils.py
```diff
@@ -0,0 +1,144 @@
...
+from torch.utils.data import DataLoader, Dataset
+
+
+def build_dataloader(dataset: Dataset,
+                     batch_size: int,
+                     num_workers: Optional[int] = None) -> DataLoader:
+    if num_workers is None:
+        # Multiple workers is only supported on linux machines
+        if 'linux' in platform.platform().lower():
+            num_workers = max(1, psutil.cpu_count())  # type: ignore
```
dakinggg, line 21: is this type ignore removable?

tests/test_convert_text_to_mds.py
```diff
@@ -0,0 +1,185 @@
...
+                              tokenizer_name=tokenizer_name)
+
+    # Check call counts
+    assert download_and_convert.call_count == processes * 2  # called once per process
+    assert mock_object_store.download_object.call_count == n_text_files + 1  # text files + done file
+    assert mock_object_store.upload_object.call_count == processes + 2  # shard per process + done file + index.json
+
+    # Check that the dataset can be used and produces samples with the original text_content
+    tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
+    dataset = StreamingDataset(local=remote_folder)
```
dakinggg, line 160: can we iter the whole thing and assert that we have the right amount of data?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+                        type=str,
+                        required=True,
+                        help='The folder with text files to convert to mds')
+    parser.add_argument('--compression',
+                        type=str,
+                        default='zstd',
+                        help='The compression algorithm to use for MDS writing')
+
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument(
```
dakinggg, line 49: I think this group thing is no longer correct. `--concat_tokens` is a required arg for this script, since this script doesn't support the no concat mode.

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+        default=1,
+        help='The number of processes to use to download and convert the dataset'
+    )
+    parser.add_argument(
+        '--reprocess',
+        type=bool,
+        required=False,
+        default=False,
+        help=
+        'If true, reprocess the input_folder to mds format. Otherwise, only reprocess upon changes to the input folder.'
```
dakinggg, line 90:
```suggestion
        'If true, reprocess the input_folder to mds format. Otherwise, only reprocess upon changes to the input folder or dataset creation parameters.'
```

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+            if name.endswith('.txt')
+        ]
+    else:
+        # input_folder is a local folder
+        names = [
+            text_file for dirpath, _, _ in os.walk(input_folder)
+            for text_file in glob(os.path.join(dirpath, '*.txt'))
+        ]
+    # return names, sizes
+    print(f'Found {len(names)} text files at {input_folder}')
```
dakinggg, line 124: given the logging PR has gone in, switch this to use logging instead of print

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+            max_length=concat_tokens,
+            tokenizer=tokenizer,
+            eos_text=eos_text,
+            bos_text=bos_text,
+            no_wrap=no_wrap,
+        )
+
+        columns = {'tokens': 'bytes'}
+
+        print(f'Converting to MDS format...')
```
dakinggg, line 198: same

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+def write_done_file(folder: str, file_name: str, args_str: str,
+                    object_names: List[str]):
+    with open(os.path.join(folder, file_name), 'w') as done_file:
+        done_file.write('\n'.join([args_str] + object_names) + '\n')
+
+
+def get_done_file_name() -> str:
+    return '.text_to_mds_conversion_done'
```
dakinggg, line 258: maybe just make this a constant at the top of the file?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+    max_mds_writer_workers: int,
+    compression: str,
+    processes: int,
+    args_str: str,
+    reprocess: bool,
+):
+    done_file_name = get_done_file_name()
+    is_remote_output = is_remote_path(output_folder)
+
+    object_names = get_object_names(input_folder)
```
dakinggg, line 278: should we raise an error here if no files are found?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+
+    # Check if the text files in the bucket have already been processed.
+    if not reprocess and is_already_processed(output_folder, done_file_name,
+                                              args_str, object_names):
+        print(
```
dakinggg, line 283: log

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+def _args_str(original_args: Namespace) -> str:
+    """Create a string from the args to determine whether to reprocess.
+
+    Args:
+        original_args (Namespace): args to transform
+    """
+    args = copy.deepcopy(original_args)
+
+    # Remove args that do not affect the final result.
+    delattr(args, 'max_mds_writer_workers')
```
dakinggg, line 334: even though i think this is safe given the flow of this script, can we make this function non destructive? seems unnecessary

dakinggg (commented):
LGTM pending a few small comments. Also I requested review from Karan to make sure the streaming stuff looks ok

tests/test_convert_text_to_mds.py
irenedea: good idea!

llmfoundry/utils/data_prep_utils.py
```diff
@@ -0,0 +1,144 @@
+# Copyright 2022 MosaicML LLM Foundry authors
```
karan6181, line 1: nit: should we update the license to `2023` ? Applicable to all the files in this repository. May not be part of this PR but worth addressing in a separate PR.

llmfoundry/utils/data_prep_utils.py
```diff
@@ -0,0 +1,144 @@
...
+from composer.utils import ObjectStore
+from torch.utils.data import DataLoader, Dataset
+
+
+def build_dataloader(dataset: Dataset,
+                     batch_size: int,
+                     num_workers: Optional[int] = None) -> DataLoader:
+    if num_workers is None:
+        # Multiple workers is only supported on linux machines
+        if 'linux' in platform.platform().lower():
```
karan6181, line 20: Can MAC run this code ? In that case, I would recommend do a `os.name == 'posix'` check. 

macOS and Linux are both POSIX-compliant. Hence, an application written for macOS should run on Linux.

llmfoundry/utils/data_prep_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    else:
+        prefetch_factor = max(1, 2 * batch_size //
+                              num_workers) if num_workers > 0 else 2
+
+    return DataLoader(
+        dataset=dataset,
+        sampler=None,
+        batch_size=batch_size,
+        num_workers=num_workers,
+        prefetch_factor=prefetch_factor,
```
karan6181, line 41: Any reason not using `persistent_workers=True` and `pin_memory=True` ?

scripts/data_prep/convert_text_to_mds.py
irenedea: should be log.info, yeah?

llmfoundry/utils/data_prep_utils.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Returns:
+        str: New basename of file.
+    """
+    parts = basename.split('.')
+    parts[1] = f'{shard_id:05}'
+    return '.'.join(parts)
+
+
+def merge_shard_groups(root: str) -> None:
```
karan6181, line 62: This method removes the sub-directory folder, rename the shard files, and put everything in a root folder. This method would work, but I recommend a below method which actually create a new `index.json` in a root directory and the `basename` field in `index.json` points to the shard filepath of a sub-directory. Also, with this method, you won't need `with_id()` method.

```
def merge_index(root: str) -> None:
    """Create a new index.json in a root directory which points to the sub-directory shard files.

    Args:
        root (str): Root directory.
    """
    shard_id = 0
    infos = []
    for sub_dir in os.listdir(out_root):
        index_filename = os.path.join(root, sub_dir, 'index.json')
        obj = json.load(open(index_filename))
        for info in obj['shards']:
            old_basename = info['raw_data']['basename']
            new_basename = os.path.join(sub_dir, old_basename)
            info['raw_data']['basename'] = new_basename

            if info['zip_data'] is not None:
                old_basename = info['zip_data']['basename']
                new_basename = os.path.join(sub_dir, old_basename)
                info['zip_data']['basename'] = new_basename

            shard_id += 1
            infos.append(info)

    index_filename = os.path.join(root, 'index.json')
    obj = {
        'version': 2,
        'shards': infos,
    }
    text = json.dumps(obj, sort_keys=True)
    with open(index_filename, 'w') as out:
        out.write(text)
```

We will update the jupyter notebook with the above method as well. 

Please verify it by running your script end-to-end.

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+from llmfoundry.data import ConcatTokensDataset
+from llmfoundry.utils.data_prep_utils import (DownloadingIterable,
+                                              merge_shard_groups)
+
+
+def parse_args() -> Namespace:
+    """Parse commandline arguments."""
+    parser = ArgumentParser(
+        description=
+        'Convert text files into MDS format, optionally concatenating and tokenizing'
```
karan6181, line 31: tip: add a trailing comma at the end so that lint re-structure the code uniformly. For example
```suggestion
        'Convert text files into MDS format, optionally concatenating and tokenizing',
```

Please do the same for other argparse as well.

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+        default=1,
+        help='The number of processes to use to download and convert the dataset'
+    )
+    parser.add_argument(
+        '--reprocess',
+        type=bool,
+        required=False,
+        default=False,
+        help=
+        'If true, reprocess the input_folder to mds format. Otherwise, only reprocess upon changes to the input folder or dataset creation parameters.'
```
karan6181, line 92: You can do something like to ensure the code is within certain character limits.

```
help=f'If true, reprocess the input_folder to mds format. Otherwise, ' +
     f'only reprocess upon changes to the input folder or dataset creation parameters.'
```

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+    # now that we have validated them, change BOS/EOS to strings
+    if parsed.bos_text is None:
+        parsed.bos_text = ''
+    if parsed.eos_text is None:
+        parsed.eos_text = ''
+    return parsed
+
+
+def get_object_names(input_folder: str) -> List[str]:
+    object_store = maybe_create_object_store_from_uri(input_folder)
```
karan6181, line 112: Method name is missing

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+            if name.endswith('.txt')
+        ]
+    else:
+        # input_folder is a local folder
+        names = [
+            text_file for dirpath, _, _ in os.walk(input_folder)
+            for text_file in glob(os.path.join(dirpath, '*.txt'))
+        ]
+    # return names, sizes
+    print(f'Found {len(names)} text files at {input_folder}')
```
karan6181: Is `print` a common thing in llm-foundry or one should use `logging.info()` ?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+    n_groups: int,
+    tokenizer_name: str,
+    concat_tokens: int,
+    eos_text: str,
+    bos_text: str,
+    no_wrap: bool,
+    compression: str,
+    max_mds_writer_workers: int,
+) -> Iterable:
+    num_objects = len(object_names)
```
karan6181, line 144: Missing method comments. Please check other methods too!

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+            bos_text=bos_text,
+            no_wrap=no_wrap,
+        )
+
+        columns = {'tokens': 'bytes'}
+
+        print(f'Converting to MDS format...')
+        with MDSWriter(out=output_folder,
+                       columns=columns,
+                       max_mds_writer_workers=max_mds_writer_workers,
```
karan6181, line 203: [MDSWriter()](https://docs.mosaicml.com/projects/streaming/en/stable/api_reference/generated/streaming.MDSWriter.html#streaming.MDSWriter) does not have a `max_mds_writer_workers` parameter !!!.

You should instantiate one `MDSWriter` per process and ensure each process has a unique `output_folder`. Happy to explain you offline on how it works.

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+    done_file_name = get_done_file_name()
+    is_remote_output = is_remote_path(output_folder)
+
+    object_names = get_object_names(input_folder)
+
+    # Check if the text files in the bucket have already been processed.
+    if not reprocess and is_already_processed(output_folder, done_file_name,
+                                              args_str, object_names):
+        print(
+            f'Input folder {input_folder} is already processed at {output_folder} and reprocess is set to False. Set reprocess to True if you would like to force reprocessing.'
```
karan6181, line 282: too long line. please split it.

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -0,0 +1,352 @@
...
+    eos_text: str,
+    bos_text: str,
+    no_wrap: bool,
+    compression: str,
+    max_mds_writer_workers: int,
+):
+    object_store = maybe_create_object_store_from_uri(input_folder)
+
+    # Download file_names
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
karan6181, line 181: Please also validate that `tmp_dir` is different when running with multiple workers !! If you already did, then its fine.

llmfoundry/utils/data_prep_utils.py
dakinggg: IIRC it only worked on linux specifically...but not certain.

llmfoundry/utils/data_prep_utils.py
dakinggg: neither should have any impact here I think...the dataset is only being iterated once and no gpus involved

scripts/data_prep/convert_text_to_mds.py
dakinggg: we _just_ merged a PR to foundry to do proper logging. already commented this on this pr :)

scripts/data_prep/convert_text_to_mds.py
dakinggg: @karan6181 seems like this should have crashed. Why does mdswriter accept arbitrary kwargs?

llmfoundry/utils/data_prep_utils.py
dakinggg: ah yeah, maybe, will figure this out.

scripts/data_prep/convert_text_to_mds.py
dakinggg: `tempfile.TemporaryDirectory()` gives a different dir every time you call it. so should be fine in multiple workers.

scripts/data_prep/convert_text_to_mds.py
irenedea: Yep, I've already removed all prints instead of logging.info

scripts/data_prep/convert_text_to_mds.py
irenedea: Yup! I have verified this.

scripts/data_prep/convert_text_to_mds.py
irenedea: ah yes, good catch! that was from a refactor. MDSWriter takes arbitrary kwargs, so it wasn't caught at runtime. My intention was for that to be max_workers=max_mds_writer_workers. Is this correct? It seems like that might not be necessary because MDSWriter aways writes to a local output folder here and max_workers is stated in the docstring to be for remote uploads. 

> You should instantiate one MDSWriter per process and ensure each process has a unique output_folder. Happy to explain you offline on how it works.


Currently there is one MDSWriter and a unique output folder per process. In `get_task_args` we assign unique subdirectory per process.

llmfoundry/utils/data_prep_utils.py
irenedea: I'm actually going to remove this code altogether. It's from a refactor, but this refactor is now out of scope for this PR since the code no longer uses build_dataloader.

llmfoundry/utils/data_prep_utils.py
irenedea: @karan6181 Okay if I do this in a follow up refactor? I can do that immediately after getting this in, but this change requires a few changes across files to handle paths differently since originally we assume that the output directory is flat.

scripts/data_prep/convert_text_to_mds.py
dakinggg: yeah info seems good

llmfoundry/utils/data_prep_utils.py
karan6181: Sure works for me. Thank You!

scripts/data_prep/convert_text_to_mds.py
karan6181: > seems like this should have crashed. Why does mdswriter accept arbitrary kwargs?

MDSWriter uses kwargs `progress_bar` and `max_workers`. Looks like streaming does not do a strict validation on the incorrect kwargs. I will take a look at it. 

>Currently there is one MDSWriter and a unique output folder per process. In get_task_args we assign unique subdirectory per process.

Make sense. Thank You!

irenedea: Removed the max_workers argument because it is only used for remote uploads and we are writing to a local dir.

karan6181 (commented):
LGTM. Thank You!

dakinggg (approved):
LGTM! Lets start trying it out!

PR #573, opened 2023-09-01
added sampling_method to StreamingTextDataset

snarayan21:
Previously, StreamingTextDataset didn't have the sampling_method arg. Now it does 🤯🤯


llmfoundry/data/text_data.py
```diff
@@ -88,6 +89,7 @@ def __init__(self,
                  shuffle_algo: str = 'py1b',
                  shuffle_seed: int = 9176,
                  shuffle_block_size: int = 1 << 18,
+                 sampling_method: str = 'balanced',
```
dakinggg, line 92: just to double check, this is the same as the default in streaming, right?

dakinggg (commented):
while you're here...could you pipe through here also? https://github.com/mosaicml/llm-foundry/blob/68448b2764cf6988c830e4d55796e6e28cdac20e/llmfoundry/data/finetuning/dataloader.py#L119-L133. I think there actually a few streaming args missing along that code path.

llmfoundry/data/text_data.py
snarayan21: yep same default!

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -120,16 +120,22 @@ def build_finetuning_dataloader(cfg: DictConfig,
             tokenizer=tokenizer,
             local=cfg.dataset.local,
             remote=cfg.dataset.get('remote', None),
-            split=cfg.dataset.get('split'),
-            shuffle=cfg.dataset.get('shuffle', False),
-            predownload=cfg.dataset.get('predownload', 100_000),
-            keep_zip=cfg.dataset.get('keep_zip', False),
+            split=cfg.dataset.get('split', None),
```
dakinggg, line 123: I think we require split to be specified in this code, so lets not change that right now. would need to look at the code more closely and decide.

snarayan21: okay so i'll leave out the split arg then

dakinggg (approved):
LGTM

PR #574, opened 2023-09-01
Add regression tests

dakinggg:
This PR adds a set of regression tests for llm foundry. Right now they can just be run manually with some arguments passed along at the command line. Once the regression runner infra is ready we can connect it with that, but this is still useful for now.

Most recent run group is at: https://wandb.ai/mosaic-ml/llm-foundry-regressions/groups/08-31-2023%3A%3A16%3A30%3A02%3A%3A9ff9dc15978a9f940c0b2a67607301fe9baeac77/workspace?workspace=user-danielking

Along for the ride: Fixed a couple issues with one of the existing mcli yamls.


.github/workflows/regressions.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+import subprocess
+
+dir_path = os.path.dirname(os.path.abspath(__file__))
+regressions_dir = os.path.join(dir_path, 'regression_yamls')
+
+from mcli import RunConfig, create_run
+
+
+def get_configs(cluster: str, mpt_7b_ckpt_path: str, wandb_entity: str,
```
j316chuck, line 15: can `mpt_7b_ckpt_path` live inside the yaml? 

Or is the idea that you might want to pass in separate version of  the mpt_7b_ckpt_path

.github/workflows/regressions.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import datetime
+import os
+import subprocess
+
+dir_path = os.path.dirname(os.path.abspath(__file__))
+regressions_dir = os.path.join(dir_path, 'regression_yamls')
```
j316chuck, line 10: Is there a style rule that global variables should be capital case?

.github/workflows/regressions.py
```diff
@@ -0,0 +1,68 @@
...
+
+    commit_hash = subprocess.check_output(['git', 'rev-parse',
+                                           'HEAD']).strip().decode('utf-8')
+    timestamp = datetime.datetime.now().strftime('%m-%d-%Y::%H:%M:%S')
+    wandb_group = f'{timestamp}::{commit_hash}'
+
+    # make general changes
+    loggers = {
+        'wandb': {
+            'entity': wandb_entity,
```
j316chuck, line 44: this will be hardcoded to a specific user for future regression tests but is fine for now :)

.github/workflows/regressions.py
```diff
@@ -0,0 +1,68 @@
...
+            'project': wandb_project,
+            'group': wandb_group
+        }
+    }
+    for config in all_configs:
+        config.cluster = cluster
+        config.parameters['loggers'] = loggers
```
j316chuck, line 51: won't this override all other loggers? Maybe try appending to loggers and also checking if wandb logger exists to override?

j316chuck (approved):
Thanks for this PR! Will try to build off of this/have a similar structure in the regression `testing` framework.

.github/workflows/regression_yamls/llama2-finetune.yaml
```diff
@@ -0,0 +1,121 @@
...
+  tokenizer_name: meta-llama/Llama-2-13b-hf
+  max_seq_len: 4096
+  global_seed: 17
+
+  run_name: llama2-finetune-regression
+  max_split_size_mb: 512
+
+  model:
+    name: hf_causal_lm
+    init_device: mixed
```
irenedea, line 29: q: what is init_device and why does it vary across these yamls?

.github/workflows/regressions.py
dakinggg: The idea was to not put our cloud storage paths in a public repo. Probably not the most important thing in the world, but would be nice to parametrize it.

.github/workflows/regressions.py
dakinggg: not sure. This file might not get linted, not sure. can make caps tho

.github/workflows/regressions.py
dakinggg: indeed it will. seems reasonable to append instead of overwrite

.github/workflows/regression_yamls/llama2-finetune.yaml
dakinggg: `init_device` controls what device the model is created on before being passed to composer to get FSDP wrapped. The options are:
- `cpu`: init model on cpu on all ranks. 8 copies of model in CPU RAM before being passed to composer
- `meta`: init model as [meta tensors](https://pytorch.org/torchdistx/latest/fake_tensor_and_deferred_init.html). this is very fast, and then allows init to happen directly on GPU during FSDP wrapping (or whenever it is materialized).
- `mixed`: init on cpu on local rank 0, and meta on all other ranks. this allows us to load weights from the huggingface hub and then FSDP wrap without creating 8 copies of the model in CPU RAM. The loaded weights are synced between all ranks during FSDP wrapping.

PR #575, opened 2023-09-01
add fusedce to metrics

dakinggg:


dakinggg:
@nik-mosaic first of all its a draft that I probably won't merge :) but to your answer your question its because the implementation in flash attention doesn't support `sum` reduction for some reason. I added it here to support the use of it in our metric class.

PR #576, opened 2023-09-05
Add handling for various types of malformed finetuning data

dakinggg:
This prevents hangs and confusing errors later on in training.

Along for the ride, pins fsspec due to some bug in huggingface datasets that causes duplicated data: https://github.com/huggingface/datasets/pull/6210

CI will pass whenever github actions comes back up.


tests/test_dataloader.py
```diff
@@ -339,3 +391,85 @@ def test_finetuning_dataloader_small_data(dataset_size: int,
...
+@pytest.mark.parametrize('add_bad_data_error', [True, False])
+def test_malformed_data(
+    add_bad_data_dropped: bool,
+    add_bad_data_error: bool,
+):
+    tokenizer_name = 'mosaicml/mpt-7b'
+    max_seq_len = 2048
+    dataset_size = 5
+    device_batch_size = 5
+    tiny_dataset_folder_path = os.path.join(os.getcwd(), 'test-ift-data-small')
```
irenedea, line 406: nit: maybe use tmp_path fixture if possible here, so you don't need to worry about removing the folder

llmfoundry/data/finetuning/tasks.py
```diff
@@ -52,6 +52,14 @@ def _tokenize_formatted_example(example: Dict[str, Any],
...
             f'from {example=}.'
         )
+    if not isinstance(example['prompt'], str):
+        raise TypeError(
+            f'Unable to tokenize example because "prompt" was not a string. {example=}'
+        )
+    if not isinstance(example['response'], str):
+        raise TypeError(
+            f'Unable to tokenize example because "response" was not a string. {example=}'
+        )
```
alextrott16, line 62: Smart. Thanks for adding this check.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -306,7 +314,18 @@ def dataset_mapper(example: Dict):
                 f'Dropped {examples_removed} examples where the prompt was longer than {max_seq_len}.'
             )
 
-        return prompt_length_filtered_dataset
+        empty_examples_dropped_dataset = prompt_length_filtered_dataset.filter(
+            lambda example: len(example['input_ids']) > 0 and len(example[
+                'labels']) > 0 and not all(token_id == tokenizer.pad_token_id
```
alextrott16, line 319: If `tokenizer.pad_token_id` is not set yet will this be a problem? (I'm guessing no because I think that it will just return `None` but worth considering.)

And should there be a similar check for EOS? Maybe if all the label tokens are EOS or PAD, that should be filtered.

tests/test_dataloader.py
```diff
@@ -339,3 +391,85 @@ def test_finetuning_dataloader_small_data(dataset_size: int,
...
+        expected_num_batches = (dataset_size + 5) // device_batch_size
+
+        actual_num_batches = 0
+        for _ in dl:
+            actual_num_batches += 1
+
+        assert actual_num_batches == expected_num_batches
+
+    if dist.get_global_rank() == 0:
+        shutil.rmtree(tiny_dataset_folder_path)
```
alextrott16, line 475: Will this line be reached if the test fails? If no, is there a way to guarantee that the cleanup still happens?

alextrott16 (approved):
LGTM. Thanks for catching this and adding tests!

Looks good to merge once comments are addressed and tests pass.

irenedea (approved):
LGTM

llmfoundry/data/finetuning/tasks.py
dakinggg: @alextrott16 the critical issue was that if it was all PAD, the collate fn strips pads and then you end with empty target and it crashes. This is not true if its all EOS/BOS, it will train, its just kind of a weird example. wdyt?

tests/test_dataloader.py
dakinggg: it will not, will fix, as Irene also suggested.

tests/test_dataloader.py
dakinggg: yeah makes sense, not sure why I did it the other way.

llmfoundry/data/finetuning/tasks.py
alextrott16: Filtering error-causing examples is definitely a must, and that base is covered. The rest is a judgement call. And maybe you're right that it is better to avoid trying to catch examples that just seem weird. I guess I don't have any super strong feelings there, so I am happy to go with your judgement call.

PR #581, opened 2023-09-06
typecast shuffle_block_size because of issue

codestar12:
There is an issue where yamls sometimes interpret shuffle_block_size as a float. Its not clear why this happens and in what cases but this should at least allow working from fresh branches.


sashaDoubov (approved):
LGTM, this was the same workaround that I had made in the past. We should root cause this going forward tho.

PR #588, opened 2023-09-08
Fix huggingface custom split path issue

dakinggg:
Hugging Face is very particular about how you name things when using a custom split.

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -294,7 +294,7 @@ def _build_hf_dataset_from_remote(
     finetune_dir = os.path.join(
         os.path.dirname(
             os.path.dirname(os.path.dirname(os.path.realpath(__file__)))),
-        'downloaded_finetuning', {cfg.dataset.split})
+        'downloaded_finetuning', cfg.dataset.split)
     os.makedirs(finetune_dir, exist_ok=True)
     for extension in supported_extensions:
         name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -298,7 +298,7 @@ def _build_hf_dataset_from_remote(
     for extension in supported_extensions:
         name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
         destination = str(
-            os.path.abspath(f'{finetune_dir}/{cfg.dataset.split}.{extension}'))
+            os.path.abspath(f'{finetune_dir}/data/{cfg.dataset.split}-00000-of-00001.{extension}'))
         # Since we don't know exactly what the extension will be, since it is one of a list
         # use a signal file to wait for instead of the desired file
         signal_file_path = os.path.join(finetune_dir, '.the_eagle_has_landed')
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -292,13 +292,16 @@ def _build_hf_dataset_from_remote(
     """
     supported_extensions = ['jsonl', 'csv', 'parquet']
     finetune_dir = os.path.join(
-        os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__)))),
+        os.path.dirname(
+            os.path.dirname(os.path.dirname(os.path.realpath(__file__)))),
         f'downloaded_finetuning/{cfg.dataset.split}')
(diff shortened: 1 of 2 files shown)


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -292,13 +292,16 @@ def _build_hf_dataset_from_remote(
     """
     supported_extensions = ['jsonl', 'csv', 'parquet']
     finetune_dir = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)),
-        f'downloaded_finetuning_data/{cfg.dataset.split}')
+        os.path.dirname(
+            os.path.dirname(os.path.dirname(os.path.realpath(__file__)))),
+        f'downloaded_finetuning/{cfg.dataset.split}')
```
irenedea, line 297:
```suggestion
        os.path.dirname(
            os.path.dirname(os.path.dirname(os.path.realpath(__file__)))),
            downloaded_finetuning,
            cfg.dataset.split')
```

irenedea: Add a comment or just make a descriptive var for the base path like:
```
llmfoundry_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -292,13 +292,16 @@ def _build_hf_dataset_from_remote(
...
     os.makedirs(finetune_dir, exist_ok=True)
     for extension in supported_extensions:
         name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
         destination = str(
-            os.path.abspath(f'{finetune_dir}/{cfg.dataset.split}.{extension}'))
+            os.path.abspath(
```
irenedea, line 302: Use os.path.join?
```
os.path.abspath(os.path.join(finetune_dir, 'data', f'{cfg.dataset.split}-00000-of-00001.{extension}'))
```

llmfoundry/data/finetuning/dataloader.py
dakinggg: done.

tests/test_dataloader.py
```diff
@@ -394,6 +394,50 @@ def test_finetuning_dataloader_small_data(dataset_size: int,
...
+
+    cfg = om.create(cfg)
+
+    tokenizer = build_tokenizer(
+        tokenizer_name=tokenizer_name,
+        tokenizer_kwargs={'model_max_length': max_seq_len},
+    )
+
+    expected_keys = ['input_ids', 'attention_mask', 'labels']
+    expected_keys += ['bidirectional_mask']
```
irenedea, line 434: What does this do? Can we remove it?

tests/test_dataloader.py
```diff
@@ -394,6 +394,50 @@ def test_finetuning_dataloader_small_data(dataset_size: int,
...
+
+    _ = build_finetuning_dataloader(cfg, tokenizer, 4)
+
+    if dist.get_global_rank() == 0:
+        shutil.rmtree(tiny_dataset_folder_path)
```
irenedea, line 439: Remove this. tmp_path gets automatically cleaned up

tests/test_dataloader.py
```diff
@@ -394,6 +394,50 @@ def test_finetuning_dataloader_small_data(dataset_size: int,
         shutil.rmtree(tiny_dataset_folder_path)
```
irenedea, line 395: Can we add a test for the build_hf_dataset_from_remote function?

irenedea: I think you can call `_build_hf_dataset_from_remote` with get_file mocked to call `make_tiny_ft_dataset`

irenedea: Or alternatively, we can refactor so that we have a separate function that creates the paths (like finetune_dir, etc.) for`_build_hf_dataset_from_remote`. We can then call that function in a unit test and check to see that the paths created are as expected.

dakinggg: added

irenedea (approved):

tests/test_dataloader.py
```diff
@@ -394,6 +394,87 @@ def test_finetuning_dataloader_small_data(dataset_size: int,
...
+    }
+
+    cfg = om.create(cfg)
+
+    tokenizer = build_tokenizer(
+        tokenizer_name=tokenizer_name,
+        tokenizer_kwargs={'model_max_length': max_seq_len},
+    )
+
+    with monkeypatch.context() as m:
```
irenedea, line 472: nit: I think instead you can just add this to the top of your function:
```
@patch('llmfoundry.data.finetuning.dataloader.get_file',
       new=mock_get_file)
```
and you don't need to have an extra function parameter also

PR #589, opened 2023-09-10
Fix some type ignores

hanlint:
This PR removes some `type: ignore` and fixes the underlying typing issues.

Please check carefully for any functional impacts.

BEFORE
```
Found 139 Results of "type: ignore"
```

AFTER
```
Found 50 Results of "type: ignore"
```

### DIFF
--- a/llmfoundry/data/packing.py
+++ b/llmfoundry/data/packing.py
@@ -360,9 +360,13 @@ if __name__ == '__main__':
     # build tokenizer
     if 'tokenizer' not in cfg:
         raise ValueError('config must define tokenizer')
-    tokenizer_cfg: Dict[str,
-                        Any] = om.to_container(cfg.tokenizer,
-                                               resolve=True)  # type: ignore
+
+    resolved_tokenizer_cfg = om.to_container(cfg.tokenizer, resolve=True)
+    if not isinstance(resolved_tokenizer_cfg, Dict):
+        raise ValueError(
+            'tokenizer config needs to be resolved by omegaconf into a Dict.')
+    tokenizer_cfg: Dict[Any, Any] = resolved_tokenizer_cfg
+
     tokenizer_name = tokenizer_cfg['name']
     tokenizer_kwargs = tokenizer_cfg.get('kwargs', {})
     tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
--- a/llmfoundry/data/text_data.py
+++ b/llmfoundry/data/text_data.py
@@ -341,7 +341,9 @@ if __name__ == '__main__':
     tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
 
     loader = build_text_dataloader(cfg, tokenizer, device_batch_size)
-    tokenizer = loader.dataset.tokenizer  # type: ignore
+    assert isinstance(loader.dataset, StreamingTextDataset)
+    tokenizer = loader.dataset.tokenizer
+
     for batch_ix, batch in enumerate(islice(loader, 5)):
         print('\n')
         print('#' * 20, f'Batch {batch_ix}', '#' * 20)
--- a/llmfoundry/data/text_data.py
+++ b/llmfoundry/data/text_data.py
@@ -5,7 +5,8 @@
 
 import os
 from itertools import islice
-from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Union
+from typing import (Any, Callable, Dict, List, Mapping, Optional, Sequence,
+                    Union, cast)
 
 import numpy as np
 import torch
@@ -193,11 +194,12 @@ class ConcatenatedSequenceCollatorWrapper:
                 '`bos_token_id` if sequences start with a BOS token.'
             )
 
-        self.split_token_id = eos_token_id
-        self.bos_mode = False
         if eos_token_id is None:
-            self.split_token_id = bos_token_id
+            self.split_token_id = cast(int, bos_token_id)
             self.bos_mode = True
+        else:
+            self.split_token_id = eos_token_id
+            self.bos_mode = False
 
     def __call__(self, examples: List[Any]) -> Dict[str, torch.Tensor]:
         batch = self.base_collator(examples)
@@ -206,8 +208,7 @@ class ConcatenatedSequenceCollatorWrapper:
 
     def get_sequence_id_from_batch(
             self, batch: Dict[str, torch.Tensor]) -> torch.Tensor:
-        is_separator = torch.eq(batch['input_ids'],
-                                self.split_token_id)  # type: ignore
+        is_separator = torch.eq(batch['input_ids'], self.split_token_id)
         cumulative_sep = torch.cumsum(is_separator,
                                       dim=1).to(batch['input_ids'].dtype)
         # If separator token is bos, we're already done
--- a/llmfoundry/models/hf/hf_fsdp.py
+++ b/llmfoundry/models/hf/hf_fsdp.py
@@ -94,7 +94,12 @@ def hf_get_hidden_layers(model: PreTrainedModel):
         'model.layers',  # LLaMa
         'transformer.blocks',  # MPT
     )
-    return findattr(model, hidden_layers_attrs)
+    layers = findattr(model, hidden_layers_attrs)
+    if layers is None:
+        raise ValueError(
+            f'Unable to find hidden layer for {model}. Model must have one of the following attributes: {hidden_layers_attrs}'
+        )
+    return layers
 
 
 def hf_get_init_device(init_device: Optional[str]):
@@ -136,7 +141,7 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
     # OPT has an extra layer of wrapping, so special case here
     if isinstance(causal_base_model, OPTDecoder):
         model.model._fsdp_wrap = False
-    model_block = hf_get_hidden_layers(model)  # type: ignore
+    model_block = hf_get_hidden_layers(model)
     lm_head = model.get_output_embeddings()
     # some models (OPT) implement .get_input_embeddings for the causal subclass
     # but all of them implement it for the base model
@@ -153,7 +158,7 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
             raise ValueError(
                 f'Unable to FSDP-wrap this model! `{mod_name}` does not ' +
                 'follow common layer/weight naming conventions.')
-    block_type = type(model_block[0])  # type: ignore
+    block_type = type(model_block[0])
     if init_device == 'mixed':
         # For FSDP with models with different device initializations, `mixed`, which
         # initializes the model on rank 0 on `cpu` and on all other ranks on `meta,``
@@ -186,9 +191,9 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
         # These lines ensures that both modules stay together in the top-most block when
         # the model has this tying enabled (almost all do; this property defaults to True)
         if model.config.tie_word_embeddings:
-            causal_base_model._fsdp_wrap = False  # type: ignore
-            tied_embeddings._fsdp_wrap = False  # type: ignore
-            lm_head._fsdp_wrap = False  # type: ignore
+            causal_base_model._fsdp_wrap = False
+            tied_embeddings._fsdp_wrap = False
+            lm_head._fsdp_wrap = False
 
     # FSDP Wrap and Activation Checkpoint every model block
     model.fsdp_wrap_fn = lambda module: isinstance(module, block_type)
@@ -228,15 +233,15 @@ def prepare_hf_enc_dec_model_for_fsdp(model: PreTrainedModel,
             raise ValueError(
                 f'Unable to FSDP-wrap this model! `{mod_name}` does not ' +
                 'follow common layer/weight naming conventions.')
-    decoder_block_type = type(decoder_block[0])  # type: ignore
-    encoder_block_type = type(encoder_block[0])  # type: ignore
+    decoder_block_type = type(decoder_block[0])
+    encoder_block_type = type(encoder_block[0])
 
     if model.config.tie_word_embeddings:
         # it is possible to train an enc/dec without tied embeddings, hence the check
-        tied_embeddings._fsdp_wrap = False  # type: ignore
-        encoder._fsdp_wrap = False  # type: ignore
-        decoder._fsdp_wrap = False  # type: ignore
-        lm_head._fsdp_wrap = False  # type: ignore
+        tied_embeddings._fsdp_wrap = False
+        encoder._fsdp_wrap = False
+        decoder._fsdp_wrap = False
+        lm_head._fsdp_wrap = False
 
     # FSDP Wrap and Activation Checkpoint every decoder block
     model.fsdp_wrap_fn = lambda module: isinstance(module, decoder_block_type)
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -177,7 +177,7 @@ def flash_attn_fn(
     try:
         from flash_attn import bert_padding, flash_attn_interface  # type: ignore # yapf: disable # isort: skip
     except:
-        raise RuntimeError('Please install flash-attn via pip install .[gpu]')
+        raise RuntimeError('Please install flash-attn==1.0.3.post0')
 
     check_valid_inputs(query, key, value)
 
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -177,7 +177,7 @@ def flash_attn_fn(
     try:
         from flash_attn import bert_padding, flash_attn_interface  # type: ignore # yapf: disable # isort: skip
     except:
-        raise RuntimeError('Please install flash-attn==1.0.3.post0')
+        raise RuntimeError('Please install flash-attn via pip install .[gpu]')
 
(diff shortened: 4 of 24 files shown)


llmfoundry/data/packing.py
```diff
@@ -360,9 +360,13 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
     # build tokenizer
     if 'tokenizer' not in cfg:
         raise ValueError('config must define tokenizer')
-    tokenizer_cfg: Dict[str,
-                        Any] = om.to_container(cfg.tokenizer,
-                                               resolve=True)  # type: ignore
+
+    resolved_tokenizer_cfg = om.to_container(cfg.tokenizer, resolve=True)
```
hanlint, line 364: The return types of `om.to_container` cause many issues, and is the source of many other unfixed `type: ignore` in the codebase

I'm unsure how to solve the type coercion more broadly, are there cases where the provided `cfg` could cause a return of a `List, Any, str`, such that we legitimately need a type guard here?

llmfoundry/data/text_data.py
```diff
@@ -193,11 +194,12 @@ def __init__(
                 '`bos_token_id` if sequences start with a BOS token.'
             )
 
-        self.split_token_id = eos_token_id
-        self.bos_mode = False
         if eos_token_id is None:
-            self.split_token_id = bos_token_id
+            self.split_token_id = cast(int, bos_token_id)
```
hanlint, line 198: `cast` is safe because of the mutually exclusive checks above

llmfoundry/models/utils/meta_init_context.py
```diff
@@ -80,21 +80,27 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
...
+            param_cls = type(parameter)
+            kwargs = parameter.__dict__
+
+            self._parameters[name] = param_cls(parameter.to(device), **kwargs)
+
+    def register_empty_buffer(self: torch.nn.Module,
+                              name: str,
+                              tensor: Optional[torch.Tensor],
+                              persistent: bool = True):
+        old_register_buffer(self, name, tensor, persistent=persistent)
```
hanlint, line 99: New functionality here is to pass `persistent` argument to the old register buffer method. Otherwise, the patching below would be incorrect.

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -94,7 +94,12 @@ def hf_get_hidden_layers(model: PreTrainedModel):
         'model.layers',  # LLaMa
         'transformer.blocks',  # MPT
     )
-    return findattr(model, hidden_layers_attrs)
+    layers = findattr(model, hidden_layers_attrs)
```
hanlint, line 97: Functional change: this method will now raise an error if no hidden layers found.

Seems safe given the all it's current invocations make this assumption anyway, but please vet.

llmfoundry/models/layers/llama_attention_monkeypatch.py
```diff
@@ -9,7 +9,8 @@
 from typing import Callable, Optional, Tuple
 
 import torch
-import torch.functional as F
+import torch.nn.functional as F
```
hanlint, line 12:
```
>>> import torch.functional as F
>>> F.linear
  File "<stdin>", line 1, in <module>
AttributeError: module 'torch.functional' has no attribute 'linear'
```

Was this code ever used / still maintained? If not, we should remove completely.

llmfoundry/models/utils/hf_prefixlm_converter.py
```diff
@@ -330,7 +333,7 @@ def _build_alibi_tensor(
     # and one new argument (`bidirectional_mask`) is added to the signature.
     KeyValueT = Tuple[torch.Tensor, torch.Tensor]
 
-    def forward(  # type: ignore
+    def transformer_forward(
```
hanlint, line 336: From what I understand, this method was created just to patch the transformer's forward method, and the `forward` method defined later in the code is meant for the parent method. Please confirm: cc: @alextrott16

tests/test_hf_v_mpt.py
```diff
@@ -165,7 +165,7 @@ def test_compare_hf_v_mpt(attn_impl: str, dropout: float, alibi: bool,
     print(f'{hf_model_fwd.mean().item() = }\n{model_fwd.mean().item() = }')
     if hf_model_fwd.mean().allclose(model_fwd.mean()):
         warn_msg = f'WARNING: model_fwd ({model_fwd}) and hf_model_fwd ({hf_model_fwd}) are very close at init.'
-        raise warnings.warn(warn_msg)  # type: ignore
+        raise RuntimeError(warn_msg)
```
hanlint, line 168: The previous line `raise warnings.warn(..)` is ambigious as to the author's intent. Do we want to raise an exception here or emit a warning? Since this is a test, I assume exception.

llmfoundry/data/packing.py
irenedea: Ah, that's unfortunate that there's no type narrowing...What if you cast to a `Dict[str, Any]`? Or you could assert that tokenizer_cfg is a `Dict`?

irenedea: Another thought: It seems like most of the usages are for building a tokenizer. Perhaps a shared utility function to create a tokenizer from configs is called for? That way, we could do assertion and casting all in a single place without bloating the codebase. 

I'm thinking something like:
```
def tokenizer_from_cfg(cfg: DictConfig) -> Dict[str, Any]:
    # build tokenizer
    if 'tokenizer' not in cfg:
        raise ValueError('config must define tokenizer')
    tokenizer_cfg = om.to_container(cfg.tokenizer, resolve=True)
    assert isinstance(tokenizer_cfg, Dict)
    tokenizer_name = tokenizer_cfg['name']
    tokenizer_kwargs = tokenizer_cfg.get('kwargs', {})
    tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
    return tokenizer

# other stuff...

assert isinstance(DictConfig, cfg)
tokenizer_from_cfg(cfg)
```

dakinggg: What about the way that chuck handled everything in `train.py`? https://github.com/mosaicml/llm-foundry/blob/dc1374816a007488ca9885c92e83e89e6fe933ef/scripts/train/train.py#L220-L234

scripts/inference/convert_hf_to_onnx.py
```diff
@@ -138,9 +138,9 @@ def export_to_onnx(
         with torch.no_grad():
             orig_out = model(**sample_input)
 
-        import onnx  # type: ignore
-        import onnx.checker  # type: ignore
-        import onnxruntime as ort  # type: ignore
+        import onnx
```
dakinggg, line 141: for my own education, how did you get this to pass without these ignores?

tests/test_hf_v_mpt.py
dakinggg: @vchiley Which was this supposed to be?

llmfoundry/models/hf/hf_fsdp.py
dakinggg: yeah, seems good to me.

llmfoundry/models/layers/llama_attention_monkeypatch.py
dakinggg: woah, thats on me, thanks for catching. The file is copied and modified from the huggingface file. Somehow I must have copied the import statement incorrectly. `F` is only used on the `if self.config.pretraining_tp > 1:` codepaths, and the llama models uploaded to huggingface all have `self.config.pretraining_tp == 1:`. So that code path has not been tested within this repo, but is part of the upstream huggingface implementation, so I would be in favor of leaving it in even though it is currently untested.

llmfoundry/models/layers/llama_attention_monkeypatch.py
```diff
@@ -123,9 +122,11 @@ def llama_attention_patch_torch(
     if past_key_value is not None:
         kv_seq_len += past_key_value[0].shape[-2]
     cos, sin = self.rotary_emb(value_states, seq_len=kv_seq_len)
-    query_states, key_states = apply_rotary_pos_emb(
-        query_states, key_states, cos, sin,
-        position_ids)  # type: ignore (thirdParty)
+
+    if position_ids is None:
```
dakinggg, line 126: I think `position_ids` actually can be `None` and I typed `apply_rotary_pos_emb` incorrectly when I added types to the huggingface code.

e.g.
```
In [7]: d = torch.randn(2048, 768)

In [8]: d[None]
Out[8]: 
tensor([[[-0.3377, -0.0054,  0.1669,  ..., -0.9896, -0.1753,  0.4629],
         [ 0.6908, -0.6488,  1.0811,  ..., -0.9895, -1.0824,  2.3419],
         [-1.7552, -0.3778,  1.7064,  ..., -0.0380,  0.7781,  0.1165],
         ...,
         [-0.4624, -0.2032,  0.9854,  ..., -0.6765, -1.9464, -0.7511],
         [ 0.6446, -2.3273,  0.0222,  ..., -0.4254, -0.5582, -0.6867],
         [ 2.4439,  0.6429, -0.3800,  ...,  0.1195,  0.0516,  0.7610]]])
```

not entirely sure if the [huggingface code](https://github.com/huggingface/transformers/blob/5af2c6269672cda01c24ad48fab13f14a3ffb746/src/transformers/models/llama/modeling_llama.py#L180-L188) silently assumes non None since it isn't typed.

llmfoundry/models/layers/llama_attention_monkeypatch.py
```diff
@@ -190,6 +192,9 @@ def llama_attention_patch_triton(
         raise NotImplementedError(
             'output_attentions is not supported when patching Llama attention with triton attention.'
         )
+    if position_ids is None:
```
dakinggg, line 195: same comment about pos ids

llmfoundry/models/utils/hf_prefixlm_converter.py
dakinggg, line 1: @alextrott16 could you double check this file?

llmfoundry/models/utils/meta_init_context.py
dakinggg: I am hesitant to change this code because I believe this file is just fully copied from accelerate and im not sure if they have code making some unexpected assumptions about this code: https://github.com/huggingface/accelerate/blob/4a1d5b1fb651c1c408c39f9ddaab37c54cd878ba/src/accelerate/big_modeling.py#L85-L158, however the change looks fine. wdyt?

llmfoundry/data/packing.py
irenedea: oh whoa, how does that type check correctly if pop_config returns `Any` ?

dakinggg: I think you're telling it that you _know_ the types don't match, but you're doing it anyway. similar to `assert isinstance` but without the assert part? I think this is probably ok for this usage...

llmfoundry/models/utils/meta_init_context.py
hanlint: Looking at the link you provided, looks like they do pass `persistent` over, as of four months ago:
```
  def register_empty_buffer(module, name, buffer, persistent=True):
 ```
 
So maybe it's OK? Defer to you all, I do not know this codebase well enough.

File: scripts/inference/convert_hf_to_onnx.py (hunk 8 above)
Comment (hanlint) [reply scripts/inference/convert_hf_to_onnx.py:141]: I didn't need to fix anything -- seems like the type ignores were superfluos, since `onnx` is part of the default install?

File: llmfoundry/models/utils/meta_init_context.py (hunk 3 above)
Comment (dakinggg) [reply llmfoundry/models/utils/meta_init_context.py:99]: good point haha, ok should be fine. thanks!

File: scripts/inference/convert_hf_to_onnx.py (hunk 8 above)
Comment (dakinggg) [reply scripts/inference/convert_hf_to_onnx.py:141]: interesting. when i wrote this i was getting type errors on those imports. oh well, glad it works now!

File: llmfoundry/models/layers/llama_attention_monkeypatch.py (hunk 9 above)
Comment (hanlint) [reply llmfoundry/models/layers/llama_attention_monkeypatch.py:126]: OK I changed the type on `apply_rotary_pos_emb`

File: llmfoundry/data/packing.py (hunk 1 above)
Comment (hanlint) [reply llmfoundry/data/packing.py:364]: I could use `pop_config`... it feels purely performative though, since that method doesn't actually enforce `isinstance(cfg.tokenizer, DictConfig)` -- I might as well have just done `cast(Dict, om.to_container(...))` and removed a level of indirection. Perhaps we can huddle quickly on this tomorrow.

File: tests/test_hf_v_mpt.py (hunk 7 above)
Comment (vchiley) [reply tests/test_hf_v_mpt.py:168]: Was I dumb enough to raise a warning. lololol

This is supposed to be a warning since it could cause a "false pass" of the test.

File: llmfoundry/data/text_data.py [hunk 11]
```diff
@@ -193,11 +194,12 @@ def __init__(
...
             self.bos_mode = True
+        else:
+            self.split_token_id = eos_token_id
```
vchiley, line 201:
```suggestion
            self.split_token_id = int(eos_token_id)
```

if `bos_token_id` needs to be cast, shouldn't `eos_token_id` be cast as well?
Alt: neither need to be cast.

llmfoundry/data/text_data.py
vchiley, line 198: Why use `cast(int, bos_token_id)` instead of just `int(bos_token_id)`?

llmfoundry/models/layers/attention.py
```diff
@@ -177,7 +177,7 @@ def flash_attn_fn(
     try:
         from flash_attn import bert_padding, flash_attn_interface  # type: ignore # yapf: disable # isort: skip
     except:
-        raise RuntimeError('Please install flash-attn==1.0.3.post0')
+        raise RuntimeError('Please install flash-attn via pip install .[gpu]')
```
vchiley, line 180: This tells the user to run the install instructions in README.md
But if they are running a file from `llm-foundry/scripts`, then the correct instruction might be `pip install ..[gpu]`?
Can we just direct the user to follow the install instructions in the readme?

llmfoundry/data/text_data.py
hanlint: The intent here is to force the type given the safeguards above (thus using `cast`), not try to convert to an `int`

llmfoundry/data/text_data.py
hanlint: No, the `if eos_token_id is None:` typeguard above already narrows the type here to an `int`.

llmfoundry/models/layers/attention.py
hanlint: We use `pip install .[gpu]` elsewhere in the file, e.g. see https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/layers/attention.py#L309

Given that these errors are being raised by the `llmfoundry` package, I think the right message is to do `pip install llmfoundry[gpu]` since users may not even have our repository cloned. 

Lets keep the message for now, and I'll do a follow-on PR to fix this issue globally.

dakinggg: I think this is a bit more complicated because pretty much all of our users use llm foundry directly (can look at pypi download stats) as opposed to as a pip package. Add the pip package is missing a couple dependencies due to our custom installs of a few things (https://github.com/mosaicml/llm-foundry/blob/dc1374816a007488ca9885c92e83e89e6fe933ef/.github/workflows/release.yaml#L39-L41). But definitely not needed to resolve on this PR.

llmfoundry/data/packing.py
dakinggg: Yeah, agreed its performative. I've just asserted dictconfig in other places and it seemed to work. not sure why it isn't here. but we can discuss tomorrow.

llmfoundry/models/layers/attention.py
dakinggg: ah sorry. this statement is actually for anyone using MPT from huggingface, so it was originally written to be independent of usage in llmfoundry. and to resolve it, only flash attn as specified is needed, not all of llmfoundry. so maybe we leave it?

dakinggg (approved):
LGTM apart from the open omegaconf discussion. Thanks!

llmfoundry/models/utils/hf_prefixlm_converter.py
alextrott16: Yes, this understanding is correct. I don't think the exact naming of this function should matter as long as it gets set as the `forward` method, and these changes handle that.
Also, very much agree that renaming this is the right move, as having two `forward` definitions back to back was needlessly confusing. My bad.

llmfoundry/models/utils/hf_prefixlm_converter.py
alextrott16: Changes look good to me!

llmfoundry/data/packing.py
hanlint: If it's OK, I'll merge as-is, and then open a separate PR on omegaconf-related typing ignores.

dakinggg: yup, works for me. feel free to merge when you are ready.

llmfoundry/models/layers/attention.py
dakinggg: @hanlint could you change this one back? Its actually kind of important for anyone using mpt not inside of llm foundry.

hanlint: OK sure will revert

PR #594, opened 2023-09-12
Add a callback to write huggingface checkpoints during the training run

dakinggg:
This adds a callback that does the huggingface checkpoint conversion during the training job to avoid further difficulty after the job is complete, and take advantage of backgrounded uploads while training is happening.

- [x] add precision
- [x] manual mpt test (`7b-mpt-hf-ckpt-4-jc3uSG`)
- [x] manual llama2 test (`l7b-hf-ckpt-4-HpZUAg`)


mvpatel2000 (commented):
Should this be in Composer?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -0,0 +1,164 @@
...
+        else:
+            self.remote_ud = None
+
+        self.last_checkpoint_batch: Optional[Time] = None
+
+    def run_event(self, event: Event, state: State, logger: Logger) -> None:
+        if state.get_elapsed_duration() is not None and self.check_interval(
+                state,
+                event) and self.last_checkpoint_batch != state.timestamp.batch:
+            self._save_checkpoint(state, logger)
```
mvpatel2000, line 76: why does this run on every event?

mvpatel2000 (commented):
Will defer to Evan on this one

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: the `check_interval` function handles only returning `True` for the appropriate events. Could've also implemented `batch_checkpoint`, `fit_end`, `epoch_checkpoint`, and `init`, but this seemed cleaner.

mvpatel2000: Hm... can you leave a comment on how this works? It's a bit confusing what's going on here

tests/test_hf_conversion_script.py
```diff
@@ -48,6 +176,183 @@ def get_config(
...
+        model_cfg = get_config(
+            conf_path='scripts/train/yamls/pretrain/gpt-neo-125m.yaml')
+        model_cfg['model']['config_overrides']['hidden_size'] = 36
+    elif model == 'llama2':
+        if 'HUGGING_FACE_HUB_TOKEN' not in os.environ:
+            pytest.skip(
+                'The CI cluster does not have access to the Llama models, so skip this test.'
+            )
+        model_cfg = get_config(
+            conf_path='scripts/train/yamls/pretrain/gpt-neo-125m.yaml')
```
irenedea, line 219: Can you put a yaml in the test folder that matches what you want to do? We shouldn't rely on other configs in very different folders, and that way you also won't need to modify the config programmatically

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -0,0 +1,165 @@
...
+        elif event == Event.INIT:
+            if not isinstance(state.model, HuggingFaceModel):
+                raise ValueError(
+                    f'`HuggingFaceCheckpointer` is only compatible with `HuggingFaceModel`s. '
+                    + f'Got {type(state.model)} instead.')
+            if self.upload_to_object_store and self.remote_ud is not None:
+                self.remote_ud.init(state, logger)
+                state.callbacks.append(self.remote_ud)
```
eracah, line 85: is it appended to callbacks and not loggers because the engine is just given callbacks, which has all the loggers?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -0,0 +1,165 @@
...
+                    json.dump(edited_config, f, indent=4)
+
+                if self.upload_to_object_store:
+                    assert self.remote_ud is not None
+                    # TODO change to log after other pr
+                    print(
+                        f'Uploading HuggingFace formatted checkpoint to {self.backend}://{self.bucket_name}/{save_dir}'
+                    )
+                    for filename in os.listdir(temp_save_dir):
+                        self.remote_ud.upload_file(
```
eracah, line 157: oooh clever. not needing to use logger.upload_file

eracah (commented):
Nice looks pretty good. Def doable to make a SaveForInferenceCallback I think in the future. Also some good nuggets here for refactoring CheckpointSaver

eracah:
> Should this be in Composer?

At some point yes

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -0,0 +1,165 @@
...
+
+        print('Saving HuggingFace formatted checkpoint'
+             )  #TODO change to log after other pr
+
+        from transformers.models.auto.configuration_auto import CONFIG_MAPPING
+        CONFIG_MAPPING._extra_content['mpt'] = MPTConfig
+        MPTConfig.register_for_auto_class()
+        MPTForCausalLM.register_for_auto_class('AutoModelForCausalLM')
+
+        assert isinstance(state.model, HuggingFaceModel)
```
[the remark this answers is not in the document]
dakinggg: not possible to convert. must use our `HuggingFaceModel` interface or I can't be sure of the behavior.

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: I copied it from the monolithic saver :P

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: again, copied from the monolithic saver haha

tests/test_hf_conversion_script.py
```diff
@@ -48,6 +176,183 @@ def get_config(
     return cast(DictConfig, test_cfg)
 
 
+@pytest.mark.world_size(2)
+@pytest.mark.gpu
+@pytest.mark.parametrize('model', ['mpt', 'neo', 'llama2'])
```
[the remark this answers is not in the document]
dakinggg: I think this is the minimal set I am comfortable testing. MPT is our model, so need to test. neo is a reasonable generic model, and llama uses sentence piece tokenizer. converting to huggingface occasionally breaks for some specific model due to some unaccounted for detail, so its good to have some amount of coverage of different models here. I think its pretty fast, but can always make things smaller/shorter duration if needed.

tests/test_hf_conversion_script.py
dakinggg: I decided to just create the config dictionaries in the test and not rely on yamls. I also made a jira to add useful fixtures to foundry so we can stop copying these configs around in different tests.

dakinggg:
@mvpatel2000 can't you just specify 1dur as the save interval?

mvpatel2000:
> @mvpatel2000 can't you just specify 1dur as the save interval?

🤯 ur so big brained

llmfoundry/callbacks/hf_checkpointer.py
eracah: haha forgetting my own code lol

eracah (approved):
Aight LGTM

germanjke:
@eracah @dakinggg hi guys! This thing supports no-MPT models, like LLaMA?

PR #598, opened 2023-09-14
Skip flaky lion8b test

dblalock:
The test that checks that fused LION8b is faster than unfused LION8b can apparently (if rarely) fail in the CI. This PR adds one line to skip this test.

I'd ideally like to skip it only in CI but I don't think that's a thing unless we add more pytest marks. Probably fine to just lean on manual speed testing since only one or two people would touch the underlying kernels.


mvpatel2000 (approved):

tests/test_lion8b.py
```diff
@@ -498,6 +498,7 @@ def _set_state_dict_type(model: nn.Module):
             torch.testing.assert_close(d_orig['errors'], d_new['errors'])
 
 
+@pytest.mark.skip('Avoiding rare flakiness; be sure to run manually!')
```
dakinggg, line 501: can you jira for adding retries to this test/making it not flaky? can put in the foundry backlog epic

dakinggg (approved):

tests/test_lion8b.py
dblalock: I can just fix it right now if you have ideas about how to do that. Do you think a few retries would be sufficient? I was just kinda erring on the side of caution since I can't guarantee execution time.

dakinggg: since we hadn't noticed it for a while i had assumed that a few retries would be sufficient. im fine if you just add some retries and then we revisit if it is still flaky.

dblalock: added retries + relaxed the atol when checking timing

dblalock:
@dakinggg @mvpatel2000 gonna merge this in like an hour (with retries and relaxed atol instead of skipping) unless you object.

dakinggg:
merge away

PR #599, opened 2023-09-15
Allow MPT models to return attention weights

lorabit110:
Previously, output_attentions is not propagated to the attention layer. Even when it's set to True, no attention weights were returned.


dakinggg (commented):
Could you please add a simple unit test for this?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -434,6 +434,7 @@ def forward(
                 attn_bias=attn_bias,
                 attention_mask=attention_mask,
                 is_causal=self.is_causal,
+                output_attentions=output_attentions == True,
```
dakinggg, line 437:
```suggestion
                output_attentions=bool(output_attentions),
```
any reason this version doesn't work? the type is `Optional[bool]`, so should either be `None` or a `bool` and `bool(None)` is `False`

vchiley:
[This error](https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/mpt/modeling_mpt.py#L321C22-L321C22) should be noted if you plan to use it.

lorabit110:
> Could you please add a simple unit test for this?

Updated an existing unit test to check attention.

lorabit110:
> [This error](https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/mpt/modeling_mpt.py#L321C22-L321C22) should be noted if you plan to use it.

Yeah. I am aware of that flash attention won't return attention weights.

tests/test_model.py
```diff
@@ -1341,6 +1341,7 @@ def test_forward_with_output_attentions_and_output_hidden_states(
 
         if output_attentions:
             assert len(outputs.attentions) == n_layers
+            assert outputs.attentions[0] is not None
```
dakinggg, line 1344:
```suggestion
            assert all(attn.shape == (1, 4, 2048, 2048) for attn in outputs.attentions)
```

dakinggg (approved):
One comment to make sure we're returning the right shape of stuff, but otherwise lgtm! Thanks for the PR!

tests/test_model.py
```diff
@@ -1341,6 +1341,7 @@ def test_forward_with_output_attentions_and_output_hidden_states(
 
         if output_attentions:
             assert len(outputs.attentions) == n_layers
+            assert all(attn.shape == (1, 4, 2048, 2048) for attn in outputs.attentions)
```
dakinggg, line 1344:
```suggestion
            assert all(attn.shape == (1, 4, 3, 3) for attn in outputs.attentions)
```

dakinggg:
@lorabit110 could you run `precommit run --all-files` locally? Thanks!

tests/test_model.py
```diff
@@ -1341,6 +1341,7 @@ def test_forward_with_output_attentions_and_output_hidden_states(
 
         if output_attentions:
             assert len(outputs.attentions) == n_layers
+            assert all(attn.shape == (1, 4, 3, 3) for attn in outputs.attentions)
```
dakinggg, line 1344:
```suggestion
            assert all(
                attn.shape == (1, 4, 3, 3) for attn in outputs.attentions)
```

PR #600, opened 2023-09-15
Fix streaming dataset setup for fine-tuning

boomanaiden154:
Currently, when a `StreamingFinetuningDataset` is created using the `build_finetuning_dataloader` method, a failure is returned as some incorrect parameters are passed through to the constructor of `StreamingFinetuningDataset`. This patch fixes the parameter mismatch and adds test coverage for this case.

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -125,17 +125,11 @@ def build_finetuning_dataloader(cfg: DictConfig,
             download_timeout=cfg.dataset.get('download_timeout', 60),
             validate_hash=cfg.dataset.get('validate_hash', None),
             keep_zip=cfg.dataset.get('keep_zip', False),
-            epoch_size=cfg.dataset.get('epoch_size', None),
             predownload=cfg.dataset.get('predownload', None),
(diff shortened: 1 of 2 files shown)


dakinggg:
@snarayan21 to take a look. I think we may have just forgotten to update a version pin for the streaming library. Thanks for finding!

boomanaiden154:
> I think we may have just forgotten to update a version pin for the streaming library. Thanks for finding!

That was one of my thoughts too. I didn't see anything in the constructors in the streaming library, but I could just be missing something or didn't look in the right place. Either way, something here needs to be patched, either removing the parameters at the stage I did, or fully passing them through to a constructor in the streaming library.

snarayan21:
Hey, so the issue was with the `StreamingFinetuningDataset` constructor [here](https://github.com/mosaicml/llm-foundry/blob/c308d1026a152c2990a64868d53fc33b8360a200/llmfoundry/data/finetuning/tasks.py#L70). It's supposed to take the arguments that were listed, along with some new ones that came in with Streaming v0.6.0. [Here's](https://github.com/mosaicml/llm-foundry/pull/602) a PR that addresses this. Thanks for pointing this out!

dakinggg:
Closing in favor of #602

PR #601, opened 2023-09-15
Add support for pre-tokenized streaming dataset finetuning

boomanaiden154:
This PR adds in support for performing fine-tuning training with a streaming dataset that has already been pre-tokenized.
The test passing is dependent on the fix in https://github.com/mosaicml/llm-foundry/pull/600.


dakinggg:
Thanks! Will take a look after we resolve #600

alextrott16 (changes requested):
I am not opposed to supporting pre-tokenized streaming datasets. But, I am concerned that the tokens/labels structure used to identify pre-tokenized formats is too arbitrary. It's an issue mostly because we don't have any tools that someone can use to build a streaming dataset with the tokens/labels structure.

Rather than having a "backdoor" in `StreamingFinetuningDataset` that recognizes pre-tokenized formats from the tokens/labels structure, I'd prefer that we keep with the prompt/response structure and simply allow the code to determine whether tokenization is required based on whether the prompt/response values are strings or bytes. I think this will make it easier to simply add a pre-tokenization option to our existing tooling.

Please let me know if this ask is unclear. Thanks for helping to grow our codebase!

boomanaiden154:
That's a pretty reasonable request, especially given there's nothing upstream using this! I'll change the implementation to follow your suggestion.

I'm planning on writing up some tooling eventually designed for upstream use that uses this format (probably as a part of https://github.com/mosaicml/llm-foundry/issues/611), but that'll come later.

alextrott16 (approved):
Thanks for adding this and incorporating the suggestions! LGTM!!

tests/test_dataloader.py
```diff
@@ -474,6 +494,48 @@ def test_finetuning_dataloader_streaming(tmp_path: pathlib.Path):
...
+        tokenizer_kwargs={'model_max_length': 2048},
+    )
+
+    ft_dataloader = build_finetuning_dataloader(cfg, tokenizer, 4)
+
+    expected_keys = ['input_ids', 'attention_mask', 'labels']
+
+    for batch in ft_dataloader:
+        for k in expected_keys:
+            assert k in batch
```
dakinggg, line 536: could you please test that these are exactly what you expect, and not just that the keys are present? Thanks!

boomanaiden154: This test is much more detailed now and checks explicitly for the values of `input_ids`, `labels`, and `attention_mask`. Some of the indices/checks end up a little odd because we get the batch here after the collate function has been run on it, but I've written some comments describing everything.

dakinggg:
This has been done by #945

PR #602, opened 2023-09-15
Updated streaming args for StreamingDataset subclasses

snarayan21:
Previously, the constructor for `StreamingFinetuningDataset` was not updated with the correct streaming args, but the `build_finetuning_dataloader` function that called it had some updated streaming args. This was causing errors upon dataset instantiation. This has been addressed.

Additionally, the new streaming args `batching_method` and `sampling_granularity` have been added to `StreamingFinetuningDataset` and `StreamingTextDataset`.


dakinggg:
@snarayan21 do you need to update the pin in `setup.py`? This stuff only works with latest streaming, right?

llmfoundry/data/text_data.py
```diff
@@ -66,7 +66,14 @@ class StreamingTextDataset(StreamingDataset):
         shuffle_algo (str): Which shuffling algorithm to use. Defaults to ``py1b``.
         shuffle_seed (int): Seed for Deterministic data shuffling. Defaults to ``9176``.
         shuffle_block_size (int): Unit of shuffle. Defaults to ``1 << 18``.
-        sampling_method (str): Which sampling method to use, either ``balanced`` or ``fixed``. Defaults to ``balanced``.
+        sampling_method (str): Which sampling method to use, either ``balanced`` or ``fixed``.
```
dakinggg, line 69: to double check, the defaults here are the same as the implementation prior to these new args, right?

snarayan21: Yepo

dakinggg (commented):
Thanks @snarayan21! Going to run regression tests just to make sure the upgrade doesn't break something unexpected.

boomanaiden154:
Thanks for the patch! Probably should've checked the constructors in `Streaming` a little bit more closely.

dakinggg (approved):

PR #603, opened 2023-09-16
Multi eval dataset logging

snarayan21:
Previously, only one eval dataloader was supported in llm-foundry (although multiple ICL eval tasks and the gauntlet were supported). Now users can add custom eval datasets by turning `eval_dataloader` into a list of dataloaders in their yaml. It would look like this:

Before:
```
eval_loader:
  name: text
  dataset: ...
  drop_last: false
  num_workers: 8
```
After:
```
eval_loader:
- label: first
  name: text
  dataset: ...
  drop_last: false
  num_workers: 8
- label: second
  name: text
  dataset: ...
  drop_last: false
  num_workers: 8
```
Users must specify a label for each eval dataloader so that these metrics are logged separately. This functionality was tested on wandb, mlflow, and tensorboard, shown below.

wandb:
Screenshot 2023-09-16 at 10 07 02 AM

mlflow:
Screenshot 2023-09-16 at 10 06 48 AM

tensorboard:
Screenshot 2023-09-16 at 10 29 53 AM


scripts/train/train.py
```diff
@@ -466,15 +471,29 @@ def main(cfg: DictConfig) -> Trainer:
...
-        eval_dataloader = build_dataloader(eval_loader_config, tokenizer,
-                                           device_eval_batch_size)
-        eval_loader = Evaluator(
-            label='eval',
-            dataloader=eval_dataloader,
-            metric_names=[],  # we will add these after model is created
-        )
+        if isinstance(eval_loader_config, ListConfig):
+            for eval_config in eval_loader_config:
+                assert eval_config.label is not None
```
dakinggg, line 478: raise nice error not assert

scripts/train/train.py
```diff
@@ -466,15 +471,29 @@ def main(cfg: DictConfig) -> Trainer:
...
+                eval_dataloader = build_dataloader(eval_config, tokenizer,
+                                            device_eval_batch_size)
+                eval_loader = Evaluator(
+                    label='eval/'+eval_config.label,
```
dakinggg, line 482: use fstring, bit cleaner

scripts/train/train.py
```diff
@@ -514,11 +533,12 @@ def main(cfg: DictConfig) -> Trainer:
 
     # Now add the eval metrics
     if eval_loader_config is not None:
-        assert eval_loader is not None
+        assert len(eval_loaders) != 0
```
dakinggg, line 536: the other assert was just for pyright. i don't think you need this anymore

dakinggg (commented):
Implementation looks good, thanks! Please remove the accidentally committed files, and add a simple unit test.

scripts/train/train.py
```diff
@@ -466,15 +469,31 @@ def main(cfg: DictConfig) -> Trainer:
...
+                eval_loaders.append(eval_loader)
+        else:
+            eval_dataloader = build_dataloader(eval_loader_config, tokenizer,
+                                               device_eval_batch_size)
+            eval_loader = Evaluator(
+                label='eval',
+                dataloader=eval_dataloader,
+                metric_names=[],  # we will add these after model is created
+            )
+            eval_loaders.append(eval_loader)
```
irenedea, line 496: nit:
Suggested refactor:
```
is_multi_eval = isinstance(eval_loader_config, ListConfig)
eval_configs = eval_loader_config if is_multi_eval else [eval_loader_config]
eval_loaders = []
for eval_config in eval_configs:
    eval_dataloader = build_dataloader(...)
    eval_loader = Evaluator(
       label=f'eval/{eval_config.label}' if is_multi_eval else 'eval'
    )
   eval_loaders.append(eval_loader)
```

This makes it a little clearer that it's really just the label that is conditional when creating our Evaluator objects

tests/test_train_inputs.py
```diff
@@ -158,3 +158,24 @@ def test_invalid_name_in_scheduler_cfg_errors(self,
...
+        cfg.train_loader.dataset.local = data_local
+        # Set up multiple eval datasets
+        first_eval_loader = cfg.eval_loader
+        first_eval_loader.label = 'eval_1'
+        second_eval_loader = copy.deepcopy(first_eval_loader)
+        cfg.eval_loader = om.create([first_eval_loader, second_eval_loader])
+        for loader in cfg.eval_loader:
+            loader.dataset.local = data_local
+        # Set the first eval dataloader to have no label
+        cfg.eval_loader[0].label = None
```
irenedea, line 175:
```suggestion
        # Set up multiple eval datasets
        first_eval_loader = cfg.eval_loader
        first_eval_loader.dataset.local = data_local
        second_eval_loader = copy.deepcopy(first_eval_loader)
        # Set the first eval dataloader to have no label
        first_eval_loader.label = None
        second_eval_loader.label = 'eval_1'
        cfg.eval_loader = om.create([first_eval_loader, second_eval_loader])
        
```
I think deepcopy can also handle setting data_local, and it's more clear to set the labels directly

irenedea (approved):
LGTM with some minor comments

tests/test_training.py
```diff
@@ -149,3 +150,51 @@ def test_train_gauntlet(set_correct_cwd: Any):
         inmemorylogger.data['icl/metrics/eval_gauntlet/average'][-1], tuple)
 
     assert inmemorylogger.data['icl/metrics/eval_gauntlet/average'][-1][-1] == 0
+
+
+def test_train_multi_eval(set_correct_cwd: Any):
+    """Test training run with a small dataset."""
```
irenedea, line 156: Update this doc string

irenedea (changes requested):
Can we add a unit test that tests two eval loaders with two different datasets?

scripts/train/train.py
snarayan21: done, ty for the whole refactor!

scripts/train/train.py
```diff
@@ -33,7 +33,12 @@ def validate_config(cfg: DictConfig):
     """Validates compatible model and dataloader selection."""
     loaders = [cfg.train_loader]
     if 'eval_loader' in cfg:
-        loaders.append(cfg.eval_loader)
+        eval_loader = cfg.eval_loader
+        if isinstance(eval_loader, ListConfig):
+            for loader in eval_loader:
+                loaders.append(loader)
```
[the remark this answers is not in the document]
snarayan21: I'm going to keep it as `loaders.append(loader)` so that I can check for the "label" field in each loader in the `eval_loader` list.

tests/test_train_inputs.py
snarayan21: Ah true thanks for the clear edits!

tests/test_training.py
```diff
@@ -52,6 +55,30 @@ def create_c4_dataset_xsmall(prefix: str) -> str:
     return c4_dir
 
 
+def create_arxiv_dataset(prefix: str) -> str:
+    """Creates an arxiv dataset."""
+    arxiv_dir = os.path.join(os.getcwd(), f'my-copy-arxiv-{prefix}')
+    shutil.rmtree(arxiv_dir, ignore_errors=True)
```
irenedea, line 61: Let's use `tmp_path` here. We can pass it as an argument to this function, and we should make arxiv_dir live in `tmp_path` not `os.getcwd()`. That way, we can also remove L61.

The above function should use tmp_path too, so let's just make the change for the whole file?

(`tmp_path` is a pytest fixture, and automatically gets cleaned up, so it's nice to use here so we don't have tests that create folders and that worry about removing them.)

tests/test_data_prep_scripts.py
```diff
@@ -37,13 +37,13 @@ def test_download_script_from_api():
 
 def test_json_script_from_api():
     # test calling it directly
-    path = os.path.join(os.getcwd(), 'my-copy-c4-3')
+    path = os.path.join(os.getcwd(), 'my-copy-arxiv-1')
```
irenedea, line 40: why do we have these changes?

snarayan21: The test is creating a sample arxiv dataset but the folder is named c4. just found the naming mistake when I was looking for a way to create another dataset for the multi-eval dataset tests.

irenedea: oh cool! thanks for explaining and fixing!

irenedea (approved):
LGTM! Just a minor comment about using tmp_path instead

tests/test_training.py
snarayan21: done!

PR #608, opened 2023-09-18
Add default processes in text to mds conversion

irenedea:


scripts/data_prep/convert_text_to_mds.py
```diff
@@ -87,7 +88,7 @@ def parse_args() -> Namespace:
         '--processes',
         type=int,
         required=False,
-        default=1,
+        default=min(max(psutil.cpu_count() - 2, 1), 16),
```
irenedea, line 91:
```suggestion
        default=min(max(psutil.cpu_count() - 2, 1), 32),
```

dakinggg (approved):

PR #609, opened 2023-09-18
Fix eval yamls

irenedea:
Fixes some issues in eval yamls

# Manual Testing
Here are the completed eval runs with no errors:
- mpt-eval-xFtLW5
- hf-eval-stwFhv
- hf-8bit-eval-206Zp5
- hf-lora-eval-JHvNUf


scripts/eval/yamls/hf_lora_eval.yml
```diff
@@ -5,7 +5,7 @@ precision: amp_fp16
 # If you are using one model, put it here:
 model_name_or_path: EleutherAI/gpt-neo-125m
 # If you are using a seperated lora weight, put it here:
-lora_id_or_path: nathan0/lora-gpt-neo-125m-alpaca
+lora_id_or_path: edbeeching/gpt-neo-125M-imdb-lora
```
dakinggg, line 8: maybe leave a note that this path is just an example that works with gpt-neo-125M, but the LoRA would be model specific

scripts/eval/yamls/mpt_eval.yaml
```diff
@@ -2,31 +2,37 @@ max_seq_len: 1024
...
     name: ${tokenizer_name}
     kwargs:
       model_max_length: ${max_seq_len}
   model:
     name: mpt_causal_lm
-    init_device: meta
-    d_model: 768
-    n_heads: 12
-    n_layers: 12
+    init_device: mixed
```
dakinggg, line 18: imo the values here should match the _default_ values from configuration_mpt, even though you had to change things to have this yaml run against the test checkpoint. Because this yaml is just an example. but maybe leave a note saying that the model config section must match the checkpoint you are loading in exactly

irenedea: @dakinggg Makes sense. All config values have defaults, do you think we should explicitly state all of them or just the ones that existed prior?

irenedea: also init_device by default is 'cpu', but will use 'mixed' for all eval yamls

dakinggg: I think leave out the ones that match the defaults, just to reduce verbosity here.

dakinggg: also stuff should fail if we change the defaults, cause we should not do that :)

dakinggg (approved):

PR #610, opened 2023-09-19
Support for using tiktoken tokenizers

dakinggg:


llmfoundry/callbacks/monolithic_ckpt_callback.py
```diff
@@ -74,6 +74,9 @@ def _save_checkpoint(self, state: State, logger: Logger) -> None:
         ) if self.upload_to_object_store else contextlib.nullcontext(
             enter_result=save_dir)
         with dir_context_mgr as temp_save_dir:
+            assert isinstance(temp_save_dir,
+                              str)  # pyright doesn't know about enter_result
```
mvpatel2000, line 78:
```suggestion
            assert isinstance(temp_save_dir, str)  # pyright typing
```
Nit: i'd like to keep it one-line

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,275 @@
...
+        if isinstance(ids, int):
+            if ids in self.added_tokens_decoder:
+                return self.added_tokens_decoder[ids]
+            else:
+                return self._convert_id_to_token(ids)
+
+        tokens = []
+        current_stream = []
+        for index in ids:
+            index = int(index)
```
mvpatel2000, line 157: why does this need cast?

bmosaicml (approved):
lgtm!

mvpatel2000 (commented):
I don't really understand how tokenizers work. I'd like someone from research to review

mvpatel2000 (changes requested):
Blocking for now until everyone signs off

llmfoundry/tokenizers/tiktoken.py
dakinggg: hmm, i think i copied this from some huggingface code. i don't think it should be necessary. will remove and see if tests still pass

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def _tokenize(self, text: str):
+        """Returns a tokenized string.
+
+        Note: We have slightly redefined the expected contract between this method and
+        the _convert_token_to_id method. Normally, this method turns a string, into a list of strings,
+        and then the _convert_token_to_id method turns that list of strings into a list of integers.
+        However, not all vocab indices can be decoded into a string, so instead we just return the integers
+        from this function, and have adjusted the _convert_token_to_id method to handle integers as well as strings.
+        The only use of _tokenize that I could find was in this way, so this _should_ be safe.
```
irenedea, line 110: Could we add some logging or even assertion to track this assumption? In case something breaks it'd be great to easily be able to see whether or not these functions are being used as expected

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+                # need to try this first, so that we get a proper KeyError,
+                # otherwise it crashes in the rust code
+                _ = self.encoding.decode_single_token_bytes(i)
+                vocab[self.encoding.decode([i])] = i
+            except KeyError:
+                pass
+
+        return vocab
+
+    def _tokenize(self, text: str):
```
irenedea, line 102: Please add return types to these functions

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+        However, not all vocab indices can be decoded into a string, so instead we just return the integers
+        from this function, and have adjusted the _convert_token_to_id method to handle integers as well as strings.
+        The only use of _tokenize that I could find was in this way, so this _should_ be safe.
+        """
+        tokens = [t for t in self.encoding.encode(text, allowed_special='all')]
+
+        return tokens
+
+    def _convert_token_to_id(self, token: Union[int, str]):
+        """Converts a token (str) in an id using the vocab."""
```
irenedea, line 117: => "converts a token (str) to an id"

A similarly below "in a" => "to a"

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+
+            if index in self.added_tokens_decoder:
+                tokens.append(self.encoding.decode(current_stream))
+                current_stream = []
+                tokens.append(self.added_tokens_decoder[index])
+            else:
+                current_stream.append(index)
+
+        if len(current_stream) > 0:
+            tokens.append(self.encoding.decode(current_stream))
```
irenedea, line 168: Is tokens a list of single tokens or a list of token sequences?

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+
+        return vocab
+
+    def _tokenize(self, text: str):
+        """Returns a tokenized string.
+
+        Note: We have slightly redefined the expected contract between this method and
+        the _convert_token_to_id method. Normally, this method turns a string, into a list of strings,
+        and then the _convert_token_to_id method turns that list of strings into a list of integers.
+        However, not all vocab indices can be decoded into a string, so instead we just return the integers
```
irenedea, line 108: Why is this the case?

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+                         **kwargs)
+
+    @property
+    def vocab_size(self):
+        """Returns vocab size."""
+        return self.encoding.n_vocab
+
+    @property
+    def is_fast(self) -> bool:
+        return False
```
mvpatel2000, line 86: lol

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def get_vocab(self):
+        """Returns vocab as a dict."""
+        vocab = {}
+        for i in range(self.vocab_size):
+            try:
+                # need to try this first, so that we get a proper KeyError,
+                # otherwise it crashes in the rust code
+                _ = self.encoding.decode_single_token_bytes(i)
```
mvpatel2000, line 95: is this standard?

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def _tokenize(self, text: str):
+        """Returns a tokenized string.
+
+        Note: We have slightly redefined the expected contract between this method and
+        the _convert_token_to_id method. Normally, this method turns a string, into a list of strings,
+        and then the _convert_token_to_id method turns that list of strings into a list of integers.
+        However, not all vocab indices can be decoded into a string, so instead we just return the integers
+        from this function, and have adjusted the _convert_token_to_id method to handle integers as well as strings.
+        The only use of _tokenize that I could find was in this way, so this _should_ be safe.
```
mvpatel2000, line 110: Is this because special chars cause problems? A little weird to change

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+        tokens = [t for t in self.encoding.encode(text, allowed_special='all')]
+
+        return tokens
+
+    def _convert_token_to_id(self, token: Union[int, str]):
+        """Converts a token (str) in an id using the vocab."""
+        if isinstance(token, int):
+            return token
+        else:
+            return self.encoding.encode(token, allowed_special='all')[0]
```
mvpatel2000, line 121:
```suggestion
            return token
        return self.encoding.encode(token, allowed_special='all')[0]
```

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+                Whether or not to remove special tokens in the decoding.
+
+        Returns:
+            `str` or `List[str]`: The decoded token(s).
+        """
+        if isinstance(ids, int):
+            if ids in self.added_tokens_decoder:
+                return self.added_tokens_decoder[ids]
+            else:
+                return self._convert_id_to_token(ids)
```
mvpatel2000, line 152:
```suggestion
            if ids in self.added_tokens_decoder:
                return self.added_tokens_decoder[ids]
            return self._convert_id_to_token(ids)
```

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+        tokens = []
+        current_stream = []
+        for index in ids:
+            if skip_special_tokens and index in self.all_special_ids:
+                continue
+
+            if index in self.added_tokens_decoder:
+                tokens.append(self.encoding.decode(current_stream))
+                current_stream = []
+                tokens.append(self.added_tokens_decoder[index])
```
mvpatel2000, line 163: What does this do? I don't get why there's two appends

llmfoundry/tokenizers/tiktoken.py
dakinggg: "token" here means "string", so its list of strings

llmfoundry/tokenizers/tiktoken.py
dakinggg: Its a utf-8 thing
```
In [1]: import tiktoken

In [2]: gpt4t = tiktoken.encoding_for_model('gpt-4')

In [3]: gpt4t.encode('Å')
Out[3]: [127, 227]

In [4]: gpt4t.decode([127, 227])
Out[4]: 'Å'

In [5]: gpt4t.decode([127])
Out[5]: '�'

In [6]: gpt4t.decode([227])
Out[6]: '�'
```

llmfoundry/tokenizers/tiktoken.py
dakinggg: define "standard" it is how the tiktoken tokenizer functions.

llmfoundry/tokenizers/tiktoken.py
dakinggg: its because of
```
In [1]: import tiktoken

In [2]: gpt4t = tiktoken.encoding_for_model('gpt-4')

In [3]: gpt4t.encode('Å')
Out[3]: [127, 227]

In [4]: gpt4t.decode([127, 227])
Out[4]: 'Å'

In [5]: gpt4t.decode([127])
Out[5]: '�'

In [6]: gpt4t.decode([227])
Out[6]: '�'
```

llmfoundry/tokenizers/tiktoken.py
dakinggg: first append clears the current "stream", second append adds the "added token" in its own "stream"

llmfoundry/tokenizers/tiktoken.py
irenedea: Let's also add a check or log in _convert_token_to_id to match this expectation as well

llmfoundry/tokenizers/tiktoken.py
irenedea: Ohhh, I see. Is it accurate to say this is an issue because the huggingface mechanics assume that a single token is maps to a single int?

llmfoundry/tokenizers/tiktoken.py
irenedea: Got it, can we add a comment above to say that current_stream is used to collect potentially multiple ids for a single token

tests/test_tiktoken.py
```diff
@@ -0,0 +1,139 @@
...
+    # Batched tokenization
+    wrapped_output = wrapped_tokenizer(
+        ['Hello world!', 'Hello world but longer!'])
+    original_output = original_tokenizer.encode_batch(
+        ['Hello world!', 'Hello world but longer!'])
+    reloaded_wrapped_output = reloaded_wrapped_tokenizer(
+        ['Hello world!', 'Hello world but longer!'])
+    assert wrapped_output['input_ids'] == original_output
+    assert set(wrapped_output.keys()) == {'input_ids', 'attention_mask'}
+    assert reloaded_wrapped_output == wrapped_output
```
irenedea, line 89:
```suggestion
    assert 'input_ids' in wrapped_output and 'attention_mask' in wrapped_output
    assert wrapped_output['input_ids'] == original_output
    assert reloaded_wrapped_output == wrapped_output
```

scripts/data_prep/convert_dataset_hf.py
```diff
@@ -316,7 +324,11 @@ def main(args: Namespace) -> None:
 
     if args.concat_tokens is not None:
         mode = ConcatMode.CONCAT_TOKENS
-        tokenizer = AutoTokenizer.from_pretrained(args.tokenizer)
+        if args.tokenizer == 'tiktoken':
```
irenedea, line 327: can you call build_tokenizer here instead of inlining this logic?

llmfoundry/utils/builders.py
```diff
@@ -168,12 +167,12 @@ def build_scheduler(name: str,
 def build_tokenizer(
         tokenizer_name: str,
         tokenizer_kwargs: Dict[str, Any]) -> PreTrainedTokenizerBase:
-    if tokenizer_name == 'openai':
-        return OpenAITokenizerWrapper(**tokenizer_kwargs)
-    else:
-        os.environ['TRANSFORMERS_NO_ADVISORY_WARNINGS'] = '1'
-        os.environ['TOKENIZERS_PARALLELISM'] = 'false'
+    os.environ['TRANSFORMERS_NO_ADVISORY_WARNINGS'] = '1'
+    os.environ['TOKENIZERS_PARALLELISM'] = 'false'
```
irenedea, line 171: hm, a little weird to be setting env vars in this function...out of scope for this PR, but we should probably be setting env vars another place

dakinggg:
wait for #616 before merging this

llmfoundry/tokenizers/tiktoken.py
dakinggg: i made that function support both input types

irenedea (commented):
LGTM, pending minor comments + pyright.import_or_skip() is skipping the tests on CI. 

feel free to address in a follow up PR

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,276 @@
...
+                List of IDs.
+            token_ids_1 (`List[int]`, *optional*):
+                Optional second list of IDs for sequence pairs.
+            already_has_special_tokens (`bool`, *optional*, defaults to `False`):
+                Whether or not the token list is already formatted with special tokens for the model.
+
+        Returns:
+            `List[int]`: A list of integers in the range [0, 1]: 1 for a special token, 0 for a sequence token.
+        """
+        if already_has_special_tokens:
```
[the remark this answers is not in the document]
dakinggg: I definitely fully copied this code from gpt2tokenizer...so I'm inclined to leave it.

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,281 @@
...
+        return len(token_ids_0 + sep) * [0] + len(token_ids_1 + sep) * [1]
+
+    def save_vocabulary(self,
+                        save_directory: str,
+                        filename_prefix: Optional[str] = None) -> Tuple[str]:
+
+        # ignore the below type to keep the original signature
+        # we are knowingly breaking the signature here, although not 100% certain
+        # it doesn't have side effects
+        return (None, None)  # type: ignore
```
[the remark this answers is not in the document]
dakinggg: There is some code in huggingface that calls this function to get the vocab files, but it doesn't seem to access them (or at least checks for their existence before accessing them)

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,281 @@
...
+
+        Add the missing ones to the vocabulary if needed.
+
+        Return:
+            `int`: The number of tokens added in the vocabulary during the operation.
+        """
+        actual_new_tokens = []
+        for token in self.all_special_tokens_extended:
+            encoded = self.encoding.encode(token, allowed_special='all')
+            if len(encoded) > 1:
```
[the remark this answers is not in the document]
dakinggg: yeah, because if the (string) token translates to one input id, it is already in the vocab

tests/test_tiktoken.py
dakinggg: I actually want to test that there are no _other_ keys, so I will leave it as is.

llmfoundry/utils/builders.py
dakinggg: I think this is actually the right place. You don't need to set them if you are not using huggingface tokenizers, so we set them right before we instantiate one of those.

scripts/data_prep/convert_dataset_hf.py
dakinggg: yes, yes I can 🤦

tests/test_tiktoken.py
```diff
@@ -0,0 +1,139 @@
...
+                (key, original_tokenizer.encode(key,
+                                                allowed_special='all'), value))
+
+    # Decode is lossy because some bytes are not representable in utf-8
+    # see https://github.com/openai/tiktoken/blob/39f29cecdb6fc38d9a3434e5dd15e4de58cf3c80/tiktoken/core.py#L245-L247
+    # This means that the str: int vocab mapping doesn't work. Would have to look more into how other HF tokenizers handle this.
+    if model_name in ['gpt-4, gpt-3.5-turbot']:
+        assert len(didnt_match) == 77
+    elif model_name in ['text-davinci-003']:
+        assert len(didnt_match) == 14
```
abhi-mosaic, line 136: These are hardcoded to each test right? Maybe we should put them as kwargs to the test?

tests/test_tiktoken.py
```diff
@@ -0,0 +1,139 @@
...
+            continue
+        else:
+            didnt_match.append(
+                (key, original_tokenizer.encode(key,
+                                                allowed_special='all'), value))
+
+    # Decode is lossy because some bytes are not representable in utf-8
+    # see https://github.com/openai/tiktoken/blob/39f29cecdb6fc38d9a3434e5dd15e4de58cf3c80/tiktoken/core.py#L245-L247
+    # This means that the str: int vocab mapping doesn't work. Would have to look more into how other HF tokenizers handle this.
+    if model_name in ['gpt-4, gpt-3.5-turbot']:
```
abhi-mosaic, line 133:
```suggestion
    if model_name in ['gpt-4, gpt-3.5-turbo']:
```

llmfoundry/tokenizers/tiktoken.py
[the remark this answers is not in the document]
irenedea: Got it, let's add a comment about this

abhi-mosaic (approved):
Reviewed the tests carefully, and added some comments.

Also, do we need to add `tiktoken` as a dependency in `setup.py`?

tests/test_tiktoken.py
dakinggg: yeah, let me redefine somewhere else

tests/test_tiktoken.py
dakinggg: oops. let me rewrite this section so it fails if unexpected name

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -0,0 +1,290 @@
...
+        try:
+            import tiktoken
+        except:
+            raise ImportError(
+                'You need to install tiktoken to use TiktokenTokenizerWrapper.')
+
+        if model_name is not None and encoding_name is not None:
+            raise ValueError(
+                'You need to specify either model_name or encoding_name, not both.'
+            )
```
vchiley, line 53: Would it make sense to instead set a `encoding_name` default if both are None?

dakinggg:
@abhi-mosaic tiktoken is added as an optional dep group already

tests/test_tiktoken.py
```diff
@@ -0,0 +1,150 @@
...
+    'gpt-3.5-turbo': 77,
+    'text-davinci-003': 14,
+    'cl100k_base': 77,
+}
+
+
+@pytest.mark.parametrize('model_name,encoding_name',
+                         [('gpt-4', None), ('gpt-3.5-turbo', None),
+                          ('text-davinci-003', None), (None, 'cl100k_base')])
+def test_tiktoken(model_name: Optional[str], encoding_name: Optional[str],
```
vchiley, line 40: nit: does it make sense to split these into like 4 tests?

vchiley (approved):

llmfoundry/tokenizers/tiktoken.py
dakinggg: mm, i'd rather force the user to consciously pick a tokenizer.

tests/test_tiktoken.py
dakinggg: mmmm yeah ok

mvpatel2000 (approved):

PR #614, opened 2023-09-20
simba

b-chu:


llmfoundry/utils/builders.py
```diff
@@ -147,7 +147,7 @@ def build_optimizer(model: torch.nn.Module, name: str,
         return DecoupledClipLion(model.parameters(), **optimizer_config)
     elif name == 'adalr_lion':
         return DecoupledAdaLRLion(model.parameters(), **optimizer_config)
-    elif name == 'decoupled_lionw_8b':
+    elif name == 'decoupled_simbaw':
```
dakinggg, line 150: this would be an API breaking change, so we would need to support both names here

dblalock:
Why rename it? I don't actually object to having this as an alternate name since it's a backwards-compatible one-liner. But I am concerned about telling customers this name / making this the public branding, since:

1. It suggests a qualitatively different optimizer, not a drop-in replacement for LION
2. There's a good chance we'll have a similar Adam implementation at some point, and there's no analogous name for that. We might want to brand these as something like "MiniLION" and "MiniAdam", "MosaicLION" and "MosaicAdam", etc.

PR #616, opened 2023-09-20
Run CPU tests on a new dep group `all-cpu`

dakinggg:
Previously they were running on the `dev` dep group which might skip tests for optional dependencies.


mvpatel2000 (approved):
LGTM! Great catch by Irene

irenedea (approved):

PR #618, opened 2023-09-21
Add support for automatically registering models to UC at the end of training

dakinggg:
Adds support for directly logging models to mlflow during training for easy deployment on databricks model serving. Tested and was able to deploy both an MPT model and a llama model succesfully.

- [x] Confirm can log MPT and llama and deploy
- [x] Measure time of final checkpoint for different sizes
- [x] wait for next composer patch release with mlflow model logging changes
- [x] write tests
- [x] add support for using UC
- [x] refactor to skip the extra load to memory.

along for the ride: fixing the hf checkpointer to work with non fsdp models as well

- [x] redo manual tests once ready for review and unit tests pass (successfully logged and deployed mpt and llama models)


mvpatel2000 (commented):
What scale of models can this take? What happens if I try shoving in a 30b?

dakinggg:
@mvpatel2000 large models work. at least 30b, haven't had cluster space to run 70b yet. this does call the mlflow logger `log_model` from composer. the work in this callback is just preparing the inputs to that function call.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -39,6 +40,12 @@ class HuggingFaceCheckpointer(Callback):
         huggingface_folder_name (str): Folder to save each checkpoint under (can be a format string). Default is ``ba{batch}``.
         precision: The precision to save the model in. Default is ``float32``. Options are ``bfloat16``, ``float16``, or ``float32``.
         overwrite (bool): Whether to overwrite previous checkpoints.
+        log_to_mlflow (bool): Whether to log and register the checkpoint to MLFlow. Default is ``False``.
+        mlflow_task (str): The MLFlow task to log the checkpoint under. Only used if ``log_to_mlflow`` is ``True``. Default is ``text-generation``.
+        mlflow_metadata (Optional[dict]): The MLFlow metadata to log the checkpoint with. Only used if ``log_to_mlflow`` is ``True``. Default is ``None``.
+        uc_prefix: (Optional[str]): Prefix to use for the MLFlow registered model. If specified, the model will be logged to UC rather than
+            the workspace model registry. If specified, the prefix must be of the form ``{catalog}.{schema}``
+            The model will be registered at ``{catalog}.{schema}.{model name}``. Only used if ``log_to_mlflow`` is ``True``. Default is ``None``.
```
eracah, line 48: I think you should make these arguments to the MLflowLogger not to this callback? Seems like these fall under configuring how you log to mlflow not configuring your huggingface checkpoint. 
Pros: less mlflow set up logic in this callback and less for anyone else that wants to log stuff to mlflow

dakinggg: hmm, i sort of agree...however, the task and metadata are tied to the model type....

dakinggg: ideally this would all be in your "inference checkpointer" callback in composer :)

dakinggg: also these args are _specific_ to the transformers model flavor

eracah, line 48: move to mlflowlogger?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -39,6 +40,12 @@ class HuggingFaceCheckpointer(Callback):
         huggingface_folder_name (str): Folder to save each checkpoint under (can be a format string). Default is ``ba{batch}``.
         precision: The precision to save the model in. Default is ``float32``. Options are ``bfloat16``, ``float16``, or ``float32``.
         overwrite (bool): Whether to overwrite previous checkpoints.
+        log_to_mlflow (bool): Whether to log and register the checkpoint to MLFlow. Default is ``False``.
+        mlflow_task (str): The MLFlow task to log the checkpoint under. Only used if ``log_to_mlflow`` is ``True``. Default is ``text-generation``.
+        mlflow_metadata (Optional[dict]): The MLFlow metadata to log the checkpoint with. Only used if ``log_to_mlflow`` is ``True``. Default is ``None``.
```
eracah, line 45: change these to `mlflow_config` dict args

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -87,6 +106,23 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
...
+                if len(self.mlflow_loggers) == 0:
+                    raise ValueError(
+                        f'`log_to_mlflow` was set to `True` but no `MLFlowLogger` was found in the `logger.destinations` list. '
+                        +
+                        'Please add an `MLFlowLogger` or set `log_to_mlflow` to `False`.'
+                    )
+
+                import mlflow
+                mlflow.environment_variables.MLFLOW_HUGGINGFACE_MODEL_MAX_SHARD_SIZE.set(
+                    '5GB')
```
smurching, line 124: I ran into issues too with the default MLflow HF shard size being too small (which slowed down file upload/download). Curious how 5GB worked for you - we might want to update the default in MLflow to something like 1GB at least, if we can confirm it won't regress/OOM any existing workloads

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -164,4 +215,32 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
 
-        dist.barrier()
+                elapsed_duration = state.get_elapsed_duration()
+                if self.log_to_mlflow and elapsed_duration is not None and elapsed_duration >= 1.0:
+                    components = {'model': new_model_instance}
+                    if original_tokenizer is not None:
+                        components['tokenizer'] = original_tokenizer
+
+                    log.debug('Logging Hugging Face model to MLFlow')
+                    registered_model_name = f'{state.run_name}_{os.path.basename(save_dir)}'
```
smurching, line 225: Looks like this will produce a new registered model on each run (?) since `save_dir` includes a timestamp. I wonder if there's a better way we can group together the model versions produced by these runs - it's not recommended to create a new registered model per run since this complicates deployment workflows (you can't leverage features like model aliases to manage which model version out of a group of models is deployed for serving/inference).

Is there some user-provided name we can reuse as the registered model, or require as input to enable model registration? 

For tracking model metadata like the run timestamp etc of the model, I'd recommend using [set_model_version_tag ](https://mlflow.org/docs/latest/python_api/mlflow.client.html?highlight=set_model_version_tag#mlflow.client.MlflowClient.set_model_version_tag)(basically setting metadata tags on the created model version) via something like:

```
model_version =    mlflow_logger.register_model(
                            model_uri=local_save_path,
                            name=registered_model_name,
                            await_registration_for=3600,
)

# We should use the MlflowClient instance associated with the logger, but including this as an example
MlflowClient().set_model_version_tag(name=registered_model_name, version=model_version.version, key=..., value=...)
```

Tags are also searchable etc (although not yet in UC...)

dakinggg: hmm, I can remove the timestamp info from the run and then it would just be the run name. This would still be different for each training run. We could force the user to provide a model name to register under, but I'd be a bit worried about this getting confusing (the less arguments the better, and users tend to forget to change stuff between runs, and so might end up registering things under the same model that they didn't mean to). Maybe we can start here by just using the run name and see what the usage feedback is? I don't think I have a good understanding of how users' workflows are going to look around deploying to databricks yet.

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: yeah, 5GB seemed alright. HF's default is 10. any particular reason y'all are deviating from that?

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: ok, i think i'll just make the registered model name a required arg

smurching: Lol, sorry for the slow response - yes I think that would make sense, I was gonna post this:

> Ah thanks, got it, I can share some context on model deployment workflows in Databricks (for just ML models in general - my knowledge isn't LLM-specific but this stuff seems somewhat general). In general the pattern is to use a single registered model per business problem, e.g. group model versions for `revenue_forecasting` or `customer_support_bot` under the same registered model.

smurching: In the end, I'd hope customers can automate production model training to catch / prevent issues like training a model for business problem A and registering it to a registered model for business problem B

dakinggg: ah ok, that is all helpful context, thanks!

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -164,4 +218,31 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+
+                        # TODO: Remove after mlflow fixes the bug that makes this necessary
+                        import mlflow
+                        mlflow.store._unity_catalog.registry.rest_store.get_feature_dependencies = lambda *args, **kwargs: ''
+                        mlflow_logger.save_model(
+                            flavor='transformers',
+                            transformers_model=components,
+                            path=local_save_path,
+                            **self.mlflow_logging_config,
+                        )
```
eracah, line 243: wait so save_model only happens if `registered_model_name` is not None? Isn't it possible to save model and not register it?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -58,6 +65,20 @@ def __init__(
...
+        self.mlflow_registered_model_name = mlflow_registered_model_name
+        if self.mlflow_registered_model_name is not None:
+            if mlflow_logging_config is None:
+                mlflow_logging_config = {}
+            if 'metadata' not in mlflow_logging_config:
+                mlflow_logging_config['metadata'] = {
+                    'task': 'llm/v1/completions'
+                }
+            if 'task' not in mlflow_logging_config:
+                mlflow_logging_config['task'] = 'text-generation'
```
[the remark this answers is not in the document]
dakinggg: will add a comment. the code is written as intended. both items are necessary

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -114,44 +151,61 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+            from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+
+            if state.is_model_ddp:
+                original_model: PreTrainedModel = state.model.module.model
+                original_tokenizer = state.model.module.tokenizer
+            elif isinstance(state.model.model, FSDP):
+                original_model: PreTrainedModel = state.model.model.module
+                original_tokenizer = state.model.tokenizer
+            else:
+                original_model: PreTrainedModel = state.model.model
```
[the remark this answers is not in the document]
dakinggg: 😅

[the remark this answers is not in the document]
dakinggg: would be useful for state to define a "get actual model module" function

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: so, `save_model` only saves locally. it is different from `log_model`. we're doing it this way to avoid uploading the model twice (which is what would happen if you did `log_model` and register). so yeah, we're not really enabling mlflow log model here, only register.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -114,44 +151,61 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
-                                                  state_dict=state_dict)
-
-                if state.model.tokenizer is not None:
-                    assert isinstance(state.model.tokenizer,
+                log.debug('Saving Hugging Face checkpoint to disk')
+
+                copied_config = copy.deepcopy(original_model.config)
+                if copied_config.model_type == 'mpt':
+                    copied_config.attn_config['attn_impl'] = 'torch'
+                    copied_config.init_device = 'cpu'
```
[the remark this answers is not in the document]
dakinggg: this config item is very specific to mpt. other HF models don't have it.

eracah (approved):
ok lgtm

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -13,11 +13,11 @@
 from composer.callbacks.utils import create_interval_scheduler
```
dakinggg, line 13: use the non deprecated import path

PR #626, opened 2023-09-25
Replace dashes with underscores in split name

irenedea:
HF doesn't support dashes in the split name 😅 

Error that is fixed with this change:
```
/opt/homebrew/lib/python3.11/site-packages/datasets/splits.py:361: in __init__
    raise ValueError(f"Split name should match '{_split_re}' but got '{split_name}'.")
E   ValueError: Split name should match '^\w+(\.\w+)*$' but got 'custom-dash'.
```

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -299,6 +299,9 @@ def _build_hf_dataset_from_remote(
         'downloaded_finetuning',
         cfg.dataset.split if cfg.dataset.split != 'data' else 'data_not',
     )
+    # HF datasets does not support a split with dashes, so we replace dashes
+    # with underscores in the destination split.
+    destination_split = cfg.dataset.split.replace('-', '_')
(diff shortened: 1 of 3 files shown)


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -299,14 +299,19 @@ def _build_hf_dataset_from_remote(
...
     os.makedirs(finetune_dir, exist_ok=True)
     for extension in supported_extensions:
         name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
         destination = str(
             os.path.abspath(
                 os.path.join(
                     finetune_dir, 'data',
-                    f'{cfg.dataset.split}-00000-of-00001.{extension}')))
+                    f'{destination_split}-00000-of-00001.{extension}')))
+        print('HERE!!!', destination_split)
```
irenedea, line 313:
```suggestion
```

dakinggg (approved):

PR #627, opened 2023-09-25
Propagate bias through model

mvpatel2000:
Propagate bias through model. This is needed for fp8 and transformer engine as the current implementation errors

Along for ride: bump flash attn xentropy

### DIFF
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -419,6 +419,7 @@ class GroupedQueryAttention(nn.Module):
         norm_type: str = 'low_precision_layernorm',
         fc_type: str = 'torch',
         device: Optional[str] = None,
+        no_bias: bool = False,
     ):
         super().__init__()
 
@@ -450,7 +451,9 @@ class GroupedQueryAttention(nn.Module):
             self.softmax_scale = 1 / math.sqrt(self.d_model / self.n_heads)
         self.attn_dropout_p = attn_pdrop
 
-        fc_kwargs = {}
+        fc_kwargs = {
+            'bias': not no_bias,
+        }
         if fc_type != 'te':
             fc_kwargs['device'] = device
         self.Wqkv = FC_CLASS_REGISTRY[fc_type](
@@ -557,6 +560,7 @@ class MultiheadAttention(GroupedQueryAttention):
         norm_type: str = 'low_precision_layernorm',
         fc_type: str = 'torch',
         device: Optional[str] = None,
+        no_bias: bool = False
     ):
         super().__init__(
             d_model=d_model,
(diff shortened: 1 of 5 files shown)


llmfoundry/models/layers/blocks.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         norm_type: str = 'low_precision_layernorm',
         fc_type: str = 'torch',
         device: Optional[str] = None,
+        no_bias: bool = False,
```
irenedea, line 29: why do we use `no_bias` here but `bias` elsewhere?

mvpatel2000: Our original config uses `no_bias`. I want to switch to `bias` everywhere since that is what everyone else does. But changing in cfg (which is passed directly into block) is a breaking change. Imo we should just do breaking change since we're on an early version of foundry...

mvpatel2000: We can't break API bc it breaks ckpts. :(

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -150,6 +150,11 @@ def __init__(self, config: MPTConfig):
                     log.info(f'Removing bias ({module.bias}) from {module}.')
                     module.register_parameter('bias', None)
 
+                # For transformer engine
```
dakinggg, line 153: do you still need the above code removing bias after your change?

llmfoundry/models/layers/attention.py
```diff
@@ -545,19 +548,18 @@ class MultiheadAttention(GroupedQueryAttention):
...
+                 n_heads: int,
+                 attn_impl: str = 'triton',
+                 clip_qkv: Optional[float] = None,
+                 qk_ln: bool = False,
+                 softmax_scale: Optional[float] = None,
+                 attn_pdrop: float = 0.0,
+                 norm_type: str = 'low_precision_layernorm',
+                 fc_type: str = 'torch',
+                 device: Optional[str] = None,
+                 bias: bool = True):
```
vchiley, line 562:
```suggestion
    def __init__(
        self,
        d_model: int,
        n_heads: int,
        attn_impl: str = 'triton',
        clip_qkv: Optional[float] = None,
        qk_ln: bool = False,
        softmax_scale: Optional[float] = None,
        attn_pdrop: float = 0.0,
        norm_type: str = 'low_precision_layernorm',
        fc_type: str = 'torch',
        device: Optional[str] = None,
        bias: bool = True,
    ):
```

llmfoundry/models/mpt/modeling_mpt.py
vchiley: guarantee that bias gets nerf'd???

mvpatel2000: I need it :(

mvpatel2000: I don't know why I need it, but I tried without it and it failed

vchiley (approved):
lgtm

PR #628, opened 2023-09-25
Change `repeat` to `expand` in GQA

sashaDoubov:
This reduces memory usage and improves MFU. Note that there may be some risk of NaNs with small d_head (this was previously seen for MQA). However, there is significant perf benefits for standard d_head.


vchiley (approved):

madhavatreplit:
Will this be backwards compatible with any MPT model using GQA, trained using the code from a commit before this?

sashaDoubov:
@madhavatreplit yes, this is backwards compatible, as the model weights don't change, this is simply an optimization for the attention function to avoid allocating new memory.

Issue #504, opened 2023-08-03
No such file or directory: '.local_rank0_completed_autoresume' while doing multi-node pretraining using hf_casual_lm (LLAMA2)


palash04:
## Environment

```
transformers: 4.30.2
llm-foundry: 0.2.0
mosaicml: 0.15.1
mosaicml-cli: 0.4.16
mosaicml-streaming: 0.5.1
```

## To reproduce

YAML config:
```
data_local: /data/olaai/mosaicml/llm-foundry/scripts/data_prep/c4_mds_llama_2048_cache_
data_remote: /data/olaai/mosaicml/llm-foundry/scripts/data_prep/c4_mds_llama_2048_copy
max_seq_len: 2048
global_seed: 17

# Run Name
run_name: # If left blank, will be read from env var $RUN_NAME

# Model
model:
  name: hf_causal_lm
  pretrained_model_name_or_path: meta-llama/Llama-2-7b-hf
  init_device: cpu
  pretrained: true
  use_auth_token: true  

# Tokenizer
tokenizer:
  name: meta-llama/Llama-2-7b-hf
  kwargs:
    model_max_length: ${max_seq_len}

# Dataloaders
train_loader:
  name: text
  dataset:
    local: ${data_local}
    remote: ${data_remote}
    split: train_small
    shuffle: true
    max_seq_len: ${max_seq_len}
    shuffle_seed: ${global_seed}
  drop_last: true
  num_workers: 8

eval_loader:
  name: text
  dataset:
    local: ${data_local}
    remote: ${data_remote}
    split: val_small
    shuffle: false
    max_seq_len: ${max_seq_len}
    shuffle_seed: ${global_seed}
  drop_last: false
  num_workers: 8

# Optimization
scheduler:
  name: cosine_with_warmup
  t_warmup: 100ba
  alpha_f: 0.1

optimizer:
  name: decoupled_adamw
  lr: 8.0e-5
  betas:
  - 0.9
  - 0.95
  eps: 1.0e-08
  weight_decay: 0.0

algorithms:
  gradient_clipping:
    clipping_type: norm
    clipping_threshold: 1.0

max_duration: 333800ba  # ~ 1.4T tokens
eval_interval: 10000ba
eval_first: false
eval_subset_num_batches: -1
global_train_batch_size: 16

# System
seed: ${global_seed}
device_eval_batch_size: 1
device_train_microbatch_size: 1
# device_train_microbatch_size: auto
precision: amp_bf16

[...]

PR #629, opened 2023-09-26
Add node rank to signal paths

mvpatel2000:
Add node rank to signal paths. Otherwise, this breaks for multi-node with shared mounts. 

Corresponding issue: https://github.com/mosaicml/llm-foundry/issues/504


dakinggg (approved):
Thanks!

PR #631, opened 2023-09-27
Point to composer.callback.Generate

aspfohl:
This removes the llm-foundry `Generate` callback in favor of `composer.callback.Generate`, which has newer features like logging to MLflow. I've deleted the existing `Generate` callback logic but kept it around just in case someone is directly importing from llm-foundry. The builder callback also supports converting `batch_log_interval` to `interval` for backwards compatibility

In a future llmfoundry release, should remove this callback entirely

example of logging to mlflow & wandb
Screenshot 2023-09-29 at 10 20 22 AM
Screenshot 2023-09-29 at 10 20 28 AM


llmfoundry/utils/builders.py
```diff
@@ -89,7 +89,10 @@ def build_callback(name: str, kwargs: Dict[str, Any]) -> Callback:
             'log_optimizer_metrics', True),)
     elif name == 'generate_callback':
         prompts = kwargs.pop('prompts')
-        return Generate(prompts=list(prompts), **kwargs)
+        # Generate callback used to be batch_log_interval, so this is for backwards compatibility
+        interval = kwargs.pop(
```
dakinggg, line 93: won't this be a keyerror?

dakinggg (commented):
Can you please manually test this? both with old args and new args. Also I know there isn't a unit test for this callback right now...but could you please add one?

llmfoundry/utils/builders.py
vchiley: yeah i think you'll want to do
```
kwargs.pop('interval', None) or f"{kwargs.pop('batch_log_interval')}ba"
```
or else the first one will KeyError

llmfoundry/callbacks/generate_callback.py
```diff
@@ -1,119 +1,24 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
-"""Periodically log generations to wandb from a set of prompts."""
-from typing import Any, List, Union, cast
+"""Deprecated Generate callback. Please use composer.callbacks.Generate instead."""
```
irenedea, line 4: I think most, if not all, users will use this callback via the yaml config. Is that correct @dakinggg ? If so, I think we should just remove this class altogether.

irenedea: Discussed offline with Daniel. We should keep it but will remove in a later release.

irenedea:
> Although since the Generate callback should now be in composer, not sure it makes sense to add tests here?

There are [unit tests for Generate in composer](https://github.com/mosaicml/composer/blob/f5986ea7604b494695d5ba632150374c3fc503d1/tests/callbacks/test_generate.py#L4). I think it should be sufficient to just test building the callback as we are doing now.

irenedea:
> Although since the Generate callback should now be in composer, not sure it makes sense to add tests here?

There are [unit tests for Generate in composer](https://github.com/mosaicml/composer/blob/f5986ea7604b494695d5ba632150374c3fc503d1/tests/callbacks/test_generate.py#L4) that test for functionality. Is there something else we had in mind to unit test in foundry? cc: @dakinggg

dakinggg:
@irenedea @aspfohl the "new" thing that foundry should test is that the callback works e2e with MPT

llmfoundry/callbacks/generate_callback.py
```diff
@@ -1,119 +1,27 @@
...
-        To compare previous iterations of the generations, you need to
-        1. Click on the run
-        2. Click on "artifacts" in the menu on the left side of the screen
-        3. Click on one of the artifacts called "predictions"
-        4. Click on the "files" tab
-        5. Click on "predictions.table.json"
-        6. On the left hand side, there are different versions of the table produced throughout training. Select one of these.
-        7. Now, when you hover over other versions, there will be a "compare" button, which will allow you to compare the currently
-            selected version to the version you add via compare.
+        warnings.warn(
```
hanlint, line 21: Could we add a `DeprecationWarning` here? e.g. `warnings.warn('Accessing ...', DeprecationWarning)`

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,109 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+            }
+        }),
+        resolve=True,
+    )  # type: ignore
+    tokenizer = build_tokenizer(tokenizer_cfg['name'],
+                                tokenizer_cfg.get('kwargs', {}))
+
+    # build mpt model
+    model_config = DictConfig({
+        'name': 'hf_causal_lm',
```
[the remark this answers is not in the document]
aspfohl: Updated!

llmfoundry/callbacks/generate_callback.py
```diff
@@ -1,119 +1,30 @@
...
-                    rows.append([prompt, output_text])
 
-                text_table = wandb.Table(data=rows,
-                                         columns=['prompt', 'generation'])
-                artifact.add(text_table, 'predictions')
-                wandb.log_artifact(artifact)
-                wandb.log({'generations': text_table},
-                          step=state.timestamp.batch.value)
+        warnings.warn(
+            ('Accessing llmfoundry.callbacks.generate_callback.Generate will '
```
dakinggg, line 23:
```suggestion
            ('Accessing llmfoundry.callbacks.generate_callback.Generate '
```

dakinggg (approved):

llmfoundry/utils/builders.py
```diff
@@ -90,7 +90,15 @@ def build_callback(name: str, kwargs: Dict[str, Any]) -> Callback:
             'log_optimizer_metrics', True),)
     elif name == 'generate_callback':
         prompts = kwargs.pop('prompts')
-        return Generate(prompts=list(prompts), **kwargs)
+        interval = kwargs.pop('interval', None)
+        # Generate callback used to be batch_log_interval, so this is for backwards compatibility
+        if interval is None:
+            if 'batch_log_interval' in kwargs:
+                interval = f"{kwargs.pop('batch_log_interval')}ba"
```
irenedea, line 97: Let's raise a deprecation error here.

dakinggg: i assume you mean deprecation warning? but yeah, good call.

irenedea: yup, warning!

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+        resolve=True,
+    )  # type: ignore
+    tokenizer = build_tokenizer(tokenizer_cfg['name'],
+                                tokenizer_cfg.get('kwargs', {}))
+
+    # build mpt model
+    model_config = DictConfig({
+        'name': 'mpt_causal_lm',
+        'pretrained_model_name_or_path': 'mosaicml/mpt-7b',
+        'pretrained': False,
```
irenedea, line 139: this can be removed

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+        'config_overrides': {
+            'd_model': 128,
+            'n_heads': 4,
+            'n_layers': 2,
+            'expansion_ratio': 2,
+            'attn_config': {
+                'attn_impl': 'triton',
```
irenedea, line 146: this can also be removed

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+            },
+        },
+    })
+    model = COMPOSER_MODEL_REGISTRY['mpt_causal_lm'](model_config, tokenizer)
```
irenedea, line 150: nit: use 'model_config.name' instead of the string directly, just want to keep this programatic with no string literals if we can

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+        },
+        'drop_last': False,
+        'num_workers': 4,
+        'pin_memory': False,
+        'prefetch_factor': 2,
+        'persistent_workers': False,
+        'timeout': 0
+    }
+
+    dataloader_cfg = om.create(dataloader_cfg)
```
irenedea, line 120: nit: instead of using om.create I think we can use DictConfig to wrap the dictionary construction as we do for model_config on L136

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
 
 def test_init_hfhub_mpt_cpu():
     test_init_hfhub_mpt(device='cpu', attn_impl='torch')
+
+
+@pytest.mark.gpu
+@patch.object(ComposerGenerate,
+              'generate',
+              side_effect=ComposerGenerate.generate,
```
irenedea, line 86: oo this is my first time seeing side_effect...jw, what's the difference between using side_effect=ComposerGenerate.generate and wraps=ComposerGenerate.generate?

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+    model = composer_device.module_to_device(model)
+
+    # generate callback
+    prompts = [
+        'The best banana bread recipe is',
+        '2+2=',
+        'how much wood could a woodchuck chuck',
+    ]
+    gen_interval = 1
+    generate_cb = ComposerGenerate(
```
irenedea, line 160: Instead of patching generate in the function header, wdyt about wrapping ComposerGenerate in a Mock like so:
```
generate_cb = Mock(wraps=ComposerGenerate(...), autospec=True)

# other code...

#Assertion
generate_cb.generate.assert_called_once()
```

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+        'pin_memory': False,
+        'prefetch_factor': 2,
+        'persistent_workers': False,
+        'timeout': 0
+    }
+
+    dataloader_cfg = om.create(dataloader_cfg)
+
+    # build tokenizer
+    tokenizer_cfg: Dict[str, Any] = om.to_container(
```
irenedea, line 123: is this om.to_container necessary? it is just turning the DictConfig back to a dict..?

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+        DictConfig({
+            'name': 'EleutherAI/gpt-neox-20b',
+            'kwargs': {
+                'model_max_length': max_seq_len,
```
irenedea, line 127: i think it might be unnecessary to pass max_seq_len here for testing purposes? cc @dakinggg

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+            }
+        }),
+        resolve=True,
+    )  # type: ignore
+    tokenizer = build_tokenizer(tokenizer_cfg['name'],
```
irenedea, line 132: I think L123-131 are unncessary. We can just inline the tokenizer name and kwargs.

tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,108 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+
+    # generate callback
+    prompts = [
+        'The best banana bread recipe is',
+        '2+2=',
+        'how much wood could a woodchuck chuck',
+    ]
+    gen_interval = 1
+    generate_cb = ComposerGenerate(
+        prompts,
```
irenedea, line 161: nit: explicitly passing in the batch_size here might be more clear and help us avoid issues if we change the default batch_size in composer. Maybe batch_size=1 and then the number of generate calls will be `len(prompts)`, or vice versa

tests/test_hf_mpt_gen.py
irenedea: also TIL about autospec=True..wish i knew about it sooner loll

aspfohl: autospec is awesome 🙂 

Looks like theres a lot of questions about this (see comment) https://stackoverflow.com/questions/73086280/is-there-any-difference-between-using-mock-patch-object-with-wraps-vs-side-effec
🤷‍♀️ I think I can copied this from another example we had

tests/test_hf_mpt_gen.py
aspfohl: it passed without it, I think so

PR #640, opened 2023-10-02
Change gauntlet avging

bmosaicml:
Enables gauntlet to specify different category aggregations (as opposed to simply averaging all categories together).

This is necessary because I introduced lite categories and lm task subscore categories, but still wish to retain the original categories

run: `mpt-eval-kn9FBB`


scripts/eval/yamls/eval_gauntlet.yaml
```diff
@@ -2,6 +2,27 @@ eval_gauntlet:
   weighting: EQUAL
   subtract_random_baseline: true
   rescale_accuracy: true
+  averages:
```
hanlint, line 5: While this is extremely configurable, the challenge is still that "core_average" as it shows up in w&b could still mean many things if someone accidentally excluded (or added) a category in the yaml construction.

I wonder if we should be hard-coding some of these?

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -120,6 +129,17 @@ def compute_averages(self, state: State) -> Dict[str, float]:
 
         return {k: sum(v) / len(v) for k, v in results.items()}
 
+    def calculate_averages(self, composite_scores):
```
samhavens, line 132: this has a confusingly similar name to `compute_averages`, could we add some explanation?

Also I think the other methods have type hints, `composite_scores: Dict[str, float]`?

hanlint: May also want to avoid a method that has the side effect of updating composite_scores, and also requires it be a class method. May be cleaner to make this a static method that just does one thing -- compute the named averages:
```
named_averages = compute_named_averages(composite_scores, averages)
composite_scores.update(named_averages)
```

by the way, this still has an issue if the `averages` contains the same name as a category in composite_scores, it gets overridden, so we need some protection there?

Maybe the named averages should be stored in a separate variable name?

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -95,6 +96,14 @@ def __init__(self,
                 assert weight is not None
                 benchmark['weighting'] = weight
 
+
+        self.averages = {}
+        if averages is not None:
+            self.averages = dict(averages)
+        else:
+            # if no averages spec provided, simply average everything
+            self.averages['average'] = list(self.categories.keys())
```
sashaDoubov, line 105: for disambiguity, should this be `global_average`? I worry about `average` being misleading, tho this is a minor point.

scripts/eval/yamls/eval_gauntlet.yaml
bmosaicml: I think this is something we should discuss for future re-factoring @dakinggg

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -22,6 +22,22 @@ class Weighting(Enum):
     LOG_SAMPLE_SZ = 3
 
 
+def calculate_named_averages(average_names: Dict[str, list],
```
dakinggg, line 25: Please add a docstring

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -22,6 +22,22 @@ class Weighting(Enum):
...
+                             category_scores: Dict[str, float]):
+    average_scores = {}
+    for avg_name, category_list in average_names.items():
+        composite_subset = {
+            category: score
+            for category, score in category_scores.items()
+            if category in category_list
+        }
+        average_scores[avg_name] = sum(composite_subset.values()) / len(
```
dakinggg, line 34: please break this line out into an if statement for readability

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -43,6 +59,7 @@ class EvalGauntlet(Callback):
         rescale_accuracy (bool): Flag determining whether to rescale the accuracy on each benchmark
                                  by (1-random_baseline_accuracy) before aggregating. Using this ensures that all benchmarks max out at 1.0.
         benchmark_sizes (Optional[dict]): Optional data on benchmark sizes, used when not relying on equal weighting.
+        averages (Optional[dict]): Optional specification on how to which categories to average together and the name under which to report each average.
```
dakinggg, line 62: not sure what this is supposed to say exactly, think it has a typo or two. please update and also include an example of what this looks like

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -66,13 +84,12 @@ def __init__(self,
             )
 
         self.categories = categories
+        self.category_names = [conf.get('name', '') for conf in self.categories]
```
dakinggg, line 87: was trying to understand this line and realized the docstring doesn't mention categories. can you update please?

dakinggg, line 87: what does it mean for a category not to have a name?

llmfoundry/callbacks/eval_gauntlet_callback.py
```diff
@@ -95,7 +112,20 @@ def __init__(self,
                 assert weight is not None
                 benchmark['weighting'] = weight
 
-    def compute_averages(self, state: State) -> Dict[str, float]:
+        self.averages = {}
+        if averages is not None:
+            self.averages = dict(averages)
```
dakinggg, line 117: why do you need the dict cast?

tests/test_eval_gauntlet.py
```diff
@@ -106,4 +106,5 @@ def test_gauntlet_callback():
         name = f'icl/metrics/eval_gauntlet/{category}'
         assert result[name] == pytest.approx(0.25)
 
-    assert result['icl/metrics/eval_gauntlet/average'] == pytest.approx(0.25)
```
dakinggg, line 109: I think it is time to add more unit tests here. This test tests one particular setup, but there are many configurable and optional args to the eval gauntlet. Can you please add more tests that test a variety of the options? I would recommend using `pytest.mark.parametrize` to specify different combinations of the various args to the callback.

samhavens (approved):
All feedback seems to be addressed.

PR #642, opened 2023-10-02
Add flag to disable train metrics

mvpatel2000:
Add flag to disable train metrics

### DIFF
--- a/llmfoundry/models/hf/hf_causal_lm.py
+++ b/llmfoundry/models/hf/hf_causal_lm.py
@@ -66,7 +66,8 @@ class ComposerHFCausalLM(HuggingFaceModelWithZLoss):
                  tokenizer: PreTrainedTokenizerBase):
         # set up training and eval metrics
         use_train_metrics = om_model_config.get('use_train_metrics', True)
-        train_metrics = [LanguageCrossEntropy(), LanguagePerplexity()] if use_train_metrics else []
+        train_metrics = [LanguageCrossEntropy(),
+                         LanguagePerplexity()] if use_train_metrics else []
         eval_metrics = [
             LanguageCrossEntropy(),
             LanguagePerplexity(),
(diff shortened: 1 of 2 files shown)


llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -65,10 +65,7 @@ def __init__(self, om_model_config: Union[DictConfig,
                                               nn.Module],
                  tokenizer: PreTrainedTokenizerBase):
         # set up training and eval metrics
-        train_metrics = [
-            LanguageCrossEntropy(),
-            LanguagePerplexity(),
-        ]
+        train_metrics = [LanguageCrossEntropy(), LanguagePerplexity()]
```
vchiley, line 68:
```suggestion
```

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -92,6 +89,9 @@ def __init__(self, om_model_config: Union[DictConfig,
                     'which is not significantly slower and not compatible with the LLM foundry training code, rather than the code release by MosaicML.'
                 )
 
+            if not om_model_config.get('use_train_metrics', True):
+                train_metrics = []
+
```
vchiley, line 94:
```suggestion
            use_train_metrics = om_model_config.get('use_train_metrics', True)
        else:
            use_train_metrics = True

        train_metrics = [LanguageCrossEntropy(),
                         LanguagePerplexity()] if use_train_metrics else []
```

vchiley (approved):
not sure if my suggestion helps make it more consistent or is just more confusing...

feel free to ignore. otherwise lgtm

llmfoundry/models/hf/hf_causal_lm.py
mvpatel2000: The spacing here doesn't quite work since `train_metrics` needs to be scoped outside the `if` (which is a very long `if`). This is due to the cases where the config is not a DictConfig (the typing and codepath here needs to be cleaned up...). So I will resolve

PR #643, opened 2023-10-02
Handle flash_attn version update in transformers main

irenedea:
Transformers has a [change](https://github.com/huggingface/transformers/pull/25598) that updates flash_attn to flash_attn 2 for llama models, but this version of flash attn is not compatible with foundry. 

manual test: codellama-34b-r1z1-8-4096-1-Hafr3R


dakinggg:
Closing in favor of #672

PR #644, opened 2023-10-03
Fix overriding of rope_scaling config

dakinggg:


llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, om_model_config: Union[DictConfig,
                     )
 
                 attr = getattr(config, k)
+                # attempt to disallow typos in nested configs
                 if isinstance(attr, Mapping):
```
vchiley, line 113: can a nested dict have defaults which are not part of the config?
what happens in instances where something **like** `rope_scaling` is in the config, but it doesn't have all possible config settings.

tests/test_hf_config.py
vchiley, line 1: looks like your tests fails due to not having hf token.

dakinggg: oops, forgot to skip that test on CI

llmfoundry/models/hf/hf_causal_lm.py
dakinggg: Isn't that the case you wrote which already exists? It will use any defaults of the current value of the attr. the new case is just because the default for rope scaling is `None`, so the user must fully specify it. model should crash if its not fully specified.

vchiley: yeah I'm saying what i wrote is probably wrong.
jk maybe just that it makes a strong assumption, an assumption that makes sense for MPT, but not necessarily in general.
either way, its a bit outside the scope of the PR.

vchiley (approved):

PR #652, opened 2023-10-07
Do not update past_key_values in place

irenedea:
Fixes the forward pass in MPTModel to update not `past_key_values` in place and instead use a new list, `presents`. 🎁 Naming is inspired by [gpt2](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L860). 

This addresses the following error:
```
 RuntimeError: attn_bias (shape: torch.Size([1, 4, 1, 2])) is expected to broadcast to shape: torch.Size([1, 4, 1, 3]).
```
which occurs during multi-gpu generation for mpt models when one rank hits EOS before the others.


tests/test_hf_mpt_gen.py
```diff
@@ -72,3 +78,86 @@ def test_init_hfhub_mpt(device: str, attn_impl: str):
...
+            # Rank 0 hits EOS immediately.
+            result.logits[:, :, EOS_TOKEN_ID] = torch.inf
+        else:
+            # Other ranks do not hit EOS.
+            result.logits[:, :, EOS_TOKEN_ID] = -torch.inf
+        return result
+
+def mock_from_config(config: MPTConfig, **_):
+    config_dict = config.to_dict()
+    config = MPTConfig.from_dict(config_dict)
```
irenedea, line 113: this seemed really odd to me that I would need to do this extra step of re-converting the config from a dictionary...Otherwise, it resulted in slightly different model behavior.

irenedea: Actually, the original config here is the MPTConfig class downloaded by transformers...so different from the one in the repo. So maybe there's been some differences between these classes?

dakinggg: you likely need to do something like this https://github.com/mosaicml/llm-foundry/blob/aa2ba9f139582348840637ae69945e3140181557/tests/test_onnx.py#L31-L33

dakinggg (approved):
LGTM apart from my comment on the test. Also wait for @vchiley to look please.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -422,6 +422,7 @@ def forward(
         )
 
         # initialize the past key values cache if it should be used
+        presents = () if use_cache else None
         if use_cache and past_key_values is None:
             past_key_values = [() for _ in range(self.config.n_layers)
```
vchiley, line 427: typing nit:
past_key_values is a list
presents is a tuple

vchiley (approved):
nit: `presents` implies the current x, not the past...
the tensor its referring to is the past and presents.
no clue why gpt2 names it like this. 
this is a nit because its just a naming convention comment. 

otherwise lgtm

dakinggg:
@irenedea is this ready to merge?

llmfoundry/models/mpt/modeling_mpt.py
irenedea: BaseModelOutputWithPast expects a tuple of tuples, but past_key_values is expected to be a list of tuples 😅

tests/test_mpt_gen.py
```diff
@@ -0,0 +1,97 @@
...
+            'alibi': use_alibi
+        },
+    })
+
+    # build tokenizer
+    tokenizer = build_tokenizer('EleutherAI/gpt-neox-20b', {})
+
+    # build model
+    model = COMPOSER_MODEL_REGISTRY[model_config.name](model_config, tokenizer)
+    model = composer_device.module_to_device(model)
```
dakinggg, line 87: shouldn't matter functionally for the purposes of this test, but you should put the model in eval mode here (`model.eval()`)

dakinggg (approved):
one nit

PR #653, opened 2023-10-07
Only log "changing autoresume" when actually changing

aspfohl:
Small update, but as we've learned this log can be pretty confusing otherwise. Right now, this logs for every single autoresume run (and runs that explicitly call to not autoresume)


j316chuck (approved):
Thanks for this!

PR #656, opened 2023-10-08
Fix lion8b error correction with torch 2.1

dblalock:
Un-disables error correction and instead works around the new FSDP limitations.


dakinggg (approved):
Thanks Davis!

PR #657, opened 2023-10-09
Inverse Square Root LR Schedule

mansheej:
Adds the Inverse Square Root LR Scheduler.

This scheduler is meant to easily enable continual learning. It consists of three components:
1. A linear LR warmup.
2. A component where the LR decays as an inverse square root in the number of steps to a constant value at infinite time.
3. An optional linear cooldown.

The image below show two examples of the LR schedule with a 10 step warmup, with either no cooldown (orange) or a 20 step cooldown (blue) run for 100 steps total.

W B Chart 10_11_2023, 12_01_34 AM


dakinggg (commented):
Will review in full as well, but is there any reason this should be in LLM foundry? It seems generically useful and I would probably prefer it is in composer directly.

mansheej:
> Will review in full as well, but is there any reason this should be in LLM foundry? It seems generically useful and I would probably prefer it is in composer directly.

This implementation and hyperparameters are a little different from how people typically do it. I wanted to first get it into LLM foundry, run some more experiments and have other people use it so we can work out some kinks and learn about some best practices, and then eventually upstream it into composer with the best practices documented.

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+        \tau_c = (t - t_{max} + t_{cooldown}) / t_{cooldown}
+
+    Where :math:`t_{warmup}` represents the warmup time, :math:`t_{scale}` represents the time scale,
+    :math:`t_{cooldown}` represents the cooldown time, :math:`t_{max}` represents the duration of this scheduler,
+    :math:`\alpha_{f,decay}` represents the learning rate multiplier that the inverse square root decays to at infinite time,
+    and :math:`\alpha_{f,cooldown}` represents the learning rate multiplier that the linear cooldown decays to.
+
+    Note, :math:`\alpha_{f,decay} >= \alpha_{f,cooldown}` to ensure that the learning rate is monotonically decreasing after warmup.
+
+    Also note, ``t_warmup``, ``t_scale``, and ``t_cooldown`` cannot be specified in units of duration; since this schedule is designed for continual learning,
```
dakinggg, line 56: Can you raise an error for this constraint?

dakinggg (commented):
Could you please add some unit tests testing this schedule produces what you expect?

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+        alpha_f_cooldown (float): Learning rate multiplier to decay linear cooldown to. Default = ``0.0``.
+        scale_warmup (bool): SSR also scales the warmup period. Default = ``False``.
+    """
+
+    def __init__(self,
+                 t_warmup: Union[str, Time],
+                 t_scale: Union[str, Time],
+                 t_cooldown: Union[str, Time],
+                 t_max: Union[str, Time] = '1dur',
+                 alpha_f_decay: float = 0.0,
```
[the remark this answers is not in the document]
mansheej: `alpha_f_decay` is the learning rate multiplier that **inverse square root decay** would decay to if trained for an infinite duration (line 51)--the inverse square root does not decay in finite time--which enables us to use this for continual learning/infinite training. So if set to `0.0`, in any finite time it will still be positive. If set to a non-zero value, this parameter allows us to effectively set a constant learning rate we want to train the model at for the continual learning portion. 

`alpha_f_cooldown` is the learning rate multiplier the **linear cooldown** portion of the learning rate will actually decay to at the end of training if we turn on cooldown.

llmfoundry/optim/scheduler.py
mansheej: I am already raising an error (line 81).

mansheej:
> Could you please add some unit tests testing this schedule produces what you expect?

Do you have any suggestions for reasonable unit tests for the LR scheduler?

llmfoundry/optim/scheduler.py
dakinggg: I think that is for a different case? The error you link is for if alpha f decay < alpha f cooldown. I'm suggesting an error if any of these params are specified in units of duration.

dakinggg:
I'd suggest:
1. the "build" function can create it successfully
2. set up a couple simple schedules with known values that you expect and test that your scheduler code produces exactly the schedule you expect.

llmfoundry/optim/scheduler.py
mansheej: Ah yes, that would be lines 96, 97, 98.

dakinggg: Ah I see. Can we also raise an error on init if any of them are in dur? It will fail faster that way instead of waiting until the run starts.

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+                 alpha_f_cooldown: float = 0.0,
+                 scale_warmup: bool = False):
+        if alpha_f_decay < alpha_f_cooldown:
+            raise ValueError('Required: alpha_f_decay >= alpha_f_cooldown.')
```
b-chu, line 90: Use an f-string to display values for both

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+        self.t_max = t_max
+        self.alpha_f_decay = alpha_f_decay
+        self.alpha_f_cooldown = alpha_f_cooldown
+        self.scale_warmup = scale_warmup
+        self.warmup_scheduler = LinearScheduler(alpha_i=0.0,
+                                                alpha_f=1.0,
+                                                t_max=t_warmup)
+
+    def __call__(self, state: State, ssr: float = 1.0) -> float:
+        assert state.max_duration is not None, 'max_duration should be set whenever schedulers are invoked'
```
b-chu, line 106: Raise an exception
https://google.github.io/styleguide/pyguide.html#244-decision

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+
+    def __init__(self,
+                 t_warmup: Union[str, Time],
+                 t_scale: Union[str, Time],
+                 t_cooldown: Union[str, Time],
+                 t_max: Union[str, Time] = '1dur',
+                 alpha_f_decay: float = 0.0,
+                 alpha_f_cooldown: float = 0.0,
+                 scale_warmup: bool = False):
```
b-chu, line 88: Return type hint for `__init__` should be `None`
https://peps.python.org/pep-0484/#the-meaning-of-annotations

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+            To change this behavior, set ``scale_warmup=True``.
+
+    Args:
+        t_warmup (str | Time): Warmup time.
+        t_scale (str | Time): Time scale.
+        t_cooldown (str | Time): Cooldown time.
+        t_max (str | Time): The duration of this scheduler. Default = ``"1dur"``.
+        alpha_f_decay (float): Learning rate multiplier to decay inverse square root decay to. Default = ``0.0``.
+        alpha_f_cooldown (float): Learning rate multiplier to decay linear cooldown to. Default = ``0.0``.
+        scale_warmup (bool): SSR also scales the warmup period. Default = ``False``.
```
b-chu, line 78:
```suggestion
        t_warmup (str | Time): The warmup time.
        t_scale (str | Time): The time scale.
        t_cooldown (str | Time): The cooldown time.
        t_max (str | Time): The duration of this scheduler. Default = ``"1dur"``.
        alpha_f_decay (float): The learning rate multiplier to decay inverse square root decay to. Default = ``0.0``.
        alpha_f_cooldown (float): The learning rate multiplier to decay linear cooldown to. Default = ``0.0``.
        scale_warmup (bool): If ``True``, SSR also scales the warmup period. Default = ``False``.
```
Nit

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+        _raise_if_units_dont_match(self.t_scale, state.max_duration)
+        _raise_if_units_dont_match(self.t_cooldown, state.max_duration)
+
+        t_warmup = _convert_time(self.t_warmup, state)
+        if t_warmup.value == 0:
+            warnings.warn(
+                textwrap.dedent("""\
+                The warmup duration is 0. If you specified warmup as a fraction of total
+                training duration, take note that the warmup duration is calculated in the
+                same unit as the trainer's max_duration parameter."""))
```
b-chu, line 117:
```suggestion
                The warmup duration is 0. If warmup was specified as a fraction of the total
                training duration, the warmup duration is calculated in the
                same unit as the trainer's max_duration parameter."""))
```
Nit: I didn't fix whitespace when I reworded this.

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+        self.t_cooldown = t_cooldown
+        self.t_max = t_max
+        self.alpha_f_decay = alpha_f_decay
+        self.alpha_f_cooldown = alpha_f_cooldown
+        self.scale_warmup = scale_warmup
+        self.warmup_scheduler = LinearScheduler(alpha_i=0.0,
+                                                alpha_f=1.0,
+                                                t_max=t_warmup)
+
+    def __call__(self, state: State, ssr: float = 1.0) -> float:
```
b-chu, line 105: Write a docstring for this function

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+        time = Time.from_timestring(time)
+    if isinstance(t_max, str):
+        t_max = Time.from_timestring(t_max)
+    if time.unit != t_max.unit:
+        raise ValueError(
+            'All time units must be the same as max_duration units.')
+
+
+class InverseSquareRootWithWarmupScheduler(ComposerScheduler):
+    r"""Inverse square root LR decay with warmup and optional linear cooldown.
```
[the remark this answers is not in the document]
b-chu: Can you also post a screenshot of this docstring?

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+        if state.timestamp < t_warmup:
+            if self.scale_warmup:
+                return self.warmup_scheduler(state, ssr)
+            return self.warmup_scheduler(state)
+
+        t_scale = _convert_time(self.t_scale, state, ssr=ssr)
+        t_cooldown = _convert_time(self.t_cooldown, state, ssr=ssr)
+        t_max = _convert_time(self.t_max, state, ssr=ssr)
+        current_time = state.timestamp.get(t_scale.unit)
+
```
b-chu, line 128: Add some comments explaining the math here

b-chu (changes requested):
Add some unit tests to test functionality

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,152 @@
...
+        raise ValueError(
+            'All time units must be the same as max_duration units.')
+
+
+def _raise_if_units_dur(time: Union[str, Time]) -> None:
+    if isinstance(time, str):
+        time = Time.from_timestring(time)
+    if time.unit == TimeUnit('dur'):
+        raise ValueError(
+            't_warmup, t_scale, and t_cooldown cannot be in units of "dur".')
```
b-chu, line 31: Raise the error in a way that the variable name and its value are printed

llmfoundry/optim/scheduler.py
dakinggg: This one is just for pyright

llmfoundry/optim/scheduler.py
mansheej: Added tests: lines 91, 92, 93.

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,141 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
mansheej: Done

llmfoundry/optim/scheduler.py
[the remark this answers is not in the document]
mansheej: Just following the conventions in composer schedulers. Not sure how to generate formatted docstrings, is there documentation for LLM foundry?

llmfoundry/optim/scheduler.py
mansheej: Also formatted them as @b-chu requested.

llmfoundry/optim/scheduler.py
mansheej: This is what I typically do but almost nothing in composer or LLM foundry does this so wasn't sure if I should break convention.

llmfoundry/optim/scheduler.py
mansheej: I was keeping the convention of schedulers in composer where only the protocol has a docstring but none of the schedulers do.

llmfoundry/optim/scheduler.py
mansheej: Done

llmfoundry/optim/scheduler.py
mansheej: Done

llmfoundry/optim/scheduler.py
mansheej: Done

mansheej:
Added unit tests and tried to address all the comments. Failing the Code Quality Checks, but I'm not sure why from the Error message.

dakinggg:
@mansheej try running `pre-commit run --all-files` locally

llmfoundry/optim/scheduler.py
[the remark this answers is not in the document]
codestar12: image

tests/test_scheduler.py
```diff
@@ -0,0 +1,139 @@
...
+import torch
+from composer.core import State, Time, TimeUnit
+from composer.devices import DeviceCPU, DeviceGPU
+from composer.models import ComposerClassifier
+from composer.optim.scheduler import ComposerScheduler
+
+from llmfoundry.optim.scheduler import InverseSquareRootWithWarmupScheduler
+
+MAX_DURATION = '100ba'
+STEPS_PER_EPOCH = 100
```
b-chu, line 17: Prepend private global constants with `_`

tests/test_scheduler.py
```diff
@@ -0,0 +1,139 @@
...
+        self.fc1 = fc1
+        self.fc2 = fc2
+
+    def param_init_fn(self, module: torch.nn.Module):
+        init_fn = partial(torch.nn.init.normal_, mean=0.0, std=0.1)
+
+        if isinstance(module, torch.nn.Linear):
+            init_fn(module.weight)
+            if module.bias is not None:
+                torch.nn.init.zeros_(module.bias)
```
b-chu, line 71: Can this be simplified?

b-chu: @dakinggg may have some comments on best practices for tests in this repo

tests/test_scheduler.py
b-chu, line 1: Add tests for the errors that get raised.

b-chu, line 1: Make sure the tests hit all the if statements
```
        if state.timestamp < t_warmup:
            if self.scale_warmup:
```
```
if t_cooldown_start < t_warmup:
```
```
if state.timestamp < t_cooldown_start:
```
etc

llmfoundry/optim/scheduler.py
hanlint: The assert is used for type narrowing to satisfy pyright since in state, `max_duration` is `Optional[]`. This is for internal correctness (max_duration should already be set when a schedule is stepped).

llmfoundry/optim/scheduler.py
hanlint: Probably OK to follow the convention in composer (https://github.com/mosaicml/composer/blob/dev/composer/optim/scheduler.py) of documenting the class, but not the `__call__`. 

The `__call__` is already documented in the base class `ComposerScheduler`.

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,161 @@
...
+                 t_warmup: Union[str, Time],
+                 t_scale: Union[str, Time],
+                 t_cooldown: Union[str, Time],
+                 t_max: Union[str, Time] = '1dur',
+                 alpha_f_decay: float = 0.0,
+                 alpha_f_cooldown: float = 0.0,
+                 scale_warmup: bool = False) -> None:
+        if alpha_f_decay < alpha_f_cooldown:
+            raise ValueError(
+                f'Required: alpha_f_decay >= alpha_f_cooldown. Current: alpha_f_decay={alpha_f_decay}, alpha_f_cooldown={alpha_f_cooldown}'
```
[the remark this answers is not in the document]
mansheej: Done

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,161 @@
...
+
+        t_scale = _convert_time(self.t_scale, state, ssr=ssr)
+        t_cooldown = _convert_time(self.t_cooldown, state, ssr=ssr)
+        t_max = _convert_time(self.t_max, state, ssr=ssr)
+        current_time = state.timestamp.get(t_scale.unit)
+
+        t_shift = t_scale - t_warmup
+        t_cooldown_start = t_max - t_cooldown
+        if t_cooldown_start < t_warmup:
+            t_cooldown_start = t_warmup
```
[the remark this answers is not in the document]
mansheej: This is pretty straightforward behavior. What this should be is max(t_warmup, t_max - t_cooldown) but the Time abstraction does not seem to play well with max so I implemented it with an if.

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,161 @@
...
+            alpha_i = self.alpha_f_decay + coeff * (1.0 - self.alpha_f_decay)
+
+            if t_cooldown.value == 0:
+                return alpha_i
+
+            # linearly decay the LR from its value at the step at which cooldown
+            # started to alpha_f_cooldown over t_cooldown time
+            frac_of_cooldown = ((current_time - t_cooldown_start) /
+                                t_cooldown).value
+            frac_of_cooldown = min(1.0, frac_of_cooldown)
```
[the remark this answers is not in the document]
mansheej: Always greater than 0.

tests/test_scheduler.py
dakinggg: this model isn't actually used beyond passing it to the state constructor afaict, so i think you can just pass like `torch.nn.Linear(5, 5)` as the model arg and avoid this whole large class.

tests/test_scheduler.py
dakinggg: This is a test file, so I don't think super strict on public/private things is particularly important. That being said, I don't think you actually need a real dataloader.

tests/test_scheduler.py
```diff
@@ -0,0 +1,139 @@
...
+
+    state = dummy_schedulers_state
+    assert state.dataloader_len is not None
+    assert state.max_duration is not None
+    state.max_duration = Time(value=int(state.max_duration.value * ssr),
+                              unit=state.max_duration.unit)
+    for test_time, expected_lr in zip(test_times, expected_lrs):
+        parsed_time = Time.from_timestring(test_time)
+        assert parsed_time.unit in [TimeUnit.EPOCH, TimeUnit.BATCH]
+        if parsed_time.unit == TimeUnit.EPOCH:
```
dakinggg, line 123: did you intend to test epochs as well? If not, can just delete this code branch.

dakinggg (commented):
Basically LGTM, left a few comments on simplifying the tests.

tests/test_scheduler.py
mansheej: Did not intend to test epochs so I deleted this code branch. This function was lifted directly from composer scheduler tests so I don't want to change it too much so that it is straightforward to move the tests over when we upstream to composer.

tests/test_scheduler.py
mansheej: Changed to torch.nn.Linear.

tests/test_scheduler.py
mansheej: done

tests/test_scheduler.py
mansheej: This is perhaps something that can be filed for an improvement/added during upstreaming to composer?

mansheej: I have tested the important branches. The rest (such as scale_warmup) are things that aren't really relevant for LLMs but I kept it there for compatibility with the rest of the Composer Schedulers. 
This is perhaps something that can be filed for an improvement/added during upstreaming to composer?

mansheej:
Addressed most of the comments. The rest can perhaps be filed for improvement/addressed when we upstream to Composer?

dakinggg (approved):
I am ok with the level of test coverage here. Before merging, could you please make the PR description more complete?

In this case, I think an example schedule produced by this code (e.g. a wandb graph) and a one sentence description of the gist of the schedule would suffice. Thanks!

tests/test_scheduler.py
mansheej: Done

mansheej: Done.

llmfoundry/optim/scheduler.py
```diff
@@ -0,0 +1,153 @@
...
+        if state.timestamp < t_warmup:
+            return self.warmup_scheduler(state)
+
+        t_scale = _convert_time(self.t_scale, state, ssr=ssr)
+        t_cooldown = _convert_time(self.t_cooldown, state, ssr=ssr)
+        t_max = _convert_time(self.t_max, state, ssr=ssr)
+        current_time = state.timestamp.get(t_scale.unit)
+
+        t_shift = t_scale - t_warmup
+        # t_cooldown_start = max(t_warmup, t_max - t_cooldown)
```
b-chu, line 125: Delete this comment

tests/test_scheduler.py
```diff
@@ -0,0 +1,109 @@
...
+                                                 t_scale=f'1{scale_unit}',
+                                                 t_cooldown=f'1{cooldown_unit}')
+
+def test_unit_dur_error():
+    with pytest.raises(ValueError, match='alpha_f_decay >= alpha_f_cooldown.'):
+        _ = InverseSquareRootWithWarmupScheduler(t_warmup='10ba',
+                                                 t_scale='10ba',
+                                                 t_cooldown='10ba',
+                                                 alpha_f_decay=0.0,
+                                                 alpha_f_cooldown=0.1)
```
b-chu, line 109: Add a newline at end of file

b-chu (approved):
Two small changes, thanks for the tests!

tests/test_scheduler.py
mansheej: done

llmfoundry/optim/scheduler.py
mansheej: done

PR #670, opened 2023-10-12
Adding Mosaic logger + logging data validated event

jjanezhang:
Adding Mosaic Logger + data validated log #670

1. Added mosaic logger to log foundry information for each run.
2. Logged the timestamp for data validation complete to mosaic logger

Testing
1. Started run from MCLI using this [YAML](https://github.com/mosaicml/testing/blob/c015a605779fb807b1f0fecdc8b15462fc582c7e/regression/llm-foundry/mpt-125m-sharded-resumption.yaml#L14). Checked run metadata after run finished and confirmed that data_validated timestamp is logged in the run metadata as well as all other composer logged metrics.

image


llmfoundry/callbacks/__init__.py
```diff
@@ -19,13 +20,7 @@
...
-    'MonolithicCheckpointSaver',
-    'GlobalLRScaling',
-    'LayerFreezing',
-    'ScheduledGarbageCollector',
-    'EvalGauntlet',
-    'ModelGauntlet',
-    'HuggingFaceCheckpointer',
+    'FDiffMetrics', 'Generate', 'MonolithicCheckpointSaver', 'GlobalLRScaling',
+    'LayerFreezing', 'ScheduledGarbageCollector', 'EvalGauntlet',
+    'ModelGauntlet', 'HuggingFaceCheckpointer', 'RunEventsCallback'
```
dakinggg, line 25: add trailing comma back

scripts/train/train.py
```diff
@@ -449,10 +451,17 @@ def main(cfg: DictConfig) -> Trainer:
...
-        for name, logger_cfg in logger_configs.items()
-    ] if logger_configs else None
+    loggers = None
+    if logger_configs:
+        loggers = [
+            build_logger(str(name), logger_cfg)
+            for name, logger_cfg in logger_configs.items()
+        ]
+
+    mosaicMlLogger = None
```
dakinggg, line 461:
```suggestion
    mosaicml_logger = None
```

scripts/train/train.py
```diff
@@ -449,10 +451,17 @@ def main(cfg: DictConfig) -> Trainer:
     scheduler = build_scheduler(scheduler_name, scheduler_config)
 
     # Loggers
-    loggers = [
-        build_logger(str(name), logger_cfg)
-        for name, logger_cfg in logger_configs.items()
-    ] if logger_configs else None
+    loggers = None
```
dakinggg, line 454: This setup isn't quite what I meant before.

I think the cleanest setup is to just copy the same if statement as in composer to conditionally make the mosaic logger

```
        if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower() == 'true' and os.environ.get(
                MOSAICML_ACCESS_TOKEN_ENV_VAR) is not None and not any(isinstance(x, MosaicMLLogger) for x in loggers):
            log.info('Detected run on MosaicML platform. Adding MosaicMLLogger to loggers.')
            mosaicml_logger = MosaicMLLogger()
            loggers.append(mosaicml_logger)
```

and then just call `mosaicml_logger.log_metrics` with whatever info you need. Everything else should just work after that I think.

scripts/train/train.py
```diff
@@ -454,6 +458,14 @@ def main(cfg: DictConfig) -> Trainer:
         for name, logger_cfg in logger_configs.items()
     ] if logger_configs else None
 
+    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower(
+    ) == 'true' and os.environ.get(MOSAICML_ACCESS_TOKEN_ENV_VAR):
+        mosaicml_logger = MosaicMLLogger()
+        if loggers:
+            loggers.append(mosaicml_logger)
+    else:
```
dakinggg, line 466: nit: prefer just defining `mosaicml_logger = None` above the if statement

scripts/train/train.py
```diff
@@ -454,6 +458,14 @@ def main(cfg: DictConfig) -> Trainer:
         for name, logger_cfg in logger_configs.items()
     ] if logger_configs else None
 
+    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower(
+    ) == 'true' and os.environ.get(MOSAICML_ACCESS_TOKEN_ENV_VAR):
+        mosaicml_logger = MosaicMLLogger()
+        if loggers:
+            loggers.append(mosaicml_logger)
```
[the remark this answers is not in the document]
dakinggg: +1

scripts/train/train.py
```diff
@@ -473,6 +485,10 @@ def main(cfg: DictConfig) -> Trainer:
         tokenizer,
         device_train_batch_size,
     )
+
+    if mosaicml_logger:
```
dakinggg, line 489:
```suggestion
    if mosaicml_logger is not None:
```

dakinggg (commented):
Could you please update the PR description based on the current implementation?

scripts/train/train.py
irenedea, line 489: Let's move all the mosaicml logger logic here instead of instantiating in above and using it here. It'll be easier to maintain in one place.

scripts/train/train.py
```diff
@@ -454,6 +458,14 @@ def main(cfg: DictConfig) -> Trainer:
         for name, logger_cfg in logger_configs.items()
     ] if logger_configs else None
 
+    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower(
```
irenedea, line 461: nit: Add a comment here that describes this boolean check because the boolean check is quite verbose

scripts/train/train.py
dakinggg: we're probably going to end up adding more events and moving the init back to the top shortly...so maybe just leave it there? either way is fine with me tho

scripts/train/train.py
irenedea: Alternatively: add a logger.info('...') call that describes this.

scripts/train/train.py
irenedea: oh good point, i just thought it was this one event. If we add more than one event, I'd prefer having the instantiation separate actually

scripts/train/train.py
irenedea: Just saw daniel's comment from yest: https://github.com/mosaicml/llm-foundry/pull/670#discussion_r1361318147

I think we should also include that check to see if a MosaicMLLogger is already specified as Daniel does in his code comment

dakinggg: +1

jjanezhang: Is it possible to have an instance of `MosaicMLLogger` in `loggers` because `MosaicMLLogger` isn't defined in `build_logger` above ?

dakinggg: That is a fair point. I think it is not possible today, but I'd rather include the check for safety against future code.

irenedea (approved):
LGTM! 🚀 🔥

scripts/train/train.py
```diff
@@ -454,6 +458,16 @@ def main(cfg: DictConfig) -> Trainer:
         for name, logger_cfg in logger_configs.items()
     ] if logger_configs else None
 
+    # Adds mosaicml logger to composer if the run was sent from Mosaic platform, access token is set, and mosaic logger wasn't previously added
+    mosaicml_logger = None
+    if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower(
+    ) == 'true' and os.environ.get(MOSAICML_ACCESS_TOKEN_ENV_VAR):
+        mosaicml_logger = MosaicMLLogger()
```
dakinggg, line 465: can we construct this inside the if statement too?

dakinggg (commented):
Before merging, can you please paste into the PR description evidence of a manual test that successfully logged the metric you are interested in?

scripts/train/train.py
```diff
@@ -454,6 +458,16 @@ def main(cfg: DictConfig) -> Trainer:
...
+        if loggers and not any(isinstance(x, MosaicMLLogger) for x in loggers):
+            loggers.append(mosaicml_logger)
+        else:
+            loggers = [mosaicml_logger]
```
irenedea, line 469: We will hit this statement if `loggers` exists and there is a preexisting MosaicMLLogger logger. This would be incorrect because we'd lose all the loggers and use a fresh mosaicmllogger

irenedea: i'd rewrite to something like:
```
loggers = [
        build_logger(str(name), logger_cfg)
        for name, logger_cfg in logger_configs.items()
    ] if logger_configs else []  # Make this a list instead of None so we don't have to do extra checks

if < mosaic env var checks, etc.>:
# Get the mosaicml_logger from loggers if possible
    mosaicml_logger = next(logger in loggers  if  isinstance(logger, MosaicMLLogger)), None)
    if mosaicml_logger is None:
        mosaicml_logger = MosaicMLLogger()
        loggers.append(mosaicml_logger)

```

PR #672, opened 2023-10-13
Allow flash attention 2 and upgrade to transformers 4.34.1

dakinggg:
This PR does a few things that come together:

- [x] Upgrades to transformers 4.34
- [x] Small compatibility fixes for 4.34
- [x] Transformers 4.34 breaks when flash attn <2 is installed, so some monkeypatching is added to get around that
- [x] Transformers 4.34 adds support for using flash attention 2 with some models, so we add support for that arg
- [x] Tests for the new mistral model are added
- [x] Adds documentation about the new flash attention options
- [x] test coverage for all iterations of flash version, patch type, flash2, etc
- [x] Wait for 4.34.1 and pin to that (and remove some of the above patches)
- [x] Manual tests for llama and mistral. confirm that both train with reasonable loss curves, and model print out shows using flash attention 2
- [x] Wait for https://github.com/mosaicml/composer/pull/2635 to be merged (doesn't need to be released, just want to make sure we have confidence in the new transformers version)


tests/test_huggingface_flash.py
```diff
@@ -0,0 +1,202 @@
...
+@pytest.mark.parametrize('patch_fn_name', ['torch', 'triton'])
+@pytest.mark.parametrize('explicit_mask', [True, False])
+@pytest.mark.parametrize(
+    'model_name', ['meta-llama/Llama-2-7b-hf', 'meta-llama/Llama-2-70b-hf'])
+@pytest.mark.gpu
+def test_patch_equivalence(patch_fn_name: str, explicit_mask: bool,
+                           model_name: str):
+    if 'HUGGING_FACE_HUB_TOKEN' not in os.environ:
+        pytest.skip(
+            'The CI cluster does not have access to the Llama models, so skip this test.'
```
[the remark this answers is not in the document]
dakinggg: yes, Chuck was working on that. Will check back with him next week on the status.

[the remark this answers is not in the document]
dakinggg: In the meantime though, I ran all the Llama tests on both flash1 and flash2 images.

scripts/train/README.md
```diff
@@ -332,6 +333,52 @@ train_loader:
...
+    name: hf_causal_lm
+    use_flash_attention_2: True # Will be automatically set to True if Flash Attention V2 is installed and the model supports it
+    ...
+```
+HuggingFace models currently only support Flash Attention V2.
+
+For Llama specifically, we have another option if you would like to use the Triton implementation of Flash Attention. You can specify this in your YAML like so:
+```yaml
+model:
+    name: llama
```
irenedea, line 378: does this mean llama is in the composer model registry?
should this be `name: hf_causal_lm`?

scripts/train/README.md
```diff
@@ -332,6 +333,52 @@ train_loader:
...
+        attn_config:
+            attn_impl: flash
+        ...
+```
+
+For any HuggingFace model that supports Flash Attention (e.g. Llama and Mistral), you can specify Flash Attention in your YAML like so:
+```yaml
+model:
+    name: hf_causal_lm
+    use_flash_attention_2: True # Will be automatically set to True if Flash Attention V2 is installed and the model supports it
```
irenedea, line 370: why do we have a separate flag for flash attention 2? why not have this be part of the attn_impl flag?

Like:
flash => v1 or v2 depending on what is installed
flash1 => v1
flash2 => v2

irenedea: Is this a HF thing?

irenedea (commented):
looks good! just had a few q's

tests/test_huggingface_flash.py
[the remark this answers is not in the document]
irenedea: sg as we just need to remember to remove the pytest.skip when the CI HF support is merged in

scripts/train/README.md
dakinggg: Oops, typo

scripts/train/README.md
dakinggg: Yeah, this is what the hf flag is called. They don't support flash v1, only v2.

tests/test_huggingface_flash.py
```diff
@@ -0,0 +1,199 @@
...
+    attn_output, _, _ = attention(
+        hidden_states=hidden_states,
+        attention_mask=causal_mask if explicit_mask else None,
+        position_ids=None,
+        past_key_value=None,
+        use_cache=False,
+    )
+
+    reproducibility.seed_all(42)
+    LlamaAttention.forward = patch_fn
```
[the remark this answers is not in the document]
dakinggg: 👍

scripts/train/train.py
```diff
@@ -216,6 +217,11 @@ def main(cfg: DictConfig) -> Trainer:
         os.environ[
             'PYTORCH_CUDA_ALLOC_CONF'] = f'max_split_size_mb:{max_split_size_mb}'
 
+    # Set CUDA lazy loading
```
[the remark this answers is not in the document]
dakinggg: 👍

scripts/train/README.md
```diff
@@ -332,6 +333,52 @@ train_loader:
...
+    ...
+```
+HuggingFace models currently only support Flash Attention V2.
+
+For Llama specifically, we have another option if you would like to use the Triton implementation of Flash Attention. You can specify this in your YAML like so:
+```yaml
+model:
+    name: llama
+    attention_patch_type: triton
```
irenedea, line 379: whats the diff between using attention_patch_type and 
```
    config_overrides:
        attn_config:
            attn_impl: triton
        ...
```

tests/test_huggingface_flash.py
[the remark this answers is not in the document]
irenedea: oops i submitted that comment early-- i was just gonna add that i don't think you'll need to do the resets with this, so it'll be safer

irenedea (approved):
Lgtm! thank you! just a couple of super minor things

scripts/train/README.md
dakinggg: the latter only applies to MPT because it is specific to our config class. The former only applies to Llama because we only implemented the path for llama (although we were going to add support for more HF models, but now this is not necessary because HF supports flash v2 directly).

PR #673, opened 2023-10-13
Add fixtures

irenedea:
Adds fixtures for models and boiler plate testing code.

TODO:

### DIFF
--- a/tests/fixtures/autouse.py
+++ b/tests/fixtures/autouse.py
@@ -1,0 +1,28 @@
+import pytest
+from composer.utils import reproducibility, dist, get_device
+import torch
+import gc
+
+
+@pytest.fixture(autouse=True)
+def initialize_dist(request: pytest.FixtureRequest):
+    """Initialize the default PyTorch distributed process group for tests."""
+    # should we just always initialize dist like in train.py?
+    _default = pytest.mark.world_size(1).mark
+    world_size = request.node.get_closest_marker('world_size', _default).args[0]
+    gpu = request.node.get_closest_marker('gpu')
+    if world_size > 1:
+     dist.initialize_dist(get_device('gpu' if gpu is not None else 'cpu'))
+
+@pytest.fixture(autouse=True)
+def clear_cuda_cache(request: pytest.FixtureRequest):
+    """Clear memory between GPU tests."""
+    marker = request.node.get_closest_marker('gpu')
+    if marker is not None and torch.cuda.is_available():
+        torch.cuda.empty_cache()
+        gc.collect()  # Only gc on GPU tests as it 2x slows down CPU tests
+
+@pytest.fixture(autouse=True)
+def seed_all():
+    """Sets the seed for reproducibility."""
+    reproducibility.seed_all(17)
--- a/tests/fixtures/models.py
+++ b/tests/fixtures/models.py
@@ -12,6 +12,7 @@ from llmfoundry.utils.builders import build_tokenizer
 def _build_model(config: DictConfig, tokenizer: PreTrainedTokenizerBase, device: Device):
     model = COMPOSER_MODEL_REGISTRY[config.name](config, tokenizer)
     model = device.module_to_device(model)
+    return model
 
 @fixture
 def tokenizer():
--- a/tests/fixtures/models.py
+++ b/tests/fixtures/models.py
@@ -1,0 +1,53 @@
+from typing import Any, Dict, Optional, Union
+from omegaconf import DictConfig
+from pytest import fixture
+from composer.devices import Device
+
+from llmfoundry.models.model_registry import COMPOSER_MODEL_REGISTRY
+from transformers import PreTrainedTokenizerBase
+
+from llmfoundry.utils.builders import build_tokenizer
+
+
+def _build_model(config: DictConfig, tokenizer: PreTrainedTokenizerBase, device: Device):
+    model = COMPOSER_MODEL_REGISTRY[config.name](config, tokenizer)
+    model = device.module_to_device(model)
+
+@fixture
+def tokenizer():
+    return build_tokenizer('EleutherAI/gpt-neox-20b', {})
+
+@fixture
+def build_mpt(tokenizer: PreTrainedTokenizerBase):
+    def build(device: Optional[Union[str, Device]], **kwargs: Dict[str, Any]):
+        config = DictConfig({
+            'name': 'mpt_causal_lm',
+            'd_model': 128,
+            'n_heads': 4,
+            'n_layers': 2,
+            'expansion_ratio': 2,
+        })
+        config.update(kwargs)
+        model = _build_model(config, tokenizer, device)
+        return model
+    return build
+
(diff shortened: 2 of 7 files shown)


tests/fixtures/autouse.py
```diff
@@ -0,0 +1,34 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import gc
+
+import pytest
+import torch
+from composer.utils import dist, get_device, reproducibility
+
+
+@pytest.fixture(autouse=True)
```
irenedea, line 11: nit: either use naming pytest.fixture or just fixture across all fixture files

irenedea:
hmm having trouble reproducing the CI code quality error locally, and the types it's complaining about seem to match correctly 🤷‍♀️ will take a closer look

tests/fixtures/data.py
```diff
@@ -0,0 +1,54 @@
...
+from llmfoundry.data.finetuning.dataloader import build_finetuning_dataloader
+from tests.data_utils import make_tiny_ft_dataset
+
+
+@fixture
+def tiny_ft_dataset_path(tmp_path: Path, dataset_size: int = 4) -> str:
+    """Creates a tiny dataset and returns the path."""
+    tiny_dataset_path = os.path.join(tmp_path, 'test-ift-data-small')
+    os.mkdir(tiny_dataset_path)
+    tiny_dataset_file = os.path.join(tiny_dataset_path, 'train.jsonl')
```
aspfohl, line 21: Sort of a nit, but paths shouldn't require os at all
```suggestion
    tiny_dataset_path = tmp_path / 'test-ift-data-small'
    tiny_dataset_path.mkdir()
    tiny_dataset_file = tiny_dataset_path / 'train.jsonl'
```

tests/fixtures/autouse.py
```diff
@@ -0,0 +1,34 @@
...
+    """Clear memory between GPU tests."""
+    marker = request.node.get_closest_marker('gpu')
+    if marker is not None and torch.cuda.is_available():
+        torch.cuda.empty_cache()
+        gc.collect()  # Only gc on GPU tests as it 2x slows down CPU tests
+
+
+@pytest.fixture(autouse=True)
+def seed_all():
+    """Sets the seed for reproducibility."""
```
aspfohl, line 33: Nice, could all of the other seed calls get removed since this is autoused?

aspfohl (approved):
🌮

tests/fixtures/autouse.py
irenedea: Yeah!! I'll do a pass to remove them

irenedea: [8848934](https://github.com/mosaicml/llm-foundry/pull/673/commits/8848934a3fbc91b581074af79ab45ff9d3f45859)

tests/fixtures/models.py
```diff
@@ -0,0 +1,73 @@
...
+    return model
+
+
+@fixture
+def mpt_tokenizer():
+    return build_tokenizer('EleutherAI/gpt-neox-20b', {})
+
+
+@fixture
+def build_mpt(
```
dakinggg, line 30:
```suggestion
def build_tiny_mpt(
```

tests/fixtures/models.py
```diff
@@ -0,0 +1,73 @@
...
+from pytest import fixture
+from transformers import PreTrainedTokenizerBase
+
+from llmfoundry.models.hf.hf_causal_lm import ComposerHFCausalLM
+from llmfoundry.models.model_registry import COMPOSER_MODEL_REGISTRY
+from llmfoundry.models.mpt.modeling_mpt import ComposerMPTCausalLM
+from llmfoundry.utils.builders import build_tokenizer
+
+
+def _build_model(config: DictConfig, tokenizer: PreTrainedTokenizerBase,
```
dakinggg, line 17: I'm not sure the move to device should be contained in this fixture. Code that gets the model from the constructor should not actually assume that the model is moved to the right device. For testing, we often do move it, but I'm wondering if doing it automatically might hide some tricky bugs

tests/fixtures/models.py
```diff
@@ -0,0 +1,73 @@
...
+        config.update(kwargs)
+        model = _build_model(config, mpt_tokenizer, device)
+        assert isinstance(model, ComposerMPTCausalLM)
+        return model
+
+    return build
+
+
+@fixture
+def build_hf_mpt(
```
dakinggg, line 51:
```suggestion
def build_tiny_hf_mpt(
```

tests/fixtures/data.py
```diff
@@ -0,0 +1,56 @@
...
+from pytest import fixture
+from torch.utils.data import DataLoader
+from transformers import PreTrainedTokenizerBase
+
+from llmfoundry.data.finetuning.dataloader import build_finetuning_dataloader
+from tests.data_utils import make_tiny_ft_dataset
+
+
+@fixture
+def tiny_ft_dataset_path(tmp_path: Path, dataset_size: int = 4) -> Path:
```
dakinggg, line 16: can we make this fixture clean up after itself?

tests/test_hf_mpt_gen.py
```diff
@@ -1,68 +1,34 @@
...
-                        Any] = om.to_container(test_cfg.tokenizer,
-                                               resolve=True)  # type: ignore
-    tokenizer_name = tokenizer_cfg['name']
-    tokenizer_kwargs = tokenizer_cfg.get('kwargs', {})
-    tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
-
-    # build model
-    model = COMPOSER_MODEL_REGISTRY[test_cfg.model.name](test_cfg.model,
-                                                         tokenizer)
-    test_cfg.n_params = sum(p.numel() for p in model.parameters())
```
dakinggg, line 65: any idea why this line was here?

dakinggg (approved):
Just nits, definitely a step forward. Thanks for doing this!

tests/test_hf_mpt_gen.py
irenedea: yeah, not sure. I just removed because it was unused.

tests/fixtures/data.py
irenedea: do you mean remove the dataset after the test is done? If so, I think the tmp_path fixture we are using will do that automatically

dakinggg: oh yeah, good point. Can you go remove the place where I explicitly `rmtree` the dataset?

irenedea: yup, can do that! also we should add a check to only create the dataset on rank 0, right?

irenedea: added logic below to only create the dataset on rank 0 if worldsize != 1

tests/fixtures/models.py
irenedea: ah gotcha, let's stay on the safe side and move it out then.

PR #676, opened 2023-10-15
Convert to DataSpec and add token counts that include padding

dakinggg:
This PR changes DataLoader to DataSpec in order to provide a non default token counting function that takes into account padding.

Two runs that are the same except that one of them counts tokens with padding and one counts them without.
Screenshot 2023-10-14 at 8 20 59 PM
Screenshot 2023-10-14 at 8 21 17 PM

- [x] Fix all the places we assume that dataloaders are returned not dataspecs (mostly tests)
- [x] support denoising dataloader too (maybe)

### DIFF
--- a/llmfoundry/data/text_data.py
+++ b/llmfoundry/data/text_data.py
@@ -307,6 +307,15 @@ def build_text_dataloader(
 
 
 def get_tokens_per_batch_func(pad_token_id: int) -> Callable[[Batch], int]:
+    """Returns a callable that counts the number of tokens in a batch.
+
+    Args:
+        pad_token_id (int): The id of the padding token.
+
+    Returns:
+        Callable[[Batch], int]: A callable that counts the number of tokens in a batch.
+    """
+
     def get_num_samples_in_batch(batch: Batch) -> int:
         if not isinstance(batch, Mapping) or 'input_ids' not in batch:
             raise ValueError(
--- a/llmfoundry/data/text_data.py
+++ b/llmfoundry/data/text_data.py
@@ -378,7 +378,7 @@ if __name__ == '__main__':
     tokenizer_kwargs = {'model_max_length': args.max_seq_len}
     tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
 
-    loader = build_text_dataloader(cfg, tokenizer, device_batch_size)
+    loader = build_text_dataloader(cfg, tokenizer, device_batch_size).dataloader
     assert isinstance(loader.dataset, StreamingTextDataset)
     tokenizer = loader.dataset.tokenizer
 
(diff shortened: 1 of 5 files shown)


llmfoundry/data/text_data.py
```diff
@@ -293,6 +295,45 @@ def build_text_dataloader(
...
+        Callable[[Batch], int]: A callable that counts the number of tokens in a batch.
+    """
+
+    def get_num_samples_in_batch(batch: Batch) -> int:
+        if not isinstance(batch, Mapping) or 'input_ids' not in batch:
+            raise ValueError(
+                'get_tokens_per_batch_func() requires a batch with an input_ids key'
+            )
+        
+         # count number of non padding tokens in batch
```
mvpatel2000, line 325:
```suggestion
         # Count number of non padding tokens in batch
```
nit

llmfoundry/data/text_data.py
```diff
@@ -293,6 +295,45 @@ def build_text_dataloader(
...
+        input_ids_tokens = int(torch.sum(batch['input_ids'] != pad_token_id).item())
+
+        # for encoder decoder models only
```
mvpatel2000, line 328:
```suggestion
        # For encoder decoder models only
```
nit

mvpatel2000 (commented):
LGTM but want research approval

llmfoundry/data/text_data.py
```diff
@@ -293,6 +295,47 @@ def build_text_dataloader(
         timeout=cfg.get('timeout', 0),
     )
 
+    # If we pretokenized, we may not have padding, in which the
+    # tokenizer may not have a pad_token_id. In this case, we can
+    # just use the default token counting function.
+    token_counting_func = None
+    if tokenizer.pad_token_id is not None:
```
alextrott16, line 302: This may lead to inaccurate counts if the tokenizer config during pretokenization doesn't match the tokenizer at this point.

llmfoundry/data/text_data.py
```diff
@@ -293,6 +295,47 @@ def build_text_dataloader(
...
+        token_counting_func = get_tokens_per_batch_func(
+            pad_token_id=tokenizer.pad_token_id)
+
+    return DataSpec(dataloader=dl, get_num_tokens_in_batch=token_counting_func)
+
+
+def get_tokens_per_batch_func(pad_token_id: int) -> Callable[[Batch], int]:
```
alextrott16, line 309: As a suggestion, you may want to add a `decoder_only` flag as an input.

I think you can accurately set that flag in all the places where `get_tokens_per_batch_func` gets called. And the resulting function will not have to infer enc-dec from the presence of `decoder_input_ids`. Instead the flag will just determine whether the function should expect `decoder_input_ids` in addition to `input_ids`.

alextrott16 (approved):
LGTM. Left a couple comments for you to consider, but I think it's fine either way.

llmfoundry/data/text_data.py
dakinggg: That is true. I believe we don't support training in this fashion (that is, if you put "padding" tokens into your pretokenized data, they will simply be trained on like normal tokens. But I will leave a note behind here with a bit more explanation.

llmfoundry/data/text_data.py
dakinggg: ah, yes, I like this idea, will do.

PR #679, opened 2023-10-17
Add test coverage for lion and lion8b checkpoint interop

dblalock:
Makes sure that we can switch between `DecoupledLionW` and `DecoupledLionW_8bit` seamlessly. Just touches `test_lion8b.py`.


mvpatel2000 (approved):
It's fine to ignore the dealloc thing.

Thanks for adding this!

PR #680, opened 2023-10-17
Small changes to HF repo update script

dakinggg:
Just adds a model list to make running it against all our models slightly easier. Also adds an extra print statement in the logs.

### DIFF
--- a/scripts/misc/update_hub_code.py
+++ b/scripts/misc/update_hub_code.py
@@ -29,7 +29,7 @@ _ALL_MODELS = [
 
 
 def main(hf_repos_for_upload: List[str]):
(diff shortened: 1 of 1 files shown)


scripts/misc/update_hub_code.py
```diff
@@ -62,6 +78,7 @@ def main(hf_repos_for_upload: List[str]):
             )
 
             print(f'PR opened: {result}')
+            print()
```
mvpatel2000, line 81:
```suggestion
            print(f'PR opened: {result}\n')
```

mvpatel2000 (approved):

PR #683, opened 2023-10-19
Add support for auto packing ratio

irenedea:
Adds 'auto' packing ratio for finetuning.

If 'auto' is specified, packing will be profiled for the dataloader configuration across a maximum of 20 packing ratios until the highest packing ratio with 0 waste is found.

If there are multiple ranks, we take the minimum 'auto' packing ratio across all ranks.

## Testing
- Manual tests
     -  1 epoch on mosaicml/dolly_hhrlhf, auto packing had 1.3% waste overall
         - finetune-auto-pack-test-CHlwSQ https://wandb.ai/mosaic-ml/irene-test/runs/7ccl7k7a 
             - time/token: 9390338, time/sample: 4896, time/batch: 51, time/epoch: 1
         - finetune-auto-pack-test-baseline2-zptNpN https://wandb.ai/mosaic-ml/irene-test/runs/b0xqxmlj
              - time/token: 9518016, time/sample: 59288, time/batch: 618, time/epoch: 1
     - No packing run are the same before and after PR 
              - before: finetune-auto-pack-test-baseline2-zptNpN , after: finetune-ratio-no-pack-branch-test-8N5I1y
     - Packing ratio of 3.0 is same before and after PR
              - before: finetune-ratio-3-pack-test-kW8gp6 , after: finetune-ratio-3-pack-with-auto-test-YVP00E 
- Unit tests
     - Tests packing on small tensors
     - Tests packing on small tensors with leftovers and waste calculation
     - Tests auto packing ratio selection for single and multi ranks
     - Tests auto packing with dataloader 
     - Adds 'auto' packing_ratio parameterization to existing dataloader tests


llmfoundry/data/packing.py
```diff
@@ -260,124 +258,79 @@ def pad_tensor(tensor: torch.Tensor, pad_value: int):
...
+    Args:
+        dataloader_cfg (DictConfig): The dataloader configuration for profiling.
+        tokenizer (PreTrainedTokenizerBase): The tokenizer for profiling.
+        device_batch_size (int): The size of the batches (number of examples) per device.
+
+    Returns:
+        A packing ratio that minimizes padding while maintaining zero waste.
+    """
+    min_ratio = 1
+    max_ratio = dataloader_cfg.dataset.max_seq_len / 100
```
irenedea, line 275: This assumes ~100 tokens for the "smallest" examples. Would like to get some feedback on whether or not this is reasonable.

future optimization: profile for this as well?

llmfoundry/data/dataloader.py
```diff
@@ -0,0 +1,46 @@
...
+from transformers import PreTrainedTokenizerBase
+
+from llmfoundry.data.text_data import build_text_dataloader
+
+from llmfoundry.data.denoising import build_text_denoising_dataloader
+
+from llmfoundry.data.finetuning.dataloader import build_finetuning_dataloader
+
+
+def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
```
irenedea, line 17: refactored here to share this code because we were essentially copying the same code in packing.py

llmfoundry/data/packing.py
```diff
@@ -260,124 +259,89 @@ def pad_tensor(tensor: torch.Tensor, pad_value: int):
...
-        cfg = om.to_container(cfg.parameters)
-        cfg = om.create(cfg)
-    device_batch_size = cfg.global_train_batch_size // args.num_devices
+    # Turn off packing for the dataloader (we want raw, pre-packed examples)
+    dataloader_cfg = copy.deepcopy(dataloader_cfg)
+    dataloader_cfg.dataset.packing_ratio = None
+    dataloader_cfg.dataset.max_leftovers_to_keep = None
+    dataloader_cfg.drop_last = False
+    dataloader_cfg.num_workers = 0
+    dataloader_cfg.prefetch_factor = None
```
irenedea, line 327: Set this to 0 workers because [a single worker is responsible for a single batch](https://discuss.pytorch.org/t/guidelines-for-assigning-num-workers-to-dataloader/813). Because we load one big batch only, there's no need for multiple workers. (Additionally, having more than one worker here caused the cpu tests to fail in CI, but I couldn't repro in interactive.)

Alternatively instead of just loading a single big batch, we could use the dataloader with multiple workers to get multiple smaller batches.

llmfoundry/data/packing.py
alextrott16: Yeah, I agree that this is probably a pretty good heuristic for getting an upper bound on the packing ratio.

If there's an issue, it's probably that with this heuristic, `num_packing_ratios = 20` might not have enough granularity. It might be worth coming up with a less brute force approach that better guarantees that you'll get fine-grained search in the most relevant range of packing ratios.

Off the top of my head, one idea is to start from 1, and double packing ratio until you hit one that is too high (has waste), then next look at half-way between the highest safe packing ratio you've checked and the lowest >0 waste packing ratio you've checked. Do that until the half-way point is rounded into a `raw_batch_size` that you've already checked. Then pick the best packing ratio from the set you've covered.

scripts/train/finetune_example/mpt-7b-arc-easy--gpu.yaml
```diff
@@ -41,7 +41,7 @@ train_loader:
     shuffle: true
     max_seq_len: ${max_seq_len}
     decoder_only_format: true
-    # # Use `python llmfoundry/data/packing.py --yaml-path /path/to/this/yaml/ ...`
+    # # Use `python llmfoundry/scripts/misc/profile_packing.py --yaml-path /path/to/this/yaml/ ...`
```
alextrott16, line 44: Should we instead update the YAMLs to recommend the 'auto' option?

tests/test_dataloader.py
```diff
@@ -248,10 +249,10 @@ def test_denoising_dataloader(decoder_only_format: bool, pretokenize: bool,
 
 @pytest.mark.parametrize('decoder_only_format', [True, False])
 @pytest.mark.parametrize('allow_pad_trimming', [True, False])
-@pytest.mark.parametrize('packing_ratio', [10.0, None])
+@pytest.mark.parametrize('packing_ratio', [10.0, None, 'auto'])
 def test_finetuning_dataloader(decoder_only_format: bool,
                                allow_pad_trimming: bool,
-                               packing_ratio: Optional[float]):
+                               packing_ratio: Optional[Any]):
```
alextrott16, line 255: Nit (sorry): Should the signature be `Optional[Union[int, float, str]]`?

alextrott16 (commented):
LGTM. Thanks for adding this!!
Will hold off on approval so this can get another set of eyes!

Also left a couple comments (nothing major) along with a suggestion for an alternative way to search for the optimal packing ratio. I think re-using my brute force approach might not be the best way to go. If you agree, that will require more of an overhaul of this code, unfortunately. But potentially better to do that than simply inherit my hasty decision :)

llmfoundry/data/packing.py
```diff
@@ -58,7 +57,7 @@ def efficiency(self) -> float:
 
     def __call__(
             self,
-            examples: List[Dict[str, torch.Tensor]]) -> Dict[str, torch.Tensor]:
+            examples: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
```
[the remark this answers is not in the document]
irenedea: ah yes that's right! thanks for catching that

tests/test_dataloader.py
irenedea: good point! i think it would be `Optional[Union[float, str]]`?

llmfoundry/data/packing.py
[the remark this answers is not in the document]
irenedea: hmmm I think the type isn't necessarily consistent because during profiling the type of examples is `Dict[str, torch.Tensor]` while self.base_collator is `lambda x: x`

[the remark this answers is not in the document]
irenedea: maybe we can refactor this out into two functions. I think that'll be cleaner also:
```
def __call__(self, examples: List[Dict[str, torch.Tensor]]):
    batch = self.base_collator(examples)
    return self.pack(batch)

def pack(self, examples: Dict[str, torch.Tensor]):
    # assertions, packing logic
```
and then in profiling we can just called BinPackCollator.pack

llmfoundry/data/packing.py
```diff
@@ -393,11 +357,11 @@ def split_big_batch(raw_batch_size: int) -> List:
                 batches[idx].update({key: split})
         return batches
 
-    def profile_packing(raw_batch_size: int) -> Tuple[float, float]:
-        packer = BinPackWrapper(
+    def profile(raw_batch_size: int) -> Tuple[float, float]:
+        packer = BinPackCollator(
             collator=lambda x: x,
```
irenedea, line 362: @alextrott16 Do you think we should be passing in dataloader.collate_fn here instead of `lambda x: x` ?

llmfoundry/data/packing.py
irenedea: Thanks for the insight! Getting better granularity sounds like a good next step. 
Another idea: If we feel good about the the max heuristic we could also do a binary search that starts between the min and max.

Separately, I think it would be appropriate to do this as a follow up PR since this PR already has a lot of other refactors and new tests. What do you think? @alextrott16 @dakinggg

llmfoundry/data/packing.py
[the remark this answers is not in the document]
irenedea: Updated to do this refactor. (It was needed to appease the pyright gods anyway)

llmfoundry/data/packing.py
alextrott16: I don't think we need to, unless you prefer to refactor the surrounding code. The dataloader that is set up in this script already applies the seq2seq collator function, so, in the context of this script, having `BinPackCollator` wrap an identify function should be correct in the context of the current logic.

llmfoundry/data/packing.py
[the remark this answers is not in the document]
alextrott16: Sorry for not chiming in sooner. I'm on board with any refactor that makes sense to you and whatever aligns this code with the auto packing script such that pyright is appeased.

llmfoundry/data/packing.py
alextrott16: Fine with me if it's a separate PR, but I would advocate for it being a fast follow-up since the 'auto' implementation will potentially change the packing ratio that it comes up with.

llmfoundry/data/packing.py
```diff
@@ -260,124 +261,95 @@ def pad_tensor(tensor: torch.Tensor, pad_value: int):
...
+
+    # Obtain the maximum packing_ratio/minimum padding that has no waste.
+    packing_ratio = 1
+    for packing_ratio_candidate, _, waste in profiling_results:
+        if waste > 0:
+            break
+        packing_ratio = packing_ratio_candidate
+
+    # Select the minimum packing ratio across all ranks.
+    if torch.cuda.is_available() and dist.is_available(
```
dakinggg, line 293: I think the way to do this generically is

```
if dist.is_available() and dist.is_initialized():
    device = get_device(None)
    packing_ratio_tensor = device.tensor_to_device(torch.tensor(packing_ratio))
    dist.all_reduce(packing_ratio_tensor, reduce_operation='MIN')
    packing_ratio = packing_ratio_tensor.item()
```

llmfoundry/data/packing.py
```diff
@@ -260,124 +261,95 @@ def pad_tensor(tensor: torch.Tensor, pad_value: int):
...
+    min_ratio = 1
+    max_ratio = dataloader_cfg.dataset.max_seq_len / 100
+    num_packing_ratios = 20
+    profiling_results = profile_packing(dataloader_cfg, tokenizer, min_ratio,
+                                        max_ratio, num_packing_ratios,
+                                        device_batch_size)
+
+    # Obtain the maximum packing_ratio/minimum padding that has no waste.
+    packing_ratio = 1
+    for packing_ratio_candidate, _, waste in profiling_results:
```
dakinggg, line 287: are the profiling results already sorted?

llmfoundry/data/packing.py
```diff
@@ -260,124 +261,95 @@ def pad_tensor(tensor: torch.Tensor, pad_value: int):
...
         raw_batch_size = int(packing_ratio * device_batch_size)
         if raw_batch_size not in raw_batch_sizes:
             packing_ratios.append(packing_ratio)
             raw_batch_sizes.append(raw_batch_size)
 
-    # Fetch a bunch of raw examples once, which we'll re-use
-    if 'train_loader' not in cfg:
-        raise ValueError('config must define train_loader')
-    dataloader_cfg = cfg.train_loader
+    n_profile_examples = max(raw_batch_sizes) * 100
```
dakinggg, line 348: This controls how many samples we are going to use for profiling, yeah? So this says we will profile `device_batch_size * max_ratio * 100 samples`. I think you may have already run this, but could you comment here for long this takes to run, and how changing this to 200, 300, 400 changes how long it takes to run? Would help to make an informed decision about this number.

dakinggg (commented):
Looking good! 

Two more manual tests I would like to see before merging:
(1) a short training run without packing, before and after this PR. These should be identical.
(2) a short training run with a set packing ratio, before and after this PR. These should be identical.

llmfoundry/data/dataloader.py
irenedea: circuilar dependencies, iirc it was llmfoundry/data is dependent on llmfoundry/utils and vice versa (if I added this in builders.py)

llmfoundry/data/packing.py
irenedea: yes! i'll include a comment about this

llmfoundry/data/packing.py
irenedea: when I use dist without cuda I get errors--is it supposed to work with other backends? (I think yes, but I haven't gotten it to work...)

scripts/misc/profile_packing.py
```diff
@@ -0,0 +1,103 @@
...
+        cfg = om.to_container(cfg.parameters)
+        cfg = om.create(cfg)
+    device_batch_size = cfg.global_train_batch_size // args.num_devices
+
+    # Fetch a bunch of raw examples once, which we'll re-use
+    if 'train_loader' not in cfg:
+        raise ValueError('config must define train_loader')
+    dataloader_cfg = cfg.train_loader
+
+    max_leftovers_to_keep = dataloader_cfg.dataset.get('max_leftovers_to_keep',
```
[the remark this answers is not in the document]
irenedea: the max number of "leftover" examples that didn't get packed. They get kept and the algorithm can try again to fit them. Right now it defaults to 10 * device batch size . Should we default to keeping all of them for auto to try to reduce waste as much as possible?  cc @alextrott16

tests/test_dataloader.py
irenedea: It can be, but I think for typing purposes float also accepts int (even though it's not a proper subclass)

irenedea: will use `Literal['auto']`, that's better than just `str`

llmfoundry/data/packing.py
dakinggg: yes, it should work with gloo backend as well. but this is not a big deal to investigate. can you make a jira to fix this and put it in the llm foundry backlog epic?

llmfoundry/data/packing.py
irenedea: I think we should keep it at 100 because it scales linearly and with 100, we're looking at about several minutes to profile. (Based on profiling on dolly)

irenedea: If anything I think we could optimize by using the dataloader with more workers etc and not get a single large batch but mutiple smaller batches. I can create a jira to investigate optimizations here

llmfoundry/data/packing.py
irenedea: done

llmfoundry/data/packing.py
irenedea: actually nvm, i'm not sure what i was doing before, but i tested it out again with gloo just now and it works fine. I suspect I was running nccl x cpu previously. 

```
from composer.utils import dist, get_device
import torch

torch.distributed.init_process_group('gloo')

if dist.is_available() and dist.is_initialized(): 
    device = get_device(None) 
        packing_ratio_tensor = device.tensor_to_device(torch.tensor(5.0)) 
    else:
        packing_ratio_tensor = device.tensor_to_device(torch.tensor(4.0)) 

    dist.all_reduce(packing_ratio_tensor, reduce_operation='MIN') 
    packing_ratio = packing_ratio_tensor.item()
    print(packing_ratio)
```

irenedea: will update to generic

llmfoundry/data/packing.py
dakinggg: hmm ok. sounds fine for now. please make the jira for further optimization and put it in foundry backlog epic

scripts/misc/profile_packing.py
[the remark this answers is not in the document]
dakinggg: yeah, lets keep em all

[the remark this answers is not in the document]
dakinggg: but also, my question was more that the variable doesn't seem to be used anywhere?

llmfoundry/data/packing.py
```diff
@@ -260,14 +261,169 @@ def pad_tensor(tensor: torch.Tensor, pad_value: int):
...
+                                     endpoint=True):
+        packing_ratio = np.round(10 * packing_ratio) / 10
+        raw_batch_size = int(packing_ratio * device_batch_size)
+        if raw_batch_size not in raw_batch_sizes:
+            packing_ratios.append(packing_ratio)
+            raw_batch_sizes.append(raw_batch_size)
+
+    n_profile_examples = max(raw_batch_sizes) * 100
+
+    train_dataspec = build_dataloader(dataloader_cfg, tokenizer,
```
dakinggg, line 370: I'm contemplating some determinism stuff here. What happens if we change the number of nodes...? I think you could end up with a different packing ratio, right? Should we turn off the distributed sampler for the profiling so that it is deterministic? This would negate any profiling benefit of having many processes, but maybe we should be on the safe side for now?

If we do this, please redo the profiling run that tested waste percent, in case this is a really bad decision :)

dakinggg (commented):
LGTM except for one determinism related comment

llmfoundry/data/packing.py
irenedea: hmmm yeah, also:
if you change the number of nodes without adjusting global batch size, then device batch size will decrease. This will also affect the profiling results...

irenedea: Is this accurate:
- changing the device batch size can result in a different packing ratio
- changing global batch size can result in resuming at the incorrect point in the dataset

irenedea: resolved offline, we don't need to support elastic resumption for now

llmfoundry/data/packing.py
```diff
@@ -260,14 +261,169 @@ def pad_tensor(tensor: torch.Tensor, pad_value: int):
...
+
+    Returns:
+        A packing ratio that minimizes padding while maintaining zero waste.
+    """
+    from composer.utils import dist, get_device, reproducibility
+
+    # Stash the rng state to restore later.
+    rng_state = reproducibility.get_rng_state()
+    # Set the seed so that auto packing is deterministic.
+    reproducibility.seed_all(0)
```
dakinggg, line 293: should we set this to a different value depending on rank so that each rank profiles different data?

dakinggg (approved):
LGTM, with one more question relating to the randomness

llmfoundry/data/packing.py
irenedea: I think the distributed sampler/streaming dataset that gets created with the dataloader should handle this case. Is there something I am missing?

scripts/misc/profile_packing.py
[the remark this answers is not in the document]
irenedea: oh yup, good catch. It's not necessary to pop it here

llmfoundry/data/packing.py
dakinggg: oh yeah, i confused myself. all good.

PR #687, opened 2023-10-20
Support inputs_embeds

samhavens:
This allows users to pass in embeddings directly instead of looking them up based on `input_ids`. This is useful for PEFT (prompt/prefix-tuning) and multimodal models.


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -361,7 +356,16 @@ def forward(
             S <= self.config.max_seq_len
         ), f'Cannot forward input with seq_len={S}, this model only supports seq_len<={self.config.max_seq_len}'
 
-        tok_emb = self.wte(input_ids)
+        if input_ids is not None and inputs_embeds is not None:
+            raise ValueError(
+                'You cannot specify both input_ids and inputs_embeds.')
+        elif input_ids is not None:
```
vchiley, line 362: you'll want to make `input_ids` optional and give it None as the default.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -554,22 +558,34 @@ def forward(
...
+                input_ids=input_ids,
+                past_key_values=past_key_values,
+                attention_mask=attention_mask,
+                prefix_mask=prefix_mask,
+                sequence_id=sequence_id,
+                return_dict=return_dict,
+                output_attentions=output_attentions,
+                output_hidden_states=output_hidden_states,
+                use_cache=use_cache,
+            )
```
vchiley, line 588: If you make input_ids default to None, then you can probably just pass fwd both and you wont need to make this cased.

vchiley (commented):
Should `input_ids` default to `None` with this change?
```
input_ids: Optional[torch.LongTensor] = None,
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -554,13 +560,12 @@ def forward(
...
                      if use_cache is not None else self.config.use_cache)
 
-        # if input_embeds is not none, raise a not implemented error
-        if inputs_embeds is not None:
-            raise NotImplementedError(
-                'inputs_embeds has to be None (for hf/peft support).')
-        # decoder outputs consists of (dec_features, layer_state, dec_hidden, dec_attn)
+        if input_ids is not None and inputs_embeds is not None:
+            raise ValueError(
+                'You cannot specify both input_ids and inputs_embeds.')
```
vchiley, line 565: isn't this already checked [here](https://github.com/mosaicml/llm-foundry/pull/687/files#diff-d99c63c3cfe678d1d5133c1a5f7cd670f20887c97add390e5cc021bc0a0b9127R365)

samhavens, line 565: Just realized this might be wrong. I need to look at how HF does this, but I think that during autoregressive generation we might get both of these and that's ok..?

samhavens:
> @samhavens here is what HF expects wrt these two args: https://github.com/huggingface/transformers/blob/9333bf0769561c048700377c2e0813221ab9d2c9/src/transformers/models/llama/modeling_llama.py#L955-L963

@dakinggg That validation happens in the case model, but it the CausalLM model they can both be there https://github.com/huggingface/transformers/blob/9333bf0769561c048700377c2e0813221ab9d2c9/src/transformers/models/llama/modeling_llama.py#L1140-L1150

I think this is because in `prepare_inputs_for_generation` they use them on the first decoding step then ignore them https://github.com/huggingface/transformers/blob/9333bf0769561c048700377c2e0813221ab9d2c9/src/transformers/models/llama/modeling_llama.py#L1209-L1213

dakinggg (approved):
Approving cause LGTM, but please add a test for the edge cases:
(1) both input_ids and input_embeds are None
(2a) both input_ids and input_embeds are specified without kv cache
(2b) both input_ids and input_embeds are specified with kv cache

samhavens:
> Approving cause LGTM, but please add a test for the edge cases: (1) both input_ids and input_embeds are None (2a) both input_ids and input_embeds are specified without kv cache (2b) both input_ids and input_embeds are specified with kv cache

@dakinggg do you mean new tests or add these edge cases to the 2 tests that have input embeds

dakinggg:
@samhavens add these edge cases to existing tests is fine

PR #689, opened 2023-10-24
Add ChatML special tokens to GPT-4 tokenizer

samhavens:
It might be possible to do this in general for arbitrary additional special tokens... but `tiktoken` does NOT make it easy. So I add this special case. LMK if I need to approach this another way.


bcui19 (commented):
LGTM, but will wait for Daniel to approve

samhavens:
closing in lieu of #695 which fixes the underlying issue preventing one from adding special tokens at instantiation time

PR #690, opened 2023-10-24
Add tensor parallelism for attention QKVO

linden-li:
This PR implements tensor parallelism using PyTorch's new DTensor library. This can be either used standalone or in 2D parallel fashion along with other parallelism strategies like FSDP. It partitions the QKV projections and output projections Megatron style within a GPU node, producing one synchronization step at the end.

The implementation currently supports multihead and grouped query attention. I was not able to find a good way to parallelize the attention bias with AliBi in this setting - would like some advice with this.

To turn on tensor parallelism, the attention config now has two new variables:

1. `tensor_parallel_qkvo`: a boolean flag that configures TP
2. `tp_world_size`: the nunmber of GPUs to tensor parallelize over


mvpatel2000:
@linden-li  can you please add tests

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -202,6 +254,71 @@ def __init__(self, config: MPTConfig):
...
+                    tp_mesh_dim=1,
+                )
+
+                # Call to parallelize_module moves params to gpu if they are cpu params.
+                # Move them back to cpu so that FSDP wrapping sees all params on cpu.
+                # Othewise FSDP wrapping fails as it sees some params on cpu and others on gpu.
+                assert config.init_device == 'cpu'
+                if config.init_device == 'cpu':
+                    block = block.to('cpu')
+                new_blocks.append(block)
```
mvpatel2000, line 313: Why assert + if statement?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -202,6 +254,71 @@ def __init__(self, config: MPTConfig):
...
+            self.blocks = new_blocks
+            print('Tensor parallelism initialized...')
```
mvpatel2000, line 315: logger?

tests/models/test_model.py
```diff
@@ -1930,3 +1933,99 @@ def test_head_dim_8_triton_mqa_attn(batch_size: int = 2):
         output = model(batch)
 
     assert not torch.isnan(output.logits).any()
+
+@pytest.mark.world_size(2)
```
vchiley, line 1937: Shouldn't this be tested on 4 GPUs minimum?
The TP'd layers need to be fsdp'd in the other dim.

tests/models/test_model.py
```diff
@@ -1930,3 +1933,99 @@ def test_head_dim_8_triton_mqa_attn(batch_size: int = 2):
...
+@pytest.mark.gpu
+def test_tp_qkvo():
```
vchiley, line 1939: This test checks that the model weights are TP'd against taking a model and TP-ing its weights with the same TP-ing tool (which itself is NOT tested).
Can we create an un-TP tool and test un-TP-ing the model against the full model?

Can we add an end to end test?
ie fwd pass a sample through both models and verify that outputs are the same; then bwd to compute gradients and un-TP grads to verify they are the same?

vchiley:
> The implementation currently supports multihead and grouped query attention. I was not able to find a good way to parallelize the attention bias with AliBi in this setting - would like some advice with this.

@sashaDoubov do you still have that implementation we designed for TP-ing ALiBi bias for TP with DeepSpeed?

sashaDoubov:
This needs more thorough testing, but this is what we had for chunking the alibi bias: 
diff:
https://github.com/mosaicml/llm-foundry/compare/main...sashaDoubov:llm-foundry:benchmarking

PR #691, opened 2023-10-24
Long context evals

maxisawesome:
Adds three long context eval tasks that are padded up to 2k, 4k, and 8k token context length.
- HotPotQA - 10+ concatenated "documents" (3-4 sentences from a wikipedia article) with a single question at the end. Sometimes this requires two "documents" (multi-hop QA) to properly answer the question. While 1 or 2 docs are needed to answer the question, the rest are randomly chosen "distractor" docs that are not relevant to the question.
- WikiQA (WikiQA-Altered_Numeric_QA) - Single documents from wikipedia at different context lengths. There are significantly more HTML tags in this dataset. All answers are some sort of number.
- KV_Pairs - Constructed JSONs with Key Value pairs. The question will be a single key and and the expected answer is the corresponding value that was listed in the JSON.

Caveats: 
- Fewshot versions of these tasks should be pregenerated (because right now _every_ example is approximately the full context length, so fewshot = 5 for the 4k context length task would be 20k total tokens)
- generation scripts for these datasets are not included


maxisawesome:
Opening a new pull request that pulls from HF datasets and never has our long context data checked into git. Closing this one because adding large files to your git history can be really annoying as git will pull them then delete them and it'll make pulls take very long, and redoing git history can also be a huge pain.

PR #692, opened 2023-10-25
Fix mlflow model logging bug

dakinggg:
If the save interval for the hf checkpointer is set to 1dur _and_ the max duration is set in epochs, the mlflow model logging would never happen, because the last save event happens on `BATCH_CHECKPOINT`, which is before the epoch is incremented, meaning that elapsed_duration would not be >= 1.0. This adds a special case for this setting to check if we are on the last batch.

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -136,6 +136,21 @@ class HuggingFaceCheckpointer(Callback):
                 mlflow.environment_variables.MLFLOW_HUGGINGFACE_MODEL_MAX_SHARD_SIZE.set(
                     '5GB')
 
+    def _is_last_batch(self, state: State):
+        elapsed_duration = state.get_elapsed_duration()
+        if elapsed_duration is not None and elapsed_duration >= 1.0:
+            return True
+
+        assert state.max_duration is not None  # for pyright
+        # If the save interval is specified as 1dur, and the max duration is in epoch units
+        # we need a special case to identify we are on the last batch and should write the mlflow checkpoint
+        if self.save_interval.unit == TimeUnit.DURATION and self.save_interval.value == 1 and state.max_duration.unit == TimeUnit.EPOCH:
+            assert state.dataloader_len is not None  # for pyright
+            return int(state.timestamp.batch) % math.ceil(
+                state.max_duration.value * state.dataloader_len) == 0
+
+        return False
+
     def _save_checkpoint(self, state: State, logger: Logger):
         del logger  # unused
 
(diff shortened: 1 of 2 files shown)


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -225,7 +233,18 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                # we need a special case to identify we are on the last batch and should write the mlflow checkpoint
+                is_last_batch = False
+                assert state.max_duration is not None  # for pyright
+                if self.save_interval.unit == TimeUnit.DURATION and self.save_interval.value == 1 and state.max_duration.unit == TimeUnit.EPOCH:
+                    assert state.dataloader_len is not None  # for pyright
+                    is_last_batch = int(state.timestamp.batch) % math.ceil(
+                        state.max_duration.value * state.dataloader_len) == 0
+                if self.mlflow_registered_model_name is not None and (
+                    (elapsed_duration is not None and
+                     elapsed_duration >= 1.0) or is_last_batch):
```
irenedea, line 247: nit: can we use a helper function to define is_last_batch? I think it'll be more clear that way. The conditional here can be updated to`if self.mlflow_registered_model_name is not None and _is_last_batch(state, self.save_interval)` 

```
def _is_last_batch(state, save_interval):
     elapsed_duration = state.get_elapsed_duration()
     if elapsed_duration is not None and elapsed_duration >= 1.0:
         return True
     assert state.max_duration is not None
     if self.save_interval.unit == TimeUnit.DURATION and self.save_interval.value == 1 and state.max_duration.unit == TimeUnit.EPOCH:
                    assert state.dataloader_len is not None  # for pyright
                     return int(state.timestamp.batch) % math.ceil(
                        state.max_duration.value * state.dataloader_len) == 0
      return False
```

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -225,7 +233,18 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
-                if self.mlflow_registered_model_name is not None and elapsed_duration is not None and elapsed_duration >= 1.0:
+
+                # If the save interval is specified as 1dur, and the max duration is in epoch units
+                # we need a special case to identify we are on the last batch and should write the mlflow checkpoint
+                is_last_batch = False
+                assert state.max_duration is not None  # for pyright
+                if self.save_interval.unit == TimeUnit.DURATION and self.save_interval.value == 1 and state.max_duration.unit == TimeUnit.EPOCH:
+                    assert state.dataloader_len is not None  # for pyright
+                    is_last_batch = int(state.timestamp.batch) % math.ceil(
+                        state.max_duration.value * state.dataloader_len) == 0
```
irenedea, line 244: do we need to cache dataloader_len? see mihir's comment: https://github.com/mosaicml/composer/pull/2633#discussion_r1369656775

irenedea (approved):
lgtm! one nit and one question about caching dataloader len

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: no, this is fine because it can only get triggered once per batch, and the first time a batch is encountered is always during training.

PR #695, opened 2023-10-26
Pad tiktoken vocab so that additional_special_tokens works

dakinggg:


samhavens:
This is in lieu of #689, which was a work around for this specific issue

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -90,7 +90,11 @@ def is_fast(self) -> bool:
         return False
 
     def get_vocab(self) -> Dict[str, int]:
-        """Returns vocab as a dict."""
+        """Returns vocab as a dict.
+
+        Note: This function does not work properly due to difference in assumptions between tiktoken and Hugging Face tokenizers.
```
samhavens, line 95: I think we should print a warning here in case someone calls this, just in case

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -26,7 +26,7 @@ def __init__(self,
                  eos_token: Optional[str] = '<|endoftext|>',
                  bos_token: Optional[str] = '<|endoftext|>',
                  pad_token: Optional[str] = None,
-                 **kwargs: Dict[str, Any]):
+                 **kwargs: Any):
```
samhavens, line 29: Aren't `**kwargs` always a dict of str to Any??

samhavens (approved):
I lean on the side of wanting to add a warning to `get_vocab` just in case, but I also think it's a very unlikely edge case

llmfoundry/tokenizers/tiktoken.py
dakinggg: You're supposed to type **kwargs withthe type of the values only

samhavens: TIL. Thanks!

llmfoundry/tokenizers/tiktoken.py
dakinggg: @samhavens lmk what you think of the warning i added

samhavens (approved):
As the person who will probably forget about this and be saved by the warning: thank you. LGTM

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -101,6 +112,24 @@ def get_vocab(self) -> Dict[str, int]:
...
+        candidate_extra_id = f'<extra_id_{extra_id_index}>'
+        indices_to_fill_in = {i for i in range(self.vocab_size)} - set(
+            vocab.values())
+
+        # Add enough indices to make get_vocab() the right length
+        for index_to_add in indices_to_fill_in:
+            # Make sure we don't overwrite a token that already exists
+            while candidate_extra_id in vocab:
+                extra_id_index += 1
+                candidate_extra_id = f'<extra_id_{extra_id_index}>'
```
rajammanabrolu, line 128: So there's a big vocab gap in the public tiktoken between a bunch of the special tokens and the ENDOFPROMPT token which is likely where the private special tokens are used
https://github.com/openai/tiktoken/blob/main/tiktoken_ext/openai_public.py#L72 (Pointed out by @samhavens ) - should we add to the end of this by appending or overwrite the intermediate tokens in that gap?

dakinggg: I chose to just go off the end (after ENDOFPROMPT) because it seemed more straightforward and with no risk 🤷‍♂️ open to a different opinion though.

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -101,6 +112,24 @@ def get_vocab(self) -> Dict[str, int]:
             except KeyError:
                 pass
 
+        # As far as I can tell, we don't require get_vocab to completely work,
+        # but when using additional_special_tokens, Hugging Face determines the next
+        # token index to add with len(self.get_vocab()) so we need the _size_ of this dictionary to be correct.
+        extra_id_index = 0
+        candidate_extra_id = f'<extra_id_{extra_id_index}>'
+        indices_to_fill_in = {i for i in range(self.vocab_size)} - set(
```
irenedea, line 120: for my understanding, self.vocab_size > vocab.values() ?

llmfoundry/tokenizers/tiktoken.py
rajammanabrolu: sgtm

llmfoundry/tokenizers/tiktoken.py
dakinggg: tbh i don't know exactly what each is supposed to correspond to (and may be different per tokenizer), but in this case, vocab_size is what tiktoken reports the index of its largest index token to be, and vocab.values() just contains all the tokens we could decode on their own. so it excludes the gap in token indices with tiktoken, and any token indices that cannot be converted to a string on their own.

PR #702, opened 2023-10-27
Async eval callback

aspfohl:
Here's an example of a finetuning run that I've split icl light eval async:
Screenshot 2023-11-09 at 4 12 25 PM

Check out the run yaml by describing the run: `meow-dVKYjP`
And then subsequent eval run yamls: `eval0-meow-W8dG0M`

Wandb: https://wandb.ai/mosaic-ml/llm-foundry-scripts_train/runs/lgrecstd/overview?workspace=user-aspfohl


scripts/train/train.py
```diff
@@ -518,6 +518,9 @@ def main(cfg: DictConfig) -> Trainer:
         for name, callback_cfg in callback_configs.items()
     ] if callback_configs else []
 
+    use_async_eval = any(
+        isinstance(callback, Evaluator.Async) for callback in callbacks)
```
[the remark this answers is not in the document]
aspfohl: yep, typo 👍

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,156 @@
+# Copyright 2023 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+import logging
```
mvpatel2000, line 3:
```suggestion
# SPDX-License-Identifier: Apache-2.0

import logging
```
nit

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,156 @@
...
+MAX_RUN_NAME_LENGTH = 40
+
+# Note: train parameter names. See comments if they are different from eval
+REQUIRED_PARAMS_FOR_EVAL = {
+    'device_eval_batch_size',
+    'icl_tasks',  # only required for eval
+    'max_seq_len',
+    'model',  # models
+    'save_folder',  # required, but used as load_path
+}
```
mvpatel2000, line 27: Is it possible to set which version of foundry you want to run with which may affect this?

Can we add a test to ensure this stays in-sync?

mvpatel2000 (commented):
requires bumping / requiring mcli pin

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,156 @@
...
+            launched. If an integer, it will be assumed to be in :attr:`.TimeUnit.EPOCH`.
+            Otherwise, the unit must be either :attr:`.TimeUnit.EPOCH`, :attr:`.TimeUnit.BATCH`,
+            :attr:`.TimeUnit.TOKEN`, or :attr:`.TimeUnit.SAMPLE`.
+    """
+
+    def __init__(
+        self,
+        interval: Union[str, int, Time],
+        compute: Optional[ComputeConfig] = None,
+    ):
```
aspfohl, line 69: TODO: arbitrary kwargs can be passed into eval run

llmfoundry/callbacks/async_eval_callback.py
aspfohl: For now, could we require the eval and training run require the same foundry version? Otherwise this gets a little hairy, this callback would need to be backwards compatible for all eval and train param changes

dakinggg: yeah i think its fair to require same version

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,207 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+import logging
```
mvpatel2000, line 3:
```suggestion
# SPDX-License-Identifier: Apache-2.0

import logging
```
nit

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,207 @@
...
+from typing import Any, Dict, List, Optional, Union
+
+from composer.core import Callback, Event, State, Time
+from composer.loggers import Logger
+from composer.loggers.mosaicml_logger import (MOSAICML_PLATFORM_ENV_VAR,
+                                              RUN_NAME_ENV_VAR)
+from composer.utils import dist
+from composer.utils.misc import create_interval_scheduler
+from mcli.api.runs import \
+    ComputeConfig  # TODO: available in root in mcli 0.5.27+
```
mvpatel2000, line 14: Should pin be higher because of this...?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,207 @@
...
+    'precision',
+    'python_log_level',
+    'seed',
+}
+
+
+def get_run_name(previous_run_name: str, count: int) -> str:
+    *name_without_uuid_suffix, _ = previous_run_name.split('-')
+    name_suffix = ('-'.join(name_without_uuid_suffix))[:MAX_RUN_NAME_LENGTH]
+    return f'eval{count}-{name_suffix}'
```
mvpatel2000, line 45: should we emit a warning if truncating?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,207 @@
...
+            image=self.current_run.image,
+            compute=self.compute or default_compute,
+            command=command,
+            integrations=cfg.integrations,
+            env_variables=cfg.env_variables,
+            metadata=cfg.metadata,
+            parameters=params,
+        )
+
+        new_run = create_run(c, timeout=60)
```
mvpatel2000, line 203: why timeout arg?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,207 @@
...
+            'cluster': self.current_run.cluster,
+        }
+        params = self.get_eval_parameters(cfg.parameters or {},
+                                          self.current_run.name)
+
+        # TODO: This just runs an eval run, but we also want to attach the
+        # deployment, which would require a hf conversion and parametrizing the
+        # dependent_deployment in the run config
+        command = 'cd llm-foundry/scripts \n composer eval/eval.py $PARAMETERS'
+        c = RunConfig(
```
mvpatel2000, line 192:
```suggestion
        run_config = RunConfig(
```
rename?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,207 @@
...
+                'AsyncEval callback is only supported when running on the MosaicML platform'
+            )
+
+        run_name = os.environ.get(RUN_NAME_ENV_VAR, None)
+        if not run_name:
+            raise RuntimeError(
+                'RUN_NAME environment variable must be set to use the AsyncEval callback'
+            )
+
+        # allows the MapiException to be raised if the run doesn't exist
```
mvpatel2000, line 136:
```suggestion
        # Allows the MapiException to be raised if the run doesn't exist
```

mvpatel2000 (commented):
a couple minor comments

llmfoundry/callbacks/async_eval_callback.py
aspfohl, line 46: Instead of count, thinking this count be the interval?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,207 @@
...
+            f'Initialized AsyncEval callback. Will generate runs at interval {interval}'
+        )
+
+    def run_event(self, event: Event, state: State, logger: Logger) -> None:
+        del logger
+        if all([
+                state.get_elapsed_duration() is not None,
+                self.check_interval(state, event),
+                self.last_launch != state.timestamp.batch,
+                dist.get_global_rank() == 0
```
aspfohl, line 117: TODO: I think I may need to consider multi-node in this condition too? (Eg the first node rank 0 will launch a run)

llmfoundry/callbacks/async_eval_callback.py
aspfohl: Yea will do :+1: Originally added the todo before I had made the mcli change

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,208 @@
...
+            raise Exception(
+                f'Missing the following required parameters for async eval: {looking_for}'
+            )
+
+        # Convert the save_folder to a load_path
+        subset_keys['load_path'] = get_load_path(
+            subset_keys.pop('save_folder'),
+            parameters.get('save_latest_filename', None))
+
+        # Update the loggers to use the training run name
```
aspfohl, line 166: Should it be logged as separate runs or the same run? Debating, it could be confusing because the eval runs could sit in pending

llmfoundry/callbacks/async_eval_callback.py
aspfohl: added comment. could likely just keep default..

.github/workflows/pytest-gpu.yaml
```diff
@@ -52,7 +52,6 @@ jobs:
       run: |
         set -ex
         python -m pip install mosaicml-cli
-        mcli init --mcloud
```
dakinggg, line 55: was this just not necessary?

tests/callbacks/test_async_eval_callback.py
```diff
@@ -0,0 +1,237 @@
...
+from mcli import Run, RunConfig, RunStatus
+
+RUN_NAME = 'foo_bar-1234'
+
+
+def test_get_run_name():
+    a = get_run_name('foo-1234', 0)
+    assert a == 'eval0-foo'
+
+    b = get_run_name(50 * 'foo' + '-1234', 1)
```
dakinggg, line 19: what is this testing? something about name length?

scripts/train/train.py
```diff
@@ -545,14 +549,14 @@ def main(cfg: DictConfig) -> Trainer:
 
     eval_gauntlet_callback = None
 
-    if icl_tasks_config is not None:
```
dakinggg, line 548: I think you should include the eval loader section above in the async run? It is a bit strange to only do some of the eval async.

dakinggg: realizing that you did this because the eval.py script doesn't support a normal eval loader, only ICL. we should fix that along with adding support for it to async eval. doesn't necessarily need to be the same PR though.

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,214 @@
...
+        name_suffix = name_suffix[:MAX_RUN_NAME_LENGTH]
+
+    return f'eval{count}-{name_suffix}'
+
+
+def get_load_path(save_folder: str,
+                  save_latest_filename: Optional[str] = None) -> str:
+    # TODO: check that the prefix is remote and not a local file (not supported of course)
+
+    if not save_latest_filename:
```
dakinggg, line 58:
```suggestion
    if save_latest_filename is None:
```

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,214 @@
...
+        rank = dist.get_global_rank()
+        save_latest_filename = f'latest-rank{rank}.pt'
+
+    return f'{save_folder}/{save_latest_filename}'
```
dakinggg, line 62: So this will just use whatever the last checkpoint that has arrived in object store is? Even if it has already evaled on that checkpoint?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def get_eval_models_dict(
+    model: Dict[str, Any],
+    tokenizer: Dict[str, Any],
+) -> List[Dict[str, Any]]:
+    name = model.get('name')
+
+    cfg_overrides = model.pop('cfg_overrides', {})
```
dakinggg, line 71: this doesn't seem quite right, could you explain what you are trying to do with these overrides?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,214 @@
...
+    """Run the eval loop asynchronously as part of a MosaicML platform run.
+
+    Args:
+        interval: Union[str, int, Time]: The interval describing how often eval runs should be
+            launched. If an integer, it will be assumed to be in :attr:`.TimeUnit.EPOCH`.
+            Otherwise, the unit must be either :attr:`.TimeUnit.EPOCH`, :attr:`.TimeUnit.BATCH`,
+            :attr:`.TimeUnit.TOKEN`, or :attr:`.TimeUnit.SAMPLE`.
+        compute: Optional[Union[ComputeConfig, Dict[str, Any]]]: The compute configuration to
+            use for the eval run. If not provided, the same cluster as the current run and a
+            single GPU node will be used.
```
dakinggg, line 93:
```suggestion
            single, full GPU node will be used.
```

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,214 @@
...
+    ):
+        self.check_interval = create_interval_scheduler(interval)
+        self.compute = compute
+        self.count = 0
+        self.last_launch: Optional[Time] = None
+
+        # Run these during init to fail fast in any of the error cases
+        self.current_run = self._get_current_run()
+        self.get_eval_parameters(
+            self.current_run.submitted_config.parameters or {},
```
dakinggg, line 109: What happens if the current run params is empty?

llmfoundry/callbacks/async_eval_callback.py
dakinggg: wdym? the condition as you have it written will only launch the run on global rank 0 (i.e. once). I think that is the correct behavior?

llmfoundry/callbacks/async_eval_callback.py
dakinggg: Yeah, I think it should be the same run. but then, as we were discussing, we need a way to propagate through what "step" to log the eval for.

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,214 @@
...
+            'gpus': 8,
+            'cluster': self.current_run.cluster,
+        }
+        params = self.get_eval_parameters(cfg.parameters or {},
+                                          self.current_run.name)
+
+        # TODO: This just runs an eval run, but we also want to attach the
+        # deployment, which would require a hf conversion and parametrizing the
+        # dependent_deployment in the run config
+        command = 'cd llm-foundry/scripts \n composer eval/eval.py $PARAMETERS'
```
dakinggg, line 197: I suppose this doesn't work if I don't use the parameters section. It also doesn't work if I command line override any of the parameters. Should we be using the config that gets passed to `train.py` instead?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,214 @@
...
+
+    new_model = {'model_name': name, 'model': model}
+
+    if tokenizer:
+        new_model['tokenizer'] = tokenizer
+
+    return [new_model]
+
+
+class AsyncEval(Callback):
```
dakinggg, line 83: Lets note in the docstring/at the top of the file that this callback is experimental, and the API is subject to change at any time.

llmfoundry/callbacks/async_eval_callback.py
aspfohl: Had made this comment I think before fixing it, yea makes sense to me!

.github/workflows/pytest-gpu.yaml
aspfohl: yep - you just need an api key in mcli 0.5

scripts/train/train.py
aspfohl: Can add as follow up! Do you think we want to run the eval loader at the same frequency as the icl tasks / gauntlet? I had assumed we may want to run the normal eval more frequently, and only larger eval tasks at major checkpoints

tests/callbacks/test_async_eval_callback.py
aspfohl: If the name is too long run creation will fail. Can add a small comment

llmfoundry/callbacks/async_eval_callback.py
aspfohl: This was a slight typo on the key name (->`config_overrides`). I wasn't seeing `config_overrides` being used by the eval yamls?

For example, random training yaml
```yaml
model:
  name: hf_causal_lm
  pretrained_model_name_or_path: mosaicml/mpt-7b
  pretrained: true  # false: only use the architecture; true: initialize with pretrained weights
  config_overrides:
    max_seq_len: ${max_seq_len}
    attn_config:
      attn_impl: triton
      # Set this to `true` if using `train_loader.dataset.packing_ratio` below
      attn_uses_sequence_id: false
```

model configs in eval yamls:
```yaml
  models:
  -
    model_name: mpt1b
    tokenizer:
      name: EleutherAI/gpt-neox-20b
      kwargs:
    model:
      name: mpt_causal_lm
      init_device: mixed
      d_model: 2048
      n_heads: 16 # Modified 24->16 so that d_head == 128 to satisfy FlashAttention
      n_layers: 24
      expansion_ratio: 4
      max_seq_len: ${max_seq_len}
      vocab_size: 50368
      attn_config:
        attn_impl: triton
```

llmfoundry/callbacks/async_eval_callback.py
aspfohl: yea - do you think it would make more sense to somehow pull this from state? It sort of assumes you know what you're doing when you configure the checkpointing and async eval intervals

llmfoundry/callbacks/async_eval_callback.py
aspfohl: Updated in b58ccf9c33d6d1b3227762fb7a8c1552a350a56d!

llmfoundry/callbacks/async_eval_callback.py
aspfohl: Updated in b58ccf9c33d6d1b3227762fb7a8c1552a350a56d! missing parameters will raise an error on callback initialization

llmfoundry/callbacks/async_eval_callback.py
dakinggg: ok, thats fair. one separate eval run is way better I think. Although as long as you use a wandb group for your run, they would all show up in the group and maybe its not so bad.

scripts/train/train.py
dakinggg: yeah, right now `train.py` just runs all eval at the same interval. `eval.py` only runs eval once, so interval is not a relevant piece of this. So it would make sense to just keep the existing behavior of running at the same interval I think. We can add an optional interval arg to each individual ICL task if we want/it becomes worthwhile for some reason, but doesn't seem like a priority to me.

llmfoundry/callbacks/async_eval_callback.py
dakinggg: Unfortunately we don't really have a way to know when the checkpoint actually arrives in object store, which would be the ideal time run the asynceval callback. I previously suggested something along the lines of the eval callback keeping track of what checkpoints it has evaled and not doing the same one more than once, but that would require the eval job to be able to communicate back to the training job or something. hmmm

llmfoundry/callbacks/async_eval_callback.py
dakinggg: it will depend on which model type you are instantiating. `hf_causal_lm` and `mpt_causal_lm` yaml sections look different, but each type should persist from the original training config straight to the eval config. Might be easier to explain more in person.

llmfoundry/callbacks/async_eval_callback.py
aspfohl: Summarizing conversation with @dakinggg, and iterating over possible solutions to async eval checkpointing:
1. Callback checks in "micro intervals" if any checkpoints are ready, and if so schedules new runs
    * Would need to know when the checkpoint is ready. Reporting on when the checkpoint has finished uploading is difficult (but not impossible) because of complexity of sharded checkpoints. 
    * Completed checkpoints need to be accessible in state
2. Sidecar CPU run manages eval run creation (no async eval callback at all)
    * Need dedicated cpu instances on each cluster (one per expected hero / finetuning run)
    * Need to know when checkpoint is ready
3. Eval "wait for checkpoint" parameter and custom logic to wait until checkpoint is ready
    * GPUs sit idle until checkpoint is ready
    * Timeout needed to keep these runs not running forever. What happens on failure case?
4. Cloud support for "scheduled runs" and wait to run eval runs for fixed amount of time 
    * Checkpoint completion highly dependent on model size & network quality, "fixed" time not a great solution
5. Async callback called with latest checkpoint (pull from state in CheckpointSaver callback)
    * Intermediate checkpoints may fail if checkpoint is not ready. These eval runs can have watchdog enabled or be manually retried. It might be ok if not all intermediate checkpoint evals successfully run
    * The final checkpoint must be eval-ed. Checkpointsaver will block on fit end - Async eval should trigger after this

Going with (5), unless anyone has major feedback, ack @dakinggg @eracah @mvpatel2000

dakinggg: ya, 5 is good for me for prototyping. i would guess that 2 is the long term solution though.

scripts/train/train.py
aspfohl: Pushing as followup, assigned this to me https://databricks.atlassian.net/browse/RESEARCH-870

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,274 @@
...
+from composer.loggers.mosaicml_logger import (MOSAICML_PLATFORM_ENV_VAR,
+                                              RUN_NAME_ENV_VAR)
+from composer.utils import dist
+from composer.utils.misc import create_interval_scheduler
+
+from mcli import ComputeConfig, Run, RunConfig, create_run, get_run
+
+log = logging.getLogger(__name__)
+
+MAX_RUN_NAME_LENGTH = 40
```
dakinggg, line 25: is this importable from mcli?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,274 @@
...
+    'eval_gauntlet',
+    'fsdp_config',
+    'icl_subset_num_batches',
+    'loggers',
+    'precision',
+    'python_log_level',
+    'seed',
+}
+
+NAME_PREFIX = 'eval'
```
dakinggg, line 45: not clear from this variable name what this is, can you make more descriptive?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,274 @@
...
+    for logger, config in subset_keys.get('loggers', {}).items():
+        if logger == 'wandb':
+            config['group'] = config.pop('name', training_run_name)
+
+        # mlflow currently does not support grouping, so this will just launch
+        # a new mlflow run
+
+    # Create new eval models list
+    model = subset_keys.pop('model')
+    new_models = {'model_name': model.get('name'), 'model': model}
```
dakinggg, line 144: is `model_name` optional? if so this is fine, otherwise it should raise a nice error if `model.get('name')` gives `None`

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,274 @@
...
+            parameters=self.training_config,
+            checkpoint=checkpoint,
+            training_run_name=self.current_run.name,
+        )
+        params['run_name'] = run_name
+
+        # TODO: This just runs an eval run, but we also want to attach the
+        # deployment, which would require a hf conversion and parametrizing the
+        # dependent_deployment in the run config
+        command = 'cd llm-foundry/scripts \n composer eval/eval.py $PARAMETERS'
```
dakinggg, line 259: this command assumes that the llm foundry integration is set up a specific way. may need to do a bit more parsing to make sure these paths are valid

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,274 @@
...
+        get_eval_parameters(
+            parameters=training_config,
+            checkpoint='test',
+            training_run_name=self.current_run.name,
+        )
+        log.info(
+            f'Initialized AsyncEval callback. Will generate runs at interval {interval}'
+        )
+
+    def run_event(self, event: Event, state: State, logger: Logger) -> None:
```
dakinggg, line 200: did you check that this waits for the final checkpoint to finish uploading before triggering on `FIT_END`? I believe it would have to run after `RemoteUploaderDownloader` in order for this to hold true. I think this is actually explicitly not true because of the order callbacks are run in (https://github.com/mosaicml/composer/blob/66f412dcc5965d8ea5d5e4d46469893ada06071e/composer/core/engine.py#L278-L294)

dakinggg: we could maybe launch the final run on `close`... @eracah got a better way?

scripts/train/train.py
```diff
@@ -503,10 +504,13 @@ def main(cfg: DictConfig) -> Trainer:
 
     # Callbacks
     callbacks: List[Callback] = [
-        build_callback(str(name), callback_cfg)
+        build_callback(str(name), callback_cfg, om.to_container(logged_cfg))
```
dakinggg, line 507: I think you need to set `resolve=True` in the `to_container` call? having trouble finding the default. or is `logged_cfg` already resolved?

scripts/train/train.py
```diff
@@ -545,14 +549,14 @@ def main(cfg: DictConfig) -> Trainer:
 
     eval_gauntlet_callback = None
 
-    if icl_tasks_config is not None:
+    if icl_tasks_config is not None and not use_async_eval:
```
dakinggg, line 552: maybe make a jira to move the non ICL eval to this callback too after your other PR lands?

llmfoundry/utils/builders.py
```diff
@@ -76,8 +76,11 @@ def build_icl_data_and_gauntlet(
     return icl_evaluators, logger_keys, eval_gauntlet_cb
 
 
-def build_callback(name: str, kwargs: Union[DictConfig, Dict[str,
-                                                             Any]]) -> Callback:
+def build_callback(
+    name: str,
+    kwargs: Union[DictConfig, Dict[str, Any]],
+    config: Dict[str, Any],
```
dakinggg, line 82: should we default `config` to `None` since only some callbacks (i.e. yours) needs it? and raise error if not provided when making async eval? fine either way, i think this api is not critical to keep backwards compat on

scripts/train/train.py
aspfohl: `logged_cfg` is already resolved (ln 188) - needed the deep copy  of it to have the config parameters before they were popped

llmfoundry/utils/builders.py
aspfohl: Yea it didn't feel too important to keep backwards compat since it is only used for training. Tests are already updated, would prefer to keep as is

llmfoundry/callbacks/async_eval_callback.py
aspfohl: eek, yea this is required for eval

scripts/train/train.py
aspfohl: Added eval loader in add7fbbee444f8434f5069a9defd00fcbb56b017 - actually a pretty small change

llmfoundry/callbacks/async_eval_callback.py
eracah: @dakinggg, it is true that if this runs on fit_end it will run before the RUD fit_end, but this callback seems to operate on locally saved checkpoints, so that means:
1. No need to worry that this callback will run `FIT_END` before RUD does
2. we need a way for it to come after `CheckpointSaver` in the callbacks list
3. We need to figure out a way to maybe edit num_checkpoints_to_keep to be nonzero in order for the callback to work. Otherwise the local checkpoints it needs will be gone!

aspfohl: Synced offline, sounds like we'll just want 
1. configure check_interval, so it doesn't run at the last batch_end
2. make a close function in the callback that calls the launch function

aspfohl:
Example of latest successful runs (trivial 1b ex - async every ba, trains for 3ba)

mpt-1b-demo-d2tV6B
eval-2ba-mpt-1b-demo-tYzJlk (Actually batch 1)
eval-3ba-mpt-1b-demo-N83r5n (Actually batch 2)
eval-final-mpt-1b-demo-m2Pf1K (Batch 3 - latest/final)

Messed up the experiment tracker config so rerunning now once there's compute availability..

scripts/train/train.py
```diff
@@ -528,16 +532,16 @@ def main(cfg: DictConfig) -> Trainer:
     print('Building eval loader...')
     eval_icl_seq_len: int = icl_seq_len if icl_seq_len else max_seq_len
     evaluators, _, eval_gauntlet_callback = build_evaluators(
-        eval_loader_config,
-        icl_tasks_config,
-        eval_gauntlet_config,
+        eval_loader_config,  # TODO: async eval should not even call eval loader
```
dakinggg, line 535: whats the todo here? don't quite understand

llmfoundry/utils/builders.py
```diff
@@ -157,8 +157,11 @@ def build_icl_data_and_gauntlet(
     return icl_evaluators, logger_keys, eval_gauntlet_cb
 
 
-def build_callback(name: str, kwargs: Union[DictConfig, Dict[str,
-                                                             Any]]) -> Callback:
+def build_callback(
+    name: str,
+    kwargs: Union[DictConfig, Dict[str, Any]],
+    config: Any = None,
```
dakinggg, line 163: this type is `Dict[str, Any]` right?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,380 @@
...
+    if looking_for:
+        raise Exception(
+            f'Missing the following required parameters for async eval: {looking_for}'
+        )
+
+    for logger, config in subset_keys.get('loggers', {}).items():
+        if logger == 'wandb':
+            config['group'] = config.pop('name', training_run_name)
+
+            config['init_kwargs'] = config.pop('init_kwargs', {})
```
dakinggg, line 139: I think it would be a bit nicer to do this through `metadata` arg (currently only present in `train.py` but easy to copy over to `eval.py`). That way it will automatically end up on all the trackers (wandb, mlflow, and run metadata). wdyt?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,380 @@
...
+            config['init_kwargs']['config'] = config['init_kwargs'].pop(
+                'config', {})
+            config['init_kwargs']['config']['eval_interval'] = interval.value
```
dakinggg, line 142: I think calling this `eval_timestamp` would be more clear.

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,380 @@
...
+            config['init_kwargs']['config'][
```
dakinggg, line 143: and similarly `eval_timestamp_units`

llmfoundry/callbacks/async_eval_callback.py
dakinggg: id propagate this change up to the args to this function too. this isn't really an "interval" right? its the timestamp we're currently doing eval at?

dakinggg (approved):
LGTM, few remaining comments

scripts/train/train.py
aspfohl: in async eval we don't pass the icl / gauntlet configs into `build_evaluators`. There's a little bit more work needed to integrated into https://github.com/mosaicml/llm-foundry/pull/742 

Imagining we wouldn't even call build_evaluators in this case

dakinggg: ohhh got it, makes sense. Yeah, please update the comment, thanks!

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -0,0 +1,375 @@
...
+        # dependent_deployment in the run config
+        command = f'cd {installation_path}/scripts \n composer eval/eval.py $PARAMETERS'
+        run_config = RunConfig(
+            name=run_name,
+            image=self.current_run.image,
+            compute=self.compute or default_compute,
+            command=command,
+            integrations=integrations,
+            env_variables=cfg.env_variables,
+            metadata=cfg.metadata,
```
dakinggg, line 368: In order for the metadata to make it to wandb, you'll need to copy https://github.com/mosaicml/llm-foundry/blob/5fdcc43bb31beccbd65bc0f18c5b12b7940445be/scripts/train/train.py#L469-L475 and https://github.com/mosaicml/llm-foundry/blob/5fdcc43bb31beccbd65bc0f18c5b12b7940445be/scripts/train/train.py#L613 from `train.py` to `eval.py`

PR #704, opened 2023-10-30
Remove prefixlm support for OPT and Bloom

### LINKED ISSUE #703: MPT models on the Hub not working with `transformers` main
Hi there! 

Currently with transformers main loading MPT models from the Hub fails because it tries to import some private method (such as `_expand_mask` ) that has been recently removed: https://github.com/huggingface/transformers/pull/27086

The simple loading script below should work

```python
from accelerate import init_empty_weights
from transformers import AutoModelForCausalLM, AutoConfig

model_id = "mosaicml/mpt-7b"
config = AutoConfig.from_pretrained(
    model_id, trust_remote_code=True
)
with init_empty_weights():
    model = AutoModelForCausalLM.from_config(
        config, trust_remote_code=True
    )
```

dakinggg:
Current code is broken on transformers main due to private functions being removed. We are just removing support for these models types as we don't think these code paths are used, and we would need to do further investigation to update to the latest masking code in transformers.

Closes #703


tests/test_model.py
```diff
@@ -438,36 +437,6 @@ def test_loss_fn():
                                     atol=1e-4), f'differed at step {i}'
 
 
-@pytest.mark.parametrize('prefixlm', [False, True])
-def test_opt_wrapping(prefixlm: bool):
```
alextrott16, line 442: I don't know that we want to axe this test, unless we're giving up on OPT support entirely. I'm pretty sure this test actually is more to do with FSDP wrapping being correct, so we'd just want to slice away the Prefix LM part of it.

alextrott16 (commented):
The changes to `hf_prefixlm_converter.py` look good, but I think you're throwing out more test code than you want.

alextrott16 (approved):
Welcome back, `test_opt_wrapping`. Welcome back...

Approve!

tests/test_model.py
dakinggg: you are correct. fixed.

PR #705, opened 2023-10-30
Fix attention patch compatibility for llama2

irenedea:
manual test with triton patch: llama2-finetune-regression-XpDBf9


tests/test_huggingface_flash.py
```diff
@@ -122,6 +164,7 @@ def test_flash2(model_name: str, use_flash_attention_2: bool):
             'config_overrides': {
                 'num_hidden_layers': 2,
                 'intermediate_size': 64,
+                'use_cache': False,
```
irenedea, line 167: cache needs to be false when using patch https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/layers/llama_attention_monkeypatch.py#L83

llmfoundry/models/layers/llama_attention_monkeypatch.py
```diff
@@ -78,6 +78,8 @@ def llama_attention_patch_torch(
     past_key_value: Optional[Tuple[torch.Tensor]] = None,
     output_attentions: bool = False,
     use_cache: bool = False,
+    # Temporary fix for llama2 transformers compatibility, padding_mask will be removed in the next transformers release >4.32.1.
```
dakinggg, line 81: >=4.34.1 would be the next release, right?

tests/test_huggingface_flash.py
dakinggg: discussed offline, this should be removed

llmfoundry/models/layers/llama_attention_monkeypatch.py
irenedea: oops typo, yeah 4.34.1 is the latest, so anything >4.34.1

dakinggg (approved):

PR #706, opened 2023-10-30
Add num_proc to map and filter calls

dakinggg:
Adds the `num_proc` arg to the calls to huggingface dataset map/filter.

before: `num-proc-test-baseline-1-cDb1Vk`
after: `num-proc-test-1-4nqb1g`

dataset map goes from ~27 seconds to ~4 seconds for dolly_hhrlhf. Should be a bigger win for bigger datasets.

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -347,7 +347,7 @@ class DatasetConstructor:
             dataset_mapper,
             batched=False,
             remove_columns=columns_to_remove,
-            num_proc=num_cpus_to_use),
+            num_proc=num_cpus_to_use,
         )
         prompt_length_filtered_dataset = tokenized_dataset.filter(
             lambda example: len(example['input_ids']) < max_seq_len,
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -340,17 +340,18 @@ class DatasetConstructor:
             return _tokenize_formatted_example(example, tokenizer)
 
         detected_cpu_count = os.cpu_count() or 1
+        num_cpus_to_use = max(1, detected_cpu_count - 4)
 
         columns_to_remove = list(dataset[0].keys())
         tokenized_dataset = dataset.map(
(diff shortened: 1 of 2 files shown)


dakinggg:
@alextrott16 I vaguely recall you had an issue with adding this at one point. Do you remember this? Seems to work fine for me now.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -339,14 +339,19 @@ def dataset_mapper(example: Dict):
...
             return _tokenize_formatted_example(example, tokenizer)
 
+        detected_cpu_count = os.cpu_count() or 1
+
         columns_to_remove = list(dataset[0].keys())
         tokenized_dataset = dataset.map(
             dataset_mapper,
             batched=False,
             remove_columns=columns_to_remove,
+            num_proc=max(1, detected_cpu_count - 4),
```
irenedea, line 349: why minus 4?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -339,14 +339,19 @@ def dataset_mapper(example: Dict):
...
         )
         prompt_length_filtered_dataset = tokenized_dataset.filter(
-            lambda example: len(example['input_ids']) < max_seq_len)
+            lambda example: len(example['input_ids']) < max_seq_len,
+            num_proc=max(1, detected_cpu_count - 4),
```
irenedea, line 353: super, super nit:  can we use a shared var for L353 and L349

irenedea (approved):
LGTM! question and super nit

llmfoundry/data/finetuning/tasks.py
dakinggg: totally arbitrary choice to not use all the cpus 🤷‍♂️

alextrott16 (approved):
LGTM. Thanks for doing this.

@dakinggg I don't have a particularly clear memory of running into any issues with using `num_proc`. If you feel confident that our test coverage would catch any issues resulting from this leading to wonkiness, then I'm satisfied.

PR #711, opened 2023-11-02
Add Model sharding support with deepspeed

vivekgoe:
In order to execute large models (MPT-30B, MPT-70B) with Deepspeed Zero-3 only, model parameters need to be sharded across devices during loading itself. Added call to deepspeed.zero.Init() to achieve this parameter sharding.

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -406,14 +406,22 @@ def main(cfg: DictConfig):
 
     # Build Model
     print('Initializing model...')
-    with init_context:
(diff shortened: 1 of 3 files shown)


vivekgoe:
@abhi-mosaic we need a fix in Habana SW to go along with this change, so we need to hold on merging this until we figure out how to share that change with you.

vivekgoe:
@abhi-mosaic  I would still like to merge these changes, shall I create another PR on habana_alpha branch and share it with you for review? Please let me know. Thanks.

PR #713, opened 2023-11-03
Fix passed metadata to mlflow logging

wenfeiy-db:
Two fixes:
1. The passed `mlflow_logging_config` is an omegaconfig DictConfig, which is not serializable by yaml and will cause mlflow logging to fail if metadata is passed there. This PR convert it to python native dict in `build_callbacks`.
2. The current implementation doesn't set `task` key in `metadata` as long as `metatdata` is passed in `mlflow_logging_config`. However, it's possible for the client to pass a `metatdata` in `mlflow_logging_config` that doesn't include `task` key. In this case, we should keep other metadata and set the `task`.


dakinggg (commented):
Thanks! Could you take this opportunity to improve our test coverage too? :D I'm thinking just checking that things are called with the correct args [here](https://github.com/mosaicml/llm-foundry/blob/6c412412da951b22bfaabb1d0f0333bc4b31a6d0/tests/test_hf_conversion_script.py#L423-L427) instead of just testing that the function is called

wenfeiy-db:
@dakinggg Could you take another look? Tests added and CI is good!

dakinggg (approved):

PR #714, opened 2023-11-04
Allow logging metadata

dakinggg:
Adds a new `metadata` field to `train.py` that will log metadata to the Mosaic platform and any experiment trackers present

Tested that a `run_group` key can be added to mosaic, wandb, and mlflow

Screenshot 2023-11-04 at 12 01 08 AM
Screenshot 2023-11-04 at 12 02 41 AM
Screenshot 2023-11-04 at 12 03 50 AM


mvpatel2000 (approved):
LGTM

PR #716, opened 2023-11-05
Run HF dataset processing on local rank 0 first

dakinggg:
Adjust the HF dataset processing code to only process the data on local rank 0. Other ranks will use the cached arrow dataset. This is both _much_ faster than processing on all ranks, and seems to resolve various hangs and crashed we have seen for large datasets.

Manual test that this doesn't affect loss curve in anyway
Screenshot 2023-11-06 at 1 26 25 PM

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -38,6 +38,7 @@ import warnings
 from typing import Any, Callable, Dict, List, Optional, Union
 
 import datasets as hf_datasets
+from composer.utils import dist
 from omegaconf import DictConfig
 from streaming import StreamingDataset
 from transformers import PreTrainedTokenizerBase
@@ -340,7 +341,9 @@ class DatasetConstructor:
             return _tokenize_formatted_example(example, tokenizer)
 
         detected_cpu_count = os.cpu_count() or 1
-        num_cpus_to_use = max(1, detected_cpu_count - 4)
+        detected_cpus_with_margin = detected_cpu_count - 4
+        cpus_per_rank = detected_cpus_with_margin // dist.get_world_size()
+        num_cpus_to_use = max(1, cpus_per_rank)
 
         columns_to_remove = list(dataset[0].keys())
         tokenized_dataset = dataset.map(
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -367,7 +367,7 @@ class DatasetConstructor:
         empty_examples_dropped_dataset = prompt_length_filtered_dataset.filter(
             lambda example: len(example['input_ids']) > 0 and len(example[
                 'labels']) > 0 and any(token_id != tokenizer.pad_token_id
-                                       for token_id in example['labels']))
+                                       for token_id in example['labels']), num_proc=num_cpus_to_use)
         empty_examples_removed = len(prompt_length_filtered_dataset) - len(
             empty_examples_dropped_dataset)
         if empty_examples_removed > 0:
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -333,6 +333,13 @@ class DatasetConstructor:
             preprocessing_fn = self.get_preprocessing_fn_from_str(
                 proto_preprocessing_fn, dataset_name)
 
+        signal_file_path = f'.node_{dist.get_node_rank()}_local_rank0_data_prep_completed'
+
+        if dist.get_local_rank() != 0:
+            log.debug("Waiting for local_rank 0 to finish data prep")
+            with dist.local_rank_zero_download_and_wait(signal_file_path):
+                dist.barrier()
+
         dataset = hf_datasets.load_dataset(dataset_name, split=split, **kwargs)
 
         def dataset_mapper(example: Dict):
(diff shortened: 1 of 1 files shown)


xiaohanzhan-db (approved):
Nice! lgtm!

llmfoundry/data/finetuning/tasks.py
```diff
@@ -332,6 +333,13 @@ def build_from_hf(
             preprocessing_fn = self.get_preprocessing_fn_from_str(
                 proto_preprocessing_fn, dataset_name)
 
+        signal_file_path = f'.node_{dist.get_node_rank()}_local_rank0_data_prep_completed'
+
+        if dist.get_local_rank() != 0:
```
karan6181, line 338: Question: Do you need this line since `local_rank_zero_download_and_wait()` is already waiting for non-local ranks?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -332,6 +333,13 @@ def build_from_hf(
...
+            log.debug('Waiting for local_rank 0 to finish data prep')
+            with dist.local_rank_zero_download_and_wait(signal_file_path):
+                dist.barrier()
```
karan6181, line 341: How would local rank 0 would reach here if you keep the line `if dist.get_local_rank() != 0`?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -340,18 +348,21 @@ def dataset_mapper(example: Dict):
             return _tokenize_formatted_example(example, tokenizer)
 
         detected_cpu_count = os.cpu_count() or 1
-        num_cpus_to_use = max(1, detected_cpu_count - 4)
+        detected_cpus_with_margin = detected_cpu_count - 8
```
karan6181, line 351: Can you add a comment in the code on why we are doing this? It's a good selection since we don't want to overload all the CPUs I suppose.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -361,17 +372,32 @@ def dataset_mapper(example: Dict):
...
             warnings.warn(
                 f'Dropped {empty_examples_removed} examples where the prompt or response was empty, '
                 + 'or the response was only padding tokens.')
 
+        if dist.get_local_rank() == 0:
+            log.debug('Local rank 0 finished data prep')
+            with open(signal_file_path, 'wb') as f:
+                f.write(b'local_rank0_completed_data_prep')
+
+            dist.barrier()
```
karan6181, line 397: How would other non-local-0-rank would reach this line, or is this for local rank 0 across multiple nodes?

karan6181 (commented):
pytest runs on a single process so hard to test the functionality using unit test. However, I would recommend adding a unit test to test the functionality either in this PR or the next depending on the criticality of this PR.

llmfoundry/data/finetuning/tasks.py
xiaohanzhan-db: Good question^. At least for MPICH, all processes need to be executing the barrier() call. 
But for torch.dist in python, seem to be that as long as all processes "reach" to this block, barrier is off (in this case non-zero ranks skip this block). But this part may need some testing or comment clarification.

xiaohanzhan-db (changes requested):

llmfoundry/data/finetuning/tasks.py
dakinggg: yes, the way this works is in multiple steps. First, all non local rank 0s wait at `local_rank_zero_download_and_wait()` for the `signal_file` to be written. local rank 0s proceed with all of the dataset processing. This is a file barrier that will not time out. Then, once local rank 0 has finished processing the dataset, it writes the signal file. Then all ranks proceed to the `barrier`. Then, local rank0s exit this code while non local rank 0s load the already processed datasets (should be pretty much instant). Finally, non local rank0s exit this code.

irenedea (approved):
LGTM, couple nits:
(1) update the PR title to "only" => "first"? Or "use cached arrow processing results for non-rank-0" ? 
(2) add comment to explain the map/filter/load functions are all cached via arrow

llmfoundry/data/finetuning/tasks.py
mvpatel2000, line 341:
```suggestion
                pass
```
Instead, use one global barrier at end -- having ranks hit the same "effective" barrier at different points is very confusing

llmfoundry/data/finetuning/tasks.py
```diff
@@ -361,17 +372,32 @@ def dataset_mapper(example: Dict):
...
+            os.remove(signal_file_path)
```
mvpatel2000, line 398:
```suggestion

    dist.barrier()   # Wait for all ranks to sync after data creation
    
    if dist.get_local_rank() == 0:
            os.remove(signal_file_path)
```
Now, all ranks sync at the same barrier

mvpatel2000 (commented):
LGTM but I would refactor the barrier structure so all ranks enter the same barrier

dakinggg:
@karan6181 @mvpatel2000 @xiaohanzhan-db refactored to not have the confusingly placed barriers.

mvpatel2000 (approved):

dakinggg:
@karan6181 @irenedea added more explanatory comments

PR #717, opened 2023-11-06
Add databricks dependency

irenedea:
Adds the databricks dependency group from mosaicml (composer)

manual test run: ift-mpt-7b-regression-tXp1Ls


setup.py
```diff
@@ -81,6 +81,10 @@
     'hf_transfer==0.1.3',
 ]
 
+extra_deps['databricks'] = [
+    'mosaicml[databricks]',
```
irenedea, line 85: didn't specify a version because I assume it default to whatever composer version we're using.

dakinggg: yeah, using the composer version is ok. we only pull it out if we need to be more restrictive than composer for some reason

dakinggg (approved):

PR #720, opened 2023-11-07
add act checkpoint at sub layer level

cli99:
This PR allows users to specify what module to do activation checkpoint through fsdp when `activation_checkpointing` is `true` in fsdp_config. It checks the field `activation_checkpointing_target ` in `model_config`, if nothing is specified but `activation_checkpointing` is set to `true`, then `MPTBLOCK` is act checkpointed. To checkpoint the submodules in ATTN_CLASS_REGISTRY, FFN_CLASS_REGISTRY,NORM_CLASS_REGISTRY, add the corresponding names to `activation_checkpointing_target`.
For example, 
```
activation_checkpointing_target: 
    - grouped_query_attention
```
checkpoints the activation of GroupedQueryAttention.
If multiple submodules are up for act checkpointing, add their names to activation_checkpointing_target

### DIFF
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -45,7 +45,9 @@ from transformers.models.llama.modeling_llama import \
 from transformers.models.llama.modeling_llama import \
     LlamaRotaryEmbedding as HFRotaryEmbedding
 
-from llmfoundry.models.layers.attention import attn_bias_shape, build_attn_bias
+from llmfoundry.models.layers.attention import (ATTN_CLASS_REGISTRY,
+                                                attn_bias_shape,
+                                                build_attn_bias)
 from llmfoundry.models.layers.blocks import MPTBlock
 from llmfoundry.models.layers.custom_embedding import SharedEmbedding
 from llmfoundry.models.layers.fc import FC_CLASS_REGISTRY as FC_CLASS_REGISTRY
@@ -705,7 +707,25 @@ class MPTForCausalLM(MPTPreTrainedModel):
 
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
-        return isinstance(module, MPTBlock)
+        if not hasattr(self.config, 'activation_checkpointing_target'):
+            return isinstance(module, MPTBlock)
+        act_ckpt_str = self.config.activation_checkpointing_target
+        act_ckpt_lst = act_ckpt_str.replace(' ', '').split(',')
+        if act_ckpt_lst:
+            if 'MPTBlock' in act_ckpt_lst or 'mptblock' in act_ckpt_lst:
+                act_ckpt_lst = ['MPTBlock']
+            for mod_name in act_ckpt_lst:
+                if mod_name in ['MPTBlock', 'mptblock']:
+                    mod_type = MPTBlock
+                elif mod_name in ATTN_CLASS_REGISTRY:
+                    mod_type = ATTN_CLASS_REGISTRY[mod_name]
+                elif mod_name in FFN_CLASS_REGISTRY:
+                    mod_type = FFN_CLASS_REGISTRY[mod_name]
+                elif mod_name in NORM_CLASS_REGISTRY:
+                    mod_type = NORM_CLASS_REGISTRY[mod_name]
+                else:
+                    continue
+                return isinstance(module, mod_type)
 
     def prepare_inputs_for_generation(
         self,
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -707,14 +707,15 @@ class MPTForCausalLM(MPTPreTrainedModel):
 
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
-        if not hasattr(self.config, 'activation_checkpointing_target'):
+        if not hasattr(self.config, 'activation_checkpointing_target'
+                      ) or self.config.activation_checkpointing_target is None:
             log.info(
                 f'activation checkpointing MPTBlock as activation_checkpointing_target is not set in model_config'
             )
             return isinstance(module, MPTBlock)
-        act_ckpt_list = self.config.activation_checkpointing_target
-        if act_ckpt_list:
-            if 'MPTBlock' in act_ckpt_list or 'mptblock' in act_ckpt_list:
+        else:
+            act_ckpt_list = self.config.activation_checkpointing_target
+            if 'MPTBlock' in act_ckpt_list:
                 act_ckpt_list = ['MPTBlock']
                 warnings.warn(
                     f'activation checkpointing MPTBlock, ignoring other sub-block modules if specified'
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -715,7 +715,7 @@ class MPTForCausalLM(MPTPreTrainedModel):
             if 'MPTBlock' in act_ckpt_lst or 'mptblock' in act_ckpt_lst:
                 act_ckpt_lst = ['MPTBlock']
             for mod_name in act_ckpt_lst:
-                if mod_name in ['MPTBlock', 'mptblock']:
+                if mod_name.lower() == 'mptblock':
                     mod_type = MPTBlock
                 elif mod_name in ATTN_CLASS_REGISTRY:
                     mod_type = ATTN_CLASS_REGISTRY[mod_name]
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -709,26 +709,28 @@ class MPTForCausalLM(MPTPreTrainedModel):
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
         if not hasattr(self.config, 'activation_checkpointing_target'):
             return isinstance(module, MPTBlock)
-        act_ckpt_str = self.config.activation_checkpointing_target
-        act_ckpt_list = act_ckpt_str.replace(' ', '').split(',')
+        act_ckpt_list = self.config.activation_checkpointing_target
         if act_ckpt_list:
             if 'MPTBlock' in act_ckpt_list or 'mptblock' in act_ckpt_list:
-                act_ckpt_lst = ['MPTBlock']
-            for mod_name in act_ckpt_lst:
+                act_ckpt_list = ['MPTBlock']
+                warnings.warn(
+                    f'activation checkpointing MPTBlock, ignoring other sub-block modules if specified'
+                )
+            mod_types = ()
+            for mod_name in act_ckpt_list:
                 if mod_name.lower() == 'mptblock':
-                    mod_type = MPTBlock
+                    mod_types += (MPTBlock,)
                 elif mod_name in ATTN_CLASS_REGISTRY:
-                    mod_type = ATTN_CLASS_REGISTRY[mod_name]
+                    mod_types += (ATTN_CLASS_REGISTRY[mod_name],)
                 elif mod_name in FFN_CLASS_REGISTRY:
-                    mod_type = FFN_CLASS_REGISTRY[mod_name]
+                    mod_types += (FFN_CLASS_REGISTRY[mod_name],)
                 elif mod_name in NORM_CLASS_REGISTRY:
-                    mod_type = NORM_CLASS_REGISTRY[mod_name]
+                    mod_types += (NORM_CLASS_REGISTRY[mod_name],)
                 else:
                     warnings.warn(
                         f'module name specified in activation_checkpointing_target ({mod_name}) not recognized, available options are names in ATTN_CLASS_REGISTRY, FFN_CLASS_REGISTRY, NORM_CLASS_REGISTRY, or MPTBlock.'
                     )
-                    continue
-                return isinstance(module, mod_type)
+            return isinstance(module, mod_types)
 
     def prepare_inputs_for_generation(
         self,
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -710,9 +710,9 @@ class MPTForCausalLM(MPTPreTrainedModel):
         if not hasattr(self.config, 'activation_checkpointing_target'):
             return isinstance(module, MPTBlock)
         act_ckpt_str = self.config.activation_checkpointing_target
-        act_ckpt_lst = act_ckpt_str.replace(' ', '').split(',')
-        if act_ckpt_lst:
-            if 'MPTBlock' in act_ckpt_lst or 'mptblock' in act_ckpt_lst:
+        act_ckpt_list = act_ckpt_str.replace(' ', '').split(',')
+        if act_ckpt_list:
+            if 'MPTBlock' in act_ckpt_list or 'mptblock' in act_ckpt_list:
                 act_ckpt_lst = ['MPTBlock']
             for mod_name in act_ckpt_lst:
                 if mod_name.lower() == 'mptblock':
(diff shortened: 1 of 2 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,25 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
 
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
-        return isinstance(module, MPTBlock)
+        if not hasattr(self.config, 'activation_checkpointing_target'):
+            return isinstance(module, MPTBlock)
+        act_ckpt_str = self.config.activation_checkpointing_target
+        act_ckpt_lst = act_ckpt_str.replace(' ', '').split(',')
```
mvpatel2000, line 713: Can we use `list` instead of `lst`

mvpatel2000, line 713: why is there a `.replace`?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,25 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+        if act_ckpt_lst:
+            if 'MPTBlock' in act_ckpt_lst or 'mptblock' in act_ckpt_lst:
+                act_ckpt_lst = ['MPTBlock']
+            for mod_name in act_ckpt_lst:
+                if mod_name in ['MPTBlock', 'mptblock']:
+                    mod_type = MPTBlock
```
mvpatel2000, line 719:
```suggestion
                if mod_name.lower() == 'mptblock':
                    mod_type = MPTBlock
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,25 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+                elif mod_name in ATTN_CLASS_REGISTRY:
+                    mod_type = ATTN_CLASS_REGISTRY[mod_name]
+                elif mod_name in FFN_CLASS_REGISTRY:
+                    mod_type = FFN_CLASS_REGISTRY[mod_name]
+                elif mod_name in NORM_CLASS_REGISTRY:
+                    mod_type = NORM_CLASS_REGISTRY[mod_name]
+                else:
+                    continue
```
mvpatel2000, line 727: Can you emit a warning in this case?

llmfoundry/models/mpt/modeling_mpt.py
vchiley, line 713: this list conversion isn't needed. omconfig allows for lists eg adding this in the yaml:
```
activation_checkpointing_target:
  - mptblock
  - grouped_query_attention
```

should result in: 
```config.activation_checkpointing_target = ['mptblock', 'grouped_query_attention']```

Comment (cli99) [reply llmfoundry/models/mpt/modeling_mpt.py:713]: to remove space in the multi-name string, e,g 'grouped_query_attention,  MPTMLP'

File: llmfoundry/models/mpt/modeling_mpt.py [hunk 4]
```diff
@@ -705,7 +707,25 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
-        return isinstance(module, MPTBlock)
+        if not hasattr(self.config, 'activation_checkpointing_target'):
+            return isinstance(module, MPTBlock)
+        act_ckpt_str = self.config.activation_checkpointing_target
+        act_ckpt_lst = act_ckpt_str.replace(' ', '').split(',')
+        if act_ckpt_lst:
+            if 'MPTBlock' in act_ckpt_lst or 'mptblock' in act_ckpt_lst:
+                act_ckpt_lst = ['MPTBlock']
```
vchiley, line 716: why cancel out all other typing?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,25 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+                    mod_type = MPTBlock
+                elif mod_name in ATTN_CLASS_REGISTRY:
+                    mod_type = ATTN_CLASS_REGISTRY[mod_name]
+                elif mod_name in FFN_CLASS_REGISTRY:
+                    mod_type = FFN_CLASS_REGISTRY[mod_name]
+                elif mod_name in NORM_CLASS_REGISTRY:
+                    mod_type = NORM_CLASS_REGISTRY[mod_name]
+                else:
+                    continue
+                return isinstance(module, mod_type)
```
vchiley, line 728: this wont work on a list will it?

u might need to do something like
```suggestion
            mod_types = []
            for mod_name in act_ckpt_lst:
                if mod_name.lower() == 'mptblock':
                    mod_types += [MPTBlock]
                elif mod_name in ATTN_CLASS_REGISTRY:
                    mod_types += [ATTN_CLASS_REGISTRY[mod_name]]
                elif mod_name in FFN_CLASS_REGISTRY:
                    mod_types += [FFN_CLASS_REGISTRY[mod_name]]
                elif mod_name in NORM_CLASS_REGISTRY:
                    mod_types += [NORM_CLASS_REGISTRY[mod_name]]
            return isinstance(module, mod_types)
```

this way `activation_checkpointing_target` can actually be a list.

vchiley (changes requested):

llmfoundry/models/mpt/modeling_mpt.py
cli99: mptblock is layer-level, other module are sub-layer, if one checkpoints a layer, no need to checkpoint what's inside

llmfoundry/models/mpt/modeling_mpt.py
cli99: they are equivalent i think, the return is inside the for loop, once we found a match, then we can return

vchiley: way I'm seeing it:

if you pass in a list thats say: `[ATTN_CLASS_REGISTRY, NORM_CLASS_REGISTRY]`
in your version, we'll only ever get `return isinstance(module, ATTN_CLASS_REGISTRY)` (even if `isinstance(module, NORM_CLASS_REGISTRY)`

in my version, you'll actually get `isinstance(module, [ATTN_CLASS_REGISTRY,, NORM_CLASS_REGISTRY])`

llmfoundry/models/mpt/modeling_mpt.py
vchiley: Why not?
It might be a bad idea (we can add a warning if you want), but it should be possible.

eg you can act ckpt the MPTBlock and also the attn_block. why not?

cli99: act ckpt the MPTBlock meaning you recompute the entire block and do not keep the intermediate activations within the block. What does it mean if you act checkpoint attn_block on top of MPTBlock?

llmfoundry/models/mpt/modeling_mpt.py
cli99: I see, you are right, thanks!

llmfoundry/models/mpt/modeling_mpt.py
vchiley: a block is
-> ln -> Wqkv -> attn -> Wo -> residual_add -> ln -> W1 -> gelu -> W2 -> residual_add ->

act ckpt the block means ONLY the input to that entire block is stored durring fwd pass. Then during the bwd pass, the block does another fwd to recompute intermediate activations

If we act ckpt MPTBlock and also the attn_block, I'd expect the second fwd pass to not store any of the intermediate activations in the attn_block, and then to do a third fwd pass through the attn_block for its bwd. but maybe there this doesn't work for some reason 🤷‍♂️ 

Also image for some reason someone adds more layers (eg attn layers) outside of an MPTBlock and wants to act ckpt those...

we're talking about edge cases i guess. 
either way, either allow for bad behavior (and warn the user that they are doing something weird) or do what you're doing (but again warn the user that you are removing other modules from the act_ckpt list.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,25 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
 
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
-        return isinstance(module, MPTBlock)
+        if not hasattr(self.config, 'activation_checkpointing_target'):
+            return isinstance(module, MPTBlock)
```
dakinggg, line 711: can we add a `log.info` here saying that it is just act ckpting the mptblock?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,25 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+        act_ckpt_str = self.config.activation_checkpointing_target
+        act_ckpt_lst = act_ckpt_str.replace(' ', '').split(',')
+        if act_ckpt_lst:
+            if 'MPTBlock' in act_ckpt_lst or 'mptblock' in act_ckpt_lst:
```
dakinggg, line 715: should we just use the real class names throughout instead of having these lower cased aliases? I think that would be a more clear interface personally.

dakinggg (commented):
Two high level comments:
(1) Please add some tests. Not sure how hard it is to test that act ckpt works the way you think it does, but at least test that everything is constructed the way you expect.
(2) Should we add this to all HF models too? I don't see why not. That would go here: https://github.com/mosaicml/llm-foundry/blob/84c86e3b0a3b63c0c71e52f1f762325daa8adc64/llmfoundry/models/hf/hf_fsdp.py#L200

cli99:
> Two high level comments: (1) Please add some tests. Not sure how hard it is to test that act ckpt works the way you think it does, but at least test that everything is constructed the way you expect. (2) Should we add this to all HF models too? I don't see why not. That would go here:
> 
> https://github.com/mosaicml/llm-foundry/blob/84c86e3b0a3b63c0c71e52f1f762325daa8adc64/llmfoundry/models/hf/hf_fsdp.py#L200

(1) are there some existing act ckpt tests to leverage? Or shall we just check to make sure the module names passing works as expected?
(2) sub-block act checkpointing requires we know the module definitions inside the block , it's model dependent.

dakinggg:
(1) Nope. Probably ok to just test that the module passing works as expected. Would be great if we _actually_ tested activation checkpointing, but that does not exist right now, no.
(2) Right, so maybe you have to provide the full import path of the module you want to checkpoint? Is that reasonable?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,34 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+                if mod_name.lower() == 'mptblock':
+                    mod_types += (MPTBlock,)
+                elif mod_name in ATTN_CLASS_REGISTRY:
+                    mod_types += (ATTN_CLASS_REGISTRY[mod_name],)
+                elif mod_name in FFN_CLASS_REGISTRY:
+                    mod_types += (FFN_CLASS_REGISTRY[mod_name],)
+                elif mod_name in NORM_CLASS_REGISTRY:
+                    mod_types += (NORM_CLASS_REGISTRY[mod_name],)
+                else:
+                    warnings.warn(
```
vchiley, line 734: should this be a value or key error?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,34 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+                      ) or self.config.activation_checkpointing_target is None:
+            log.info(
+                f'activation checkpointing MPTBlock as activation_checkpointing_target is not set in model_config'
+            )
+            return isinstance(module, MPTBlock)
+        else:
+            act_ckpt_list = self.config.activation_checkpointing_target
+            if 'MPTBlock' in act_ckpt_list:
+                act_ckpt_list = ['MPTBlock']
+                warnings.warn(
```
vchiley, line 720: `log.info` instead of warning?
@dakinggg can verify if thats the correct log level.

dakinggg: I think warning is appropriate, because it is something a user can correct (and probably should)

vchiley:
@dakinggg re (2) this member method is a member method of the MPT model.

enabling this for HF models would requires a custom str to class conversion for a list for every hf model class before it is used in [this call](https://github.com/mosaicml/llm-foundry/blob/84c86e3b0a3b63c0c71e52f1f762325daa8adc64/llmfoundry/models/hf/hf_fsdp.py#L200)...
probably outside the scope of this PR 🤷‍♂️ 
(not sure how `maybe you have to provide the full import path of the module you want to checkpoint` would work)

vchiley (approved):
lgtm

I'd wait for @dakinggg approval before merging

dakinggg:
@vchiley you would do something like `import('string name of class')` to get the class. but yeah, doesn't need to be part of this PR. @cli99 can you make a jira for adding custom act ckpt to generic HF models and put it in the LLM Foundry backlog jira epic please?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,30 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
 
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
-        return isinstance(module, MPTBlock)
+        act_ckpt_list = getattr(self.config, 'activation_checkpointing_target',
+                                None) or ['MPTBlock']
+
+        if 'MPTBlock' in act_ckpt_list or 'mptblock' in act_ckpt_list:
+            log.info(
```
dakinggg, line 714: Can we make this log a bit more clear by only saying we are ignoring other subblock modules if we actually are (and specifying which ones we're ignoring)?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,30 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+                mod_types += (MPTBlock,)
+            elif mod_name in ATTN_CLASS_REGISTRY:
+                mod_types += (ATTN_CLASS_REGISTRY[mod_name],)
+            elif mod_name in FFN_CLASS_REGISTRY:
+                mod_types += (FFN_CLASS_REGISTRY[mod_name],)
+            elif mod_name in NORM_CLASS_REGISTRY:
+                mod_types += (NORM_CLASS_REGISTRY[mod_name],)
+            else:
+                raise ValueError(
+                    f'{mod_name=} (specified in activation_checkpointing_target) is not a recognized option, available options are names in ATTN_CLASS_REGISTRY, FFN_CLASS_REGISTRY, NORM_CLASS_REGISTRY, or MPTBlock.'
```
dakinggg, line 731: can we just print all the options here instead of making the user read code to learn the options?

dakinggg (commented):
Also would like basic unit tests testing the use of this param before merging

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -705,7 +707,35 @@ def fsdp_wrap_fn(self, module: nn.Module) -> bool:
...
+                mod_types += (FFN_CLASS_REGISTRY[mod_name],)
+            elif mod_name in NORM_CLASS_REGISTRY:
+                mod_types += (NORM_CLASS_REGISTRY[mod_name],)
+            else:
+                msg = ', '.join(
+                    list(ATTN_CLASS_REGISTRY.keys()) +
+                    list(FFN_CLASS_REGISTRY.keys()) +
+                    list(NORM_CLASS_REGISTRY.keys()) + ['MPTBlock'])
+                raise ValueError(
+                    f'{mod_name} (specified in activation_checkpointing_target) is not a recognized option, available options are {msg}.'
```
mvpatel2000, line 736:
```suggestion
                    f'{mod_name} (specified in activation_checkpointing_target) is not a recognized option out of available options {msg}.'
```
nit

cli99:
@dakinggg , added a unit test

tests/test_fsdp_act_checkpoint.py
```diff
@@ -0,0 +1,74 @@
...
+from composer import Trainer
+from composer.utils import get_device
+from omegaconf import OmegaConf as om
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import \
+    CheckpointWrapper
+
+from llmfoundry.models.mpt.modeling_mpt import ComposerMPTCausalLM
+
+
+# python3 -m composer.cli.launcher -n 2 --master_port 26000 -m pytest -m gpu tests/test_fsdp_act_checkpoint.py::test_fsdp_act_checkpoint  # noqa
```
dakinggg, line 14: can delete this. the test commands are specified in the makefile

dakinggg (approved):

PR #721, opened 2023-11-08
Add Unity Catalog support to HF checkpointer

dakinggg:
Uses our helper function to support all our supported uris, allowing saving to UC.

Manually tested:
- [x] Writing to s3 still works: `s3save-after-maybe-2-Fy4YhU`
- [x] Writing to UC works: `ucsave-after-maybe-3-8ZR8qh` (It doesn't actually work because we're having some issues writing large files to UC, but it mechanically works in that it tries to write to the correct place and some of the files make it.


panchalhp-db (approved):
Thank you for the quick fix!

PR #723, opened 2023-11-08
Better defaults for StreamingDataset subclasses

snarayan21:
With better defaults and new features being added to Streaming, this PR makes sure those improvements are reflected in llm-foundry. Specifically:
* Default `shuffle_algo` changed to `py1e`: More balanced downloads and cache limit performance
* Default `partition_algo` changed to `relaxed`: Elastic determinism and resumptions on many more numbers of nodes
* Default `predownload` changed to `None`: Lower predownload for more balanced download demand. Will be set by `StreamingDataset`.
* Default `num_canonical_nodes` changed to `None`: Now set to equal the number of physical nodes, unless using `py1s` and `py2s` shuffle algorithms, in which case will equal 64 * physical nodes.
* Default `shuffle_block_size` changed to `None`: Now set based on `num_canonical_nodes` to give a consistently good shuffle quality without needing a crazy amount of downloads.

Defaults were changed in Streaming v0.7.0 with this commit: https://github.com/mosaicml/streaming/commit/93bf0549bb1d36092a1488004917b594d2ad251e


karan6181 (commented):
Overall looks good to me. Do we need to change the defaults in https://github.com/mosaicml/llm-foundry/blob/84c86e3b0a3b63c0c71e52f1f762325daa8adc64/llmfoundry/data/denoising.py#L473 as well as in https://github.com/mosaicml/llm-foundry/blob/84c86e3b0a3b63c0c71e52f1f762325daa8adc64/llmfoundry/data/finetuning/dataloader.py#L127?

snarayan21:
> Overall looks good to me. Do we need to change the defaults in
> 
> https://github.com/mosaicml/llm-foundry/blob/84c86e3b0a3b63c0c71e52f1f762325daa8adc64/llmfoundry/data/denoising.py#L473
> 
> as well as in
> https://github.com/mosaicml/llm-foundry/blob/84c86e3b0a3b63c0c71e52f1f762325daa8adc64/llmfoundry/data/finetuning/dataloader.py#L127
> 
> ?

oop yes we do. Changing that now.

karan6181 (commented):
Can you please also update the `epoch_size` data type to `Optional[Union[int, str]]` which is inline with the streaming args? Thanks!

dakinggg (commented):
There are some more to change in `denoising.py`. Will approve after that, lgtm otherwise.

dakinggg (approved):
@snarayan21 please post a PSA in the research slack once you merge this. I'll also include a PSA in the next release notes, but want to make sure research knows that the defaults here are changing.

karan6181 (approved):
LGTM. Thank You!

PR #724, opened 2023-11-09
change default overwrite to True

dakinggg:
I'd like to change the default for `overwrite` to `True` in the `HuggingFaceCheckpointer`. The main usage of this callback is to save HF format checkpoints at a different interval from composer checkpoints (or if its at the same interval, the HF checkpoint is saved before the composer checkpoint. This means that if something fails your run is basically not able to be resumed (unless you go change the `overwrite` param), because it will try to HF checkpoint again to the same location. Let me know if you think there is a better way to resolve this though.


irenedea (approved):

mvpatel2000:
Fwiw I really don't like defaulting to overwrite True. But its foundry / experimental so im more ok with it

PR #725, opened 2023-11-09
Attempt to fix a very occasional hang in datasets map/filter

dakinggg:
Inspired by comments on this datasets issue https://github.com/huggingface/datasets/issues/3172, I'm trying out removing lambdas.

Also adds some extra safety to first load and cache the tokenizer on local rank 0.

Test run: `no-lambdas-1-nSBWMV`
Loss still the same
Screenshot 2023-11-08 at 9 29 11 PM

Also opened an issue on datasets: https://github.com/huggingface/datasets/issues/3172

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -362,8 +362,12 @@ class DatasetConstructor:
             num_proc=num_cpus_to_use,
             desc='Tokenizing dataset',
         )
+
+        def filter_long_prompts(example: Dict) -> bool:
+            return len(example['input_ids']) < max_seq_len
+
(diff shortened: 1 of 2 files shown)


mvpatel2000 (approved):

subercui:
Hi, has this issue been resolved already? I encounters the same message as in https://github.com/huggingface/datasets/issues/6393 . Running with datasets 2.16.0 and llm-foundry recently built from source

dakinggg:
If you're on llm foundry latest, it generally works fine. In very rare scenarios I have still hit the hang reported in that issue to huggingface datasets. If you're completely blocked, I'd recommend converting your dataset to our Streaming format and using that code path instead.

PR #727, opened 2023-11-09
Fix logging verbosity in HF model download script and repair symlinks 

jerrychen109:
This PR temporarily suppresses `InvalidCertWarning` when downloading a HF model from a cache server with `ignore_cert=True`, as the warnings are quite noisy. In addition, it also sets the default log level to `INFO` in the download script so the logging is actually visible.

Also modifies the script to call `download_from_hf_hub` even after a successful cache download for the purpose of repairing the Hugging Face symlinks to address a non-deterministic bug where the symlinks aren't properly repaired by the time composer loads the model.

Tested a run downloading MPT-7B from a cache server using the new commit `94eff5bae6210865e0a7ee60fe2ec7eba4750522`. Note the no-op for `Fetching 24 files` resulting from the download from HF after:
image

And successful model initialization:
image

### DIFF
--- a/llmfoundry/utils/model_download_utils.py
+++ b/llmfoundry/utils/model_download_utils.py
@@ -9,11 +9,13 @@ import time
 from http import HTTPStatus
 from typing import Optional
 from urllib.parse import urljoin
+import warnings
 
+from bs4 import BeautifulSoup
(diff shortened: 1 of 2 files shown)


dakinggg (approved):
LGTM, thanks!

PR #728, opened 2023-11-09
Enable `tie_word_embeddings` config setting to enable / disable weight tied embeddings

vchiley:
Enable disabling embedding weight tying
(Using huggingface `PretrainedConfig`'s `tie_word_embeddings` kwarg)

Same seed, same config (except `tie_word_embeddings: {true, false}`), produces slightly different convergence curves
Screenshot 2023-11-09 at 5 04 05 PM
and the resulting param count is different:
Screenshot 2023-11-09 at 5 06 02 PM

### DIFF
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -618,7 +618,9 @@ class MPTForCausalLM(MPTPreTrainedModel):
 
     def get_output_embeddings(
             self) -> Union[SharedEmbedding, nn.Embedding, nn.Linear]:
-        return self.lm_head or self.transformer.get_input_embeddings()
+        if self.lm_head is not None:
+            return self.lm_head
+        return self.transformer.get_input_embeddings()
 
     def set_output_embeddings(
         self, new_embeddings: Union[SharedEmbedding, nn.Embedding,
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -628,13 +628,19 @@ class MPTForCausalLM(MPTPreTrainedModel):
         if self.lm_head is not None:
             self.lm_head = new_embeddings
         else:
-            assert isinstance(new_embeddings, (SharedEmbedding, nn.Embedding))
+            if not isinstance(new_embeddings, (SharedEmbedding, nn.Embedding)):
+                raise ValueError(
+                    'new_embeddings must be an instance of SharedEmbedding ' +
+                    f'or nn.Embedding, but got {type(new_embeddings)}.')
+            warnings.warn(
+                'Using `set_output_embeddings` to set the embedding layer of ' +
+                'MPTForCausalLM with tied weights. Given weights are tied, ' +
+                'using `set_input_embeddings` is recommended over using ' +
+                '`set_output_embeddings`.')
             self.transformer.set_input_embeddings(new_embeddings)
 
     def tie_weights(self) -> None:
-        if self.lm_head is not None:
-            del self.lm_head
-            self.lm_head = None
+        self.lm_head = None
 
     def set_decoder(self, decoder: MPTModel) -> None:
         self.transformer = decoder
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -684,12 +684,13 @@ class MPTForCausalLM(MPTPreTrainedModel):
             use_cache=use_cache,
         )
 
-        out = outputs.last_hidden_state.to(self.transformer.wte.weight.device)
         if self.transformer.lm_head is not None:
-            logits = self.transformer.lm_head(out)
+            logits = self.transformer.lm_head(outputs.last_hidden_state)
         else:
             # move outputs to same device as weights for token embedding
             # needed to support HF `device_map`
+            out = outputs.last_hidden_state
+            out = out.to(self.transformer.wte.weight.device)
             logits = self.transformer.wte(out, True)
 
         if self.logit_scale is not None:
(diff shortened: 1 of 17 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -574,14 +575,20 @@ class MPTForCausalLM(MPTPreTrainedModel):
...
-        if not config.tie_word_embeddings:
-            raise ValueError(
-                'MPTForCausalLM only supports tied word embeddings')
-
         log.info(f'Instantiating an MPTForCausalLM model from {__file__}')
 
         self.transformer: MPTModel = MPTModel(config)
 
+        self.lm_head = None
+        if config.tie_word_embeddings is False:
```
sashaDoubov, line 583: was there a reason to do `if config.tie_word_embeddings is False:` instead of `if not config.tie_word_embeddings`:

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -602,19 +609,30 @@ def __init__(self, config: MPTConfig):
...
     def set_input_embeddings(
             self, value: Union[SharedEmbedding, nn.Embedding]) -> None:
-        self.transformer.wte = value
+        self.transformer.set_input_embeddings(value)
 
-    def get_output_embeddings(self) -> nn.Embedding:
-        return self.transformer.wte
+    def get_output_embeddings(
+            self) -> Union[SharedEmbedding, nn.Embedding, nn.Linear]:
+        return self.lm_head or self.transformer.get_input_embeddings()
```
sashaDoubov, line 621:
```suggestion
        if self.lm_head is not None:
            return self.lm_head
        return self.transformer.get_input_embeddings()
```
would prefer this to be more explicit

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -859,7 +879,11 @@ def flops_per_batch(self, batch: Mapping) -> int:
         # assume the backward pass is approximately 2x the forward pass
 
         bs, msl = batch['input_ids'].shape[0:2]
-        params_flops_per_token = 2 * self.n_active_params
+        params = self.n_active_params
+        if self.model.transformer.config.tie_word_embeddings is False:
```
sashaDoubov, line 883: same nit.

sashaDoubov, line 883: (potentially) same nit, this is up to you.

llmfoundry/models/mpt/modeling_mpt.py
vchiley: I do this because it seems more explicit, but 🤷‍♂️
I'll change it

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -602,19 +609,32 @@ def __init__(self, config: MPTConfig):
...
 
     def set_output_embeddings(
-            self, new_embeddings: Union[SharedEmbedding, nn.Embedding]) -> None:
-        self.transformer.wte = new_embeddings
+        self, new_embeddings: Union[SharedEmbedding, nn.Embedding,
+                                    nn.Linear]) -> None:
+        if self.lm_head is not None:
+            self.lm_head = new_embeddings
+        else:
+            assert isinstance(new_embeddings, (SharedEmbedding, nn.Embedding))
```
dakinggg, line 631: this should probably be an actual error since it is invalid input, not just convincing pyright?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -602,19 +609,32 @@ def __init__(self, config: MPTConfig):
...
+            self.transformer.set_input_embeddings(new_embeddings)
+
+    def tie_weights(self) -> None:
+        if self.lm_head is not None:
+            del self.lm_head
```
dakinggg, line 636: Is the `del` necessary?

tests/test_model.py
```diff
@@ -796,10 +805,15 @@ def test_generate(attention_impl: str, device: str, pos_emb_config: dict):
...
             **pos_emb_config,
         },
+        tie_word_embeddings=tie_word_embeddings,
     )
     mpt = MPTForCausalLM(hf_config)
-    mpt.eval()
+    if not tie_word_embeddings:
+        assert mpt.lm_head is not None
+        with torch.no_grad():
+            mpt.lm_head.weight.copy_(mpt.transformer.wte.weight)
```
dakinggg, line 814: why do you need this?

llmfoundry/models/mpt/modeling_mpt.py
vchiley: probably not, was just being a little too thorough.
will remove

tests/test_model.py
vchiley: I think test didn't pass without it.
My reasoning: given the tokens are hardcoded, the init of the lm_head needed to be the same as the input embed to match the hardcoded tokens.

llmfoundry/models/mpt/modeling_mpt.py
vchiley: Initially, I did include it just to get past pyright, but can make it more of a real error.

dakinggg (approved):

tests/test_model.py
dakinggg: wdym "the tokens are harcoded"? I think only the input is?

vchiley: found this was caused by a precision / sampling issue.
adding `precision` arg to test and skipping test that fails.

tests/test_model.py
```diff
@@ -1237,7 +1259,7 @@ def test_forward_with_cache(attn_impl: str, device: str, pos_emb_config: dict):
         torch.testing.assert_close(
             second_output.logits,
             full_output.logits[:, -1, :].unsqueeze(1),
-            atol=1e-2,
+            atol=1e-1,
```
dakinggg, line 1262: what about 1.1e-2

vchiley: that seems to pass as well.

dakinggg (approved):

PR #729, opened 2023-11-09
Combine filters into one, to avoid datasets error

dakinggg:
Combines the two `filter` calls into one `filter` call during data preprocessing. It is not clear to me why, but it resolves the issue I reported to `datasets` (https://github.com/huggingface/datasets/issues/6393). Tested many jobs, both single node and large scale to confirm the issue is resolved.

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -377,7 +377,8 @@ class DatasetConstructor:
             filtered_dataset)
         if examples_removed > 0:
             warnings.warn(
-                f'Dropped {examples_removed} examples where the prompt was longer than {max_seq_len}.'
+                f'Dropped {examples_removed} examples where the prompt was longer than {max_seq_len}, ' + 
+                'the prompt or response was empty, or the response was all padding tokens.'
             )
 
         # Now local rank 0 indicates to the other ranks that it is done
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -374,7 +374,7 @@ class DatasetConstructor:
         )
 
         examples_removed = len(tokenized_dataset) - len(
-            prompt_length_filtered_dataset)
+            filtered_dataset)
         if examples_removed > 0:
             warnings.warn(
                 f'Dropped {examples_removed} examples where the prompt was longer than {max_seq_len}.'
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -400,11 +400,11 @@ class DatasetConstructor:
         #         f'Dropped {empty_examples_removed} examples where the prompt or response was empty, '
         #         + 'or the response was only padding tokens.')
 
-        # # Now local rank 0 indicates to the other ranks that it is done
-        # if dist.get_local_rank() == 0:
-        #     log.debug('Local rank 0 finished data prep')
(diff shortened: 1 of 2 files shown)


llmfoundry/data/finetuning/tasks.py
```diff
@@ -363,43 +363,31 @@ def dataset_mapper(example: Dict):
...
-
-        empty_examples_removed = len(prompt_length_filtered_dataset) - len(
-            empty_examples_dropped_dataset)
-        if empty_examples_removed > 0:
-            warnings.warn(
-                f'Dropped {empty_examples_removed} examples where the prompt or response was empty, '
-                + 'or the response was only padding tokens.')
-
         # Now local rank 0 indicates to the other ranks that it is done
         if dist.get_local_rank() == 0:
```
j316chuck, line 392: Just wondering do we need this dist.get_local_rank() == 0 logic anymore by combining the filters?

j316chuck (approved):
Thanks for fixing this one!

llmfoundry/data/finetuning/tasks.py
```diff
@@ -363,43 +363,31 @@ def dataset_mapper(example: Dict):
...
+        def filter_long_or_empty_examples(example: Dict) -> bool:
+            less_than_max_seq_len = len(example['input_ids']) < max_seq_len
+            non_empty_input = len(example['input_ids']) > 0
+            non_empty_labels = len(example['labels']) > 0
+            non_padding_response = any(
+                token_id != pad_token_id for token_id in example['labels'])
+            return (less_than_max_seq_len and non_empty_input and
+                    non_empty_labels and non_padding_response)
+
+        filtered_dataset = tokenized_dataset.filter(
```
j316chuck, line 377: QQ: does this logic happen on every rank?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -363,43 +363,31 @@ def dataset_mapper(example: Dict):
...
 
-        examples_removed = len(tokenized_dataset) - len(
-            prompt_length_filtered_dataset)
+        examples_removed = len(tokenized_dataset) - len(filtered_dataset)
         if examples_removed > 0:
             warnings.warn(
-                f'Dropped {examples_removed} examples where the prompt was longer than {max_seq_len}.'
+                f'Dropped {examples_removed} examples where the prompt was longer than {max_seq_len}, '
+                +
+                'the prompt or response was empty, or the response was all padding tokens.'
```
mvpatel2000, line 388:
```suggestion
                f'Dropped {examples_removed} examples where the prompt was longer than {max_seq_len}, '
                'the prompt or response was empty, or the response was all padding tokens.'
```

mvpatel2000 (approved):

llmfoundry/data/finetuning/tasks.py
mvpatel2000: Should auto concat?

llmfoundry/data/finetuning/tasks.py
dakinggg: yeah, that part is unrelated. still need the data processing happening first on local rank 0.

llmfoundry/data/finetuning/tasks.py
dakinggg: yup. but all the non local rank 0 ranks will just be reading a cached dataset from disk after local rank 0 processes it.

PR #730, opened 2023-11-10
Make TiktokenTokenizerWrapper compatible with convert_composer_to_hf.py

irenedea:
Need to make it possible to do `AutoTokenizer.from_pretrained(x, trust_remote_code=True)`

test run:
test-convert-composer-to-hf-with-fix-DYmTWJ

### DIFF
--- a/llmfoundry/utils/checkpoint_conversion_helpers.py
+++ b/llmfoundry/utils/checkpoint_conversion_helpers.py
@@ -19,7 +19,8 @@ from typing import Any, Dict, Optional, Tuple, Union
 
 import numpy as np
 import sentencepiece as spm
-from transformers import AutoTokenizer, PreTrainedTokenizer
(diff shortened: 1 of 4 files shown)


llmfoundry/tokenizers/tiktoken.py
```diff
@@ -334,3 +334,14 @@ def construct_logit_tensor(self, logprobs: Dict[str,
 
 
 TiktokenTokenizerWrapper.register_for_auto_class()
+
+# Register TiktokenTokenizerWrapper with AutoTokenizer
+from transformers import AutoTokenizer, PretrainedConfig
+
+
+class TiktokenTokenizerWrapperConfig(PretrainedConfig):
```
irenedea, line 342: @dakinggg does creating a tokenizerconfig like this seem correct to you?

dakinggg (approved):

PR #732, opened 2023-11-10
Gate the dist calls in build_tokenizer

dakinggg:
The failing regression test now makes it past the point it failed at (`elastic-resumption-2-to-1-node-JoTmuj`). It still fails, but because I don't have my OCI credentials set up, not because of the dist error. And here is a normal manual run on this PR (`no-dist-1-6mxV5j`).

We don't have a CI setup that can run non in an environment with WORLD_SIZE > 1 and dist not initialized, but I tested this combination of settings manually and confirmed that it failed before this PR, and works after.

### DIFF
--- a/llmfoundry/utils/builders.py
+++ b/llmfoundry/utils/builders.py
@@ -190,9 +190,10 @@ def build_tokenizer(
 
     signal_file_path = f'.node_{dist.get_node_rank()}_local_rank0_completed_tokenizer_setup'
 
-    # Make sure the tokenizer files are downloaded and cached first by local rank 0
-    with dist.local_rank_zero_download_and_wait(signal_file_path):
-        pass
+    if dist.is_available() and dist.is_initialized() and dist.get_world_size() > 1:
+        # Make sure the tokenizer files are downloaded and cached first by local rank 0
+        with dist.local_rank_zero_download_and_wait(signal_file_path):
+            pass
 
     if tokenizer_name.startswith('tiktoken'):
         tokenizer = TiktokenTokenizerWrapper(**tokenizer_kwargs)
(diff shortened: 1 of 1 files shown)


llmfoundry/utils/builders.py
```diff
@@ -190,9 +190,11 @@ def build_tokenizer(
 
     signal_file_path = f'.node_{dist.get_node_rank()}_local_rank0_completed_tokenizer_setup'
 
-    # Make sure the tokenizer files are downloaded and cached first by local rank 0
-    with dist.local_rank_zero_download_and_wait(signal_file_path):
-        pass
+    if dist.is_available() and dist.is_initialized(
```
mvpatel2000, line 193: Do you need both is_available and is_initialized?

mvpatel2000 (approved):

llmfoundry/utils/builders.py
```diff
@@ -208,14 +210,16 @@ def build_tokenizer(
             int(1e30),
         )
 
-    if dist.get_local_rank() == 0:
-        with open(signal_file_path, 'wb') as f:
-            f.write(b'local_rank0_completed_tokenizer_setup')
+    if dist.is_available() and dist.is_initialized(
+    ) and dist.get_world_size() > 1:
+        if dist.get_local_rank() == 0:
```
j316chuck, line 215:
```suggestion
    if dist.is_available() and dist.is_initialized(
    ) and dist.get_world_size() > 1 and dist.get_local_rank() == 0:
```

llmfoundry/utils/builders.py
dakinggg: thats what we use everywhere else. Do I not need both?

mvpatel2000: Lets just leave both

llmfoundry/utils/builders.py
```diff
@@ -208,14 +210,16 @@ def build_tokenizer(
...
+            with open(signal_file_path, 'wb') as f:
+                f.write(b'local_rank0_completed_tokenizer_setup')
 
-    dist.barrier()
+        dist.barrier()
 
-    if dist.get_local_rank() == 0:
-        os.remove(signal_file_path)
+        if dist.get_local_rank() == 0:
```
j316chuck, line 221: Does this need a `if dist.is_available() and dist.is_initialized() and and dist.get_world_size() > 1`? 

Might be useful to wrap in helper function as well

j316chuck (approved):
To test more rigorously, you can also kick off the regression tests from this fork?

llmfoundry/utils/builders.py
dakinggg: its not used much...almost everything is run after initialize dist is called and we don't actually need to check this. Also the line you commented on is inside the above available/initialized check, yeah? If I'm misreading, that would def be a bug

j316chuck: oh yeah nice I misread thanks!

PR #735, opened 2023-11-14
Remove tokenizer_name field

dakinggg:
We forgot to remove the field from some yamls at some point. It is no longer used, as there is a separate `tokenizer` section responsible for building the tokenizer.

Test run using one of these yamls: `mosaic-gpt-1b-gpus-8-JBpxrz`


j316chuck (approved):
Thanks! Just wondering, did this print out in the warnings?

PR #736, opened 2023-11-14
Bump to composer 0.17

dakinggg:


tests/test_mpt_gen.py
```diff
@@ -143,3 +141,52 @@ def test_mpt_generate_callback(attn_impl: str, use_alibi: bool,
...
+        tiny_ft_dataloader: DataLoader):
+    device = get_device('gpu')
+
+    # build mpt model
+    model = build_tiny_mpt(
+        tie_word_embeddings=False,
+        attn_config={
+            'attn_impl': attn_impl,
+            'attn_uses_sequence_id': False,
+            'alibi': True,
```
j316chuck, line 159: nit: does this test fail when alibi is True and tied word embeddings False? If so, would prefer we keep it in the same function and just assert that an error is thrown in this case that we don't support atm.

tests/test_mpt_gen.py
```diff
@@ -143,3 +141,52 @@ def test_mpt_generate_callback(attn_impl: str, use_alibi: bool,
...
+        },
+    )
+    model = device.module_to_device(model)
+
+    # generate callback
+    prompts = [
+        'The best banana bread recipe is',
+        '2+2=',
+        'how much wood could a woodchuck chuck',
```
j316chuck, line 168: good prompt

j316chuck (approved):
LGTM! Left a nit for your consideration

tests/test_mpt_gen.py
dakinggg: nah, there was some flakiness associated with this test so I split it into two in case some weird pytest stuff was happening, and alibi + untied word embeddings didn't seem like a relevant different thing to test so I removed it. I can add it back though.

dakinggg: I'll add it back just to make sure

PR #737, opened 2023-11-14
Fix pairwise attention comparison in test

sashaDoubov:
Fixing attention implementation not being set correctly, thank you to @ShashankMosaicML for flagging.

### DIFF
--- a/tests/test_flash_triton_torch.py
+++ b/tests/test_flash_triton_torch.py
@@ -182,10 +182,15 @@ def test_attn_impl(attn_impl_0: str,
         tp = torch_name_param_map[n]
         assert p.grad is not None
         assert tp.grad is not None
-        assert allclose_helper(p, tp)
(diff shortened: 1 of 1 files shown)


dakinggg (approved):

ShashankMosaicML (approved):

tests/test_flash_triton_torch.py
```diff
@@ -181,8 +182,15 @@ def gen_bias(attn_impl: str):
         tp = torch_name_param_map[n]
         assert p.grad is not None
         assert tp.grad is not None
-        assert allclose_helper(p, tp)
```
ShashankMosaicML, line 184: I this line should not be removed.

ShashankMosaicML: oh, is this removed because they are initialized to be the same?

sashaDoubov: no.. that was unintentional, thank you for catching it

PR #742, opened 2023-11-16
Add eval loader to eval script

aspfohl:


llmfoundry/data/dataloader.py
```diff
@@ -41,4 +41,6 @@ def build_dataloader(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
             device_batch_size,
         )
     else:
-        raise ValueError(f'Not sure how to build dataloader with config: {cfg}')
```
aspfohl, line 44: Nit: this was not a helpful error message 😛

scripts/eval/eval.py
```diff
@@ -122,6 +124,11 @@ def evaluate_model(
         icl_tasks, eval_gauntlet_config, tokenizer, device_eval_batch_size,
         max_seq_len, icl_subset_num_batches)
 
+    if eval_loader_config is not None:
+        loader_evaluators = build_eval_loader(eval_loader_config, tokenizer,
+                                              device_eval_batch_size)
+        evaluators.extend(loader_evaluators)
```
dakinggg, line 130: I don't think this is quite complete, if you look in `train.py` we add the metrics to the evaluator later, and that needs to happen here too.

llmfoundry/utils/builders.py
```diff
@@ -39,6 +40,30 @@
...
+) -> Evaluator:
+    evaluators = []
+
+    is_multi_eval = isinstance(eval_loader_config, ListConfig)
+    eval_configs = eval_loader_config if is_multi_eval else [eval_loader_config]
+    for eval_config in eval_configs:
+        eval_dataloader = build_dataloader(eval_config, tokenizer,
+                                           device_eval_batch_size)
+
+        # For training, metrics are added after the model is created
```
dakinggg, line 56: commented elsewhere also, but I don't think this is quite right. we want to add the _train_ metrics for this "normal" evaluator, but the default of model.get_metrics() will give you _eval_ metrics if I'm reading the code correctly.

dakinggg (commented):
Could you please do a test run that does both normal eval and ICL eval to make sure it works?

scripts/eval/eval.py
```diff
@@ -68,19 +69,21 @@ def load_peft_model(model_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
                     f'Got exception {str(e)} while loading model {model_cfg.name}. {num_retries-retries} retries remaining'
                 )
 
+    assert composer_model_wrapper is not None
```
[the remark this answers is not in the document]
dakinggg: I think this one is just for pyright, so assert is appropriate. That is, the code will never reach this point while `composer_model_wrapper is None`, because it will raise an exception first.

scripts/train/train.py
```diff
@@ -579,13 +548,27 @@ def main(cfg: DictConfig) -> Trainer:
     optimizer_name: str = optimizer_config.pop('name')
     optimizer = build_optimizer(model, optimizer_name, optimizer_config)
 
-    # Now add the eval metrics
+    ## Evaluation
+    print('Building eval loader...')
```
dakinggg, line 552: this was actually written intentionally with creating the dataset before creating the model to help fast fail on dataset errors, since downloading the model can take a while (hence the weird structure of adding the metrics after the fact). I'd like to preserve that fast fail.

tests/data_utils.py
```diff
@@ -3,8 +3,23 @@
...
 from typing import Optional
 
+from omegaconf import DictConfig
+from omegaconf import OmegaConf as om
+
+from scripts.data_prep.convert_dataset_hf import main as main_hf  # noqa: E402
+from scripts.data_prep.convert_dataset_json import \
+    main as main_json  # noqa: E402
+
+# Add repo root to path so we can import scripts and test it
```
dakinggg, line 19: don't you need to do this before you import `scripts`? might be missing something

tests/data_utils.py
```diff
@@ -65,3 +80,83 @@ def make_tiny_ft_dataset(
         for sample in samples:
             _f.write(json.dumps(sample))
             _f.write('\n')
+
+
+def create_c4_dataset_xsmall(path: pathlib.Path) -> str:
+    """Creates a small mocked version of the C4 dataset."""
+    c4_dir = os.path.join(path, f'my-copy-c4')
+    downloaded_split = 'val_xsmall'  # very fast to convert
```
dakinggg, line 88: I changed this to `val_xxsmall` in a different PR. Can you keep that change please? And while you're here, update the function name to also say xxsmall. Thanks!

scripts/eval/eval.py
```diff
@@ -43,7 +44,8 @@ def load_peft_model(model_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
     }
 
     retries = 0
-    while retries < num_retries:
+    composer_model_wrapper = None
+    while retries < num_retries and composer_model_wrapper is None:
```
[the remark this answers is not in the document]
aspfohl: This was technically fine before (although maybe a little error prone with the try and except return/raise logic) , but its the return type that was annoying to deal with 😬

tests/data_utils.py
aspfohl: there may be some magic on previous tests running this already.. moved up just in case, likely want to add something like this in conftest instead but will save that for a later date

llmfoundry/utils/builders.py
```diff
@@ -39,6 +40,90 @@
...
+            'Eval loader requires metrics, either through the models defaults and/or train_metrics'
+        )
+
+    eval_metric_names = list(model.train_metrics.keys())
+    eval_loaders, other_evaluators = [], []
+    for evaluator in evaluators:
+        if evaluator.metric_names == []:
+            evaluator.metric_names = eval_metric_names
+            eval_loaders.append(evaluator)
+        else:
```
aspfohl, line 120: Is it safe to assume ICL evaluators will always be initiated with metrics names?

scripts/train/train.py
```diff
@@ -579,13 +548,27 @@ def main(cfg: DictConfig) -> Trainer:
...
+    eval_gauntlet_callback = None
+    if icl_tasks_config is not None:
+        icl_evaluators, _, eval_gauntlet_callback = build_icl_data_and_gauntlet(
+            icl_tasks_config, eval_gauntlet_config, tokenizer,
+            device_eval_batch_size, icl_seq_len if icl_seq_len else max_seq_len,
+            icl_subset_num_batches)
+        evaluators.extend(icl_evaluators)
+
+    if eval_gauntlet_callback is not None:
+        callbacks.append(eval_gauntlet_callback)
```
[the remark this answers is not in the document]
aspfohl: :+1: I think this makes it feel cleaner, see 1050deff296484c93c9a2416eee96862e6394951 (there may be a few tests I still need to fix)

scripts/train/train.py
aspfohl: This makes it a little annoying, but yea this makes sense 1050deff296484c93c9a2416eee96862e6394951

scripts/eval/eval.py
```diff
@@ -143,6 +156,10 @@ def evaluate_model(
         composer_model = load_model(model_cfg.model, tokenizer, fsdp_config,
                                     num_retries)
 
+    # Now add the eval metrics
+    if eval_loader_config is not None:
+        evaluators = add_metrics_to_eval_loaders(evaluators, composer_model)
```
dakinggg, line 161: i think we can just pass `composer_model.train_metrics` in here instead of the whole model? Any particular reason you did it this way?

dakinggg: also I think this should actually be `composer_model.get_metrics(is_train=True)`. thats the interface that is actually defined on `ComposerModel`. (this would be correcting an error from the code before your changes)

aspfohl:
@dakinggg 
> a training run with an eval loader

mpt-1b-ctx-8k-gpus-8-3IPwqf

> an eval run with an eval loader

mpt-eval-YwAj8C (stopped early)

PR #746, opened 2023-11-18
Reenable tests that were accidentally disabled

dakinggg:
We accidentally had some tests that weren't running due to a mixup between `pytest.mark.gpu` and explicitly skipping tests when cuda is not available. This PR resolves that mix up. Going forward, tests should not need to check for cuda available and skip if not available. The gpu mark should handle that.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1575,11 +1575,11 @@ def test_alibi_vs_hf():
             torch.testing.assert_close(alibi_bias_hf, alibi_bias_m)
 
 
-@pytest.mark.parametrize('attn_impl,device', [
+@pytest.mark.parametrize('attn_impl', [
     'torch',
-    pytest.param('flash', 'gpu', marks=pytest.mark.gpu),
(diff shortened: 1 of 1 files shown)


tests/test_model.py
```diff
@@ -777,14 +773,12 @@ def test_advanced_mask_building(attention_impl: str):
     },
 }])
 @pytest.mark.parametrize('tie_word_embeddings', [True, False])
-def test_generate(attention_impl: str, device: str, precision: str,
-                  pos_emb_config: dict, tie_word_embeddings: bool):
+def test_generate(attention_impl: str, precision: str, pos_emb_config: dict,
+                  tie_word_embeddings: bool):
+    device = 'gpu' if torch.cuda.is_available() else 'cpu'
```
irenedea, line 778:
```suggestion
    device = get_device(None)
```

irenedea: and similar for other places in the file

irenedea (approved):

ShashankMosaicML (approved):

PR #749, opened 2023-11-20
Catch exception raised in hf prep properly

j316chuck:
# Description
**Issue**: There is a barrier deadlock in `build_from_hf` if an Exception is raised which causes errored runs to not fail properly and burn compute. 

Concretely, the non local rank 0 processes depend on a signal file to be written but if a rank 0 process raises an exception then we never reach the `dist.barrier()` state resulting in a deadlock. 

**Solution**: We add a try catch around the dataset preparation code to ensure that local rank 0 gets to writing/removing the signal file enabling all processes to reach the `dist.barrier()` state. 

# Tests:
```
python3   -m composer.cli.launcher -n 2 --master_port 26002   -m pytest -v --durations=20 -m 'gpu' -o tmp_path_retention_policy=none  -k "test_finetuning_dataloader_small"

=============================================================================================================== slowest 20 durations ================================================================================================================
3.69s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-True-2-4]
3.09s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-True-4-8]
3.05s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-False-2-8]
2.95s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-True-2-8]
2.74s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-False-2-8]
2.73s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-False-4-8]
2.63s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-True-4-8]
2.63s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-True-2-8]
2.56s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-True-2-4]
2.55s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-False-2-4]
2.50s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-True-4-4]
2.49s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-False-4-8]
2.43s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-False-4-4]
2.36s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-False-2-4]
2.32s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-True-4-4]
2.27s call     tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-False-4-4]
1.97s setup    tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-True-2-4]
0.24s setup    tests/test_dataloader.py::test_finetuning_dataloader_small_data[True-False-2-4]
0.24s setup    tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-True-2-4]
0.24s setup    tests/test_dataloader.py::test_finetuning_dataloader_small_data[False-True-4-8]
================================================================================================= 16 passed, 1030 deselected, 24 warnings in 48.55s =================================================================================================
```
Also ported the `test_finetuning_dataloader_small` to GPU tests since it wasn't being ran


llmfoundry/data/finetuning/tasks.py
```diff
@@ -403,8 +408,11 @@ def filter_long_or_empty_examples(example: Dict) -> bool:
         if dist.get_local_rank() == 0:
             os.remove(signal_file_path)
 
+        if error:
+            log.error('Error during data prep')
+            raise error
         log.debug('All ranks finished data prep')
-        return filtered_dataset
+        return filtered_dataset # type: ignore
```
dakinggg, line 415: why type ignore?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -403,8 +408,11 @@ def filter_long_or_empty_examples(example: Dict) -> bool:
         if dist.get_local_rank() == 0:
             os.remove(signal_file_path)
 
+        if error:
```
dakinggg, line 411:
```suggestion
        if error is not None:
```

llmfoundry/data/finetuning/tasks.py
j316chuck: because filtered_dataset is of type unbound since it's declared inside the try catch statement.

dakinggg: ah, maybe a bit cleaner to declare it as `None` outside of the try/except, and then just assert its a dataset instance right before return

dakinggg (commented):
Also, lets add a test for this. world size 2 with a broken jsonl file should reproduce for the test I think?

llmfoundry/data/finetuning/tasks.py
j316chuck: fixed

llmfoundry/data/finetuning/tasks.py
```diff
@@ -307,8 +306,8 @@ def get_preprocessing_fn_from_str(
     def build_from_hf(
         self, cfg: DictConfig, max_seq_len: int,
         tokenizer: PreTrainedTokenizerBase
-    ) -> Union[hf_datasets.DatasetDict, hf_datasets.Dataset,
-               hf_datasets.IterableDatasetDict, hf_datasets.IterableDataset]:
+    ) -> Union[hf_datasets.DatasetDict, hf_datasets.Dataset, hf_datasets.
```
dakinggg, line 309: I don't think this should include `None` return type. Is the assert below not enough to convince pyright its not None?

j316chuck: assert is enough removed

PR #752, opened 2023-11-21
Pin Chat format to TiktokenTokenizerWrapper

rajammanabrolu:
Pin ChatML format to TiktokenTokenizerWrapper so that we aren't affected by default changes and can use `apply_chat_template`


tests/test_tiktoken.py
```diff
@@ -259,3 +279,21 @@ def test_additional_special_tokens(model_name: Optional[str],
...
+                         encoding_name: Optional[str], tmp_path: pathlib.Path):
+    special_tokens_to_add = ['<|im_start|>', '<im_end>']
+    wrapped_tokenizer, _, _ = get_tokenizers_for_testing(
+        model_name,
+        encoding_name,
+        tmp_path,
+        add_bos_token=False,
+        add_eos_token=False,
+        additional_special_tokens=special_tokens_to_add)
+    for i, dict_chats in enumerate(MULTI_TURN_CHAT_ML):
```
dakinggg, line 296: looks like this is a loop over a list of length 1? did you intend for it to have more items? or should it not be a list?

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -88,6 +88,9 @@ def pickle_Encoding(enc: Encoding):
         self.add_bos_token = add_bos_token
         self.add_eos_token = add_eos_token
 
+        # Pin default ChatML Format in case HF changes defaults
+        chat_template = "{% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' + '\n'}}{% endfor %}"
```
dakinggg, line 92: It looks like HF implements this is a property of the tokenizer. Is there a reason you did it this way? would prefer to follow HF if there is no particular reason.

https://github.com/huggingface/transformers/blob/38e2633f80a4924bf613b0240622492beee4cfcc/src/transformers/models/llama/tokenization_llama.py#L415

tests/test_tiktoken.py
rajammanabrolu: its so that we can easily add more cases in the future when we have default system prompts and various forms of multi turns

llmfoundry/tokenizers/tiktoken.py
rajammanabrolu: I can't get it to load the template when I do this in tests, test_chat_formatting has chat_template set to None - not too sure what I'm doing wrong

rajammanabrolu: ok fixed now

dakinggg (approved):

PR #760, opened 2023-11-22
enable param group configuration in llm-foundry

vchiley:
eg usage
```
optimizer:
  name: decoupled_lionw
  lr: 6.0e-4
  betas:
  - 0.9
  - 0.95
  weight_decay: 6.0e-4
  param_groups:
  - param_str_match: ln
    lr: 6.0e-6
    weight_decay: 6.0e-6
```

### DIFF
--- a/llmfoundry/optim/lion8b.py
+++ b/llmfoundry/optim/lion8b.py
@@ -1,7 +1,7 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Any, Callable, Dict, Iterable, Optional, Tuple
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
 
 import torch
 
@@ -58,15 +58,17 @@ class DecoupledLionW_8bit(torch.optim.Optimizer):
             device, or b) step() is executed on a non-CUDA parameter.
     """
 
-    def __init__(self,
-                 params: Iterable[torch.Tensor],
-                 lr: float = 1e-3,
-                 betas: Tuple[float, float] = (0.9, 0.99),
-                 weight_decay: float = 0,
-                 quantize: bool = True,
-                 compress_state_dict: bool = False,
-                 error_correction: bool = False,
-                 _fused: bool = True):  # XXX this flag is mostly for testing...
+    def __init__(
+            self,
+            params: Union[Iterable[torch.Tensor], Iterable[Dict[str, Any]]],
+            lr: float = 1e-3,
+            betas: Tuple[float, float] = (0.9, 0.99),
+            weight_decay: float = 0,
+            quantize: bool = True,
+            compress_state_dict: bool = False,
+            error_correction: bool = False,
+            _fused: bool = True,  # XXX this flag is mostly for testing...
+    ):
 
         if lr < 0.0:
             raise ValueError('Invalid learning rate: {}'.format(lr))
--- a/llmfoundry/utils/builders.py
+++ b/llmfoundry/utils/builders.py
@@ -4,7 +4,8 @@
 import logging
 import os
 import warnings
-from typing import Any, Dict, List, Optional, Tuple, Union
+from collections import OrderedDict
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 import torch
 from composer import algorithms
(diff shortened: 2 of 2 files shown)


vchiley:
potential users: @sashaDoubov @samhavens @b-chu @bcui19 @ShashankMosaicML

llmfoundry/utils/builders.py
```diff
@@ -155,18 +158,118 @@ def build_algorithm(name: str, kwargs: Dict[str, Any]) -> Algorithm:
...
+
+    if 'param_groups' in optimizer_config.keys():
+        params = []
+        param_dict = OrderedDict((n, p) for n, p in model.named_parameters())
+
+        for param_group_config in optimizer_config['param_groups']:
+            str_match = param_group_config.pop('param_str_match')
+            filter_fn = functools.partial(re.search, str_match)
+            param_names = [n for n in param_dict.keys() if filter_fn(n)]
+            group_params = {'params': [param_dict.pop(n) for n in param_names]}
```
j316chuck, line 242: Should we allow the user the property to make this exclusive or not? E.g: is there a world where a parameter could exist in two parameter groups?

j316chuck (approved):
Nice, left a comment but thanks for the clean implementation.

llmfoundry/utils/builders.py
vchiley: > is there a world where a parameter could exist in two parameter groups
does pytorch enable this? technically yes
Is it at all common? no

We can potentially enable it in the future in some way if ppl find a use-case & good design pattern for it.
(note on design pattern: At first glance, it seems like if param group assignment is not exclusive, the user would need to define the configuration in a VERY meticulous manner (effectively manually assigning every single param to a param group in the yaml or something like that; seems laborious given it'll probably never be used.))

llmfoundry/utils/builders.py
```diff
@@ -155,18 +158,118 @@ def build_algorithm(name: str, kwargs: Dict[str, Any]) -> Algorithm:
...
+        "name": "decoupled_lionw",
+        "lr": 1e-3,
+        "weight_decay": 1e-2,
+        "betas": [0.9, 0.999],
+        "eps": 1e-8,
+        "disable_grad": ["norm", "bias"]
+    }
+    ```
+
+    To create modify the optimizer parameters for all parameters that contain the
```
dakinggg, line 192:
```suggestion
    To create and modify the optimizer parameters for all parameters that contain the
```

llmfoundry/utils/builders.py
```diff
@@ -155,18 +158,118 @@ def build_algorithm(name: str, kwargs: Dict[str, Any]) -> Algorithm:
...
+    }
+    ```
+
+    Args:
+        model (torch.nn.Module): model to extract parameters from
+        optimizer_config (Dict[str, Any]): optimizer config
+
+    Returns:
+        Union[Iterable[torch.Tensor], Iterable[Dict[str, Any]]]: an iterable of
+            torch.Tensor's or dict's. Specifies what Tensors should be optimized.
```
dakinggg, line 222:
```suggestion
            torch.Tensor's or dict's. Specifies what Tensors should be optimized and their param groupings.
```

llmfoundry/utils/builders.py
```diff
@@ -155,18 +158,118 @@ def build_algorithm(name: str, kwargs: Dict[str, Any]) -> Algorithm:
...
+            group_params = {'params': [param_dict.pop(n) for n in param_names]}
+            group_params.update(param_group_config)
+
+            params.append(group_params)
+
+        optimizer_config.pop('param_groups')
+
+        params.insert(0, {'params': param_dict.values()})
+
+        log.debug(f'Optimizer param_groups: {params}.')
```
dakinggg, line 251: what does this actually print out? is it horrible/can we print something more human readable?

dakinggg (approved):

llmfoundry/optim/lion8b.py
```diff
@@ -58,15 +58,17 @@ class DecoupledLionW_8bit(torch.optim.Optimizer):
...
-                 lr: float = 1e-3,
-                 betas: Tuple[float, float] = (0.9, 0.99),
-                 weight_decay: float = 0,
-                 quantize: bool = True,
-                 compress_state_dict: bool = False,
-                 error_correction: bool = False,
-                 _fused: bool = True):  # XXX this flag is mostly for testing...
+    def __init__(
+            self,
+            params: Union[Iterable[torch.Tensor], Iterable[Dict[str, Any]]],
```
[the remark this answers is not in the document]
vchiley: sure BUT I think this would break with eg torch 1.13? which would then break regressions.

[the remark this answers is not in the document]
dakinggg: oh the type is new? nevermind then. only import it if it is available in 1.13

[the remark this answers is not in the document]
vchiley: https://github.com/pytorch/pytorch/blob/orig/release/1.13/torch/optim/optimizer.py

not in torch 1.13

PR #765, opened 2023-11-28
Add script for doing bulk generation against an endpoint

aspfohl:
As titled. Example usage:
Screenshot 2023-11-28 at 11 16 25 AM

Current limitations:
* Could hit oom if too much data loaded 
* ~~Assumes data is pre-loaded locally~~ supports remote data too!
* Only supports text completion for now (TODO: use SDK?)

Here's an example mcli yaml that works:
```yaml
name: mpt7b-generate
image: mosaicml/pytorch:latest
# compute:
#   gpus: 1
#   cluster: TODO
command: |-
  for i in {1..20}; do curl http://0.0.0.0:8080/v2/ping && echo 'endpoint is up' && break || echo 'sleeping' && sleep 10; done
  python llm-foundry/scripts/inference/endpoint_generate.py --prompts "The lazy dog jumped over" "The best banana bread recipe is"

integrations:
- integration_type: git_repo
  pip_install: .
- integration_type: pip_packages
  packages:
    - aiohttp
    - ratelimit

dependent_deployment:
  model:
    download_parameters:
      hf_path: mosaicml/mpt-7b-instruct
  env_variables:
    - key: MODEL_BACKEND_OVERRIDE
      value: "vllm"

env_variables:
  - key: ENDPOINT_URL
    value: "http://0.0.0.0:8080/v2/completions"
```


scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+
+    parser.add_argument('--max-tokens', type=int, default=100)
+    parser.add_argument('--temperature', type=float, default=1.0)
+    parser.add_argument('--top-k', type=int, default=50)
+    parser.add_argument('--top-p', type=float, default=1.0)
+    return parser.parse_args()
+
+
+def load_prompt_string_from_file(prompt_path_str: str):
```
alextrott16, line 81: I realize that this is being re-used from `hf_generate`, so we don't want them to diverge, but it would be so useful if we could allow for files that contain multiple prompts. It could be as simple as just defining some sort of split delimiter string, using it to do `prompt_string.split(PROMPT_DELIMITER)`, and modifying the script to expect this function to return a list of strings instead of a single string.

linden-li (approved):
LGTM, thanks for adding this

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,221 @@
+# Copyright 2022 MosaicML LLM Foundry authors
```
dakinggg, line 1: Please clearly identify this script as experimental and subject to change at any time at the top of the file

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+
+    parser.add_argument('--max-tokens', type=int, default=100)
+    parser.add_argument('--temperature', type=float, default=1.0)
+    parser.add_argument('--top-k', type=int, default=50)
+    parser.add_argument('--top-p', type=float, default=1.0)
+    return parser.parse_args()
+
+
+def load_prompts_from_file(prompt_path_str: str) -> List[str]:
```
dakinggg, line 89: might as well change `hf_generate.py` to allow multiple prompts too?

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,221 @@
...
+        raise ImportError('Please install aiohttp') from e
+
+    try:
+        from ratelimit import limits, sleep_and_retry
+    except ImportError as e:
+        raise ImportError('Please install ratelimit') from e
+
+    if args.batch_size > args.rate_limit:
+        raise ValueError(
+            f'Batch size is {args.batch_size} but rate limit is set to { args.rate_limit} / s'
```
dakinggg, line 118:
```suggestion
            f'Batch size is {args.batch_size} but rate limit is set to {args.rate_limit} / s'
```

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,221 @@
...
+
+    api_key = os.environ.get(ENDPOINT_API_KEY_ENV, '')
+    if not api_key:
+        log.warning(f'API key not set in {ENDPOINT_API_KEY_ENV}')
+
+    # Load prompts
+    prompt_strings = []
+    for prompt in args.prompts:
+        if prompt.startswith('file::'):
+            prompt = load_prompts_from_file(prompt)
```
dakinggg, line 136: this can be multiple prompts right? so it should extend the below list instead of append?

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,221 @@
...
+        headers = {'Authorization': api_key, 'Content-Type': 'application/json'}
+
+        req_start = time.time()
+        async with session.post(url, headers=headers, json=data) as resp:
+            if resp.ok:
+                try:
+                    response = await resp.json()
+                except requests.JSONDecodeError:
+                    raise Exception(
+                        f'Bad response: {resp.status_code} {resp.reason}'  # type: ignore
```
dakinggg, line 162: why do these need type ignore? is there a type check missing somewhere?

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,229 @@
...
+            batch += 1
+
+        results = await asyncio.gather(*tasks)
+        res = pd.concat(results)
+
+    res.reset_index(drop=True, inplace=True)
+    log.info(f'Generated {len(res)} prompts, example data:')
+    log.info(res.head())
+
+    # save res to local output folder
```
dakinggg, line 207: usually for this kind of thing we just have one `output_path`, and if its local we write locally, if its remote we write remotely. does that work here too?

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,229 @@
...
+    parser = ArgumentParser(
+        description='Call prompts against a text completions endpoint')
+
+    #####
+    # Path Parameters
+    parser.add_argument(
+        '-p',
+        '--prompts',
+        nargs='+',
+        help='Generation prompts. Use syntax "file::/path/to/prompt.txt" to load a ' +\
```
dakinggg, line 43: update docs to reflect allowing multiple prompts

scripts/inference/endpoint_generate.py
aspfohl: The async session post handler was having issues - I can try to fight with it a bit more

scripts/inference/endpoint_generate.py
aspfohl: :+1: I can convert `local_folder` into a temp file that we don't parametrize, and then output folder should work as a local dir? Will make that required in this case

scripts/inference/endpoint_generate.py
aspfohl: Ah, yea good point now that its list[str]

scripts/inference/endpoint_generate.py
dakinggg: also, is it actually a folder? or should it just a single file path?

aspfohl: It's a folder right now - this seemed more consistent with other scripts.. Should it be a file? Doing `{output_folder}/output.csv` as the final file name

dakinggg: If you only have one file in the folder, it should be a file path imo. the others are folder because they actually require folder i think. e.g huggingface checkpoints are folders, not just single files.

scripts/inference/endpoint_generate.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def parse_args() -> Namespace:
+    """Parse commandline arguments."""
+    parser = ArgumentParser(
+        description='Call prompts against a text completions endpoint')
+
+    #####
+    # Path Parameters
+    parser.add_argument(
```
dakinggg, line 40: since this is an experimental script, im ok merging whatever version is useful to people, but would it be useful to support a remote path for the prompts file? should be pretty straightforward using composer's abstractions.

scripts/inference/endpoint_generate.py
aspfohl: yea sgtm, 16c9307

PR #766, opened 2023-11-29
Only strip object names when creating new output path

irenedea:
We should only strip for the call to `os.path.join` to prevent the following:
```
>>> os.path.join('foo', '/bar')
'/bar'
```
when we want the resulting path to be `foo/bar`


dakinggg (approved):
Can you add a simple test please?

XiaohanZhangCMU (approved):
lgtm.

irenedea:
@dakinggg added test [373e65b](https://github.com/mosaicml/llm-foundry/pull/766/commits/373e65b7dd9fcc26d6e832027bcd5319f522d038)

PR #771, opened 2023-12-01
Add "completion" alias for response key

dakinggg:


tests/test_dataloader.py
```diff
@@ -372,6 +373,39 @@ def test_finetuning_dataloader_small_data(dataset_size: int,
...
+        with pytest.raises(KeyError):
+            _tokenize_formatted_example(example, MagicMock())
+
+
+def test_tokenize_example_well_formed():
+    tokenizer = transformers.AutoTokenizer.from_pretrained('gpt2')
+
+    for prompt_key in _ALLOWED_PROMPT_KEYS:
+        for response_key in _ALLOWED_RESPONSE_KEYS:
+            example = {prompt_key: 'prompt', response_key: 'response'}
```
nancyhung, line 403: nit: do you also want to add `completion` response key given that was what this PR aimed to do?

nancyhung (approved):
lgtm, comment about adding to the test case

tests/test_dataloader.py
dakinggg: they keys are parametrized over all the options, you're probably looking at the values, which are just the actual text :)

PR #768, opened 2023-11-30
Reorganize tests to make them easier to find

aspfohl:
Moves tests into folders that roughly mimic the source code structure
Renames test files where it seemed appropriate, to match source code structure
Remove adding "scripts" to path several times - instead do it on test setup
Some adjustments to path-based logic that broke down after moving files

PR #773, opened 2023-12-02
Read UC delta table to GPU cluster

XiaohanZhangCMU:


scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,127 @@
...
+                f'A file or a folder {args.json_output_path} already exists and is not empty. Remove it and retry!'
+            )
+
+    os.makedirs(args.json_output_path, exist_ok=True)
+
+    log.info(f'Directory {args.json_output_path} created.')
+
+    stream_delta_to_json(args)
+
+    print(f'Convert delta to json is done. check {args.json_output_path}.')
```
irenedea, line 126:
```suggestion
```

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,127 @@
...
+    args = parser.parse_args()
+
+    if os.path.exists(args.json_output_path):
+        if not os.path.isdir(args.json_output_path) or os.listdir(
+                args.json_output_path):
+            raise RuntimeError(
+                f'A file or a folder {args.json_output_path} already exists and is not empty. Remove it and retry!'
+            )
+
+    os.makedirs(args.json_output_path, exist_ok=True)
```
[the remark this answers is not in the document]
irenedea: Similarly I think we can move all the log.info calls to that function as well.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,127 @@
...
+        df = pd.DataFrame.from_dict(result)
+        df.to_json(os.path.join(json_output_path,
+                                f'shard_{start+1}_{end}.json'))
+
+    cursor.close()
+    connection.close()
+
+
+if __name__ == '__main__':
+    log.info(f'Start .... Convert delta to json')
```
irenedea, line 83: nit: Can do this after argument parsing and give a more in-depth logging message that includes the delta table name and json output path, for example.

tests/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,66 @@
...
+repo_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.append(repo_dir)
+
+import unittest
+from unittest.mock import MagicMock, patch
+
+from scripts.data_prep.convert_delta_to_json import stream_delta_to_json
+
+
+class TestStreamDeltaToJson(unittest.TestCase):
```
irenedea, line 21: you shouldn't need to inherit from TestCase

```suggestion
class TestStreamDeltaToJson():
```

tests/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,66 @@
...
+        mock_connect.assert_called_once_with(server_hostname='test_host',
+                                             http_path='test_http_path',
+                                             access_token='test_token')
+        mock_to_json.assert_called()
+        mock_cursor.close.assert_called()
+        mock_connection.close.assert_called()
+
+
+if __name__ == '__main__':
+    unittest.main()
```
irenedea, line 66: Shouldn't need this

```suggestion
```

tests/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,66 @@
...
+# copyright 2022 mosaicml llm foundry authors
+# spdx-license-identifier: apache-2.0
+
+import os
+import sys
+from typing import Any
+
+# Add repo root to path so we can import scripts and test it
+repo_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.append(repo_dir)
```
irenedea, line 13: Shouldn't need this as of https://github.com/mosaicml/llm-foundry/pull/768
```suggestion
```

tests/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+# copyright 2022 mosaicml llm foundry authors
+# spdx-license-identifier: apache-2.0
```
irenedea, line 5:
```suggestion
```

tests/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2022 MosaicML LLM Foundry authors
```
irenedea, line 1: move this file to tests/a_scripts (tests have been reorganized as of https://github.com/mosaicml/llm-foundry/pull/768)

tests/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,66 @@
...
+        data_response_item = MagicMock()
+        data_response_item.asDict.return_value = {
+            'name': 'test',
+            'id': 1
+        }  # Assuming SELECT query returns this format
+        mock_cursor.fetchall.side_effect = [[count_response],
+                                            [column_response_item],
+                                            [data_response_item]]
+
+        stream_delta_to_json(mock_args)
```
irenedea, line 55: Can you add a test with a small batch_size?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,127 @@
...
+import logging
+import os
+
+import pandas as pd
+from databricks import sql
+
+log = logging.getLogger(__name__)
+
+
+def stream_delta_to_json(args: argparse.Namespace, batch_size: int = 1 << 20):
```
irenedea, line 14: Should `batch_size` be exposed via the script?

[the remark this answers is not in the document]
XiaohanZhangCMU: Done.

scripts/data_prep/convert_delta_to_json.py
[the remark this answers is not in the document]
XiaohanZhangCMU: Done.

tests/test_convert_delta_to_json.py
XiaohanZhangCMU: Why?

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: It should be automatically adjusted based on available memory and dataset. I'm still thinking how to set this value more properly. It should probably not be exposed.

tests/test_convert_delta_to_json.py
irenedea: We don't do that for any of our other tests

tests/a_scripts/data_prep/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,63 @@
...
+from unittest.mock import MagicMock, patch
+
+from scripts.data_prep.convert_delta_to_json import stream_delta_to_json
+
+
+class TestStreamDeltaToJson():
+
+    @patch('scripts.data_prep.convert_delta_to_json.sql.connect')
+    @patch('scripts.data_prep.convert_delta_to_json.pd.DataFrame.to_json')
+    def test_stream_delta_to_json(self, mock_to_json: Any, mock_connect: Any):
```
irenedea, line 21: Does this need to be in a class? Most of our tests in foundry are not in classes, they're just top level functions

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,139 @@
...
+import os
+import time
+
+import pandas as pd
+from databricks import sql
+
+log = logging.getLogger(__name__)
+
+
+def stream_delta_to_json(server_hostname: str,
```
irenedea, line 15: nit: rename to `convert_delta_to_json`?

tests/a_scripts/data_prep/test_convert_delta_to_json.py
```diff
@@ -0,0 +1,63 @@
...
+        mock_connect.assert_called_once_with(server_hostname='test_host',
+                                             http_path='test_http_path',
+                                             access_token='test_token')
+        mock_to_json.assert_called()
+        mock_cursor.close.assert_called()
+        mock_connection.close.assert_called()
+
+
+if __name__ == '__main__':
+    unittest.main()
```
irenedea, line 63:
```suggestion
```

You should be able to run tests with `pytest tests/a_scripts/data_prep/test_convert_delta_to_json.py`

tests/test_convert_delta_to_json.py
XiaohanZhangCMU: Done

tests/a_scripts/data_prep/test_convert_delta_to_json.py
XiaohanZhangCMU: @irenedea I am getting with or without these two lines now. Any suggestion why? It used to work. And I checked my composer is indeed installed. 

 ImportError while loading conftest '/Users/xiaohan.zhang/Codes/llm-foundryX/tests/conftest.py'.
tests/conftest.py:8: in <module>
    from composer.utils import reproducibility
E   ModuleNotFoundError: No module named 'composer'

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,450 @@
...
+        reader = pa.ipc.open_stream(decompressed_data)
+        table = reader.read_all()
+
+        # Convert the PyArrow table into a pandas DataFrame
+        df = table.to_pandas()
+        df.to_json(
+            os.path.join(json_output_path, 'part_' + str(ipart) + '.json'))
+
+
+def download_arrow_starargs(args: Tuple):
```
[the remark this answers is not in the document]
XiaohanZhangCMU: combined with download_json_starargs. make cf_collect_type = arrow / json a global option.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,450 @@
...
+
+def run_query(
+    q: str,
+    method: str,
+    cursor: Optional[Cursor] = None,
+    spark: Optional[SparkSession] = None,
+    collect: bool = True
+) -> Optional[Union[List[Row], DataFrame, SparkDataFrame]]:
+
+    assert method in ['dbsql', 'dbconnect'], f'Unrecognized method: {method}'
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Done.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,450 @@
...
+    print('JSON files have been combined into a JSONL file.')
+
+
+def run_query(
+    q: str,
+    method: str,
+    cursor: Optional[Cursor] = None,
+    spark: Optional[SparkSession] = None,
+    collect: bool = True
+) -> Optional[Union[List[Row], DataFrame, SparkDataFrame]]:
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Done

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,450 @@
...
+
+        if attempt == max_retry - 1:
+            raise RuntimeError(f'Failed to download after {max_retry} attempts')
+
+
+def download_json_starargs(args: Tuple):
+    return download_json(*args)
+
+
+def download_arrow(ipart: int, url: str, json_output_path: str):
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Fixed.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,450 @@
...
+# This is a monkey patch on top of the DB Connect package that allows
+# the client to fetch the results in different formats from the server. To be
+# able to use the code make sure this module is not overriden by DB Connect classes.
+
+
+def to_cf(self: SparkConnectClient, plan: pb2.Plan, type: str = 'json'):
+    """Executes plan object return as cloud fetch presigned URLS.
+
+    It can handle the current outptu formats that are supported by the server.
+    In contrast to the regular API methods of the client, this method does not
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Done.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,450 @@
...
+Result = namedtuple(
+    'Result', ['url', 'row_count', 'compressed_size', 'uncompressed_size'
+              ])  # pyright: ignore
+
+# This is a monkey patch on top of the DB Connect package that allows
+# the client to fetch the results in different formats from the server. To be
+# able to use the code make sure this module is not overriden by DB Connect classes.
+
+
+def to_cf(self: SparkConnectClient, plan: pb2.Plan, type: str = 'json'):
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Fixed.

tests/a_scripts/data_prep/test_convert_delta_to_json.py
XiaohanZhangCMU: Removed.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+import os
+import time
+import urllib.parse
+from argparse import ArgumentParser, Namespace
+from collections import namedtuple
+from concurrent.futures import ProcessPoolExecutor
+from typing import Iterable, List, Optional, Tuple, Union
+from uuid import uuid4
+
+# PB2 stuff
```
karan6181, line 14: nit: do we need this?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+# the client to fetch the results in different formats from the server. To be
+# able to use the code make sure this module is not overriden by DB Connect classes.
+
+
+def to_cf(self: SparkConnectClient,
+          plan: pb2.Plan,
+          type: str = 'json') -> Tuple[List[Result], int, bool]:
+    """Executes plan object return as cloud fetch presigned URLS.
+
+    It can handle the current outptu formats that are supported by the server.
```
karan6181, line 56:
```suggestion
    It can handle the current output formats that are supported by the server.
```

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+
+    # Add the request options
+    if type == 'json':
+        format = cloud_pb2.ResultOptions.CloudOptions.FORMAT_JSON
+    elif type == 'csv':
+        format = cloud_pb2.ResultOptions.CloudOptions.FORMAT_CSV
+    elif type == 'arrow':
+        format = cloud_pb2.ResultOptions.CloudOptions.FORMAT_ARROW
+    else:
+        raise Exception('Invalid type')
```
karan6181, line 71: Should this be more descriptive saying the supportive format plus a specific exception class?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    cloud_option.Pack(ro)
+    req.request_options.append(
+        pb2.ExecutePlanRequest.RequestOption(extension=cloud_option))
+
+    # Create the iterator
+    iterator = ExecutePlanResponseReattachableIterator(req, self._stub,
+                                                       self._retry_policy,
+                                                       self._builder.metadata())
+    # Iterate over the response
+
```
karan6181, line 89: nit: empty line

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    result = []
+    row_count = 0
+    is_overflow = False
+
+    for response in iterator:
+        if response.HasField('extension') and response.extension.Is(
+                cloud_pb2.CloudResultBatch.DESCRIPTOR):
+            batch = cloud_pb2.CloudResultBatch()
+            assert response.extension.Is(cloud_pb2.CloudResultBatch.DESCRIPTOR)
```
karan6181, line 98: Please avoid the `assert` statement in the source code.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+
+# This is a monkey patch on top of the DB Connect package that allows
+# the client to fetch the results in different formats from the server. To be
+# able to use the code make sure this module is not overriden by DB Connect classes.
+
+
+def to_cf(self: SparkConnectClient,
+          plan: pb2.Plan,
+          type: str = 'json') -> Tuple[List[Result], int, bool]:
+    """Executes plan object return as cloud fetch presigned URLS.
```
karan6181, line 54: Please also add comments for args `plan` and `type`.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+            is_overflow |= batch.truncated
+    return result, row_count, is_overflow
+
+
+SparkConnectClient.to_cf = to_cf  # pyright: ignore
+
+
+def collect_as_cf(self: DataFrame,
+                  type: str = 'json') -> Tuple[List[Result], int, bool]:
+    query = self._plan.to_proto(self._session.client)  # pyright: ignore
```
karan6181, line 114: nit: method comment is missing.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        json_directory(str): directory containing the JSON files
+        output_file(str): output JSONL file
+    """
+    json_files = [f for f in os.listdir(json_directory) if f.endswith('.jsonl')]
+    with open(output_file, 'w') as outfile:
+        for file_name in json_files:
+            with open(os.path.join(json_directory, file_name), 'r') as infile:
+                for line in infile:
+                    outfile.write(line)
+    print('JSON files have been combined into a JSONL file.')
```
karan6181, line 134: please use `log.info()` instead. Thanks! Could you check for all the occurrences?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    Args:
+        json_directory(str): directory containing the JSON files
+        output_file(str): output JSONL file
+    """
+    json_files = [f for f in os.listdir(json_directory) if f.endswith('.jsonl')]
+    with open(output_file, 'w') as outfile:
+        for file_name in json_files:
+            with open(os.path.join(json_directory, file_name), 'r') as infile:
+                for line in infile:
+                    outfile.write(line)
```
karan6181, line 133: the line that your wrote is ok or you need something like this?

```
# Parse the JSON data from the line
data = json.loads(line)
# Write the data to the output file
outfile.write(json.dumps(data) + '\n')
```

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    print('JSON files have been combined into a JSONL file.')
+
+
+def run_query(
+    q: str,
```
karan6181, line 138: better name? `sql_query`?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    if method not in ['dbsql', 'dbconnect']:
+        raise ValueError(f'Unrecognized method: {method}')
+
+    if method == 'dbsql':
+        if cursor is None:
+            raise ValueError(f'cursor cannot be None if using method dbsql')
+        cursor.execute(q)
+        if collect:
+            return cursor.fetchall()
+
```
karan6181, line 162: nit: extra line

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    elif method == 'dbconnect':
+        if spark == None:
+            raise ValueError(f'sparkSession is required for dbconnect')
+        df = spark.sql(q)
+        if collect:
+            return df.collect()
+        return df
+
+    return None
```
karan6181, line 171: so this will be called with `method=dbsql` and `collect=False`, is this the right expectation?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+
+
+def get_args(signed: List, json_output_path: str, columns: List,
```
karan6181, line 174: This method is only being used once. Do you need this method or can you just copy paste the method definition to the caller?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    if resp.status_code == 200:
+        if resp_format == 'json':
+            data = resp.json()
+            pd.DataFrame(data, columns=columns).to_json(os.path.join(
+                json_output_path, 'part_' + str(ipart) + '.jsonl'),
+                                                        orient='records',
+                                                        lines=True)
+            return
+
+        if compressed:
```
karan6181, line 206: Can you please add a comment that this will get called when `resp_format` is `arrow`? Or did I misunderstood the code!

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        # Convert the PyArrow table into a pandas DataFrame
+        df = table.to_pandas()
+        df.to_json(os.path.join(json_output_path,
+                                'part_' + str(ipart) + '.jsonl'),
+                   orient='records',
+                   lines=True,
+                   force_ascii=False)
+
+
+def download_starargs(args: Tuple) -> None:
```
karan6181, line 225: same comment as above, do you need this method?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    return download(*args)
+
+
+def fetch_data(method: str, cursor: Optional[Cursor],
+               sparkSession: Optional[SparkSession], s: int, e: int,
```
karan6181, line 230: better name than `s` and `e`. Maybe `start` and `end`?

Also method comment is missing.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        FROM
+            {tablename}
+    )
+    SELECT {columns_str}
+    FROM NumberedRows
+    WHERE rn BETWEEN {s+1} AND {e}"""
+
+    if method == 'dbconnect':
+        spark_df = run_query(query, method, cursor, sparkSession, collect=False)
+        if not spark_df:
```
karan6181, line 247: `if spark_df is None:` It would be great to specify the condition explicitly.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+            raise RuntimeError(
+                f'Expect spark dataframe with {query} but got None')
+        pdf = spark_df.toPandas()  # pyright: ignore
+    else:  #  method == 'dbsql':
+        ans = run_query(query, method, cursor, sparkSession, collect=True)
+        if not ans:
```
karan6181, line 253: same here

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        sparkSession (pyspark.sql.sparksession): spark session
+        dbsql (databricks.sql.connect): dbsql session
+    """
+    cursor = dbsql.cursor() if dbsql is not None else None
+
+    try:
+        ans = run_query(f'SELECT COUNT(*) FROM {tablename}', method, cursor,
+                        sparkSession)
+        nrows = [row.asDict() for row in ans][0].popitem()[1]  # pyright: ignore
+        log.info(f'total_rows = {nrows}')
```
karan6181, line 287: not sure if we need to log this as `info` or `debug`. Would it help user when they see this information?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        columns = [row.asDict().popitem()[1] for row in ans]  # pyright: ignore
+        order_by = columns[0]
+        columns_str = ','.join(columns)
+        log.info(f'order by column {order_by}')
+    except Exception as e:
+        raise RuntimeError(
+            f'Error in get columns from {tablename}. Restart sparkSession and try again'
+        ) from e
+
+    if method == 'dbconnect' and sparkSession:
```
karan6181, line 305: `if method == 'dbconnect' and sparkSession is not None:` ?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+
+        args = get_args(signed, json_output_path, columns, cf_collect_type)
+
+        # Stopping the SparkSession to avoid spilling connection state into the subprocesses.
+        sparkSession.stop()
+
+        with ProcessPoolExecutor(max_workers=processes) as executor:
+            list(executor.map(download_starargs, args))
+
+    elif method == 'dbsql' and cursor:
```
karan6181, line 322: `elif method == 'dbsql' and cursor is not None:`?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        order_by = columns[0]
+        columns_str = ','.join(columns)
+        log.info(f'order by column {order_by}')
+    except Exception as e:
+        raise RuntimeError(
+            f'Error in get columns from {tablename}. Restart sparkSession and try again'
+        ) from e
+
+    if method == 'dbconnect' and sparkSession:
+        print('processes = ', processes)
```
karan6181, line 306: please replace `print` with `log`

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        'Download delta table from UC and convert to json to save local')
+    parser.add_argument('--delta_table_name',
+                        required=True,
+                        type=str,
+                        help='UC table <catalog>.<schema>.<table name>')
+    parser.add_argument('--json_output_path',
+                        required=True,
+                        type=str,
+                        help='Local path to save the converted json')
+    parser.add_argument('--DATABRICKS_HOST',
```
karan6181, line 416: Since this will be run in a mosaic cluster, the `DATABRICKS_HOST` and `DATABRICKS_TOKEN` would be set by the user. Do you think we should just get this from env variable as `os.environ.get('DATABRICKS_TOKEN')`?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+            compute_id = args.cluster_id  # "1115-130834-ms4m0yv" - valid 14.1.0
+            w = WorkspaceClient()
+            res = w.clusters.get(
+                cluster_id=compute_id)  # '0704-124501-tsc2fxq' - invalid 12.2.x
+            runtime_version = res.spark_version.split('-scala')[0].replace(
+                'x-snapshot', '0').replace('x', '0')
+            if version.parse(runtime_version) < version.parse(
+                    MINIMUM_DBR_VERSION):
+                raise RuntimeError(
+                    f'You need at least {MINIMUM_DBR_VERSION} to use Databricks-connect to read delta table for FT API but got {res.spark_version}'
```
karan6181, line 386: nit: line exceed 100 character I guess

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    parser.add_argument('--batch_size',
+                        required=False,
+                        type=int,
+                        default=1 << 30,
+                        help='row chunks to transmit a time to avoid OOM')
+    parser.add_argument('--processes',
+                        required=False,
+                        type=int,
+                        default=1,
+                        help='number of processes allowed to use')
```
karan6181, line 437: I was wondering if we should use the maximum cores `os. cpu_count()` as default instead of 1 for faster data processing.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    parser.add_argument(
+        '--cluster_id',
+        required=True,
+        type=str,
+        default=None,
+        help=
+        'Use serverless if not present. IMPORTANT! make sure cluster has runtime newer than 14.1.0, the databricks-connect client version'
+    )
+    parser.add_argument('--debug', type=bool, required=False, default=False)
```
karan6181, line 446: looks like you are utilizing this arg in the code. Do you have a plan to add some code when user enable `debug` mode?

setup.py
```diff
@@ -85,6 +85,9 @@
 
 extra_deps['databricks'] = [
     'mosaicml[databricks]>=0.17.1,<0.18',
+    'databricks-sql-connector>=3, <4',
```
karan6181, line 88:
```suggestion
    'databricks-sql-connector>=3,<4',
```

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+
+MINIMUM_DBR_VERSION = '14.0.0'
+
+log = logging.getLogger(__name__)
+
+Result = namedtuple(
+    'Result', ['url', 'row_count', 'compressed_size', 'uncompressed_size'
+              ])  # pyright: ignore
+
+cf_collect_type = 'arrow'  # optionally change to json if arrow fails
```
dakinggg, line 44: what is this comment about? is it a note for yourself? if it is intended to be a user facing arg, can we make an actual arg?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+
+# This is a monkey patch on top of the DB Connect package that allows
```
dakinggg, line 46: could you specify a little bit more how the monkeypatch differs from the code it is replacing? Is it a complete rewrite or just a few lines? If just a few lines, please demarcate them inside the function with comments.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+                  type: str = 'json') -> Tuple[List[Result], int, bool]:
+    query = self._plan.to_proto(self._session.client)  # pyright: ignore
+    return self._session.client.to_cf(query, type)  # pyright: ignore
+
+
+DataFrame.collect_cf = collect_as_cf  # pyright: ignore
+
+
+def iterative_combine_jsons(json_directory: str, output_file: str) -> None:
+    """Combine jsonl files in json_directory into one big jsonl file.
```
dakinggg, line 122: Should specify that this function does _not_ work for nested subdirectories (or update the function to use `os.walk` or something similar.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+
+    Args:
+        json_directory(str): directory containing the JSON files
+        output_file(str): output JSONL file
+    """
+    json_files = [f for f in os.listdir(json_directory) if f.endswith('.jsonl')]
+    with open(output_file, 'w') as outfile:
+        for file_name in json_files:
+            with open(os.path.join(json_directory, file_name), 'r') as infile:
```
dakinggg, line 131: this seems like it is written to merge many jsonl files into one jsonl file, but the docstring says it is intended to convert many json files into one jsonl files. Is the implementation wrong or the docstring wrong (or am I missing something)?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+    cursor: Optional[Cursor] = None,
+    spark: Optional[SparkSession] = None,
+    collect: bool = True
+) -> Optional[Union[List[Row], DataFrame, SparkDataFrame]]:
+    """Run SQL query via databricks-connect or databricks-sql.
+
+    Args:
+        q (str): sql query
+        method (str): select from dbsql and dbconnect
+        cursor (Cursor): connection.cursor
```
dakinggg, line 149: note in args description which ones are optional, and what the default results in

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+        spark (SparkSession): spark session
+        collect (bool): whether to get the underlying data from spark dataframe
+    """
+    if method not in ['dbsql', 'dbconnect']:
```
dakinggg, line 153: can make this error just the `else` if the `if/elif` below

scripts/data_prep/convert_delta_to_json.py
dakinggg, line 171: i don't think the return none should be here. this code is expected to never execute, right?

dakinggg: im guessing that would also change the return type to not be `Optional`

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+            ) from e
+    else:
+        method = 'dbconnect'
+        if not args.cluster_id:
+            session_id = str(uuid4())
+            sparkSession = DatabricksSession.builder.host(
+                args.DATABRICKS_HOST).token(args.DATABRICKS_TOKEN).header(
+                    'x-databricks-session-id', session_id).getOrCreate()
+        else:
+            # IMPORTANT: make sure cluster has runtime newer than 14.1.0, the databricks-connect client version.
```
dakinggg, line 376: the comment disagrees with the min DBR version constant at the top of the file, which is correct?

scripts/data_prep/convert_delta_to_json.py
dakinggg, line 416: heads up on these args, I believe we will need to get them from a databricks config file not env vars. cc @panchalhp-db knows more

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: it is not a user-facing argument. I removed this global var for clarity.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: It is a new feature, add-on to the existing package.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,451 @@
...
+from pyspark.sql.dataframe import DataFrame as SparkDataFrame
+from pyspark.sql.types import Row
+
+MINIMUM_DBR_VERSION = '14.0.0'
+
+log = logging.getLogger(__name__)
+
+Result = namedtuple(
+    'Result', ['url', 'row_count', 'compressed_size', 'uncompressed_size'
+              ])  # pyright: ignore
```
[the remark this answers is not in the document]
XiaohanZhangCMU: Got ""namedtuple" provides no types for tuple entries; use "NamedTuple" instead".

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: it is merging many jsonl files into one jsonl. docstring is updated.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: Why do we need to parse the json data?

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: Thanks. Changed all.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: changed to ``query``

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: removed the last line for clarity.

XiaohanZhangCMU: yes, it should never reach ``None``. Fixed.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: The multiprocessing functions need this iterator as a global.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: comment added.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: I'm afraid so. there has to be a global function that breaks the tuple args.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: ``debug`` is more appropriate.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: Good catch! I changed the one on the top to 14.0.0 by mistake. 14.1.0 is the minimum.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: @panchalhp-db can you suggest what changes are appropriate here? Is there any code I can reuse to get the vars from config file?

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: Good call.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: Good catch. ``debug`` is removed. This was from early development (used by myself).

scripts/data_prep/convert_delta_to_json.py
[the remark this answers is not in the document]
XiaohanZhangCMU: I got this code snippet from Martin's patch. Is it safe if I just replace it with NamedTuple?

scripts/data_prep/convert_delta_to_json.py
panchalhp-db: Hey sorry, I completely missed these github notifications. We decided that we will make the config file available at the path pointed to by `DATABRICKS_CONFIG_FILE` environment variable inside the user workload containers. You should be able to parse it to get the host and the token. @jbloxham would know the details of what the file would look like but I think it would likely be similar to what your local `~/.databrickscfg` file looks like.

panchalhp-db: You should be able to use workspace client to get these creds eg:
```
from databricks.sdk import WorkspaceClient

w = WorkspaceClient()

host = w.config.host
token = w.config.token
```

panchalhp-db (approved):
LGTM!

scripts/data_prep/convert_delta_to_json.py
jbloxham: Yup, what Harsh said :)

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,496 @@
...
+        cursor.execute(query)
+        if collect:
+            return cursor.fetchall()
+    elif method == 'dbconnect':
+        if spark == None:
+            raise ValueError(f'sparkSession is required for dbconnect')
+        df = spark.sql(query)
+        if collect:
+            return df.collect()
+        return df
```
nancyhung, line 191: nit: there are a few instances of this `method == 'dbsql` v.s. `method == 'dbconnect'` in this file. A more Pythonic way to abstract this functionality is to create an abstract class that has these common methods and instantiate method-specific subclasses that inherit from the abstract class. The equivalent would be operating on an `ObjectStore` abstraction in [Composer](https://github.com/mosaicml/composer/blob/dev/composer/utils/object_store/object_store.py#L62C1-L62C28) but having specific `UCVolumeObjectStore` implementations

e.g.
```
class JsonToDeltaConverter(abc.ABC):
  def fetch_data():
    pass
  def run_query():
    pass
 
class DbConnectJsonToDeltaConverter(JsonToDeltaConverter):
  ...
class DbSqlJsonToDeltaConverter(JsonToDeltaConverter):
  ...
```

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,496 @@
...
+            result += [
+                Result(b.url, b.row_count, b.compressed_size,
+                       b.uncompressed_size) for b in batch.results
+            ]
+            row_count += sum(result.row_count for result in batch.results)
+            is_overflow |= batch.truncated
+    return result, row_count, is_overflow
+
+
+SparkConnectClient.to_cf = to_cf  # pyright: ignore
```
nancyhung, line 117: I briefly skimmed sourcegraph and couldn't find equivalent references to this way of configuring `SparkConnectClient`. QQ: is there a precedent you're following here? Granted, most of the codebase is in Scala 😅

nancyhung (approved):
A few non-blocking comments! Could you link your reference DB notebook you're iterating with on the PR description for reference?

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: The conversation below tracks this issue. tldr, I should use workspsace client instead of env vars. 
https://github.com/mosaicml/llm-foundry/pull/773#discussion_r1443498582

XiaohanZhangCMU: Thank you @panchalhp-db @jbloxham . Can you take another look at the latest change?

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: Add the request to backlog ticket [JIRA](https://databricks.atlassian.net/issues/?jql=project+%3D+%22STR%22+ORDER+BY+created+DESC&atlOrigin=eyJpIjoiZWE2ZjNiMTAxMjg1NDdjMGFlNTIzNWQ0MjgzODI3YmMiLCJwIjoiaiJ9)

scripts/data_prep/convert_delta_to_json.py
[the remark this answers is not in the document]
XiaohanZhangCMU: Hmm, It does not work switching to NamedTuple. Let's keep it for now.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,487 @@
...
+
+# ``collect_as_cf`` is an addon new feature monkey patch on top of the DB Connect package.
+# It allows the client to fetch the results in different formats from the server.
+# To be able to use the code make sure this module is not overriden by DB Connect classes.
+
+
+def to_cf(self: SparkConnectClient,
+          plan: pb2.Plan,
+          type: str = 'json') -> Tuple[List[Result], int, bool]:
+    """Executes plan object return as cloud fetch presigned URLS.
```
dakinggg, line 52: Not sure what this sentence is supposed to say, could you clarify?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,487 @@
...
+
+    # Add the request options
+    if type == 'json':
+        format = cloud_pb2.ResultOptions.CloudOptions.FORMAT_JSON
+    elif type == 'csv':
+        format = cloud_pb2.ResultOptions.CloudOptions.FORMAT_CSV
+    elif type == 'arrow':
+        format = cloud_pb2.ResultOptions.CloudOptions.FORMAT_ARROW
+    else:
+        raise Exception(
```
dakinggg, line 80:
```suggestion
        raise ValueError(
```

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,487 @@
...
+    Args:
+       plan (pb2.Plan): The plan object to be executed by spark.
+       type (str): The output format of the result, supported formats are 'json', 'csv', and 'arrow'.
+
+    Returns:
+       Tuple[List[Result], int, bool]: A tuple containing:
+           - A list of Result namedtuples, each containing a URL, row count, compressed size,
+             and uncompressed size of the part of the result.
+           - Total row count of all parts of the result.
+           - A boolean indicating whether the result is truncated or overflowed.
```
dakinggg, line 67: are truncated and overflowed synonymous? I can't quite tell from this description.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,487 @@
...
+
+        # Stopping the SparkSession to avoid spilling connection state into the subprocesses.
+        sparkSession.stop()
+
+        with ProcessPoolExecutor(max_workers=processes) as executor:
+            list(executor.map(download_starargs, args))
+
+    elif method == 'dbsql' and cursor is not None:
+        for start in range(0, nrows, batch_size):
+            print('start = ', start)
```
dakinggg, line 369: log instead of print

setup.py
```diff
@@ -84,7 +84,11 @@
 ]
 
 extra_deps['databricks'] = [
-    'mosaicml[databricks]>=0.17.2,<0.18',
+    'mosaicml[databricks]>=0.17.1,<0.18',
```
dakinggg, line 87: this was >=0.17.2 before, should we keep that min?

dakinggg (approved):
LGTM, didn't review thoroughly, relying on others.

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,487 @@
...
+from pyspark.sql.connect.client.reattach import \
+    ExecutePlanResponseReattachableIterator
+from pyspark.sql.connect.dataframe import DataFrame
+from pyspark.sql.dataframe import DataFrame as SparkDataFrame
+from pyspark.sql.types import Row
+
+MINIMUM_DBR_VERSION = '14.1.0'
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
```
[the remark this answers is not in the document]
XiaohanZhangCMU: removed the hardcoded.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: --> "Executes the query plans and return as presigned URLS for cloud fetch"

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: They are opposite. 
is_overflow  |= batch.untruncated.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: to_cf is a patch to the original SparkConnectClient.to_cf.

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: since this is a direct copy so we don't need to load and parse. let me know if you feel otherwise..

scripts/data_prep/convert_delta_to_json.py
dakinggg: it seems a bit weird to return one bool for that. Can you either separate the bools or explain a bit in the comment why they are treated the same?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,490 @@
...
+    os.makedirs(args.json_output_path, exist_ok=True)
+
+    log.info(f'Directory {args.json_output_path} created.')
+
+    method = 'dbsql'
+    dbsql = None
+    sparkSession = None
+
+    w = WorkspaceClient()
+    res = w.clusters.get(cluster_id=args.cluster_id)
```
nancyhung, line 404: FYI this won't work if the `cluster_id` is a serverless `cluster_id`, see [my notebook](https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/187185339621026/command/187185339621075). We can pass in `None` or a special `serverless` string to recognize a serverless cluster and skip this check. Let me know how you want to default this and I can update Mcloud / the SDK accordingly @XiaohanZhangCMU 

cc @panchalhp-db

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -0,0 +1,490 @@
...
+from packaging import version
+from pyspark.sql import SparkSession
+from pyspark.sql.connect.client.core import SparkConnectClient
+from pyspark.sql.connect.client.reattach import \
+    ExecutePlanResponseReattachableIterator
+from pyspark.sql.connect.dataframe import DataFrame
+from pyspark.sql.dataframe import DataFrame as SparkDataFrame
+from pyspark.sql.types import Row
+
+MINIMUM_DBR_VERSION = '14.1.0'
```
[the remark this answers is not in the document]
XiaohanZhangCMU: This is the minimum dbr version for dbconnect. Will add another one for sqlconnect and change their names.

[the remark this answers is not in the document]
XiaohanZhangCMU: Here are the minimum required dbr versions: 

MINIMUM_DB_CONNECT_DBR_VERSION = '14.1.0'
MINIMUM_SQL_CONNECT_DBR_VERSION = '12.2.0'

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: Can you explain it further by "treated the same"? And by "separated"? 

The boolean is returned but never used in the code. It's safe to ignore. 
 
signed, _, _ = df.collect_cf('arrow')

scripts/data_prep/convert_delta_to_json.py
XiaohanZhangCMU: @nancyhung good catch. Can you check the new logic? Let's use "serverless" as the magic word to skip the check.

PR #775, opened 2023-12-04
Enable flag to not pass PAD tokens in ffwd

bcui19:
This PR does two things:

1. Modifies the `attn_bias` function to always return the attention_mask.
2. Enables us to remove pad tokens before calling `.forward` on the ffwd network then re-add in the pad tokens.

Loss curves on a fully randomly initialized network:
image

We also get slightly higher throughput from this when there are PAD tokens in our dataset (and no degradation when compared to `main` with `attn_impl: triton`:
image

wandb: https://wandb.ai/mosaic-ml/padding_check?workspace=user-bcui


llmfoundry/models/layers/ffn_padding_utils.py
vchiley, line 1: can this be imported from `https://github.com/sz128/flash-attention/blob/3c0d49532cb3bc0a36b405f34590ff82cf582853/flash_attn/bert_padding.py`?

llmfoundry/models/layers/blocks.py
vchiley, line 1: Does passing `attention_mask` into attn fwd break anything?
I don't think it should, just asking.

vchiley: https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/layers/attention.py#L436

this gets triggered

vchiley: https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/layers/attention.py#L157
This is triggered in torch attn impl

vchiley:
> We also get slightly higher throughput from this when there are PAD tokens in our dataset:

Can you run main vs your branch with use_pad_tok_in_ffwd flag vs your branch without use_pad_tok_in_ffwd flag?

llmfoundry/models/layers/__init__.py
```diff
@@ -9,24 +9,14 @@
...
-    'FC_CLASS_REGISTRY',
-    'SharedEmbedding',
-    'FFN_CLASS_REGISTRY',
-    'build_ffn',
+    'scaled_multihead_dot_product_attention', 'flash_attn_fn',
+    'triton_flash_attn_fn', 'MultiheadAttention', 'MultiQueryAttention',
+    'attn_bias_shape', 'build_attn_bias', 'build_alibi_bias',
+    'ATTN_CLASS_REGISTRY', 'MPTMLP', 'MPTBlock', 'NORM_CLASS_REGISTRY',
+    'LPLayerNorm', 'FC_CLASS_REGISTRY', 'SharedEmbedding', 'FFN_CLASS_REGISTRY',
+    'build_ffn', 'unpad_input', 'pad_input'
```
dakinggg, line 21: add trailing comma back so formatting stays nice

dakinggg (commented):
Can you add a test that tests numerical equivalence of computation with and without the flag? might be off by a bit because of numerics, but lets see.

llmfoundry/models/layers/blocks.py
dakinggg: yeah, this is not ideal right? Any way we can avoid?

llmfoundry/models/layers/ffn_padding_utils.py
dakinggg: +1

llmfoundry/models/layers/blocks.py
bcui19: It's possible, I could just be checking the `attn_bias` for this value:`torch.finfo(attn_bias.dtype).min`

taken from here: https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/models/mpt/modeling_mpt.py#L418

@dakinggg would you prefer this?

dakinggg: I think that works? @vchiley ?

llmfoundry/models/layers/blocks.py
```diff
@@ -132,6 +135,22 @@ def forward(
...
+                    'Please install flash-attn==1.0.9 or flash-attn==2.3.2')
+            m, indices, _, _ = unpad_input(m, attention_mask)
         n = self.ffn(m)
+        if not self.use_pad_tok_in_ffwd:
+            try:
+                from flash_attn.bert_padding import pad_input  # type: ignore # yapf: disable # isort: skip
+            except:
+                raise RuntimeError(
+                    'Please install flash-attn==1.0.9 or flash-attn==2.3.2')
+            n = pad_input(n, indices, batch_size, seq_len)
```
vchiley, line 154: I'd put
```
try:
    from flash_attn.bert_padding import unpad_input, pad_input  # type: ignore # yapf: disable # isort: skip
except:
    unpad_input, pad_input = None, None
```
at the top of the file

then have this as
```suggestion
        batch_size = m.size(0)
        seq_len = m.size(1)
        if not self.use_pad_tok_in_ffwd:
            if unpad_input is None:
                raise RuntimeError(
                    'Please install flash-attn==1.0.9 or flash-attn==2.3.2')
            m, indices, _, _ = unpad_input(m, attention_mask)
        n = self.ffn(m)
        if not self.use_pad_tok_in_ffwd:
            n = pad_input(n, indices, batch_size, seq_len)
```
(removed the second runtime error cuz if the first one passed, the second one should be fine)

llmfoundry/models/layers/blocks.py
dakinggg: Ok, I leave it up to you which method you prefer.

vchiley: its either we use BW to find `attn_bias.min()` or we use BW to add a mask to `attn_bias`
these use approximately the same amount of BW... 🤷‍♂️ 

I'd say leave as is

bcui19: With `attn_impl: triton` we have the same loss and same throughput:

image
image

llmfoundry/models/layers/blocks.py
```diff
@@ -132,6 +140,15 @@ def forward(
         m = x
         if self.norm_2 is not None:
             m = self.norm_2(x)
+        batch_size = m.size(0)
+        seq_len = m.size(1)
```
mvpatel2000, line 144:
```suggestion
        batch_size, seq_len= m.size[:2]
```

mvpatel2000 (commented):
Minor nits

llmfoundry/models/layers/blocks.py
dakinggg: I think this is ok. Its an optional flag that may or may not help performance, similar to many other configuration options.

dakinggg: Could you move the summary of performance differences into the PR description please? Also what setting did you originally see a throughput win on?

llmfoundry/models/layers/blocks.py
```diff
@@ -132,6 +140,15 @@ def forward(
...
+        if not self.use_pad_tok_in_ffwd:
+            if unpad_input is None:
+                raise RuntimeError(
+                    'Please install flash-attn==1.0.9 or flash-attn==2.3.2')
```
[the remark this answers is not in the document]
dakinggg: I think this should actually be in `configuration_mpt.py`. There is a big ol `validate_config` function there, which is where this kind of check belongs.

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -131,6 +132,7 @@ def __init__(
                 See llmfoundry.models.utils.param_init_fns.py for info on other param init config options
             fc_type (str): choose fc layer implementation. Options: torch and te. te layers support fp8 when using H100 GPUs.
             tie_word_embeddings (bool): Whether to tie the input embedding and output layers.
+            use_pad_tok_in_ffwd (bool): Whether to forward the pad token in the feedforwards networks.
```
dakinggg, line 135:
```suggestion
            use_pad_tok_in_ffwd (bool): Whether to forward the pad token in the feedforward networks.
```

llmfoundry/models/layers/blocks.py
dakinggg: oh wait, sorry, the attention mask thing makes it slower when you're _not_ using the new option? @vchiley wdyt about this?

bcui19: Yeah but only with `attn_impl: torch,` with `attn_impl: triton` it's faster

vchiley: i mean no one should be using the torch path
thats more for testing 

so long as the triton path isn't effected, we're good.

llmfoundry/models/layers/blocks.py
```diff
@@ -53,6 +58,7 @@ def __init__(
         fc_type: str = 'torch',
         device: Optional[str] = None,
         no_bias: bool = False,
+        use_pad_tok_in_ffn: bool = True,
```
vchiley, line 61:
```suggestion
        use_pad_tok_in_ffn: bool = False,
```
default False

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         init_config: Dict = init_config_defaults,
         fc_type: str = 'torch',
         tie_word_embeddings: bool = True,
+        use_pad_tok_in_ffn: bool = True,
```
vchiley, line 63:
```suggestion
        use_pad_tok_in_ffn: bool = False,
```

default False

tests/models/test_model.py
```diff
@@ -831,6 +832,43 @@ def test_forward_with_padding(attention_impl: str, pos_emb_config: dict,
                 batched_output[1, :],
                 atol=1e-6 if attention_impl == 'torch' else 1e-8)
 
+        try:
```
vchiley, line 835: having try except block in test seems counter-intuitive 🤔 

you probably cant import it on CPU tests. maybe gate by "if not running on gpu"?

tests/models/test_model.py
```diff
@@ -731,6 +731,7 @@ def test_forward_with_padding(attention_impl: str, pos_emb_config: dict,
             'init_std': 0.02,
         },
         tie_word_embeddings=tie_word_embeddings,
+        use_pad_tok_in_ffn=True,
```
vchiley, line 734: Isn't this the default?
Why do you need to explicitly set it?

dakinggg (approved):
Thanks!

PR #779, opened 2023-12-05
Add safe_load option to restrict HF dataset downloads to allowed file types

irenedea:


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -123,105 +124,110 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
+                    +
+                    f'has {full_dataset_size} samples, but your minimum batch size '
+                    +
+                    f'is {minimum_dataset_size} because you are running on {world_size} gpus and '
+                    +
+                    f'your per device batch size is {dataloader_batch_size}. Please increase the number '
+                    +
+                    f'of samples in your dataset to at least {minimum_dataset_size}.'
+                )
+
```
irenedea, line 184: @dakinggg Is this incorrect to have this check for both streaming and hf datasets? Originally, this check was only running for hf datasets.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -325,8 +325,8 @@ def get_preprocessing_fn_from_str(
         return preprocessing_fn
 
     def build_from_hf(
-        self, cfg: DictConfig, max_seq_len: int,
-        tokenizer: PreTrainedTokenizerBase
+        self, dataset_name_or_path: str, split: str, max_seq_len: int,
+        tokenizer: PreTrainedTokenizerBase, proto_preprocessing_fn: Union[dict, DictConfig, str], hf_kwargs: dict[str, Any]
```
irenedea, line 329: @dakinggg I want to do some refactors to make args more explicit, rather than passing around cfgs, but I'm guess this is an api breaking change? Any suggestions on how to proceed? 😬

llmfoundry/data/finetuning/dataloader.py
dakinggg: i didn't think too hard about which checks could be applied to streaming datasets safely. totally possible this should be applied in both places.

llmfoundry/data/finetuning/tasks.py
dakinggg: I am ok with that type of breaking...its so important. And it doesn't break usage at the yaml level, which is the more important API to keep intact.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -123,105 +124,110 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
-            num_workers=cfg.num_workers,
-            pin_memory=cfg.get('pin_memory', True),
-            prefetch_factor=cfg.get('prefetch_factor', 2),
-            persistent_workers=cfg.get('persistent_workers', True),
-            timeout=cfg.get('timeout', 0),
-        )
+    sampler = None
 
+    # Build streaming or HF dataset.
+    if cfg.dataset.get('remote') is not None:
```
irenedea, line 130: @dakinggg should this be `cfg.dataset.get('local') is not None` since `remote` is a optional path and the data may exist locally?

llmfoundry/data/finetuning/dataloader.py
irenedea: @karan6181 Do you have any thoughts here? Should we be doing this check for streaming dataset as well? I haven't considered data duplication that might happen automatically in streaming, which might make this check incorrect

dakinggg: one thing is that `len(dataset)` means something different for a streaming dataset (its per device) than a map style hf dataset (full dataset).

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -199,28 +159,92 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
-            prefetch_factor=cfg.get('prefetch_factor', 2),
-            persistent_workers=cfg.get('persistent_workers', True),
-            timeout=cfg.get('timeout', 0),
-        )
+            shuffle=cfg.dataset.shuffle)
+    else:
+        # Build streaming dataset
+        dataset = _build_streaming_dataset(cfg.dataset, tokenizer, device_batch_size)
+    
+    assert dataset is not None
```
karan6181, line 169: nit: Please avoid `assert` statement in the source code. It's okay to put it in the tests code. Instead, raise an Exception with the descriptive error message.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -199,28 +159,92 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
+        persistent_workers=cfg.get('persistent_workers', True),
+        timeout=cfg.get('timeout', 0),
+    )
 
     token_counting_func = get_tokens_per_batch_func(
         pad_token_id=tokenizer.pad_token_id)
 
     return DataSpec(dataloader=dl, get_num_tokens_in_batch=token_counting_func)
 
+def _build_preprocessing_function(dataset_name_or_path: str, proto_preprocessing_fn: Optional[Union[dict, DictConfig, str]]):
```
karan6181, line 188: nit: method comments are missing. Please also check other methods.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -199,28 +159,92 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
+    if isinstance(proto_preprocessing_fn, dict) or isinstance(
+            proto_preprocessing_fn, DictConfig):
+        return dataset_constructor.get_preprocessing_fn_from_dict(
+            proto_preprocessing_fn)
+    return dataset_constructor.get_preprocessing_fn_from_str(
+            proto_preprocessing_fn, dataset_name_or_path)
+
+def _build_hf_dataset(hf_name: str, split: Optional[str], safe_load: bool, max_seq_len: int, 
+                      preprocessing_fn: Optional[Callable[[dict[str, Any]], 
+                                                          dict[str,str]]], hf_kwargs: dict) -> Union[hf_datasets.DatasetDict, hf_datasets.Dataset, hf_datasets.IterableDatasetDict, hf_datasets.IterableDataset]:
```
karan6181, line 198: nit: split this line under 100 character.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -199,28 +159,92 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
+        split=split,
+        max_seq_len=cfg.dataset.max_seq_len,
+        preprocessing_fn=preprocessing_fn,
+        tokenizer=tokenizer,
+        hf_kwargs=hf_kwargs
+    )
+
+
+def _build_streaming_dataset(dataset_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase, device_batch_size: int) -> StreamingFinetuningDataset:
+    return dataset_constructor.build_from_streaming(
```
karan6181, line 224: FYI: Streaming dataset added `allow_unsafe_types` flag as part of PR https://github.com/mosaicml/streaming/pull/519. Soon, streaming dataset will have a new patch release version with this change. @knighton created the PR https://github.com/mosaicml/llm-foundry/pull/785 to add this flag.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -275,12 +299,43 @@ def _validate_config(dataset_cfg: DictConfig) -> None:
             'dataset, but both were None.'
         )
 
+def _downloaded_datasets_dir() -> str:
```
karan6181, line 302: looks like this method get the directory path rather than downloading the dataset? Maybe a different method name?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -275,12 +299,43 @@ def _validate_config(dataset_cfg: DictConfig) -> None:
...
+    return os.path.join(
+        os.path.dirname(
+            os.path.dirname(os.path.dirname(os.path.realpath(__file__)))),
+        'downloaded_finetuning')
+
+def _safe_download_hf_dataset(hf_name: str, token: Union[bool, str, None]) -> str:
+    local_dataset_dir = os.path.join(_downloaded_datasets_dir(), hf_name)
+    if len(os.listdir(local_dataset_dir)) > 0:
```
karan6181, line 310: Wondering, is `> 0` a right check? is there a case where the process still downloading the file in the background where the downloading file named as a temporary file and once it is downloaded, it's get replaced with the actual file name. I have seen this with boto3 and OCI and not sure about how HF managed the file during downloading?

llmfoundry/data/finetuning/dataloader.py
irenedea: Will do! still wip, will make sure we have some good code documentation once it's ready for review!

irenedea:
closing in favor of https://github.com/mosaicml/llm-foundry/pull/798

llmfoundry/data/finetuning/dataloader.py
irenedea: I think HF will not download in the background. When the call to snapshot download is done, it will be done downloading.

llmfoundry/data/finetuning/dataloader.py
irenedea: I think we do have some asserts like this one for pyright. dataset is assigned for each conditional branch, but pyright is unable to infer that the type is not None

PR #783, opened 2023-12-07
Auto packing fixes

irenedea:
Fixes:
1. use a temporary local directory when profiling with streaming, handles Reused local directory error
2. handle tiny datasets that are smaller than device batch size
3. handle small max_seq_len that caused max packing ratio to be less than min packing ratio

### DIFF
--- a/tests/data/test_packing.py
+++ b/tests/data/test_packing.py
@@ -174,7 +174,7 @@ def test_auto_packing_with_streaming_dataloader(tmp_path: Path):
         # Gets copied per worker and we cannot check the waste for child processes.
         'num_workers': 0,
         'pin_memory': False,
-        'prefetch_factor': None if using_torch_2() else 2,
+        'prefetch_factor': None,
         'persistent_workers': False,
         'timeout': 0,
     })
--- a/tests/data/test_packing.py
+++ b/tests/data/test_packing.py
@@ -5,7 +5,6 @@ from pathlib import Path
 from typing import Any, Dict, List
 from unittest.mock import Mock, patch
 
-import numpy as np
 import pytest
 import torch
 from composer.utils import dist, reproducibility, using_torch_2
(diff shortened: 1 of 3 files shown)


tests/data/test_packing.py
```diff
@@ -149,6 +151,44 @@ def patched_packing_ratio(*args: Any, **kwargs: Any):
...
 
 
+@patch('llmfoundry.data.finetuning.dataloader.auto_packing_ratio',
+       patched_packing_ratio)
+def test_auto_packing_with_streaming_dataloader(tmp_path: Path):
+    columns = {'prompt': 'str', 'response': 'str'}
+    tokenizer = build_tokenizer('gpt2', {})
+    remote_dir = str(tmp_path / 'remote')
+    local_dir = str(tmp_path / 'local')
+    with MDSWriter(out=remote_dir, columns=columns, compression=None) as out:
```
dakinggg, line 161: should we use the existing helper for constructing tiny ift data and just add a streaming option to it? nit, take it or leave it

dakinggg (approved):

tests/data/test_packing.py
```diff
@@ -149,6 +151,44 @@ def patched_packing_ratio(*args: Any, **kwargs: Any):
...
+        # Need to test with 0 num_workers because the packing collator object
+        # Gets copied per worker and we cannot check the waste for child processes.
+        'num_workers': 0,
+        'pin_memory': False,
+        'prefetch_factor': None if using_torch_2() else 2,
+        'persistent_workers': False,
+        'timeout': 0,
+    })
+
+    loader = build_finetuning_dataloader(cfg, tokenizer,
```
dakinggg, line 182: I removed `using_torch_2` in another PR, as I also removed support for torch version pre torch 2. so everything is torch 2 now.

irenedea: ah, cool!

PR #784, opened 2023-12-07
Chat fomating template changes

rajammanabrolu:
Changes to the jinja template for chat formatting to use the `add_generation_prompt` arg which is useful for formatting prompts at inference (ends with assistant\n) and other changes around using defined system prompts vs default ones. Wrote tests for all.

### DIFF
--- a/.github/mcp/mcp_pytest.py
+++ b/.github/mcp/mcp_pytest.py
@@ -114,6 +114,16 @@ if __name__ == '__main__':
         integrations=[git_integration],
         command=command,
         scheduling={'max_duration': args.timeout / 60 / 60},
+        env_variables=[
+            {
+                'key': 'MOSAICML_PLATFORM',
+                'value': 'False',
+            },
+            {
+                'key': 'PYTHONUNBUFFERED',
+                'value': '1',
+            },
+        ],
     )
 
     # Create run


llmfoundry/tokenizers/tiktoken.py
```diff
@@ -173,12 +173,30 @@ def default_chat_template(self):
...
+            "{{ '<|im_start|>system\n' + system_message.strip() + '\n'}}"
+            '{% endif %}'
+            "{{ '<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' }}"
+            '{% else %}'
+            "{{ '\n' + '<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' }}"
+            '{% endif %}'
+            '{% if (add_generation_prompt == true) %}'
+            "{{ '\n' + '<|im_start|>' + 'assistant' + '\n' }}"
+            "{% elif (message['role'] == 'assistant') %}"
+            '{{ eos_token }}'
```
samhavens, line 198: Is this what other tokenizer's chat templates do? It seems fine, since at generation time we are going to treat `<|im_end|>` as a stop token. But if it's non-standard maybe we should make a note about it.

samhavens: I was wondering if adding eos_token was standard. Just looked at llama2's, like you said, they do something similar

PR #786, opened 2023-12-07
Support Remote and HF promptfiles in hf_generate script

samhavens:
Small QoL improvement for testing generation. Uses composer's `get_file` to support remote prompt files, and adds a bit of syntax for pointing to huggingface hub datasets as well.

`-p file::/local/path` was already supported. This adds `-p file::s3://remote/path` and `-p dataset::mosaicml/some-hub-dataset`. For HF datasets, it defaults to looking for a column named `prompt` but will use any string passed as the `prompt_delimiter` as the column name (kind of abuses the API, but it felt understandable)

@dakinggg if this is out of scope let me know and I can close this out. Most workloads like this end up going to inference... but I can see it being useful both internally and for customers.

EDIT: updated the syntax so remote files are just `s3://` and HF datasets are `hf://`

example invocations:

```
python hf_generate.py -n mosaicml/mpt-7b -p hf://mosaicml/some-prompts --model_dtype bf16   --max_batch_size 8
```

and 

```
python hf_generate.py -n mosaicml/mpt-7b -p s3://our-bucket/path/prompts.txt --prompt-delimiter $'\n' --max_batch_size 8
```


dakinggg (commented):
The functionality seems good and I'd be happy to merge wrt that. I don't like the invention of a new URI format here tho. We use the standard like `s3://` in other places, and making it different here is pretty confusing.

To not break existing functionality, I suggest you leave the `file::` thing for local files, and then you just use `s3://` etc for remote files. `hf://` is the canonical prefix for huggingface things.

llmfoundry/utils/prompt_files.py
```diff
@@ -56,3 +68,73 @@ def load_prompts_from_file(prompt_path: str,
...
+        List of prompt string(s)
+    """
+    backend, _, _ = parse_uri(prompt_path)
+    if backend in ['', None]:
+        raise ValueError(
+            f'prompt_path_str must start with s3:// etc if using object storage'
+        )
+
+    local_path = prompt_path.split('/')[-1]
+    get_file(path=prompt_path, destination=local_path, overwrite=True)
```
dakinggg, line 93: you probably want to make `local_path` a tmp location?

llmfoundry/utils/prompt_files.py
```diff
@@ -56,3 +68,73 @@ def load_prompts_from_file(prompt_path: str,
...
+    with open(local_path, 'r') as f:
+        prompt_string = f.read()
+
+    if prompt_delimiter is None:
+        return [prompt_string]
+    return [i for i in prompt_string.split(prompt_delimiter) if i]
+
+
+def load_prompts_from_dataset(dataset_path: str,
+                              prompt_delimiter: Optional[str] = None
```
dakinggg, line 104: I'd prefer you just add a new argument to the script for the dataset column name instead of overloading the delimiter param

PR #791, opened 2023-12-08
Update gauntlet v0.2 to reflect results of calibration

bmosaicml:
In order to filter out noisy tasks and set appropriate number of few shot per task, we used the following methodology:

1. Evaluate each benchmark on several 3B models trained for [20x, 50x, 100x, 250x, 500x] chinchilla token duration for  num_fewshot=[0, 1, 3, 5, 10]
2. Determine which num_fewshot settings the models have monotonically increasing performance. If multiple num_fewshot have monotonically increasing performance we broke ties by choosing the industry standard if there is one, or `3`.
3. For benchmarks which were scoring 0 (and therefore not monotonic), we excluded them unless they are known to show good scaling at larger model scales. These tasks will average in as 0 in expectation, but they will introduce slight variance.

Later versions of this experiment will estimate the exact confidence intervals of the gauntlet sub-scores + overall average.

image

[Google sheet](https://docs.google.com/spreadsheets/d/1rkM6gBc2yX1qYbRjuVlhR9CFr2gOsKNENZgW71X5LnM/edit#gid=0)

### DIFF
--- a/scripts/eval/yamls/eval_gauntlet_v0.2.yaml
+++ b/scripts/eval/yamls/eval_gauntlet_v0.2.yaml
@@ -1,0 +1,127 @@
+eval_gauntlet:
+  weighting: EQUAL
+  subtract_random_baseline: true
+  rescale_accuracy: true
+  averages:
+    core_average:
+    - world_knowledge
+    - commonsense_reasoning
+    - language_understanding
+    - symbolic_problem_solving
+    - reading_comprehension
+  categories:
+  - name: world_knowledge
+    benchmarks:
+    - name: jeopardy
+      num_fewshot: 3
+      random_baseline: 0
+    - name: bigbench_qa_wikidata
+      num_fewshot: 3
+      random_baseline: 0
+    - name: arc_easy
+      num_fewshot: 3
+      random_baseline: 0.25
+    - name: arc_challenge
+      num_fewshot: 3
+      random_baseline: 0.25
+    - name: mmlu
+      num_fewshot: 5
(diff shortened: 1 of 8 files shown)


scripts/eval/yamls/eval_gauntlet_v0.2.yaml
```diff
@@ -0,0 +1,127 @@
...
+      random_baseline: 0.0
+    - name: simple_arithmetic_nospaces
+      num_fewshot: 5
+      random_baseline: 0.0
+    # - name: math_qa
+    #   num_fewshot: 10
+    #   random_baseline: 0.25
+    # - name: logi_qa
+    #   num_fewshot: 10
+    #   random_baseline: 0.25
```
abhi-mosaic, line 92: should we comment or should we delete?

abhi-mosaic (approved):
small comment, but otherwise looks good to me! I spot checked the most important ICLs and the k-shots look like what was discussed.

scripts/eval/yamls/eval_gauntlet_v0.2.yaml
bmosaicml: I'll delete

dakinggg (approved):
LGTM from my perspective. Please wait for whoever you want to take a look from research.

tbarton16 (approved):
LGTM

sashaDoubov (approved):
LGTM 🚀

mansheej (approved):

PR #792, opened 2023-12-08
Fix utf-8 decode errors in tiktoken wrapper

dakinggg:


llmfoundry/tokenizers/tiktoken.py
```diff
@@ -61,6 +61,7 @@ def __init__(self,
                  eos_token: Optional[str] = '<|endoftext|>',
                  bos_token: Optional[str] = '<|endoftext|>',
                  pad_token: Optional[str] = None,
+                 errors: str = 'replace',
```
irenedea, line 64: nit: rename to `decode_errors`

irenedea: Any reason to make this non configurable for now and directly do `decode('utf-8', errors='replace')`? Generally thinking it might be helpful to opt for less configurability with apis--less room for user error and what not

dakinggg: I'd like to leave for consistency because this is actually just copied from https://github.com/huggingface/transformers/blob/80377eb018c077dba434bc8e7912bcaed3a64d09/src/transformers/models/gpt2/tokenization_gpt2.py#L136. not a strong opinion though, wdyt?

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -78,6 +79,9 @@ def __init__(self,
             eos_token (Optional[str], optional): The eos token. Defaults to '<|endoftext|>'.
             bos_token (Optional[str], optional): The bos token. Defaults to '<|endoftext|>'.
             pad_token (Optional[str], optional): The pad token. Defaults to None.
+            errors (`str`, *optional*): Paradigm to follow when decoding bytes to UTF-8. See
```
irenedea, line 82:
```suggestion
            errors (str, optional): Paradigm to follow when decoding bytes to UTF-8. See
```

llmfoundry/tokenizers/tiktoken.py
irenedea: ahh got it, don't feel too strongly about it. I'm okay to leave it

irenedea (approved):

dakinggg:
Force merged while GPU CI is briefly down. This change does not touch anything GPU related.

PR #794, opened 2023-12-09
update openai wrapper to work with tiktoken interface and newest openai version

bmosaicml:
Some changes to the OpenAI tokenizer (tiktoken) made the previously working e2e OpenAI eval wrapper no longer work with the eval script.

Some changes to the newest version of OpenAI API include deprecating the  `openai.ChatCompletion.create` method in favor of first constructing a client and calling `client.chat.completion.create`. Additionally, return results are no longer dictionaries , but instead `openai.types.Completion` objects

Tested with `composer eval/eval.py eval/yamls/openai_eval.yaml`

Ran on all of lm_tasks_v0.2: `openai-eval-O5OK4O`


dakinggg (approved):
Thank you! Could you please add a note to the yaml noting which tasks are supported with which models? (some are not supported because we can't get logits, right?)

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
```diff
@@ -30,20 +31,23 @@ class OpenAIEvalInterface(InferenceAPIEvalWrapper):
...
                 conda_channel='conda-forge') from e
-        openai.api_key = os.getenv('OPENAI_API_KEY')
+        self.client = openai.OpenAI()
         self.model_name = model_cfg['version']
 
     def generate_completion(self, prompt: str, num_tokens: int):
         raise NotImplementedError()
 
-    def process_result(self, completion: Optional[dict]):
+    def process_result(self, completion):  # pyright: ignore
```
maxisawesome, line 50: I tried to import the ChatCompletion object from openai up on line 38 but it didn't like that. I assume it looks at the typing before running the code and we only import openai after running init

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
```diff
@@ -52,26 +56,30 @@ def get_next_token_logit_tensor(self, prompt: str, num_tokens: int = 1):
...
+                        e._message):  # pyright: ignore
                     raise e
-                sleep(60)
+                delay *= 2 * (1 + random.random())
+                sleep(delay)
                 continue
-            except Exception:
+            except Exception as e:
+                print(f'Found Exception: {e}')
+                # TODO: Why continue on unspecified Exception?
```
maxisawesome, line 82: Should we just error here?

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
```diff
@@ -80,23 +88,23 @@ class OpenAIChatAPIEvalWrapper(OpenAIEvalInterface):
...
+                    model_cfg.get('sytsem_role_prompt',
+                                  'Please complete the following text: ')
+            }, {
                 'role': 'user',
                 'content': prompt
             }],
             max_tokens=num_tokens,
             temperature=0.0)
 
+    # TODO: Do we still need retokenize, rebatch, and eval_forward?
```
maxisawesome, line 107: I haven't tested this all the way through and did not change any of these methods. Are these still necessary? @bmosaicml

maxisawesome: Well, I should say that I did run jeopardy_all w/ the new wrapper and it was successful and had good numbers, but I haven't manually inspected the inputs/outputs.

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
hanlint: I think you can do upfront:
```
If TYPE_CHECKING:
    import openai
    from openai import ...
```

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
hanlint: Yeah this catch all doesn't make sense to me.. @bmosaicml what was the motivation?

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
bmosaicml: Yes I double checked, chat API still strips the sentence-initial space

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
bmosaicml: Removed it to make it more specific!

maxisawesome:
with the changes, LGTM 👍

maxisawesome (approved):

dakinggg (approved):
Approving to unblock the `setup.py` change. Deferring the full review to Max.

PR #795, opened 2023-12-09
Adding a fix for Cross Entropy Loss for long sequence lengths.

ShashankMosaicML:
For longer sequence lengths (>16k), we see that the current cross entropy loss implementation throws an illegal memory access error. This PR fixes this by reverting to an older version of cross entropy loss that works for longer sequence lengths as well.

WandB link for the experiments: https://wandb.ai/mosaic-ml/longcont-CE-debug

We see the same convergence, mfu, and memory usage before and after the fix
Screenshot 2023-12-11 at 12 26 28 PM

Screenshot 2023-12-11 at 12 26 13 PM

Screenshot 2023-12-11 at 12 27 01 PM


llmfoundry/models/layers/cross_entropy_loss.py
```diff
@@ -0,0 +1,159 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copied from https://github.com/Dao-AILab/flash-attention/blob/v1.0.9/flash_attn/losses/cross_entropy.py
```
vchiley, line 4: Should we use latest commit before triton refactor:
https://github.com/Dao-AILab/flash-attention/blob/f1a73d074002226c42ce65a1df170ecff9f022c0/flash_attn/losses/cross_entropy.py

Also use [permanent link](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files)
for v1.0.9 link is: https://github.com/Dao-AILab/flash-attention/blob/6d48e14a6c2f551db96f0badc658a6279a929df3/flash_attn/losses/cross_entropy.py
for latest commit before triton: https://github.com/Dao-AILab/flash-attention/blob/f1a73d074002226c42ce65a1df170ecff9f022c0/flash_attn/losses/cross_entropy.py

ShashankMosaicML: Done.

vchiley (approved):

dakinggg (approved):

PR #796, opened 2023-12-09
Enable GLU FFN type

vchiley:
1. Enable GLU FFN
2. enable user to set `ffn_hidden_size` directly.

### DIFF
--- a/llmfoundry/models/layers/ffn.py
+++ b/llmfoundry/models/layers/ffn.py
@@ -19,7 +19,7 @@ except:
 
 log = logging.getLogger(__name__)
 
-ignore_expr_log_msg = '`expansion_ratio` (=%d) ignored when `ffn_hidden_size` (=%d) is specified.'
+_IGNORE_EXPR_LOG_MSG = '`expansion_ratio` (=%d) ignored when `ffn_hidden_size` (=%d) is specified.'
 
 
 class MPTMLP(nn.Module):
--- a/llmfoundry/models/layers/ffn.py
+++ b/llmfoundry/models/layers/ffn.py
@@ -4,8 +4,7 @@
 """MPT Blocks used for the MPT Model."""
 
 import logging
-from copy import deepcopy
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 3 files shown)


llmfoundry/models/layers/ffn.py
```diff
@@ -15,6 +17,10 @@
 except:
     te = None
 
+log = logging.getLogger(__name__)
+
+ignore_expr_log_msg = '`expansion_ratio` (=%d) ignored when `ffn_hidden_size` (=%d) is specified.'
```
dakinggg, line 22:
```suggestion
_IGNORE_EXPR_LOG_MSG = '`expansion_ratio` (=%d) ignored when `ffn_hidden_size` (=%d) is specified.'
```

llmfoundry/models/layers/ffn.py
```diff
@@ -23,34 +29,58 @@ def __init__(
...
         device: Optional[str] = None,
         bias: bool = True,
     ):
         super().__init__()
-        fc_kwargs: dict[str, Any] = {
+        if ffn_hidden_size is not None:
+            log.info(ignore_expr_log_msg % (expansion_ratio, ffn_hidden_size))
+        else:
+            ffn_hidden_size = d_model * expansion_ratio
+            if ffn_hidden_size != int(ffn_hidden_size):
```
dakinggg, line 41: does this imply that the type of `expansion_ratio` is wrong and it should really be a `float`? (I think yes)

llmfoundry/models/layers/ffn.py
```diff
@@ -23,34 +29,58 @@ def __init__(
...
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         return self.down_proj(self.act(self.up_proj(x)))
 
 
+class MPTGeGLU(MPTMLP):
+
+    def __init__(self, **kwargs: Any):
+        super().__init__(**kwargs)
+        self.gate = deepcopy(self.up_proj)
```
dakinggg, line 74: Any reason you're not just constructing a new layer here and instead doing a deepcopy? I find it a bit unintuitive with the deepcopy, because I have to think about if there are any weird interactions resulting from copying.

llmfoundry/models/layers/ffn.py
```diff
@@ -62,27 +92,34 @@ def build_ffn(
...
             d_model=d_model,
             expansion_ratio=expansion_ratio,
             fc_type=fc_type,
+            ffn_hidden_size=ffn_hidden_size,
             device=device,
             bias=bias,
         )
     elif ffn_type == 'te_ln_mlp':
         assert te is not None
+        if ffn_hidden_size is not None:
```
dakinggg, line 116: this looks like the same check as in `MPTMLP`, except less complete because it doesn't have the int checking stuff. should this check just happen once in this function instead of duplicated?

dakinggg (approved):

llmfoundry/models/layers/ffn.py
vchiley: `expansion_ratio` can be type float or int 
eg if someone puts `expansion_ratio: 3.5`, then `ffn_hidden_size == int(ffn_hidden_size)` so we're good 
but if they put something ridiculous like 3.123, then it'll probably error.

dakinggg: ya, i'm saying the type signature on this constructor is wrong cause it says `expansion_ratio: int`

llmfoundry/models/layers/ffn.py
vchiley: Just trying make as few line changes as possible (for no good reason), but I can make a new layer

llmfoundry/models/layers/ffn.py
vchiley: exp ratio only gets passed into `MPTMLP` and not into the te layer.
removing exp ratio from MPTMLP signature is a breaking change (and I'd want to have this check happen inside of MPTMLP)

vchiley: I'll create a `_resolve_ffn_hidden_and_exp_ratio` fn

PR #798, opened 2023-12-12
Add safe_load option to restrict HF dataset downloads to allowed file types

irenedea:
## Manual Tests
test-safe-load-false-valid-hvr7NF - safe_load=False + valid dataset (runs successfully)
test-safe-load-LZNNUv - safe_load=True + valid dataset (runs successfully)
test-safe-load-squad-Lku1Q4 - safe_load=True + invalid dataset (fails with proper error)
test-no-safe-load-squad-jhdtOQ  - safe_load=False + invalid dataset (loads the dataset, fails bc data formatting reasons)

safe_load does not add any noticeable overhead. Both runs (valid dataset with safe_load=True and with safe_load=False) take 7 min.


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -148,48 +154,48 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
-                max_seq_len=cfg.dataset.max_seq_len,
-                tokenizer=tokenizer,
-            )
+            dataset_name_or_path = _download_remote_hf_dataset(
+                remote_path=dataset_name_or_path, split=split)
 
-        collate_fn, dataloader_batch_size = _build_collate_fn(
-            cfg, tokenizer, device_batch_size)
+        # Initialize preprocessing function.
+        proto_preprocessing_fn = cfg.dataset.get('preprocessing_fn')
```
irenedea, line 174: initially this logic was done as part of constructor.build_from_hf. However, as part of this refactor, I'd like for all functions except `build_finetuning_dataloader` to not accept any omegaconf types like DictConfig as arguments.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -122,8 +123,13 @@ def build_finetuning_dataloader(cfg: DictConfig,
     if tokenizer.pad_token is None:
         tokenizer.pad_token = tokenizer.eos_token
 
+    collate_fn, dataloader_batch_size = _build_collate_fn(
```
irenedea, line 126: todo (for another refactor, out of scope for this PR): make `_build_collate_fn` accept not a cfg but arguments directly. This may be hard because if the packing_ratio is "auto", `build_collate_rn` will call `build_finetuning_dataloader` under the hood. Probably will be a bigger refactor than just expanding the arguments of `_build_collate_fn`

llmfoundry/data/finetuning/tasks.py
```diff
@@ -368,9 +366,46 @@ def build_from_hf(
...
+                                f'safe_load is set to True. No data files with safe extensions {_SUPPORTED_EXTENSIONS} '
+                                + f'found for dataset {dataset_name}. ')
+                    # Set dataset_name to the downloaded location.
+                    dataset_name = local_dataset_dir
+
+                # dataset_name is a local dir path. Use the abspath to prevent confusion.
+                dataset_name = os.path.abspath(dataset_name)
+
+                # Ensure that the local dir contains only allowed file types.
+                dataset_files = [
```
irenedea, line 397: This check will always run even if dataset_name is a local folder. I think it's fine to have safe_load restrict to these file types even if it's a local folder, just wanted to point that out.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -368,9 +366,46 @@ def build_from_hf(
...
+                    f for _, _, files in os.walk(dataset_name) for f in files
+                ]
+                if not all(
+                        Path(f).suffix in _SUPPORTED_EXTENSIONS
+                        for f in dataset_files):
+                    raise ValueError(
+                        f'Dataset at local path {dataset_name} contains invalid file types. '
+                        + 'Allowed file types are: .csv, .jsonl, and .parquet.')
```
irenedea, line 405:
```suggestion
                        + f'Allowed file types are: $_SUPPORTED_EXTENSIONS')
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -13,7 +12,9 @@
 from transformers import PreTrainedTokenizerBase
 
 from llmfoundry.data.finetuning.collator import Seq2SeqFinetuningCollator
-from llmfoundry.data.finetuning.tasks import dataset_constructor
+from llmfoundry.data.finetuning.tasks import (_DOWNLOADED_FT_DATASETS_DIRPATH,
```
dakinggg, line 15: these probably aren't private if we have to import them here?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -148,48 +154,51 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
+            split = split.replace('-', '_')
+            dataset_name_or_path = _download_remote_hf_dataset(
+                remote_path=dataset_name_or_path, split=split)
 
-        collate_fn, dataloader_batch_size = _build_collate_fn(
-            cfg, tokenizer, device_batch_size)
+        # Get the preprocessing function.
+        proto_preprocessing_fn = cfg.dataset.get('preprocessing_fn')
+
+        preprocessing_fn = dataset_constructor.get_preprocessing_fn_from_dict(dict(proto_preprocessing_fn)) \
```
dakinggg, line 179: i'd just break this out into if/else instead of having this multiline `\` thing

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -148,48 +154,51 @@ def build_finetuning_dataloader(cfg: DictConfig,
...
+        # Ensure dataset is large enough.
         if cfg.drop_last:
             world_size = dist.get_world_size()
             minimum_dataset_size = world_size * dataloader_batch_size
             if hasattr(dataset, '__len__'):
                 full_dataset_size = len(dataset)
                 if full_dataset_size < minimum_dataset_size:
                     raise ValueError(
-                        f'Your dataset (name={cfg.dataset.hf_name}, split={cfg.dataset.split}) '
+                        f'Your dataset (name={cfg.dataset.hf_name}, split={split}) '
```
dakinggg, line 201:
```suggestion
                        f'Your dataset (name={dataset_name_or_path}, split={split}) '
```

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -290,38 +298,26 @@ def _build_hf_dataset_from_remote(
...
-        'downloaded_finetuning',
-        destination_split if destination_split != 'data' else 'data_not',
+        _DOWNLOADED_FT_DATASETS_DIRPATH,
+        split if split != 'data' else 'data_not',
     )
     os.makedirs(finetune_dir, exist_ok=True)
-    for extension in supported_extensions:
-        name = f'{cfg.dataset.hf_name.strip("/")}/{cfg.dataset.split}.{extension}'
+    for extension in _SUPPORTED_EXTENSIONS:
+        name = f'{remote_path.strip("/")}/{split}{extension}'
```
dakinggg, line 316: was the `.` deleted intentionally?

dakinggg: I assume yes, cause you deleted it everywhere, but how come?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -49,6 +50,14 @@ def preprocessing_fn(example: Dict) -> Dict[str, str]:
 
 _ALLOWED_RESPONSE_KEYS = {'response', 'completion'}
 _ALLOWED_PROMPT_KEYS = {'prompt'}
+_DOWNLOADED_FT_DATASETS_DIRPATH = os.path.join(
```
dakinggg, line 53: instead of chained `dirname`, can you use `os.pardir` (I know I wrote it originally, but lets fix it :P)?

tests/data/test_dataloader.py
```diff
@@ -441,12 +486,13 @@ def test_finetuning_dataloader_custom_split(tmp_path: pathlib.Path, split: str):
 
 
 def mock_get_file(path: str, destination: str, overwrite: bool = False):
-    make_tiny_ft_dataset(path=destination, size=16)
+    if Path(destination).suffix == '.jsonl':
```
dakinggg, line 489: is this supposed to silently do nothing if not jsonl?

dakinggg (commented):
generally lgtm. Could you provide some manual testing? Please test both using a custom file, and using both supported and unsupported datasets from the hub.

In the custom file and support hub dataset cases, check that it runs and also that startup time is not dramatically slower than previously.

For the unsupported hub dataset case, just make sure it errors in a reasonable way (i.e. doesn't hang or something)

tests/data/test_dataloader.py
irenedea: ah, I think it should ideally throw a FileNotFoundError

llmfoundry/data/finetuning/dataloader.py
irenedea: Yeah! I kept the `.` in _SUPPORTED_EXTENSIONS so `_SUPPORTED_EXTENSIONS=['.jsonl', '.csv', '.parquet']`
It just made it a little cleaner in my other code

llmfoundry/data/finetuning/dataloader.py
irenedea: ah, I wanted to actually print out the original name.

llmfoundry/data/finetuning/dataloader.py
dakinggg: "private" is all fake in python anyway. I'm not too opinionated on which version of this you choose.

tests/data/test_dataloader.py
```diff
@@ -306,6 +310,51 @@ def test_finetuning_dataloader(decoder_only_format: bool,
...
+        'persistent_workers': False,
+        'timeout': 0
+    })
+
+    tokenizer = build_tokenizer('gpt2', {})
+
+    with expectation:
+        _ = build_finetuning_dataloader(cfg, tokenizer, 1)
+
+    # If no raised errors, we should expect downlaoded files with only safe file types.
```
dakinggg, line 346:
```suggestion
    # If no raised errors, we should expect downloaded files with only safe file types.
```

dakinggg (approved):

PR #801, opened 2023-12-13
clean up resolve_ffn_hidden_and_exp_ratio

vchiley:
Hot fix cleanup to https://github.com/mosaicml/llm-foundry/pull/796. Fixes:
- Make resolve_ffn_hidden_and_exp_ratio not private (there are other places one could use the fn; eg [here](https://github.com/mosaicml/llm-foundry-private/pull/53#discussion_r1425649919))
- `resolve_ffn_hidden_and_exp_ratio` would return `expansion_ratio` without modifying it; given its already an input, this return is superfluous.

### DIFF
--- a/llmfoundry/models/layers/ffn.py
+++ b/llmfoundry/models/layers/ffn.py
@@ -19,7 +19,7 @@ except:
 log = logging.getLogger(__name__)
 
 
-def resolve_ffn_hidden_and_exp_ratio(
+def resolve_ffn_hidden_size(
     d_model: int,
     expansion_ratio: Union[int, float],
     ffn_hidden_size: Optional[int] = None,
@@ -42,7 +42,7 @@ def resolve_ffn_hidden_and_exp_ratio(
         ffn_hidden_size = int(d_model * expansion_ratio)
         if ffn_hidden_size != d_model * expansion_ratio:
             raise ValueError(
-                f'`d_model * expansion_ratio` (={d_model * expansion_ratio}) must be an integer.'
+                f'`d_model * expansion_ratio` must be an integer ({d_model=}; {expansion_ratio=}; {d_model * expansion_ratio=}).'
             )
     return ffn_hidden_size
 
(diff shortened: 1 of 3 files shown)


dakinggg (approved):
Missed it on the last PR, but could you add a simple test that tests setting `ffn_hidden_size` directly?

llmfoundry/models/layers/ffn.py
```diff
@@ -19,11 +19,21 @@
 log = logging.getLogger(__name__)
 
 
-def _resolve_ffn_hidden_and_exp_ratio(
+def resolve_ffn_hidden_and_exp_ratio(
```
irenedea, line 22: suggestion: rename to `resolve_ffn_hidden_size` now that we only return the hidden size

llmfoundry/models/layers/ffn.py
```diff
@@ -32,9 +42,9 @@ def _resolve_ffn_hidden_and_exp_ratio(
         ffn_hidden_size = int(d_model * expansion_ratio)
         if ffn_hidden_size != d_model * expansion_ratio:
             raise ValueError(
-                f'`d_model * expansion_ratio` ({ffn_hidden_size}) must be an integer.'
+                f'`d_model * expansion_ratio` (={d_model * expansion_ratio}) must be an integer.'
```
irenedea, line 45: nit: include d_model and expansion_ratio values in the error message

llmfoundry/models/layers/ffn.py
vchiley: Thank you!

llmfoundry/models/layers/ffn.py
vchiley: Thank you!

vchiley: eg: 
```
ValueError: `d_model * expansion_ratio` must be an integer (d_model=128; expansion_ratio=1.231; d_model * expansion_ratio=157.568).
```

PR #802, opened 2023-12-13
Fix token counting to use attention mask instead of ids

dakinggg:
This fixes off by one errors when pad token and eos token are the same thing.

Manual test compare to before, there are 1440 more tokens counted for a run of length 30 batches * 48 samples per batch, as expected, because there is one eos token per sample.
Screenshot 2023-12-13 at 4 24 13 PM

### DIFF
--- a/tests/data/test_dataloader.py
+++ b/tests/data/test_dataloader.py
@@ -605,11 +605,11 @@ def test_malformed_data(
 @pytest.mark.parametrize('model_max_length', [1024, 2048])
 @pytest.mark.parametrize('padding_side', ['left', 'right'])
 @pytest.mark.parametrize('add_decoder_input_ids', [True, False])
-def test_token_counting_func(pad_token_id: Optional[int], batch_size: int,
+def test_token_counting_func(pad_token_id: int, batch_size: int,
                              model_max_length: int, padding_side: str,
                              add_decoder_input_ids: bool):
     gptt = transformers.AutoTokenizer.from_pretrained('gpt2')
-    gptt.pad_token_id = pad_token_id if pad_token_id is not None else gptt.eos_token_id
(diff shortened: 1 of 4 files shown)


alextrott16 (approved):
Thanks for doing this. Definitely makes a lot more sense to be counting based on attention mask. Approve.

tests/data/test_dataloader.py
```diff
@@ -735,8 +741,10 @@ def test_token_counting_func_dataloader_setting(
             },
             **common_args
         })
+        ds_mock = MagicMock()
```
irenedea, line 744: Why is this mocked but not for streaming-finetuning?

tests/data/test_dataloader.py
```diff
@@ -654,27 +654,33 @@ def test_token_counting_func_dataloader_setting(
...
     if dataloader_type == 'denoising':
-        batch_tokenized['decoder_input_ids'] = batch_tokenized[
-            'input_ids'].clone()
+        expected_token_count += 2 * batch_size  # for the two eos tokens
+        expected_token_count += 5 * batch_size  # for the corruption prefix tokens
+
+    if dataloader_type in {'finetuning-hf', 'finetuning-streaming'}:
+        for b in batch_tokenized:
+            b['labels'] = b['input_ids'].copy()
         expected_token_count *= 2
```
irenedea, line 682: why is the expected token count doubled for finetuning?

irenedea (approved):
LGTM! Just had a couple questions.

tests/data/test_dataloader.py
irenedea: oh, nvm it is!

irenedea: I assume we need the mock here because the streamingtextdataset has attributes that are accessed.

PR #805, opened 2023-12-14
Make the ffn activation func configurable

vchiley:


llmfoundry/models/layers/ffn.py
```diff
@@ -143,13 +171,17 @@ def build_ffn(
             expansion_ratio=expansion_ratio,
             fc_type=fc_type,
             ffn_hidden_size=ffn_hidden_size,
+            ffn_act_fn=ffn_act_fn,
```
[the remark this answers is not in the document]
irenedea: You can also make `resolve_ffn_act_fn` private since it would only be used once here.

llmfoundry/models/layers/ffn.py
```diff
@@ -143,13 +171,17 @@ def build_ffn(
...
             device=device,
             bias=bias,
         )
     elif ffn_type == 'te_ln_mlp':
         assert te is not None
         ffn_hidden_size = resolve_ffn_hidden_size(d_model, expansion_ratio,
                                                   ffn_hidden_size)
+        if ffn_act_fn is not None:
```
irenedea, line 182: super nit: do this before L180 to raise exception as early as possible

llmfoundry/models/layers/ffn.py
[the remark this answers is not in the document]
vchiley: I talk making it not private [here](https://databricks.slack.com/archives/C06A2SP722W/p1702575065778869?thread_ts=1702516600.853729&cid=C06A2SP722W)

llmfoundry/models/layers/ffn.py
```diff
@@ -18,6 +20,28 @@
 
 log = logging.getLogger(__name__)
 
+ffn_act_fn_default = {
```
irenedea, line 23: nit: signify that it's private and immutable
```suggestion
_FFN_ACT_FN_DEFAULT = {
```

llmfoundry/models/layers/ffn.py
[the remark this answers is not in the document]
irenedea: keeping public sgtm, thanks!

llmfoundry/models/layers/ffn.py
irenedea: similarly for `default_act_fn`

llmfoundry/models/layers/ffn.py
vchiley: added [here](https://github.com/mosaicml/llm-foundry/pull/805/files#diff-2f97d1b3cb814c319489ccdb2313215ee45e7f8d5ebeb57e381c5ea1353c05f4R298)

tests/models/test_model.py
```diff
@@ -351,7 +351,20 @@ def test_full_forward_and_backward_t5_small(batch_size: int = 2):
      pytest.param('flash', torch.float16, marks=pytest.mark.gpu),
      pytest.param('flash', torch.bfloat16, marks=pytest.mark.gpu)])
 @pytest.mark.parametrize('ffn_type', ['mptmlp', 'mptgeglu'])
-def test_determinism(attn_impl: str, precision: torch.dtype, ffn_type: str):
+@pytest.mark.parametrize('ffn_act_fn', [
```
sashaDoubov, line 354: could you please try testing the use of custom functions? ex. `function: torch.nn.functional.relu`

sashaDoubov (approved):

tests/models/test_model.py
vchiley: you cant pass `torch.nn.functional.relu` into a yaml

llmfoundry/models/layers/ffn.py
```diff
@@ -18,6 +20,43 @@
...
+    'approximate': 'none',
+}
+
+
+def resolve_ffn_act_fn(
+    config: Optional[dict] = None,) -> Callable[[torch.Tensor], torch.Tensor]:
+    """Resolve the activation function for the feed-forward network.
+
+    Args:
+        config (Optional[dict]): The configuration dictionary for the activation function.
```
dblalock, line 34:
```suggestion
        config (Optional[Dict]): The configuration dictionary for the activation function.
```
Would be great to mention what keys or values we're expecting it to have (in particular, that anything from torch.nn.functional works). Also that we bind all the other kv pairs to the function as a partial. Otherwise this is kinda tautological.

llmfoundry/models/layers/ffn.py
```diff
@@ -150,6 +194,9 @@ def build_ffn(
         assert te is not None
         ffn_hidden_size = resolve_ffn_hidden_size(d_model, expansion_ratio,
                                                   ffn_hidden_size)
+        if ffn_act_fn is not None:
+            raise ValueError(
+                f'te block does not support custom activation functions.')
```
dblalock, line 199:
```suggestion
                f'Transformer Engine block does not support custom activation functions.')
```

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -295,6 +295,9 @@ def _validate_config(self) -> None:
             self.ffn_config['fc_type'] = self.fc_type
         elif self.ffn_config['ffn_type'] == 'te_ln_mlp':
             self.ffn_config['bias'] = not self.no_bias
+            if 'ffn_act_fn' in self.ffn_config.keys():
+                raise ValueError(
+                    f'te block does not support custom activation functions.')
```
dblalock, line 300:
```suggestion
                    f'Transformer Engine block does not support custom activation functions.')
```

tests/models/test_model.py
```diff
@@ -351,7 +351,20 @@ def test_full_forward_and_backward_t5_small(batch_size: int = 2):
...
+    {
+        'name': 'gelu',
+        'approximate': 'none'
+    },
+    {
+        'name': 'silu'
+    },
```
dblalock, line 361: probably just autoformatting, but always looking to keep code dense
```suggestion
    {'name': 'gelu', 'approximate': 'none'},
    {'name': 'silu'},
```

tests/models/test_model.py
```diff
@@ -516,12 +530,29 @@ def test_opt_wrapping():
...
     (2, 256),
 ])
+@pytest.mark.parametrize('ffn_act_fn', [
+    {
+        'name': 'gelu',
+        'approximate': 'none'
+    },
+    {
+        'name': 'silu'
+    },
```
dblalock, line 552:
```suggestion
    {'name': 'gelu', 'approximate': 'none'},
    {'name': 'silu'},
```
same as above

dblalock (commented):
Just looked at local stuff in this review.

llmfoundry/models/layers/ffn.py
vchiley: @b-chu is a big fan of `dict` instead of `Dict`
I'll let you guys battle it out 😄 

👍 I'll add more to the docstr

tests/models/test_model.py
vchiley: pre-commit `yapf` undoes this...

tests/models/test_model.py
vchiley: pre-commit `yapf` undoes this...

not sure how to make it clean AND make `yapf` happy

llmfoundry/models/layers/ffn.py
vchiley: Brain mentioned that `dict` is more bwd compatible which makes it the send 🚀

irenedea (approved):
LGTM! Pending removal of the `function` config input

llmfoundry/models/layers/ffn.py
```diff
@@ -18,6 +20,47 @@
...
+        config (Optional[dict]): The configuration dictionary for the activation function.
+            The dict config must specify a 'function' or the 'name' of a function.
+            If 'function' is specified, a Callable function is expected. If 'name' is
+            specified, the name is expected to be the name of a `torch.nn.functional`
+            function. All of other key values pairs are bound to the function as a partial.
+
+    Returns:
+        Callable[[torch.Tensor], torch.Tensor]: The activation function.
+    """
+    config = deepcopy(config or _FFN_ACT_FN_DEFAULT)
```
irenedea, line 43:
```suggestion
    if config is None:
        return _DEFAULT_ACT_FN
    config = deepcopy(config or _FFN_ACT_FN_DEFAULT)
```

vchiley: circular dependency
generating _DEFAULT_ACT_FN uses this fn
https://github.com/mosaicml/llm-foundry/pull/805/files/96b6330b8c3b51294ef041f7c3110c48ca655e46#diff-df5fd070e28295f023dcec5a8ad79df3cfd90c6d7a141f9714164f8ceff0fc93R62

vchiley: I can do 
```suggestion
    if config is None:
        config = _FFN_ACT_FN_DEFAULT
    config = deepcopy(config)
```

irenedea: ah, I see. How about remove the optional part, so the function accepts `config: dict`. It looks like `resolve_ffn_act_fn` is never called on `None` or without args. The default is already nicely explicitly set in other places with `_DEFAULT_ACT_FN` (like in `resolve_hidden_size`)

vchiley: [default here is None](https://github.com/mosaicml/llm-foundry/pull/805/files/96b6330b8c3b51294ef041f7c3110c48ca655e46#diff-df5fd070e28295f023dcec5a8ad79df3cfd90c6d7a141f9714164f8ceff0fc93R177)
gets passed in [here](https://github.com/mosaicml/llm-foundry/pull/805/files/96b6330b8c3b51294ef041f7c3110c48ca655e46#diff-df5fd070e28295f023dcec5a8ad79df3cfd90c6d7a141f9714164f8ceff0fc93R192)
which will throw error

vchiley: not sure that works with this setup...
we set a default act_fn [here](https://github.com/mosaicml/llm-foundry/pull/805/files/96b6330b8c3b51294ef041f7c3110c48ca655e46#diff-df5fd070e28295f023dcec5a8ad79df3cfd90c6d7a141f9714164f8ceff0fc93R101) in case someone uses MPTMLP on its own without using build_ffn

we could set a default [here](https://github.com/mosaicml/llm-foundry/pull/805/files/96b6330b8c3b51294ef041f7c3110c48ca655e46#diff-df5fd070e28295f023dcec5a8ad79df3cfd90c6d7a141f9714164f8ceff0fc93R177) but all ffn types allow us to specify custom act fns
(expacially ones we don't own like transformer engine ffn)

irenedea: ah 😅 defaults with python dicts are something, your suggestion above sounds good to me.

Alternatively you could explicitly handle the case below. 
```
act_fn = _DEFAULT_ACT_FN if None else resolve_ffn_act_fn(ffn_act_fn)
```

PR #806, opened 2023-12-15
Fix openai not conditioned imports

dakinggg:
Also adds a smoketest to prevent this issue in the future

Also fixes a typo in the code itself

### DIFF
--- a/.github/workflows/smoketest.yaml
+++ b/.github/workflows/smoketest.yaml
@@ -1,0 +1,41 @@
+name: Smoketest
+on:
+  push:
+    branches:
+    - main
+    - release/*
+  pull_request:
+    branches:
+    - main
+    - release/*
+  workflow_dispatch:
+# Cancel old runs when a new commit is pushed to the same branch if not on main or dev
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/dev' }}
+defaults:
+  run:
+    working-directory: .
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/smoketest.yaml
```diff
@@ -0,0 +1,41 @@
...
+  run:
+    working-directory: .
+jobs:
+  smoketest:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        python_version:
+        - "3.9"
```
j316chuck, line 26: Do we always support 3.9?

tests/test_smoketest.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+from llmfoundry import callbacks, data, models, optim, tokenizers, utils
+
+
+# This very simple test is just to use the above imports, which check and make sure we can import all the top-level
+# modules from foundry. This is mainly useful for checking that we have correctly conditionally imported all optional
+# dependencies.
+def test_smoketest():
```
j316chuck, line 10: Would put comments as part of function docstring

j316chuck (approved):
LGTM! Left a question and a nit

.github/workflows/smoketest.yaml
dakinggg: we should, yeah.

dakinggg: we run code quality tests on 3.9 and 3.10 as well

bandish-shah (approved):
LGTM

PR #807, opened 2023-12-16
Remove redundant transposes for rope rotation

ShashankMosaicML:
Now that we use transformers v4.36.0, we do not need to transpose query and keys back and forth for apply rope rotations to them. This PR fixes this.

We can see that the loss and mfu plots are almost identical
Screenshot 2023-12-19 at 8 10 48 PM

Screenshot 2023-12-19 at 8 11 33 PM


vchiley (commented):
Should we gate on HF version? Or is that the min version of HF that we install?

ShashankMosaicML:
> Should we gate on HF version? Or is that the min version of HF that we install?

Yes, the minimum version of HF we install now is 4.36 ( https://github.com/mosaicml/llm-foundry/blob/06b9a1f37d88bf0f94bae44ee4b266123b68b447/setup.py#L52 ). Hence, we do not need to gate since the 'unsqueeze_dim' argument was added in version 4.35.

dakinggg:
@ShashankMosaicML Lets gate on transformers version please. Since this is in the model code, it'd be good to make it compatible with all (or as much as possible) transformers versions

ShashankMosaicML:
> @ShashankMosaicML Lets gate on transformers version please. Since this is in the model code, it'd be good to make it compatible with all (or as much as possible) transformers versions

Done.

llmfoundry/models/layers/attention.py
```diff
@@ -34,6 +35,10 @@ def is_flash_v1_installed():
     return version.parse(flash_attn.__version__) < version.parse('2.0.0')
 
 
+def check_transformers_version(hf_version: str):
```
dakinggg, line 38:
```suggestion
def is_transformers_version_gte(hf_version: str) -> bool:
```

dakinggg (approved):

PR #808, opened 2023-12-18
Fix typo

XiaohanZhangCMU:
quick fix

### DIFF
--- a/scripts/data_prep/convert_dataset_hf.py
+++ b/scripts/data_prep/convert_dataset_hf.py
@@ -232,7 +232,7 @@ def build_hf_dataset(
     else:
         if not isinstance(tokenizer, PreTrainedTokenizerBase):
             raise ValueError(
-                f'{tokenizer=} must be of type PreTrainedTokenizerBase')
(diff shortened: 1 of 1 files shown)


scripts/data_prep/convert_dataset_hf.py
```diff
@@ -232,7 +232,7 @@ def build_hf_dataset(
     else:
         if not isinstance(tokenizer, PreTrainedTokenizerBase):
             raise ValueError(
-                f'{tokenizer=} must be of type PreTrainedTokenizerBase')
+                f'{tokenizer} must be of type PreTrainedTokenizerBase')
```
dakinggg, line 235: I don't think this is a typo, python format strings have some nice functionality :)
```
In [2]: f'{a=} blah'
Out[2]: 'a=1 blah'
```

XiaohanZhangCMU: Uh, sorry I didn't know about that trick haha! One of my byod tests failed here, that's how I noticed it. Will close it.

PR #810, opened 2023-12-19
add memory snapshot to callbacks

cli99:
This PR adds a callback to capture memory snapshot. See https://github.com/mosaicml/composer/pull/2788.


j316chuck (approved):
Want to add a test run and share this good work on slack 🎉

cli99:
The test is on the composer side. https://github.com/mosaicml/composer/blob/dev/tests/callbacks/test_memory_snapshot.py @j316chuck

PR #813, opened 2023-12-20
Async callback: Don't skip checkpoints, reliably only launch async eval when the checkpoint is ready

aspfohl:
One limitation of https://github.com/mosaicml/llm-foundry/pull/702 was there was no guarantee every intermediate checkpoint would be evaled, and that the eval run would not fail due to the checkpoint upload not being completely finished.

This patches that by checking all saved checkpoints on the checkpointer saver and comparing that to the files actually visible in the save folder. I introduce a "check_interval" parameter that is configurable separately from the eval interval. Most of the new logic is in `_get_checkpoints_and_launch_runs` and includes handling of sharded checkpointing

Tested manually with runs below, will add unit tests on helpers after a first round of reviews

Example run with sharded checkpointing:
* async-sharded-anna-rUbAQh
* eval-2ba-async-sharded-anna-uqj855 (cancelled all other eval runs)

Example run with mono checkpoint:
* async-anna-nmEmY9
* eval-2ba-async-anna-up6sOJ


llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -71,32 +72,29 @@ def get_run_name(training_run_name: str, current_interval: str) -> str:
     return f'{RUN_NAME_PREFIX}-{current_interval}-{name_without_uuid_suffix}'
 
 
-def get_latest_checkpoint(event: Event, state: State) -> Optional[str]:
-    """Get the latest checkpoint from the training run.
+SUPPORTED_UNITS = {TimeUnit.EPOCH, TimeUnit.BATCH}
+
+
+def get_interval_from_checkpoint(checkpoint: str, unit: TimeUnit) -> Time:
```
aspfohl, line 78: This and some of the other path wrangling I'm doing feels sort of hacky. I couldn't find anything obvious in composer that I could reuse, please point me to suggestions if you have any

dakinggg: yeah this is pretty painful and fragile. What if we saved more info on the `CheckpointSaver` callback in composer instead of trying to pull stuff out of strings here?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -194,8 +216,10 @@ class AsyncEval(Callback):
         training_config: Dict[str, Any]: The config from the training run
         interval: Union[str, int, Time]: The interval describing how often eval runs should be
             launched. If an integer, it will be assumed to be in :attr:`.TimeUnit.EPOCH`.
-            Otherwise, the unit must be either :attr:`.TimeUnit.EPOCH`, :attr:`.TimeUnit.BATCH`,
-            :attr:`.TimeUnit.TOKEN`, or :attr:`.TimeUnit.SAMPLE`.
+            Otherwise, the unit must be either :attr:`.TimeUnit.EPOCH` or :attr:`.TimeUnit.BATCH`.
+        check_interval: Optional[Union[str, int, Time]]: The interval describing how often
```
dakinggg, line 220: does this actually need to be configurable? or we can we just remove this knob?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -205,79 +229,186 @@ def __init__(
...
+            Returns checkpoints that have not been evaled
+        """
+        checkpointer = None
+        for callback in state.callbacks:
+            if isinstance(callback, CheckpointSaver):
+                checkpointer = callback
+                break
+
+        if not checkpointer:
+            log.warning('No checkpoint saver callback found. Skipping eval')
```
dakinggg, line 300: this should be `warnings.warn` I think, because it is a warning the user could take corrective action to fix

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -205,79 +229,186 @@ def __init__(
...
+                # checkpoint is something like folder/ep0-ba4/__0_0.distcp
+                checkpoint_folder = '/'.join(checkpoint.split('/')[:-1])
+
+                if f'{checkpoint_folder}/.metadata' not in found_checkpoints:
+                    log.debug(
+                        f'Checkpoint {checkpoint} not fully uploaded (missing metadata), skipping'
+                    )
+                    continue
+
+                for i in range(dist.get_world_size()):
```
dakinggg, line 347: perhaps it would be a bit less fragile to just count the number of files in the folder with the right file ending? im not sure if torch might change these filenames at some point.

dakinggg (commented):
Left a few comments to start. 

Also, one other thing to consider, how should this handle runs that get resumed?
An initial idea:
- resolve https://github.com/mosaicml/composer/issues/2087 and give the `CheckpointSaver` state about what checkpoints it has saved
- give the `AsyncEval` callback state about what checkpoints it has evaled
- i think everything just works at that point

llmfoundry/callbacks/async_eval_callback.py
aspfohl: Here's my proposition: https://www.notion.so/checkpoint-saver-state-improvements-423c8b7f35ca40a0baee678ef63bf8c6

I know we're pretty aggressive with locking to the latest composer in foundry, but what would you think about a hackier string parsing approach until the next composer release / composer bump in foundry? Even if a little more brittle, I think this logic makes the async callback significantly more usable. When manually testing, the previous version was consistently missing intermediate checkpoints

llmfoundry/callbacks/async_eval_callback.py
aspfohl: fair, removed!

aspfohl:
Composer dependency: https://github.com/mosaicml/composer/pull/2819
Example end to end run (single checkpoint): async-anna-JUOSGY
Example end to end training run (sharded checkpoint) async-sharded-anna-k2D9kB
Example preempted run: clusters are a little full right now, tested this yaml interactively, cut it off at the first saved checkpoint, and then manually reran to test autoresume functionality:
```yaml
seed: ${global_seed}
autoresume: true
model:
  name: mpt_causal_lm
  d_model: 2048
  n_heads: 16
  n_layers: 24
  vocab_size: 50368
  attn_config:
    attn_impl: triton
  init_device: meta
  max_seq_len: ${max_seq_len}
  expansion_ratio: 4
run_name: null
callbacks:
  async_eval:
    interval: 2ba
  lr_monitor: {}
  speed_monitor:
    window_size: 10
icl_tasks: eval/yamls/tasks_light.yaml
optimizer:
  lr: 0.0002
  eps: 1.0e-08
  name: decoupled_adamw
  betas:
  - 0.9
  - 0.95
  weight_decay: 0
precision: amp_bf16
tokenizer:
  name: EleutherAI/gpt-neox-20b
  kwargs:
algorithms:
  gradient_clipping:
    clipping_type: norm
    clipping_threshold: 1
data_local: ./my-copy-c4
data_remote: null
eval_loader:
  name: text
  dataset:
    local: ${data_local}
    split: val_small
    remote: ${data_remote}
    shuffle: false
    max_seq_len: ${max_seq_len}
    shuffle_seed: ${global_seed}
  drop_last: false
  num_workers: 8
fsdp_config:
  mixed_precision: PURE
  limit_all_gathers: true
  sharding_strategy: FULL_SHARD
  activation_cpu_offload: false
  activation_checkpointing: false
  activation_checkpointing_reentrant: false
global_seed: 17
max_seq_len: 2048
save_folder: s3://mosaicml-internal-checkpoints-test/anna/240105
max_duration: 2ba
eval_interval: 10ba
progress_bar: false
train_loader:
  name: text
  dataset:
    local: ${data_local}
    split: train_small
    remote: ${data_remote}
    shuffle: true
    max_seq_len: ${max_seq_len}
    shuffle_seed: ${global_seed}
  drop_last: true
  num_workers: 8
save_interval: 2ba
log_to_console: true
console_log_interval: 1ba
device_eval_batch_size: 4
eval_subset_num_batches: -1
global_train_batch_size: 512
device_train_microbatch_size: 4
scheduler:
  name: cosine_with_warmup
  alpha_f: 0.1
  t_warmup: 100ba
```

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -234,76 +209,250 @@ def __init__(
...
+            # (which is the most important) is evaled after it is written
+            include_end_of_training=False,
         )
 
+        log.info('Initialized AsyncEval callback. Will generate runs at ' +
+                 f'interval {interval}, checking at {self.check_interval}')
+
+    def state_dict(self) -> Dict[str, Any]:
+        checkpoints_evaled = []
+        for i, (c, rn) in self.checkpoints_evaled.items():
```
dakinggg, line 254: Could you spell out the variable names for this loop?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -234,76 +209,250 @@ def __init__(
...
+            checkpoints_evaled.append((interval_dict, c, rn))
+
+        return {
+            'checkpoints_evaled': checkpoints_evaled,
+        }
+
+    def load_state_dict(self, state_dict: Dict[str, Any]):
+        previous_checkpoints_evaled = state_dict.get('checkpoints_evaled', [])
+        if previous_checkpoints_evaled:
+            for (i, c, rn) in previous_checkpoints_evaled:
```
dakinggg, line 268: same

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -234,76 +209,250 @@ def __init__(
...
+
+        # Validate the interval (how often to launch eval runs)
+        self.interval = validate_interval(interval,
+                                          self.training_params['save_interval'])
+
+        # Configures how often to check for new checkpoints
+        self.check_interval = Time(max(self.interval.value // 5, 1),
+                                   self.interval.unit)
+
+        # Keep track of checkpoints by interval that have already been evaled
```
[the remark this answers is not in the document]
dakinggg: As I continue through I think it would be good to rename `interval` throughout when it actually means "the timestamp of this particular checkpoint/eval. Would be much more clear for me to read. lmk if you disagree

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -234,76 +209,250 @@ def __init__(
...
+            checkpointer_checkpoints: All checkpoints from the checkpointer state
+            remote_files: List of remote files in the save folder
+
+        Returns:
+            Dict of checkpoints that are complete and ready to be evaled
+        """
+        # Count the number of shards for each checkpoint group
+        remote_file_group_counts = Counter()
+        for f in remote_files:
+            interval_path = Path(f).parts[-2]
```
dakinggg, line 296: I think this is not totally guaranteed, since we let users customize the paths. Should we instead be taking taking the full path without the filename? or does that break your logic elsewhere?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -234,76 +209,250 @@ def __init__(
...
+
+        Returns:
+            Dict of checkpoints that are complete and ready to be evaled
+        """
+        unique_remote_checkpoints = set(remote_checkpoints)
+
+        checkpoints_to_eval = {}
+        for checkpoint, interval in checkpointer_checkpoints.items():
+            # eg {save_folder}/ep0-ba1-rank0.pt
+            interval_path = Path(checkpoint).parts[-1]
```
dakinggg, line 340: also i think technically the time information doesn't have to be in the filename and could be just in the path...if this is hard its probably ok not to optimize for now.

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -234,76 +209,250 @@ def __init__(
...
+        Args:
+            state: The current state of the training run
+
+        Returns:
+            Returns checkpoints that have not been evaled
+        """
+        checkpointer = None
+        for callback in state.callbacks:
+            if isinstance(callback, CheckpointSaver):
+                checkpointer = callback
```
dakinggg, line 362: just a little bit of proofing against strange uses...can we emit a warning if there are multiple checkpointsavers and say that this callback will be using the first one?

llmfoundry/callbacks/async_eval_callback.py
aspfohl: This info doesn't actually need to be in the path because we use the checkpoint ts (interval) from the checkpointer, the comment was just an example from parsing.

For this and above, it does assume `save_filename` in train.py is unique by checkpoint ts and does not include a directory (such as `foobar/ep{epoch}-ba{batch}-rank{rank}.pt`).. Does this seem reasonable? Can add an error on callback initialization to check for this not being the case

This makes the `load_path` logic for launching the eval runs significantly easier because it will be `{save_folder}/{what is referred to as checkpoint}` from remote.

dakinggg: ok yeah as long as you hard error if save_filename has a folder, im good with it.

dakinggg (approved):
LGTM pending adding that one error

PR #814, opened 2023-12-21
Add generic flatten imports to HF checkpointer

b-chu:
When using HF checkpointer from another repo, the checkpointer does not correctly flatten all imports. This change allows the user to specify the imports that need to be flattened.

No existing behavior has changed.

Tested using a different repo.


dakinggg (approved):
I know there are no unit tests for this (just a full test of checkpoint conversion), but in the spirit of not making the situation worse, can you add a unit test for the new helper that you are adding?

PR #818, opened 2023-12-21
Fix token counting to allow there to be no attention mask

dakinggg:
When we pretokenize, we just pass raw tensors to the collator, which does not add an attention mask. This is fine, because we pretokenize without padding, but would crash on the token counting function if your tokenizer had a pad token.

IFT token count same before and after
Screenshot 2023-12-21 at 4 11 02 PM

pretokenized token count as expected (40x2048x960)
Screenshot 2023-12-21 at 4 11 13 PM


samhavens (approved):

alextrott16 (approved):
"I knew you'd come crawling back"
-- input_ids, probably

PR #820, opened 2023-12-23
Adding support for alibi when using flash attention

ShashankMosaicML:
ALiBi support has recently been added to FlashAttention2 ([PR#540](https://github.com/Dao-AILab/flash-attention/pull/540/files)) and is stable with `flash-attention=v2.4.2`. This PR allows for ALiBi slopes to be passed to flash attention, hence enabling ALiBi for Flash Attention.

Experiments on 125m and 1b models show (nearly) identical training curves for ALiBi with [Triton attention](https://github.com/mosaicml/llm-foundry/blob/2b1fa79dda4feb9eac34227359f2b279df564edc/llmfoundry/models/layers/flash_attn_triton.py) (which is the current default attention implementation when using ALiBi) and ALiBi with Flash Attention. Further, using Flash Attention results in higher MFU numbers. In the plots below, 'treat' refers to flash attention, and 'control' refers to triton attention.

[Wandb experiment link](https://wandb.ai/mosaic-ml/flash-alibi?workspace=user-shashank-rajput)
Screenshot 2024-01-01 at 2 04 21 PM

Screenshot 2024-01-01 at 2 04 57 PM

**Note**: Flash Attention seems to use more memory than Triton:

Screenshot 2024-01-01 at 2 03 04 PM

These changes would effectively also enable the use of sliding window attention and memory efficient sequence id masking with ALiBi when using Flash Attention.

---

As an aside, @Skylion007 has been able to successfully update his own version of MosaicBERT with FA2, and has found it to indeed be faster than FA1: [w&b link](https://wandb.ai/mosaic-ml/test-mosaic-bert-fa2?workspace=user-aaron-mosaicml)


llmfoundry/models/layers/attention.py
```diff
@@ -346,7 +346,8 @@ def flash_attn_fn(
             softmax_scale=softmax_scale,
             causal=reset_is_causal,
             return_attn_probs=needs_weights,
-            window_size=(sliding_window_size, sliding_window_size))
+            window_size=(sliding_window_size, sliding_window_size),
+            **alibi_kwargs)
     else:
         raise RuntimeError(
             'flash-attn==1.0.9 or flash-attn==2.3.6 is required.')
```
Skylion007, line 353:
```suggestion
            'flash-attn==1.0.9 or flash-attn==2.4.2 is required.')
```

jacobfulano:
This looks great! What remains to be done?

Skylion007:
FYI, I added a BERT FA2 PR to the examples repo: https://github.com/mosaicml/examples/pull/440

tests/models/layers/test_flash_attn.py
```diff
@@ -6,7 +6,8 @@
 import pytest
 import torch
 
-from llmfoundry.models.layers.attention import (flash_attn_fn,
+from llmfoundry.models.layers import attention
```
irenedea, line 9: nit: can we remove this import? The below import should take care of everything

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -607,6 +607,13 @@ def forward(
             attn_uses_sequence_id=self.attn_uses_sequence_id,
             attn_impl=self.attn_impl,
             attention_mask=attention_mask)
+
+        if self.alibi and self.attn_impl == 'flash':
```
irenedea, line 611: how come attn_bias needs to be set for flash attention but not for other attention types like triton?

irenedea: Also I think this is missing the check for the flash attention version

llmfoundry/models/layers/attention.py
```diff
@@ -256,8 +256,9 @@ def flash_attn_fn(
 
         past_key_value = (key, value)
 
-    if attn_bias is not None:
-        raise NotImplementedError(f'attn_bias not implemented for flash attn.')
+    if attn_bias is not None and not is_flash_v2_installed(v2_version='v2.4.2'):
```
irenedea, line 259: Does this still work if we update the fa2 version? i.e. If you have a version higher than 2.4.2, will `is_flash_v2_installed(v2_version='v2.4.2')` return False?

tests/models/layers/test_flash_attn.py
```diff
@@ -253,3 +254,107 @@ def test_sliding_window(sliding_window_size: int):
...
+        dropout_p=0.0,
+        training=False,
+        needs_weights=False,
+        multiquery=False,
+    )
+
+    output_2.sum().backward()
+
+    assert torch.allclose(output_1, output_2)
+    assert torch.norm(query_2.grad - query_1.grad  # type: ignore
```
irenedea, line 355: what's the reason for the type ignore?

tests/models/test_model.py
```diff
@@ -647,8 +647,9 @@ def test_mpt_creation(norm_type: str, no_bias: bool, tie_word_embeddings: bool,
 def test_sequence_id_based_masking(attention_impl: str, pos_emb_config: dict):
     # Testing the output of concatenated sequence with sequence id masking vs individual sequences.
     alibi = pos_emb_config['alibi']
-    if alibi and attention_impl == 'flash':
-        pytest.skip(f'alibi only implemented with torch and triton attention.')
+    if alibi and attention_impl == 'flash' and not is_flash_v2_installed(
+            v2_version='v2.4.2'):
+        pytest.skip(f'flash attention below v2.4.2 does not support alibi.')
```
irenedea, line 652: I'd make a helper function to handle this logic for all tests.

irenedea (commented):
Thanks for doing this! Added some questions and comments.

llmfoundry/models/layers/attention.py
ShashankMosaicML: Yes, this will still work `is_flash_v2_installed` checks if at least the `v2_version` provided is installed:

https://github.com/mosaicml/llm-foundry/blob/083b4b2451a60487f0ff891b4fbd671064b90a30/llmfoundry/models/layers/attention.py#L27C16-L27C16

jacobfulano (approved):
This looks good to me! Could we potentially include an example yaml in `scripts/train/pretrain` with flash attention 2 as well?

vchiley:
> This looks good to me! Could we potentially include an example yaml in `scripts/train/pretrain` with flash attention 2 as well?

if you install llm-foundry using `pip install -e . [gpu-flash2]` (instead of `pip install -e . [gpu]`) and set `attn_impl: flash, MPT will uses FA2.

tests/models/layers/test_flash_attn.py
ShashankMosaicML: We use `attention.attn_bias_shape` later in the same file.  For that we need to explicitly import attention, because `from llmfoundry.models.layers.attention import (flash_attn_fn, ...` does not import attention:

image

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Regarding attn_bias for triton and torch: For those attention types, the attn_bias gets set [here](https://github.com/mosaicml/llm-foundry/blob/083b4b2451a60487f0ff891b4fbd671064b90a30/llmfoundry/models/layers/attention.py#L788). 

Regarding the missing check: I skipped the check because I added a check for the same in `configuration_mpt.py`
Screenshot 2024-01-04 at 12 58 43 PM

dakinggg (commented):
lgtm, will leave approval for @irenedea

tests/models/layers/test_flash_attn.py
irenedea: you can do:
```
>>> from llmfoundry.models.layers.attention import attn_bias_shape
>>> attn_bias_shape
<function attn_bias_shape at 0x169318900>
>>> 
```

tests/models/layers/test_flash_attn.py
```diff
@@ -253,3 +254,107 @@ def test_sliding_window(sliding_window_size: int):
...
+    query_2 = query_1.detach().clone()
+    query_2.requires_grad = True
+    key_2 = key_1.detach().clone()
+    key_2.requires_grad = True
+    value_2 = value_1.detach().clone()
+    value_2.requires_grad = True
+
+    def gen_bias():
+        causal = True
+        bs = attention.attn_bias_shape('triton',
```
irenedea, line 313: add attn_bias_shape to the import above.
```suggestion
        bs = attn_bias_shape('triton',
```

tests/models/layers/test_flash_attn.py
```diff
@@ -6,7 +6,8 @@
 import pytest
 import torch
 
-from llmfoundry.models.layers.attention import (flash_attn_fn,
+from llmfoundry.models.layers import attention
+from llmfoundry.models.layers.attention import (flash_attn_fn, gen_slopes,
```
irenedea, line 10:
```suggestion
from llmfoundry.models.layers.attention import (flash_attn_fn, gen_slopes, attn_bias_shape
```

llmfoundry/models/mpt/modeling_mpt.py
irenedea: @ShashankMosaicML Can we set the attn_bias in the same way that it is set for triton and torch? We should unify these code paths if possible. 

I suggest adding that logic here: https://github.com/mosaicml/llm-foundry/blob/083b4b2451a60487f0ff891b4fbd671064b90a30/llmfoundry/models/layers/attention.py#L786 such that this function should handle alibi with fa2.

Additionally for my understanding, is there a reason why for fa2 we call gen_slopes directly instead of calling build_alibi_bias? Thanks!

ShashankMosaicML: After discussion with @irenedea, we decided that since the attn_bias for triton and torch was different from what it was for flash attention, it would be better to have a separate variable for flash attention, which I named `alibi_slopes`. Please check the updated PR to see the changes.

FAv2 directly takes in the slopes corresponding to ALiBi's linear biases, whereas Triton and Torch use the actual biases. Hence for FAv2 we simply call gen_slopes instead of calling build_alibi_bias.

tests/models/layers/test_flash_attn.py
ShashankMosaicML: pyright complained that `-` cannot operate between `None` type objects, because `query_1.grad` and` query_2.grad` can potentially be `None`.

tests/models/test_model.py
ShashankMosaicML: Done.

irenedea (approved):

tests/models/layers/test_flash_attn.py
```diff
@@ -253,3 +256,108 @@ def test_sliding_window(sliding_window_size: int):
...
+        needs_weights=False,
+        multiquery=False,
+    )
+
+    output_2.sum().backward()
+
+    assert torch.allclose(output_1, output_2)
+    assert torch.norm(query_2.grad - query_1.grad  # type: ignore
+                     ) <= 1e-2 + 1e-2 * torch.norm(query_2.grad)
+    assert torch.norm(key_2.grad - key_1.grad  # type: ignore
```
irenedea, line 360:
```suggestion
    assert key_2.grad is not None and key_1.grad is not None
    assert torch.norm(key_2.grad - key_1.grad 
```

and so forth

tests/models/layers/test_flash_attn.py
ShashankMosaicML: Fixed this.

dakinggg (approved):

PR #826, opened 2023-12-28
Add eos_text and bos_text defaults for convert_text_to_mds.py

irenedea:


scripts/data_prep/convert_text_to_mds.py
```diff
@@ -226,7 +221,9 @@ def download_and_convert(
         downloading_iter = DownloadingIterable(object_names=file_names,
                                                output_folder=tmp_dir,
                                                object_store=object_store)
-        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
+        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name,
+                                                  add_bos_token=False,
+                                                  add_eos_token=False)
         tokenizer.model_max_length = 5000000000  # Hack to prevent warnings from HuggingFace
```
irenedea, line 227: Added this because if a user specifies a particular `bos_text` or `eos_text` the tokenizer should not automatically also add bos or eos token

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -343,14 +340,32 @@ def convert_text_to_mds(
...
-        bos_text (str): Text to prepend to each example to separate concatenated samples
         no_wrap: (bool): Whether to let text examples wrap across multiple training examples
         compression (str): The compression algorithm to use for MDS writing
         processes (int): The number of processes to use.
         args_str (str): String representation of the arguments
         reprocess (bool): Whether to always reprocess the given folder of text files
+        bos_text (Optional[str]): Text to prepend to each example to separate concatenated samples
+            If None, use the tokenizer's bos_token if tokenizer.add_bos_token is True, otherwise use an empty string.
+        eos_text (Optional[str]): Text end to append to each example to separate concatenated samples
+            If None, use the tokenizer's eos_token.
```
irenedea, line 351: don't love that these defaults aren't mirrors of each other, but this allows the defaults to match the finetuning cases we discussed.

dakinggg (commented):
I think this is trying to put the finetuning configuration in the wrong place. I think the default for this script should be to just do whatever the tokenizer does by default. That is what I would expect as a user. I think its reasonable to error if the tokenizer adds a bos (eos) and the user also specified bos (eos). That error would occur in `ConcatTokensDataset` constructor (currently its a warning).

irenedea:
@dakinggg Hm, so with what you propose, a user could not specify an alternative bos if the tokenizer already adds one by default? OK with that being the case, i don't really have a strong opinion on that.

dakinggg:
@irenedea if we added a `tokenizer_kwargs` then they could, by turning off the auto bos adding.

And today, they _can_ add a custom bos, but it will be _in addition to_ the one the tokenizer adds by default. Which I don't think is ever what is desired.

irenedea:
closing in favor of https://github.com/mosaicml/llm-foundry/pull/843

PR #829, opened 2023-12-31
Align GLU implementation with LLaMa

vchiley:
Align [LLM-Foundry GLU implementation](https://github.com/mosaicml/llm-foundry/blob/f2b7bef416f6d0d54f1e6dbe45fee61bff821f38/llmfoundry/models/layers/ffn.py#L148) with [LLaMa implementation](https://github.com/huggingface/transformers/blob/3cefac1d974db5e2825a0cb2b842883a628be7a0/src/transformers/models/llama/modeling_llama.py#L268)
ie 
```
self.down_proj(self.act(self.gate_proj(x)) * self.up_proj(x))
```
instead of 
```
self.down_proj(self.act(self.up_proj(x)) * self.gate(x))
```

Additionally, with https://github.com/mosaicml/llm-foundry/pull/805, we are no longer constricted to using `gelu`s. Therefore it's no longer GeGLU, but ANY_ACT_FN_GLU, so I'm changing the name to generic GLU.

### DIFF
--- a/llmfoundry/models/mpt/configuration_mpt.py
+++ b/llmfoundry/models/mpt/configuration_mpt.py
@@ -107,7 +107,7 @@ class MPTConfig(PretrainedConfig):
                     factor (float): Scaling factor to use if using 'linear' or 'dynamic' as rope_scaling.type.
                 kv_n_heads (Optional[int]): For grouped_query_attention only, allow user to specify number of kv heads.
             ffn_config (Dict): A dictionary used to configure the model's ffn module:
-                ffn_type (str): type of ffn to use. Options: mptmlp, mptgeglu, te_ln_mlp
+                ffn_type (str): type of ffn to use. Options: mptmlp, mptglu, te_ln_mlp
             init_device (str): The device to use for parameter initialization.
             logit_scale (Optional[Union[float, str]]): If not None, scale the logits by this value.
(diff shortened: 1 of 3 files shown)


llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -291,7 +291,7 @@ def _validate_config(self) -> None:
                     + 'pip install flash-attn==1.0.6 --no-build-isolation \n' +
                     'pip install git+https://github.com/NVIDIA/TransformerEngine.git@144e4888b2cdd60bd52e706d5b7a79cb9c1a7156'
                 )
-        if self.ffn_config['ffn_type'] in ['mptmlp', 'mptgeglu']:
```
dakinggg, line 294: Can you add an error here if ffn type is mptgeglu that tells user the name changed and the structure changed slightly, linking to this pr for details. Just in case anyone was using it.

dakinggg (approved):

llmfoundry/models/mpt/configuration_mpt.py
vchiley: [Done](https://github.com/mosaicml/llm-foundry/pull/829/commits/52be34f1d4e92dfae75b0e6dd1e947403c167e47)

PR #830, opened 2024-01-02
Use `sync_module_states: True` when using HSDP

abhi-mosaic:
If a user is training from scratch with one of the HSDP hybrid sharding strategies, they must use `sync_module_states: True` otherwise the different data replicas will start with different randomly initialized weights.


mvpatel2000 (approved):
Great catch :)

PR #831, opened 2024-01-02
Add how to use multiple streams in loaders in yaml

cli99:
Users asks for an example to use multiple streams (https://github.com/mosaicml/streaming?tab=readme-ov-file#seamless-data-mixing) through yaml file.


scripts/train/README.md
```diff
@@ -3,17 +3,28 @@
...
-   1. [How to start single and multi-node pretraining](#howtostartpretraining)
-1. [Part 2: LLM Finetuning](#llmfinetuning)
-   1. [Using a dataset on the HuggingFace Hub](#hfdataset)
-   1. [Using a local dataset](#localdataset)
-   1. [Using a StreamingDataset (MDS) formatted dataset locally or in an object store](#mdsdataset)
-1. [Using Flash Attention](#flashattention)
-1. [FAQ: How many GPUs do I need to train a LLM?](#howmandygpus)
-1. [FAQ: Optimizing Performance](#optimizingperformance)
+- [LLM Pretraining and Finetuning](#llm-pretraining-and-finetuning)
+      - [Table of Contents](#table-of-contents)
```
dakinggg, line 7: I don't think we need the self reference to the table of contents here

scripts/train/README.md
```diff
@@ -333,6 +344,27 @@ train_loader:
         ...
 ```
 
+To specify multiple streams (e.g. as in [Seamless data mixing](https://github.com/mosaicml/streaming?tab=readme-ov-file#seamless-data-mixing)) to load data in the yaml file, update your YAML like so:
+
+```yaml
+train_loader:
+    name: finetuning
```
dakinggg, line 351: the finetuning dataloader doesn't actually support streams. not for any technical reason, just haven't update it, but this example should be in the pretraining data section and use the `text` dataloader

dakinggg (changes requested):

scripts/train/README.md
dakinggg: @cli99 mind finishing this PR real quick?

dakinggg:
Done in another PR

PR #834, opened 2024-01-05
Add support for eval_loader & eval_subset_num_batches in async callback

aspfohl:
As titled. To do this, the training and eval scripts needed to be adjusted slightly:
* eval loader is not passed into the trainer evaluators when using async eval
* eval_first will also not work (seemed non-trivial to support - kept this out of scope for this change)
* eval.py added support for subset_num_batches in eval loader 

This also includes a change in the AsyncEval inputs - I removed `compute` in favor of a more generic `eval_run_config`:
* Allows for scheduling overrides - assuming you may want eval runs to be lower priority than training runs
* Allows for command overrides - needed if your training command had any custom commands besides composer train/train.py
* Allows for image override
* All other run configurations are not supported - this is to reduce number of knobs and complexity (eg parameter resolution could get quite complex)


scripts/train/train.py
```diff
@@ -531,22 +530,26 @@ def main(cfg: DictConfig) -> Trainer:
...
-        icl_seq_len=eval_icl_seq_len,
-        icl_subset_num_batches=icl_subset_num_batches,
-    )
-
-    if eval_gauntlet_callback is not None and not use_async_eval:
-        callbacks.append(eval_gauntlet_callback)
+    if use_async_eval:
+        evaluators = []
+        if eval_first:
+            raise ValueError(
```
dakinggg, line 536: Can we just warn that it's being skipped (and set it to false)?

scripts/eval/eval.py
```diff
@@ -297,10 +299,14 @@ def main(cfg: DictConfig) -> Tuple[List[Trainer], pd.DataFrame]:
                                              'loggers',
                                              must_exist=False,
                                              default_value={})
-    icl_subset_num_batches: int = pop_config(cfg,
-                                             'icl_subset_num_batches',
-                                             must_exist=False,
-                                             default_value=None)
+    subset_num_batches: int = pop_config(cfg,
+                                         'subset_num_batches',
```
dakinggg, line 303: Any particular reason to not just keep the naming consistent with train and call this `eval_subset_num_batches`?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -185,57 +193,116 @@ def validate_interval(interval: Union[str, int, Time],
     return result
 
 
+def validate_eval_run_config(
+        eval_run_config: Optional[Dict[str, Any]]) -> Dict[str, Any]:
+
+    if isinstance(eval_run_config, DictConfig):
```
dakinggg, line 199: As with the other logger thing, I think you just want to be popping it with convert=True.

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -185,57 +193,116 @@ def validate_interval(interval: Union[str, int, Time],
     return result
 
 
+def validate_eval_run_config(
+        eval_run_config: Optional[Dict[str, Any]]) -> Dict[str, Any]:
```
dakinggg, line 197: This input type doesn't match the implementation which accepts dictconfig

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -185,57 +193,116 @@ def validate_interval(interval: Union[str, int, Time],
...
+        run_config = eval_run_config.copy()
+
+    if not run_config:
+        return {}
+
+    supported_keys = {'image', 'command', 'compute', 'scheduling'}
+    for key in run_config:
+        if key not in supported_keys:
+            raise ValueError(
+                f'Unsupported eval run config key {key}. Supported keys: {supported_keys}'
```
dakinggg, line 214: Can we do a set operation here to show all unsupported keys in one error message?

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -185,57 +193,116 @@ def validate_interval(interval: Union[str, int, Time],
...
+            )
+
+    for dict_field in ('scheduling', 'compute'):
```
dakinggg, line 217: Is there a simpler way to do this validation using anything from msdk? The very specific type checking feels a little rough. Was wondering if maybe we can just construct the config objects, or call some validate function from msdk.

llmfoundry/callbacks/async_eval_callback.py
```diff
@@ -185,57 +193,116 @@ def validate_interval(interval: Union[str, int, Time],
...
-            parameters=training_config,
+            parameters=training_params,
             checkpoint='test',
             training_run_name=self.current_run.name,
         )
         log.info(
             f'Initialized AsyncEval callback. Will generate runs at interval {interval}'
         )
 
+        # TODO: potentially support eval_first
```
dakinggg, line 304: Let's just track this in jira

dakinggg (commented):
LGTM, would like to see a manual test run that does the eval loader eval and icl eval before merging.

aspfohl:
Training run: async-anna-dxGTWJ (stopped a little early but eval loader looked ok)
Spawned eval run: eval-4ba-async-anna-ZY55m7

Manually launched eval run (just eval, no icl) eval-4ba-async-anna-SX9lGw

dakinggg (approved):

PR #835, opened 2024-01-05
use current year in license

aspfohl:
Not finding exactly what I want from this pre-commit hook, there were a few options here:
a. update the license header file with 2024 and don't mess with insert-license configuration
   * Updates all files with 2024. Would require us to update this file every year manually if we want to automatically keep it up to date

b. update insert-license to include --use-current-year
   * Updates all files with 2022-2024. Whoever makes the first commit of 2025 will change every single file with the pre commit hook
   * I can't seem to find any difference between --use-current-year by itself and --use-current-year + --allow-past-years  https://github.com/Lucas-C/pre-commit-hooks?tab=readme-ov-file#handling-years-flexibly

Went with (b) to have this be a little more automated 🤷‍♀️

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ install_requires = [
 extra_deps = {}
 
 extra_deps['dev'] = [
-    'pre-commit>=3.2.0,<3',
+    'pre-commit>=3.2.0,<4',
     'pytest>=7.2.1,<8',
     'pytest_codeblocks>=0.16.1,<0.17',
     'pytest-cov>=4,<5',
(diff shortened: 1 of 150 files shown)


setup.py
```diff
@@ -73,7 +76,7 @@
 extra_deps = {}
 
 extra_deps['dev'] = [
-    'pre-commit>=2.18.1,<3',
+    'pre-commit>=3.2.0,<3',
```
aspfohl, line 79: Required to support `allow-past-years`, a reasonable bump
https://github.com/Lucas-C/pre-commit-hooks/releases
https://github.com/pre-commit/pre-commit/releases/tag/v3.2.0

mvpatel2000 (approved):

aspfohl:
https://github.com/mosaicml/composer/pull/2825#issuecomment-1881468122

PR #838, opened 2024-01-05
Shashank/new benchmarks

ShashankMosaicML:
Adding new benchmarks for large scale pretraining (>=128 GPUs), as well as adding minor updates to the benchmarking scripts.


scripts/train/benchmarking/submit_benchmarks.py
```diff
@@ -420,11 +420,11 @@ def run_config(config: Tuple[str, int, int, str, str, int, str],
     if args.data_remote is None:
         command += f"""
             cd llm-foundry/scripts
-            python data_prep/convert_dataset_hf.py --dataset c4 --data_subset en --out_root ./my-copy-c4 --splits train_small val_small --concat_tokens {max_seq_len} --tokenizer gpt2 --eos_text '<|endoftext|>'
```
vchiley, line 423: why not allow user to set max seq len?

ShashankMosaicML: Good catch, let me revert it.

ShashankMosaicML: Done.

vchiley (approved):
👍

PR #841, opened 2024-01-07
Fix error when decoding a token in the id gap (or out of range) in a tiktoken tokenizer

dakinggg:
Previously, if you tried to decode an invalid token, the tokenizer would crash. For a trained model, this would never be an issue because it would not produce invalid tokens, but for an untrained model, it could randomly produce any token up to the embedding size. This PR has the tokenizer instead return empty string for these invalid tokens.

This behavior matches some Hugging Face fast tokenizers, but not slow tokenizers, which error on out of range indices. Even though the tiktoken wrapper is technically a slow tokenizer, the fast behavior seems better to avoid crashes on random models.

It also adds a utf-8 encoding to two file open calls. See https://github.com/mosaicml/composer/pull/2824 for more details.

### DIFF
--- a/llmfoundry/tokenizers/tiktoken.py
+++ b/llmfoundry/tokenizers/tiktoken.py
@@ -254,8 +254,8 @@ class TiktokenTokenizerWrapper(PreTrainedTokenizer):
     def _convert_id_to_token(self, index: int) -> Optional[str]:
         """Converts an index (integer) in a token (str) using the vocab."""
         # For tokens in either the gap in ids in the tokenizer, or beyond the range of the tokenizer,
-        # we return empty string. This matches the behavior of some Hugging Face tokenizers (e.g. llama, opt),
-        # but not all (e.g. gpt2 which crashes on out of range ids)
+        # we return empty string. This matches the behavior of Hugging Face fast tokenizers,
+        # but not slow tokenizers.
         return self.decoder.get(index, '')
 
     def convert_tokens_to_string(self, tokens: List[str]) -> str:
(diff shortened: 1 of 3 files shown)


sashaDoubov (approved):
LGTM! this is very useful for debugging + thank you for digging into the utf-8 encoding issue.

llmfoundry/tokenizers/tiktoken.py
```diff
@@ -253,7 +253,10 @@ def _convert_token_to_id(self, token: str) -> Optional[int]:
 
     def _convert_id_to_token(self, index: int) -> Optional[str]:
         """Converts an index (integer) in a token (str) using the vocab."""
-        return self.decoder.get(index)
+        # For tokens in either the gap in ids in the tokenizer, or beyond the range of the tokenizer,
+        # we return empty string. This matches the behavior of Hugging Face fast tokenizers,
+        # but not slow tokenizers.
+        return self.decoder.get(index, '')
```
irenedea, line 259: nit: maybe add a warning if we hit an id gap or out of range index

irenedea (approved):

llmfoundry/tokenizers/tiktoken.py
dakinggg: The other HF tokenizers don't do this so I think I'm going to leave it as is to match their behavior. Can add in a follow up if you feel strongly.

irenedea: nah that sounds good to leave as is

PR #842, opened 2024-01-08
Handle large files during text to mds conversion

irenedea:
Previously, large files are read entirely at once via `file.read()`. This reads the file and tokenizes in chunks.


llmfoundry/utils/data_prep_utils.py
```diff
@@ -107,6 +112,49 @@ def __iter__(self):
...
+        super().__init__(tokenizer, max_length, bos_text, eos_text, no_wrap)
+
+    def __iter__(self) -> Iterable[Dict[str, bytes]]:
+
+        buffer = []
+        for file in self.files:
+            with open(file, 'r') as f:
+                buffer += self.bos_tokens
+                for chunk in iter(partial(f.read, 1000), ''):
+                    encoded = self.tokenizer(chunk,
```
irenedea, line 145: todo: this logic is wrong, must turn the tokenizer's add bos token off if we're tokenizer in chunks

llmfoundry/utils/data_prep_utils.py
```diff
@@ -107,6 +112,49 @@ def __iter__(self):
...
+        self.files = files
+        super().__init__(tokenizer, max_length, bos_text, eos_text, no_wrap)
+
+    def __iter__(self) -> Iterable[Dict[str, bytes]]:
+
+        buffer = []
+        for file in self.files:
+            with open(file, 'r') as f:
+                buffer += self.bos_tokens
+                for chunk in iter(partial(f.read, 1000), ''):
```
irenedea, line 144: probably can do more than 1000 chars

PR #844, opened 2024-01-08
Validation

XiaohanZhangCMU:
JIRA: https://databricks.atlassian.net/jira/software/c/projects/STR/issues/STR-141?filter=allissues

This script is particularly useful in scenarios where there is a risk of data being malformed. It acts as a preventive measure to ensure data integrity and helps in cost assessment for the fine-tuning process.

### Note on Long-Term Solution:
- **Temporary Measure**: This script is a stop-gap solution.
- **Future Development**: We are in the process of developing a long-term data preparation service, which will eventually replace this script.

### Tasks Include:
- count_tokens
- run tokenization on the dataset
  1) For IFT task: validate tokenization by running tokenizer + filter on the entire dataset. count the number of tokens. Throws error if there are any empty responses or prompts
  2) For CPT task: call donwload_text_to_mds.py and count the resulted mds dataset. Note this could take a long time.

### DIFF
--- a/scripts/data_prep/validate_and_tokenize_data.py
+++ b/scripts/data_prep/validate_and_tokenize_data.py
@@ -64,7 +64,7 @@
 
 # COMMAND ----------
 
-dbutils.library.restartPython()
+# dbutils.library.restartPython()
 
 # COMMAND ----------
 
@@ -77,13 +77,14 @@ from streaming import StreamingDataset
 import numpy as np
 from omegaconf import OmegaConf as om
 from argparse import Namespace
-from typing import Union, Tuple 
+from typing import Union, Tuple
 from llmfoundry.utils import build_tokenizer
-import torch 
+from huggingface_hub import dataset_info
+from datasets import get_dataset_split_names
 
 # COMMAND ----------
 
-# MAGIC %md 
+# MAGIC %md
 # MAGIC ## User Defines the Cell Below
 
 # COMMAND ----------
@@ -503,7 +504,7 @@ def _args_str(original_args: Namespace) -> str:
 
 from streaming.base.storage.upload import CloudUploader
 from streaming.base.storage.download import download_file
-import json 
+import json
 
 def integrity_check(out: Union[str, Tuple[str, str]]):
     """Check if the index file has integrity.
@@ -538,11 +539,8 @@ def integrity_check(out: Union[str, Tuple[str, str]]):
         merged_index = json.load(open(local_merged_index_path, 'r'))
         n_shard_files = len({b['raw_data']['basename'] for b in merged_index['shards']})
         return n_shard_files == actual_n_shard_files
-        
+
 def check_HF_datasets(dataset_names_with_splits):
-    from huggingface_hub import dataset_info
-    from datasets import get_dataset_split_names
-    import os
     token = os.environ.get("HUGGING_FACE_HUB_TOKEN")
     for dataset_name_with_split in dataset_names_with_splits:
         dataset_name, split = os.path.split(dataset_name_with_split)
@@ -565,7 +563,7 @@ def check_HF_datasets(dataset_names_with_splits):
         if split not in splits:
             return False, f"Failed to load Hugging Face dataset {dataset_name_with_split}. Split not found."
     return True, ""
-    
+
 def is_hf_dataset_path(path):
     """Check if a given string is a dataset path used by Hugging Face.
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -66,6 +66,7 @@ repos:
     - --comment-style
     - '#'
     types: [python]
+    exclude: scripts/data_prep/validate_and_tokenize_data.py
```
nancyhung, line 69: was there a particular reason this was excluded just curious?

scripts/data_prep/validate_and_tokenize_data.py
```diff
@@ -0,0 +1,731 @@
...
+
+# COMMAND ----------
+
+# MAGIC %pip install llm-foundry
+
+# COMMAND ----------
+
+# dbutils.library.restartPython()
+
+# COMMAND ----------
```
nancyhung, line 78: Instead of having all this commented out code in this open-source repo, you might want to leave it in the Databricks notebook and document it internally. We can publish the notebook in our docs for users to clone and run on their data.

Alternatively, if you want to publish the notebook in foundry, you can export the notebook as `ipynb` format and it looks a lot nicer in an IDE.

I would also add documentation in the README for what this script is used for: https://github.com/mosaicml/llm-foundry/tree/main/scripts/data_prep#readme.

.pre-commit-config.yaml
XiaohanZhangCMU: For databricks to render a python script as a notebook, it needs the script to start with #databricks notebook source. This change asks pre-commit to skip adding the license header to the script.

scripts/data_prep/validate_and_tokenize_data.py
```diff
@@ -0,0 +1,731 @@
...
+# MAGIC     future: Literal[False] = False,
+# MAGIC }
+# MAGIC - What null checkings do we want to have?
+# MAGIC - How to map the model to its expected eos_text / bos_text format? [Ref](https://databricks.slack.com/archives/C05K29T9NBF/p1703644153357929?thread_ts=1703643155.904289&cid=C05K29T9NBF)
+# MAGIC - How to automate tokenization for CPT? it is always really standard: sequence -> concat(tok(BOS), tok(sequence), tok(EOS)), and then concatenate sequences. [Ref](https://databricks.slack.com/archives/C05K29T9NBF/p1703698056000399?thread_ts=1703643155.904289&cid=C05K29T9NBF)
+# MAGIC ```
+
+# COMMAND ----------
+
+# MAGIC %pip install llm-foundry
```
irenedea, line 72: this should install the version of foundry that we're using in finetuning, so I think this will need to git clone a specific commit and pip install from the cloned repo

scripts/data_prep/validate_and_tokenize_data.py
```diff
@@ -0,0 +1,731 @@
...
+
+# COMMAND ----------
+
+# dbutils.library.restartPython()
```
irenedea, line 76: why do we need this?

scripts/data_prep/validate_and_tokenize_data.py
XiaohanZhangCMU: Good point. I'll make it a ipynb nb. We can discuss where to put it.

scripts/data_prep/validate_and_tokenize_data.py
```diff
@@ -0,0 +1,731 @@
...
+    output_folder: str,
+    input_folder: str,
+    concat_tokens: int,
+    eos_text: str,
+    bos_text: str,
+    no_wrap: bool,
+    compression: str,
+    processes: int,
+    args_str: str,
+    reprocess: bool,
```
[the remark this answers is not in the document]
XiaohanZhangCMU: If I am putting the script somewhere else, can we figure out a way for me to import convert_text_to_mds?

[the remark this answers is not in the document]
irenedea: yeah want to meet tomorrow? we can discuss more

notebooks/validate_and_tokenize_data.ipynb
```diff
@@ -854,7 +902,7 @@
     "n_billing_tokens_in_dataset = len(mds_dataset) * FT_API_args.context_length \n",
     "print(f\"Dataset has ~{n_billing_tokens_in_dataset} tokens that will be charged for during training\")\n",
     "print(f\"By default, you'll train for {n_epochs} epochs on this dataset\")\n",
-    "print(f\"By default, you'll be charged for ~{n_epochs * n_billing_tokens_in_dataset} tokens\")"
+    "print(f\"By default, ~{n_epochs * n_billing_tokens_in_dataset} tokens will be used in training\")"
```
irenedea, line 905: instead of ~ use "approximately"
```suggestion
    "print(f\"By default, approximately {n_epochs * n_billing_tokens_in_dataset} tokens will be used in training\")"
```

PR #845, opened 2024-01-08
Disable Environment Variable Resolution

irenedea:
# Manual tests
## Before
ift-mpt-7b-zmb50m - works as expected
## After (using this branch)
ift-mpt-7b-env-disabled-nr5HmB - fails as expected with omegaconf error
ift-mpt-7b-env-disabled-with-yaml-res-mdubZr - works because only yaml res


dakinggg (commented):
Can you also manually test that this doesn't mess up the other yaml style variable interpolation?

irenedea:
@dakinggg tested and updated the PR description

dakinggg (approved):

PR #850, opened 2024-01-09
Logs upload URI

milocress:
Logs upload URI for each file uploaded.

Tested for S3, GC, and OCI backends.


b-chu (approved):
Thanks for the PR!

PR #857, opened 2024-01-11
Add model downloading from oras

irenedea:
## Manual tests
`ift-mpt-7b-harbor-download-i8vxl8`

### DIFF
--- a/scripts/misc/download_hf_model.py
+++ b/scripts/misc/download_hf_model.py
@@ -5,13 +5,12 @@
 import argparse
 import logging
 import os
+import shutil
+import subprocess
+from typing import Dict
 
 import yaml
-import subprocess
-
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
-
-import shutil
 
 from llmfoundry.utils.model_download_utils import (download_from_cache_server,
                                                    download_from_hf_hub)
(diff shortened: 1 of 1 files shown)


scripts/misc/download_hf_model.py
```diff
@@ -56,28 +129,33 @@
                 ignore_cert=args.ignore_cert,
             )
 
-            # A little hacky: run the Hugging Face download just to repair the symlinks in the HF cache file structure.
-            # This shouldn't actually download any files if the cache server download was successful, but should address
-            # a non-deterministic bug where the symlinks aren't repaired properly by the time the model is initialized.
-            log.info('Repairing Hugging Face cache symlinks')
+            if args.fallback == 'hf':
```
irenedea, line 132: @jerrychen109 is it okay to condition on this?

irenedea: should we condition on the save_dir being the HF cache? symlinks are just for the HF cache, right?

scripts/misc/download_hf_model.py
```diff
@@ -56,26 +135,33 @@
...
```
irenedea, line 138:
```suggestion
            if args.fallback == 'hf' and args.save_dir == HUGGINGFACE_HUB_CACHE:
```

irenedea:
closing in favor of https://github.com/mosaicml/llm-foundry/pull/881

PR #868, opened 2024-01-12
Delta to JSONL conversion script cleanup and bug fix

nancyhung:
This change includes 2 primary changes:
1. Fix bug in the Delta to JSONL conversion script where the `runtime_version` of the cluster still had alphabetic characters, which failed version parsing i.e. (`14.0.1-photon` instead of `14.0.1`). This change adds more regex to strip the extra alphabetic characters and trailing delineators (`-`, `.`).
2. Add some cleanups to the script to abstract the cluster validation and details into its own separate function. We also don't need the `cluster_id` argument to be mandatory now that we are supporting `serverless`.

I was able to test this e2e with a yaml generated by the finetuning API in the Databricks training environment!
Run name: `test-delta-Udjv3E`

### DIFF
--- a/scripts/data_prep/convert_delta_to_json.py
+++ b/scripts/data_prep/convert_delta_to_json.py
@@ -513,7 +513,7 @@ if __name__ == '__main__':
         required=False,
         type=str,
         default='train-00000-of-00001.jsonl',
-        help='The combined final jsonl that combines all partitioned jsonl')
+        help='The name of the combined final jsonl that combines all partitioned jsonl')
     args = parser.parse_args()
 
     from databricks.sdk import WorkspaceClient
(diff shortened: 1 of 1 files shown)


irenedea (approved):
LGTM

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -377,63 +377,60 @@ def fetch(
...
+                    databricks_host).token(databricks_token).header(
                         'x-databricks-session-id', session_id).getOrCreate()
 
             else:
                 sparkSession = DatabricksSession.builder.remote(
-                    host=args.DATABRICKS_HOST,
-                    token=args.DATABRICKS_TOKEN,
+                    host=databricks_host,
+                    token=databricks_token,
                     cluster_id=args.cluster_id).getOrCreate()
```
XiaohanZhangCMU, line 434: cluster_id=args.cluster_id).getOrCreate()

XiaohanZhangCMU (changes requested):

XiaohanZhangCMU (approved):
LGTM

PR #873, opened 2024-01-15
Enable streaming of local finetuning dataset

eldarkurtic:
Current path for streaming of finetuning datasets does not allow for streaming from local path (which works for text datasets out of the box and is also supported by `StreamingFinetuningDataset` class). Diff just updates the way `local` is considered when building finetuning dataloaders. It has been tested on MPT-7B finetuning from local streaming dolly_hhrlhf dataset.

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -128,7 +128,7 @@ def build_finetuning_dataloader(cfg: DictConfig,
 
     dataset = None  # for pyright
     sampler = None
-    if cfg.dataset.get('remote') is not None:
+    if cfg.dataset.get('remote') is not None or cfg.dataset.get('local') is not None:
         # Build streaming dataloader
         dataset = dataset_constructor.build_from_streaming(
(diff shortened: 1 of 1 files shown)


dakinggg:
Hey @eldarkurtic , thanks for the change! Could you please run `pre-commit run --all-files` locally to apply the auto formatting?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -280,6 +280,11 @@ def _validate_config(dataset_cfg: DictConfig) -> None:
                 'Using a streaming dataset requires setting both `remote` and `local`, ' +\
                 'but dataset.local is None.'
             )
+        else:
+            if dataset_cfg.get('split') not in os.listdir(dataset_cfg.local):
```
dakinggg, line 284: I don't think this is exactly the right condition. If remote is not None, then local won't contain the split folder (yet).

eldarkurtic: Good catch. Will address comments in a day or so. Thanks for the feedback

PR #876, opened 2024-01-16
Add activation monitor callback option

bilal2vec:


dakinggg:
@bilal2vec want to update this to latest foundry and merge it in? Or close?

bilal2vec:
oops may as well clsose

PR #877, opened 2024-01-17
[XS] Fix delta conversion script regex bug

nancyhung:
Small bug in the delta conversion script where we should be stripping any trailing delimiters.


XiaohanZhangCMU (approved):
lgtm!

PR #881, opened 2024-01-18
Update model download utils to support ORAS

jerrychen109:
Adapted from #857 

* Updates args + fallback command
* Replaces the original cache server downloading logic with a more generic `download_from_http_fileserver` 

Tested each download function and fallback option locally


llmfoundry/utils/model_download_utils.py
```diff
@@ -183,53 +190,98 @@ def _recursive_download(
...
+
+    secrets = {}
+    for secret in ['username', 'password', 'registry']:
+        secrets[secret] = _read_secrets_file(
+            os.path.join(credentials_dir, secret))
+
+    with open(config_file, 'r', encoding='utf-8') as f:
+        configs = yaml.safe_load(f.read())
+
+    path = configs[model]
```
irenedea, line 256:
```suggestion
    path = configs['models'][model]
```

as per our config format changes to include tokenizers as well

irenedea (commented):
LGTM, pending a change to the configs that we will need to do

irenedea (approved):

PR #883, opened 2024-01-18
fsdp wrap refac

vchiley:
fsdp wrap refac

### DIFF
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -726,10 +726,6 @@ class MPTModel(MPTPreTrainedModel):
             **self.config.init_config,
         )
 
-    # FSDP Wrap function
-    def fsdp_wrap_fn(self, module: nn.Module) -> bool:
-        return isinstance(module, MPTBlock)
-
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
         return isinstance(module, MPTBlock)
@@ -887,10 +883,6 @@ class MPTForCausalLM(MPTPreTrainedModel):
             **self.config.init_config,
         )
 
-    # FSDP Wrap function
-    def fsdp_wrap_fn(self, module: nn.Module) -> bool:
-        return isinstance(module, MPTBlock)
-
     # Activation Checkpointing
     def activation_checkpointing_fn(self, module: nn.Module) -> bool:
         act_ckpt_list = getattr(self.config, 'activation_checkpointing_target',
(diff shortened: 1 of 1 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -995,6 +987,16 @@ def _reorder_cache(
...
 
 
+def _fsdp_wrap_fn(self: Union[MPTModel, MPTForCausalLM],
+                  module: nn.Module) -> bool:
+    # FSDP Wrap function for MPT Models
+    return isinstance(module, MPTBlock)
+
+
+MPTModel.fsdp_wrap_fn = _fsdp_wrap_fn
+MPTForCausalLM.fsdp_wrap_fn = _fsdp_wrap_fn
```
mvpatel2000, line 997: nit: Can u put fn above class?

Also dont monkeypatch, call helper fn directly from `fsdp_wrap_fn`?

vchiley: refac'd so that fn is above class
Note: pyright complains that `self` types (ie `Union[MPTModel, MPTForCausalLM]`) are not defined yet, so I had to add the `# type: ignore[no-untyped-def]`

mvpatel2000 (commented):
You don't need to direct attach it either!

Just do something like:

```
def wrap_fn(module):
  return isinstance(module, MPTBlock)
  
def fsdp_wrap(self, module):
  return wrap_fn(module)
```
no hackiness

vchiley:
> You don't need to direct attach it either!
> 
> Just do something like:
> 
> ```
> def wrap_fn(module):
>   return isinstance(module, MPTBlock)
>   
> def fsdp_wrap(self, module):
>   return wrap_fn(module)
> ```
> 
> no hackiness

what if we need self in wrap fn at some point?

llmfoundry/models/mpt/modeling_mpt.py
mvpatel2000: I sitll dont really like. 

Also youc an use from future import annotations to type self. Wrap name of type in quotes eg ('MPTModel') and it will work

mvpatel2000:
> > You don't need to direct attach it either!
> > Just do something like:
> > ```
> > def wrap_fn(module):
> >   return isinstance(module, MPTBlock)
> >   
> > def fsdp_wrap(self, module):
> >   return wrap_fn(module)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > no hackiness
> 
> what if we need self in wrap fn at some point?

Then add as arg but still call instead of directly attaching to class?

mvpatel2000 (approved):

PR #886, opened 2024-01-19
Switch to the Composer integration of LoRA (works with FSDP)

### LINKED ISSUE #414: LORA + FSDP: ValueError: `FlatParameter` requires uniform `requires_grad`
## To reproduce

1. add lora configs to llm-foundry/scripts/train/finetune_example/mpt-7b-arc-easy--gpu.yaml
```
lora:
  args:
    r: 16
    lora_alpha: 32
    lora_dropout: 0.05
    target_modules: ['Wqkv']
```

2. run composer train.py finetune_example/mpt-7b-arc-easy--gpu.yaml

## Expected behavior

FSDP + LORA fine-tuning 

## Config file

No other changes to example config file

dakinggg:
Switches the PEFT integration to use the integration from composer. Note: this does not yet support registering a PEFT model to unity catalog.

- [x] wait for a composer release with the LoRA changes (currently requires using composer dev)

Full manual testing:
- [x] llama and mpt, lora and no lora. compare loss curves, memory usage, throughput, and num trainable params
Screenshot 2024-02-01 at 6 30 23 PM
Screenshot 2024-02-01 at 6 41 01 PM
Screenshot 2024-02-01 at 6 41 06 PM
Screenshot 2024-02-01 at 6 41 20 PM
Screenshot 2024-02-01 at 6 42 49 PM

- [x] llama and mpt, lora and no lora. compare size of checkpoints in object store
- [x] llama and mpt, lora and no lora. compare files in UC registered model

Closes #414


setup.py
```diff
@@ -50,11 +50,11 @@
 ]
 
 install_requires = [
-    'mosaicml[libcloud,wandb,mlflow,oci,gcs]>=0.17.2,<0.18',
+    'mosaicml[libcloud,wandb,mlflow,oci,gcs]>=0.17.2,<0.19',  # TEMPORARY
```
dakinggg, line 53: fix before merge

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -77,196 +76,209 @@ def __init__(self, om_model_config: Union[DictConfig,
...
+            from transformers.models.llama.modeling_llama import LlamaAttention
+
+            from llmfoundry.models.layers.llama_attention_monkeypatch import \
+                get_llama_attention_patch_fn
+            LlamaAttention.forward = get_llama_attention_patch_fn(
+                attention_patch_type)
+            model.config.use_cache = False
+
+        if model.config.tie_word_embeddings and resolved_init_device == 'meta':
+            model.tie_weights()
```
irenedea, line 247: what does this do?

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -86,13 +88,13 @@ def hf_get_hidden_layers(model: PreTrainedModel) -> Any:
...
-        'gpt_neox.layers',  # GPTNeoX
+        'h',  # BLOOM, GPT2, GPTJ
+        'decoder.layers',  # OPT
+        'layers',  # GPTNeoX
         'block',  # T5, BART, Pegasus (from encoder)
         'layers',  # ProphetNet, Marian (from encoder)
-        'model.layers',  # LLaMa
-        'transformer.blocks',  # MPT
+        'layers',  # LLaMa
+        'blocks',  # MPT
```
irenedea, line 97: why dont' we need the prefixes anymore?

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -77,196 +76,209 @@ def __init__(self, om_model_config: Union[DictConfig,
...
-                                          init_device=init_device)
+        composer_model = super().__init__(
+            model=model,
+            shift_labels=True,
+            tokenizer=tokenizer,
+            metrics=train_metrics,
+            eval_metrics=eval_metrics,
+            z_loss=z_loss,
+            init_device=init_device,
+            peft_config=peft_config,
```
irenedea, line 281: is it possible to refactor such that where appropriate there's one helper per named argument?

irenedea: plus also a separate helper for patching attention would be nice

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -159,41 +162,31 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
...
+    # This is a hurdle for FSDP because they need to be in the same FSDP block
+    # These lines ensures that both modules stay together in the top-most block when
+    # the model has this tying enabled (almost all do; this property defaults to True)
+    if model.config.tie_word_embeddings:
+        causal_base_model._fsdp_wrap = False
+        tied_embeddings._fsdp_wrap = False
+        lm_head._fsdp_wrap = False
+
+    if hasattr(model, 'peft_type'):
+        if model.peft_type is not None:
```
irenedea, line 178: nit
```suggestion
    if hasattr(model, 'peft_type') and model.peft_type is not None:
```

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -159,41 +162,31 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
...
+            peft_type = model.peft_type.lower()
+            active_adapters = [
+                adapter.lower() for adapter in model.active_adapters
```
irenedea, line 181: here we check against `.lower()` and previously for checking peft_type we checked with `.upper()`. let's choose one and keep consistent

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -159,41 +162,31 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
...
+            ]
+            for name, module in model.named_modules():
+                if peft_type in name.lower() and any(
+                        adapter in name.lower() for adapter in active_adapters):
+                    has_parameters = any(True for _ in module.parameters())
+                    has_buffers = any(True for _ in module.buffers())
```
irenedea, line 187: super nit: this is a little clearer imo, but there isn't really a great solution to check if generators are empty in python
```suggestion
                    has_parameters = next(module.parameters(), None) is not None
                    has_buffers = next(module.buffers(), None) is not None
```

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -159,41 +162,31 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
...
-        # the weights of the self.lm_head and self.transformer.wte are tied.
-        # This tying occurs inside the `self.post_init()` function.
-        # This is a hurdle for FSDP because they need to be in the same FSDP block
-        # These lines ensures that both modules stay together in the top-most block when
-        # the model has this tying enabled (almost all do; this property defaults to True)
-        if model.config.tie_word_embeddings:
-            causal_base_model._fsdp_wrap = False
-            tied_embeddings._fsdp_wrap = False
-            lm_head._fsdp_wrap = False
+
```
irenedea, line 165: how come we no longer need to condition on init_device?

llmfoundry/models/hf/hf_fsdp.py
irenedea: i think it is also worthy of a helper function
```suggestion
                    has_parameters = _has_next(module.parameters())
                    has_buffers = _has_next(module.buffers())
```

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -159,41 +162,31 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
...
+            peft_type = model.peft_type.lower()
+            active_adapters = [
+                adapter.lower() for adapter in model.active_adapters
+            ]
+            for name, module in model.named_modules():
+                if peft_type in name.lower() and any(
+                        adapter in name.lower() for adapter in active_adapters):
+                    has_parameters = any(True for _ in module.parameters())
+                    has_buffers = any(True for _ in module.buffers())
+                    if has_parameters or has_buffers:
```
irenedea, line 188: can you add a comment explaining this?

tests/models/hf/test_fsdp_weight_tying.py
```diff
@@ -0,0 +1,97 @@
...
+
+    tokenizer = build_tokenizer(
+        tokenizer_name=tokenizer_name,
+        tokenizer_kwargs={'model_max_length': 32},
+    )
+
+    original_model = COMPOSER_MODEL_REGISTRY[model_cfg['name']](model_cfg,
+                                                                tokenizer)
+
+    lm_head = original_model.model.lm_head if peft_config is None else original_model.model.base_model.model.lm_head
```
irenedea, line 74: use a common var for `underlying_model`

```
underlying_model = original_model.model if peft_config is None else original_model.model.base_model.model

```

tests/models/hf/test_fsdp_weight_tying.py
```diff
@@ -0,0 +1,97 @@
...
+    model_cfg = {
+        'name': 'hf_causal_lm',
+        'pretrained_model_name_or_path': 'mistralai/Mistral-7B-v0.1',
+        'config_overrides': {
+            'num_hidden_layers': 2,
+            'hidden_size': 32,
+            'intermediate_size': 64,
+            'tie_word_embeddings': True,
+        },
+        'use_auth_token': True,
```
irenedea, line 43: does it need to be True?
```suggestion
```

irenedea: if so, do we have a HF auth token for ci cd?

tests/models/hf/test_hf_peft_wrapping.py
```diff
@@ -0,0 +1,106 @@
...
+        fsdp_config=fsdp_config,
+        train_dataloader=[],
+        device_train_microbatch_size=1,
+    )
+
+    model = trainer.state.model
+    lora_A = model.model.base_model.model.model.layers[
+        0].self_attn.q_proj.lora_A['default']
+    lora_B = model.model.base_model.model.model.layers[
+        0].self_attn.q_proj.lora_B['default']
```
irenedea, line 103:
```suggestion
    model = trainer.state.model
    underlying_model = model.model.base_model.model.model
    lora_A = underlying_model.layers[
        0].self_attn.q_proj.lora_A['default']
    lora_B = underlying_model.layers[
        0].self_attn.q_proj.lora_B['default']
```

irenedea: also, what's the diff between `state.model` and `state.model.model`?

irenedea: and also `state.model.model.base_model`, and so forth...

tests/models/hf/test_fsdp_weight_tying.py
irenedea: can we use `maybe_get_underlying_model` from composer after the release?

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -206,8 +208,10 @@ def check_hf_model_equivalence(model1: PreTrainedModel,
 
     assert expected_model_config_dict == new_model_config_dict
     assert all(
-        torch.equal(p1.cpu(), p2.cpu())
-        for p1, p2 in zip(model1.parameters(), model2.parameters()))
+        torch.equal(p1.cpu(), p2.cpu()) if (
```
irenedea, line 211: nit: rewrite without all + conditional true

```
for (n1, p1), (_, p2) in zip(...):
    if not just_lora or 'lora' in n1:
         assert torch.equal(p1.cpu(), p2.cpu())
```

llmfoundry/models/hf/model_wrapper.py
```diff
@@ -45,24 +47,19 @@ def __init__(self,
...
+                 init_device: Optional[str] = None,
+                 peft_config: Optional['PeftConfig'] = None):
         super().__init__(model,
                          tokenizer,
                          use_logits=True,
                          metrics=metrics,
                          eval_metrics=eval_metrics,
-                         shift_labels=shift_labels)
+                         shift_labels=shift_labels,
+                         peft_config=peft_config)
```
irenedea, line 58: will need value passed for `should_save_peft_only`

dakinggg: My intention was not to support that configuration option via LLM Foundry, and just always save the PEFT weights only (defaults to True).

tests/models/hf/test_fsdp_weight_tying.py
dakinggg: ah, no, just copy paste, will remove.

dakinggg: there is a jira somewhere to get ci to have hf token so we can test llama in ci.

tests/models/hf/test_fsdp_weight_tying.py
dakinggg: ah yeah, should be able to update to use that helper

tests/models/hf/test_hf_peft_wrapping.py
dakinggg: its just lots of layers of wrapped models...various rounds of composer hf model, peft model, lora model, transformers causal lm, transformers base model...can make some vars

llmfoundry/models/hf/hf_causal_lm.py
dakinggg: will add comment. HF's weight tying doesn't apply properly when initing on meta device, so we have to manually tie the weights after init.

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -77,196 +76,209 @@ def __init__(self, om_model_config: Union[DictConfig,
...
+        )
 
         return composer_model
```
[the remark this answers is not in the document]
dakinggg: hm yeah i don't know why this was added...lets see if I can find when it was added.

llmfoundry/models/hf/model_wrapper.py
irenedea: gotcha, can you pass true explicitly then?

llmfoundry/models/hf/hf_causal_lm.py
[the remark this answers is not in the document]
dakinggg: ok, all good, removed.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -220,7 +220,18 @@ def _save_checkpoint(self, state: State, logger: Logger):
                 # TODO: after torch 2.1, we can load a state dict into a meta model
                 # and skip the extra model init
                 log.debug(f'Creating new model instance')
-                new_model_instance = type(original_model)(copied_config)
+
```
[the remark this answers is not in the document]
dakinggg: To avoid messing with the model on state, which is sharded and might not save correctly. tbf it might have worked, but I wanted to avoid relying on the HF save_pretrained code meshing with our FSDP context manager for getting the full state dict.

llmfoundry/models/hf/hf_fsdp.py
dakinggg: I refactored below for this func to take in `causal_base_model` instead of `model`, cause it was cleaner that way.

llmfoundry/models/hf/model_wrapper.py
dakinggg: sure

llmfoundry/models/hf/hf_fsdp.py
dakinggg: This is the main thing I fixed in https://github.com/mosaicml/composer/pull/2765. It used to hang without the special case for mixed init, but that has been resolved in composer.

llmfoundry/callbacks/hf_checkpointer.py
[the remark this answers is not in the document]
dakinggg: I think it might work calling `save_pretrained(..., is_main_process=dist.get_global_rank()==0, ...)` inside the FSDP context manager, but I try to avoid relying on Hugging Face internals wherever possible.

llmfoundry/models/hf/hf_fsdp.py
dakinggg: we are basically sticking with `.lower()` everywhere because that makes way more sense, except that you _have_ to make them upper when using them with the HF `PeftConfig` (that uppercasing is handled by composer).

llmfoundry/models/hf/hf_fsdp.py
dakinggg: I don't really think this one deserves a helper since its so small and only happening right here. Can switch to your implementation though, I think they're about equal readibility and the variable name makes it pretty clear what is happening.

llmfoundry/models/hf/hf_fsdp.py
dakinggg: updated the docstring

llmfoundry/models/hf/hf_causal_lm.py
dakinggg: did a bit of refactoring. I'm sure there is more that could be done, but I'm gonna say that is out of scope for this PR :) Please review again though as I moved a bunch of stuff around.

llmfoundry/models/hf/hf_fsdp.py
irenedea: any reason why `layers` is repeated?

dakinggg: cause i didn't notice it was the same after removing the prefixes :)

dakinggg: fixed

llmfoundry/callbacks/hf_checkpointer.py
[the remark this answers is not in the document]
dakinggg: resolving, lmk if you disagree

llmfoundry/models/hf/hf_fsdp.py
dakinggg: resolving, lmk if you disagree

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -76,197 +114,181 @@ def __init__(self, om_model_config: Union[DictConfig,
...
-                if om_model_config.pretrained:
-                    model = AutoModelForCausalLM.from_pretrained(
+                        f'Config dict override got unknown keys. ' +
+                        f'Extra keys: {extra_keys}. ' +
+                        f'Expected (a subset of) keys: {list(attr.keys())}.')
+                getattr(config, k).update(v)
+            # necessary case to allow for rope_scaling to be overriden in llama config
+            elif attr is None and isinstance(v, Mapping):
+                setattr(config, k, {})
+                getattr(config, k).update(v)
```
irenedea, line 165: is this handled by the else condition below?

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -76,197 +114,181 @@ def __init__(self, om_model_config: Union[DictConfig,
...
+
+        from llmfoundry.models.layers.llama_attention_monkeypatch import \
+            get_llama_attention_patch_fn
+        LlamaAttention.forward = get_llama_attention_patch_fn(
+            attention_patch_type)
+        model.config.use_cache = False
+
+    @staticmethod
+    def _get_peft_config(peft_config_dict: Dict[str, Any]) -> 'PeftConfig':
+        if peft_installed:
```
irenedea, line 278: nit do early fail and maybe raise ImportError instead? 
```
        if not peft_installed:
              raise ImportError(...)
```

out of scope for PR, but it would be nice to have some error utiities like composer's `MissingConditionalImportError`

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -78,21 +80,21 @@ def hf_get_causal_base_model(model: PreTrainedModel) -> Any:
...
     NOTE: Different model configurations have different hidden layer attribute names.
-        - transformer.h: (BloomForCausalLM, GPT2LMHeadModel, GPTJForCausalLM)
-        - model.decoder.layers: (OPTForCausalLM)
-        - gpt_neox.layers: (GPTNeoXForCausalLM)
-        - model.layers: (LlaMaForCausalLM)
-        - transformer.blocks: (MPTForCausalLM)
+        - h: (BloomForCausalLM, GPT2LMHeadModel, GPTJForCausalLM)
+        - decoder.layers: (OPTForCausalLM)
+        - layers: (GPTNeoXForCausalLM)
+        - layers: (LlaMaForCausalLM)
```
irenedea, line 89:
```suggestion
        - layers: (GPTNeoXForCausalLM, LlaMaForCausalLM)
```

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -159,41 +162,31 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel,
...
+    # This is a hurdle for FSDP because they need to be in the same FSDP block
+    # These lines ensures that both modules stay together in the top-most block when
+    # the model has this tying enabled (almost all do; this property defaults to True)
+    if model.config.tie_word_embeddings:
+        causal_base_model._fsdp_wrap = False
+        tied_embeddings._fsdp_wrap = False
+        lm_head._fsdp_wrap = False
+
+    # PEFT layers should be individually wrapped
+    # TODO: Revisit this if we enforce use_orig_params=True, which seems to support
```
irenedea, line 178: what benefit do we get from `use_orig_params=True`?

tests/models/hf/test_fsdp_weight_tying.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    model = trainer.state.model
+    lm_head = model.model.lm_head if peft_config is None else model.model.base_model.model.lm_head
+    embedding_layer = model.model.model.embed_tokens if peft_config is None else model.model.base_model.model.model.embed_tokens
+
+    lm_head_id = id(lm_head.weight)
+    embedding_layer_id = id(embedding_layer.weight)
+
+    assert lm_head_id == embedding_layer_id
```
irenedea, line 98: sorry, what's the purpose of this vs the checks before instantiating trainer?

tests/models/hf/test_hf_peft_wrapping.py
```diff
@@ -0,0 +1,104 @@
...
+    trainer = Trainer(
+        model=original_model,
+        device='gpu',
+        fsdp_config=fsdp_config,
+        train_dataloader=[],
+        device_train_microbatch_size=1,
+    )
+
+    model = trainer.state.model
+    underlying_model = model.model.base_model.model
```
irenedea, line 99: can you use maybe_get_underlying_model here?

tests/models/hf/test_fsdp_weight_tying.py
```diff
@@ -0,0 +1,98 @@
...
+        model=original_model,
+        device='gpu',
+        fsdp_config=fsdp_config,
+        train_dataloader=[],
+        device_train_microbatch_size=1,
+    )
+
+    model = trainer.state.model
+    lm_head = model.model.lm_head if peft_config is None else model.model.base_model.model.lm_head
+    embedding_layer = model.model.model.embed_tokens if peft_config is None else model.model.base_model.model.model.embed_tokens
```
irenedea, line 93: can you use `maybe_get_underlying_model` here?

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -590,23 +627,54 @@ def test_huggingface_conversion_callback(
...
-                                 f'ba{batches_per_epoch}'),
+                    checkpoint_path,
                     trust_remote_code=True,
                 )
 
             # Check that the loaded model has the correct precision, and then set it back
             # to the original for the equivalence check
-            assert loaded_model.config.torch_dtype == precision
-            loaded_model.config.torch_dtype = original_model.model.config.torch_dtype
+            if peft_config is None:
```
irenedea, line 675: do we need to do a precision check for the peft_config is not None case?

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -381,8 +384,25 @@ def test_huggingface_conversion_callback_interval(
 @pytest.mark.world_size(2)
 @pytest.mark.gpu
 @pytest.mark.parametrize(
-    'model,tie_word_embeddings',
-    [('mpt', True), ('mpt', False), ('neo', None), ('llama2', None)],
+    'model,tie_word_embeddings,peft_config',
```
irenedea, line 387: this test is getting kind of long, would be great to refactor. don't feel strongly if it's in this PR or as a follow up.

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -76,197 +114,181 @@ def __init__(self, om_model_config: Union[DictConfig,
...
+        # If the HuggingFace model is coming from a local folder, Hugging Face copies the modules into the
+        # transformers modules cache. On particular systems, this operation seems to cause contention between
+        # the different processes. To avoid this contention, we first create the model (on meta device) on local rank
+        # zero. This will set up the transformers model cache and avoid the future contention.
+        if dist.get_local_rank() == 0 and os.path.isdir(
+                om_model_config.pretrained_model_name_or_path):
+            with init_empty_weights(include_buffers=False):
+                with warnings.catch_warnings():
+                    warnings.simplefilter('ignore', UserWarning)
+                    AutoModelForCausalLM.from_pretrained(
```
irenedea, line 183: should this warning filter be added for other `AutoModelForCausalLM.from_pretrained` callsites? if so, a helper to do this would be nice.

irenedea (approved):
LGTM! Had some minor comments and questions.

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -57,14 +52,57 @@ class ComposerHFCausalLM(HuggingFaceModelWithZLoss):
...
-        # set up training and eval metrics
+        if not om_model_config.get('trust_remote_code',
+                                   True) and om_model_config.get(
+                                       'pretrained_model_name_or_path',
+                                       '').startswith('mosaicml/mpt'):
+            raise ValueError(
+                'trust_remote_code must be set to True for MPT models. Without this, the MPT model code will come from the transformers library, '
+                +
+                'which is not significantly slower and not compatible with the LLM foundry training code, rather than the code release by MosaicML.'
+            )
```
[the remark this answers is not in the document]
dakinggg: llama and mistral are implemented directly by huggingface, and is subject to huggingface's decisions and priorities. We want to retain control over the MPT code for usage with LLM Foundry, so we require using the code we have published.

For inference they are compatible, and most inference servers have a separate implementation optimized for inference anyway.

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -76,197 +114,181 @@ def __init__(self, om_model_config: Union[DictConfig,
...
+            train_metrics = []
+
+        # Construct the Hugging Face config to use
+        config = AutoConfig.from_pretrained(
+            om_model_config.pretrained_model_name_or_path,
+            trust_remote_code=trust_remote_code,
+            use_auth_token=use_auth_token,
+            attn_implementation=requested_attention_implementation,
+            use_cache=
+            False,  # Necessary due to https://github.com/huggingface/transformers/issues/28056
```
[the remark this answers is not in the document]
dakinggg: I don't think it was, but deferring confirming to outside the scope of this PR.

llmfoundry/models/hf/hf_causal_lm.py
dakinggg: none of this code is new, its just an unindent from before. update and setattr are different though, so no.

llmfoundry/models/hf/hf_fsdp.py
dakinggg: It is strongly preferred because the parameters names don't get changed by FSDP and its flat params. I believe pytorch is going to deprecate `use_orig_params=False` at some point.

tests/models/hf/test_fsdp_weight_tying.py
dakinggg: we want to check before (we set up the model right) and after (trainer didn't mess up the model)

tests/models/hf/test_hf_peft_wrapping.py
dakinggg: Unfortunately not easily, because that function is not meant to be applied to an FSDP wrapped model.

tests/models/hf/test_fsdp_weight_tying.py
dakinggg: same

tests/a_scripts/inference/test_convert_composer_to_hf.py
dakinggg: The config is the base model's config, and afaict the peft config doesn't have a precision option, so no, this is just for when not using peft.

tests/a_scripts/inference/test_convert_composer_to_hf.py
dakinggg: deferring :)

llmfoundry/models/hf/hf_causal_lm.py
dakinggg: No, this is just for this "pre" creation on meta device, cause it will raise a bunch of (generally correct) warnings that you're trying to load a checkpoint into a model with meta tensors which is probably not what you intended.

PR #888, opened 2024-01-19
Bump Composer version to 0.18

dakinggg:


dakinggg:
Closing because we will go to 0.19 when that is released.

PR #890, opened 2024-01-20
Fix tiktoken tokenizer add_generation_prompt

irenedea:
We should only add the generation prompt after the very last message.

### Code
```
chat = [...]
print(tokenizer.apply_chat_template(chat, tokenize=False, add_generation_prompt=True))
```
### Before Output
```
<|im_start|>user
Please summarize the goals in this text:

Going outside has benefits include reducing stress and triggering the relaxation response, which can help us not only feel better mentally, but even heal faster from physical ailments.<|im_end|>
<|im_start|>assistant

<|im_start|>assistant
You should go outside and touch grass.<|im_end|>
<|im_start|>assistant

<|im_start|>user
What else can I do?<|im_end|>
<|im_start|>assistant

```
### After Output
```
<|im_start|>user
Please summarize the goals in this text:

Going outside has benefits include reducing stress and triggering the relaxation response, which can help us not only feel better mentally, but even heal faster from physical ailments.<|im_end|>
<|im_start|>assistant
You should go outside and touch grass.<|im_end|><|endoftext|>
<|im_start|>user
What else can I do?<|im_end|>
<|im_start|>assistant
```

### DIFF
--- a/llmfoundry/tokenizers/tiktoken.py
+++ b/llmfoundry/tokenizers/tiktoken.py
@@ -198,7 +198,7 @@ class TiktokenTokenizerWrapper(PreTrainedTokenizer):
             '{% else %}'
             "{{ '\n' + '<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' }}"
             '{% endif %}'
-            '{% if (add_generation_prompt == true) %}'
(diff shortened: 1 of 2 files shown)


dakinggg (approved):
Thanks for the catch!

PR #893, opened 2024-01-22
Tiktoken chat format fix

rajammanabrolu:
small fix to add end tokens after system prompt in jinja chat template (ty @samhavens for catching this) + change tests accordingly

### DIFF
--- a/tests/tokenizers/test_tiktoken.py
+++ b/tests/tokenizers/test_tiktoken.py
@@ -73,15 +73,14 @@ Please summarize the goals in this text:
 
 Going outside has benefits include reducing stress and triggering the relaxation response, which can help us not only feel better mentally, but even heal faster from physical ailments.<|im_end|>
 <|im_start|>assistant
-You should go outside and touch grass.<|im_end|><|endoftext|>""",
-    """<|im_start|>system
+You should go outside and touch grass.<|im_end|>""", """<|im_start|>system
 You are a honest and helpful AI language model. Tell the user the truth, the whole truth, and nothing but the truth.<|im_end|>
 <|im_start|>user
 Please summarize the goals in this text:
 
 Going outside has benefits include reducing stress and triggering the relaxation response, which can help us not only feel better mentally, but even heal faster from physical ailments.<|im_end|>
 <|im_start|>assistant
-You should go outside and touch grass.<|im_end|><|endoftext|>"""
+You should go outside and touch grass.<|im_end|>"""
 ]
 
 MULTI_TURN_CHAT_STRING_SYSTEM_PROMPT = [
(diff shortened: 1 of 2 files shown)


tests/tokenizers/test_tiktoken.py
```diff
@@ -86,15 +86,15 @@
...
 MULTI_TURN_CHAT_STRING_SYSTEM_PROMPT = [
     """<|im_start|>system
-You are a helpful, respectful and honest assistant. Always answer as helpfully as possible.
+You are a helpful, respectful and honest assistant. Always answer as helpfully as possible.<|im_end|>
 <|im_start|>user
 Please summarize the goals in this text:
 
 Going outside has benefits include reducing stress and triggering the relaxation response, which can help us not only feel better mentally, but even heal faster from physical ailments.<|im_end|>
 <|im_start|>assistant
 You should go outside and touch grass.<|im_end|><|endoftext|>""",
```
irenedea, line 95: Should we also remove the eos token here?

rajammanabrolu: I think it's ok? it matches the chatml dict version https://github.com/mosaicml/llm-foundry/blob/raj/tiktoken-fixes/tests/tokenizers/test_tiktoken.py#L45-L53 so assuming that is all to the conversation tuns it should end with an eos - lmk if I'm missing something

irenedea: We don't have the eos token there for mpt-7b-8k-chat, for example. (https://huggingface.co/mosaicml/mpt-7b-8k-chat/discussions/7). I'm not familiar with the standard practice here, @samhavens?

rajammanabrolu: so looking at it more @irenedea with the mistral and llama models https://huggingface.co/HuggingFaceH4/zephyr-7b-beta/blob/main/tokenizer_config.json#L34 and https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/tokenization_llama.py#L464 - it seems that all have the eos token (at the very least) after the final assistant turn (unless add_generation_prompt is used)

rajammanabrolu: it might be good to change the mpt chat tokenizers to also do this as this seems to be standard now?

samhavens: I think that Llama-family models are using EOS similar to how we use `<|im_end|>` but not identical, it separates user/assistant pairs. 

In the transformer docs about chat templates, they have a [reference ChatML implementation](https://huggingface.co/docs/transformers/main/en/chat_templating#what-template-should-i-use) that does not include EOS.

In any use case I can think of, we'd stop generating on `<|im_end|>`, and EOS would be for separating sequences which are being packed for training. So I don't think it would hurt to append at the end of an assistant response but I don't think it is necessary and I wouldn't add it

rajammanabrolu: @samhavens Going off of your example of packing sequences for training - the eos delimiter between different conversation transcripts could be useful for Continued Pretraining (concating lots of documents of conv transcripts together) and for passing stuff into reward models, etc. - this can also be done separately by defining new templates while formatting for such use cases but seems cleaner to do it with one template?

rajammanabrolu: Apologies for the delay, after talking more with @samhavens I have removed all the eos tokens to make it consistent with the MPT chat tokenizer. We can revisit this later if having no EOS during sequence packing or if other things come up. @irenedea lmk if this is satisfactory

irenedea (approved):
LGTM! Thanks @rajammanabrolu!

PR #895, opened 2024-01-23
Add `tokenizer-only` flag to only download tokenizers from HF or oras

irenedea:
## Manual tests
Download from oras: `test-dl-tokenizer-m3T8eq`
Download from hf: `test-dl-tokenizer-inG1Zp`


scripts/misc/download_model.py
```diff
@@ -7,10 +7,11 @@
     python download_model.py hf --model mosaicml/mpt-7b --save-dir <save_dir> --token <token>
 
 Download from ORAS registry:
-    python download_model.py oras --registry <registry> --path mosaicml/mpt-7b --save-dir <save_dir>
+    python download_model.py oras --model mosaicml/mpt-7b --config-file <config_file> \
+        --credentials-dir <credentials_dir> --save-dir <save_dir>
 
 Download from an HTTP file server:
-    python download_model.py http --host https://server.com --path mosaicml/mpt-7b --save-dir <save_dir>
+    python download_model.py http --url https://server.com/path --save-dir <save_dir>
```
jerrychen109, line 14:
```suggestion
    python download_model.py http --url https://server.com/models/mosaicml/mpt-7b/ --save-dir <save_dir>
```

scripts/misc/download_model.py
```diff
@@ -56,6 +57,9 @@ def parse_args() -> argparse.Namespace:
 
     base_parser = argparse.ArgumentParser(add_help=False)
     base_parser.add_argument('--save-dir', type=str, required=True)
+    base_parser.add_argument('--tokenizer-only',
+                             default=False,
+                             action='store_true')
```
jerrychen109, line 62:
```suggestion
    base_parser.add_argument('--tokenizer-only',
                             type=bool,
                             default=False,
                             action='store_true')
```

scripts/misc/download_model.py
```diff
@@ -85,6 +89,9 @@ def parse_args() -> argparse.Namespace:
     download_from = args.download_from
 
     if download_from == 'http':
+        if args.tokenizer_only == True:
```
jerrychen109, line 92:
```suggestion
        if args.tokenizer_only:
```
If we specify the type and have the default, it'll be either true or false

llmfoundry/utils/model_download_utils.py
```diff
@@ -221,16 +232,18 @@ def download_from_oras(model: str,
                        config_file: str,
                        credentials_dir: str,
                        save_dir: str,
+                       tokenizer_only: bool,
```
jerrychen109, line 235:
```suggestion
                       tokenizer_only: bool = False,
```
for consistency with `download_from_hf_hub`

llmfoundry/utils/model_download_utils.py
```diff
@@ -57,6 +64,7 @@ def download_from_hf_hub(
         save_dir (str, optional): The local path to the directory where the model files will be downloaded.
         prefer_safetensors (bool): Whether to prefer Safetensors weights over PyTorch weights if both are
             available. Defaults to True.
+        tokenizers_only (bool): If true, only download tokenzier files.
```
jerrychen109, line 67:
```suggestion
        tokenizer_only (bool): If true, only download tokenizer files.
```
Thinking singular makes more sense?

llmfoundry/utils/model_download_utils.py
```diff
@@ -94,11 +102,14 @@ def download_from_hf_hub(
             f'No supported model weights found in repo {model}.' +
             ' Please make sure the repo contains either safetensors or pytorch weights.'
         )
+    
+    allow_patterns = TOKENIZER_FILES if tokenizers_only else None
```
jerrychen109, line 106:
```suggestion
    allow_patterns = TOKENIZER_FILES if tokenizer_only else None
```
For consistency with above suggestion

scripts/misc/download_model.py
```diff
@@ -109,7 +116,8 @@ def parse_args() -> argparse.Namespace:
         download_from_hf_hub(args.model,
                              save_dir=args.save_dir,
                              token=args.token,
+                             tokenizers_only=args.tokenizer_only,
```
jerrychen109, line 119:
```suggestion
                             tokenizer_only=args.tokenizer_only,
```
For consistency with singular `tokenizer_only`

scripts/misc/download_model.py
```diff
@@ -109,7 +116,8 @@ def parse_args() -> argparse.Namespace:
...
                              prefer_safetensors=args.prefer_safetensors)
     elif download_from == 'oras':
         download_from_oras(args.model, args.config_file, args.credentials_dir,
-                           args.save_dir, args.concurrency)
+                           args.save_dir, args.tokenizer_only, args.concurrency)
```
jerrychen109, line 123:
```suggestion
                           args.save_dir, tokenizers_only=args.tokenizer_only, args.concurrency)
```
more consistency

jerrychen109 (commented):
mostly lgtm - added some suggestions for consistency in the code. Let me know if they make sense and I'll approve after!

scripts/misc/download_model.py
irenedea: apparently, you cannot specify type with action. I initially had [that](https://github.com/mosaicml/llm-foundry/pull/895/commits/936d5e566d0db411f7f5a93e3e2bffce4194b109#diff-b91dc3e13956a84272e7cb09c99a864c10739a26142341d239ffd53596a887eeR59), but had to remove.

scripts/misc/download_model.py
irenedea: yeah with the above comment, I cannot specify the type with store_action. Maybe that's safe enough? I just preferred to be explicit in case. lmk what you think

llmfoundry/utils/model_download_utils.py
irenedea: ah yes you're right! good catch

scripts/misc/download_model.py
jerrychen109: I think it's still safe with the default value since `store_true` implies bool already so I don't think a user can provide a weird value, but not super opinionated here. I'll leave it to you!

jerrychen109 (approved):

PR #896, opened 2024-01-23
Allow for bool inputs for loggers

ngcgarcia:
`MLFlowLogger` takes in a `log_system_metrics=<bool>` parameter that we'd like to support.


ngcgarcia:
going to merge directly into main instead https://github.com/mosaicml/llm-foundry/pull/897

PR #897, opened 2024-01-23
Allow bool input for loggers

ngcgarcia:
`MLFlowLogger` takes in a `log_system_metrics=<bool>` parameter that we'd like to support.


llmfoundry/utils/builders.py
```diff
@@ -220,7 +220,7 @@ def build_callback(
 
 def build_logger(name: str, kwargs: Dict[str, Any]) -> LoggerDestination:
     kwargs_dict = {
-        k: v if isinstance(v, str) else om.to_container(v, resolve=True)
+        k: v if isinstance(v, str) or isinstance(v, bool) else om.to_container(v, resolve=True)
```
dakinggg, line 223: oops, I think this should be different though. I think it should be `k: v if not isinstance(v, (ListConfig, DictConfig)) else om.to_container(v, resolve=True). wdyt? cc @aspfohl cause blame says that you wrote this line

ngcgarcia: lgtm

aspfohl: Oof, yea I think it should be something like that. Otherwise we would need to consider other types too (int, float, list, etc)

Should we be resolving earlier (eg in train.py?) @dakinggg? I find myself fighting with typing + omegaconf more than I'd like to admit

dakinggg: Good point. I think the `pop_config` call in `train.py` for the `loggers` key should actually just have `convert=True`.

dakinggg: There is a `om.resolve(cfg)` at the top of `train.py`

ngcgarcia:
see: `logger-change-Mv4zez`

also shows E2E that changes to composer work with hiding system metrics via `log_system_metrics=False` and `ignore_metrics=['memory/*', 'throughput/*']` https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/565784758740344/runs/ccf7ee9bbbad44879ba20723aedf7c2a/model-metrics?o=3360802220363900

dakinggg (approved):

PR #904, opened 2024-01-24
Warn instead of error on tokenizer-only download with http

jerrychen109:
When download from HTTP is specified with `tokenizer-only`, let's just warn that all files will be downloaded from the file server path rather than erroring out entirely.

### DIFF
--- a/scripts/misc/download_model.py
+++ b/scripts/misc/download_model.py
@@ -90,8 +90,9 @@ if __name__ == '__main__':
 
     if download_from == 'http':
         if args.tokenizer_only:
-            raise ValueError(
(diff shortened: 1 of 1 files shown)


nancyhung (approved):
can you test that this fixes the issue before merging in and link the run name in the PR description?

irenedea (approved):
LGTM, @jerrychen109 do you still want to merge this?

jerrychen109:
Oops yeah let me quickly re-run for proof that it still works

jerrychen109:
Tested on r12z3 interactive after cloning this branch:
image

Compared to main branch without this change:
image

PR #907, opened 2024-01-24
Add new GC option

dakinggg:
Screenshot 2024-01-24 at 11 58 19 AM

### DIFF
--- a/llmfoundry/callbacks/scheduled_gc_callback.py
+++ b/llmfoundry/callbacks/scheduled_gc_callback.py
@@ -36,12 +36,6 @@ class ScheduledGarbageCollector(Callback):
         self.eval_keep_disabled = eval_keep_disabled
         self.gc_init_state = None
 
-    def batch_start(self, state: State, logger: Logger) -> None:
-        del logger
-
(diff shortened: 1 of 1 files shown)


llmfoundry/callbacks/scheduled_gc_callback.py
```diff
@@ -19,19 +20,28 @@ class ScheduledGarbageCollector(Callback):
...
         self.batch_interval = batch_interval
+        self.gen_1_batch_interval = gen_1_batch_interval
         self.eval_keep_disabled = eval_keep_disabled
         self.gc_init_state = None
 
+    def batch_start(self, state: State, logger: Logger) -> None:
+        del logger
+
+        if self.gen_1_batch_interval is not None and state.timestamp.batch.value % self.gen_1_batch_interval == 0:
+            gc.collect(1)
```
mvpatel2000, line 43: Why is this at batch_start but the other one is before_dataloader?

Can we just change functionality to always gc.collect(1) if we think it's a bug fix?

dakinggg: moved to before_dataloader. discussed the other bit offline

mvpatel2000 (approved):

PR #908, opened 2024-01-25
No symlinks at all for HF download

jerrychen109:
To stay consistent with ORAS download, disable symlinking for Hugging Face downloads so the save dir contains the real file data. 

This is also necessary if we're using the HF download function to initialize an HTTP fileserver, since they don't always play nicely with symlinks.

Before:
Screenshot 2024-01-24 at 4 44 27 PM

After:
image


nancyhung (approved):
🚀

PR #914, opened 2024-01-26
Fix fmapi_chat for instruct models and custom tokenizers

aspfohl:
🙃 https://stackoverflow.com/questions/67412925/what-is-the-difference-between-lentokenizer-and-tokenizer-vocab-size

Tested via:
```yaml
device_eval_batch_size: 4
icl_tasks:
-
  label: jeopardy
  dataset_uri: eval/local_data/world_knowledge/jeopardy_all.jsonl
  num_fewshot: [3]
  icl_task_type: language_modeling
  continuation_delimiter: "\nAnswer: " # this separates questions from answers
  has_categories: true
icl_subset_num_batches: 1
max_seq_len: 1024
models:
- model:
    local: true
    name: fmapi_chat
  model_name: mpt-example
  tokenizer:
    name: mosaicml/mpt-7b-8k-instruct
seed: 1
```

### DIFF
--- a/llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
+++ b/llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
@@ -204,7 +204,7 @@ class OpenAIChatAPIEvalWrapper(OpenAIEvalInterface):
             expected_cont_tokens = tokens[cont_idxs[0]:cont_idxs[-1] + 1]
             output_logits = torch.nn.functional.one_hot(
                 torch.tensor(tokens[1:cont_idxs[0]]),
-                num_classes=self.tokenizer.vocab_size)
+                num_classes=len(self.tokenizer))
 
             prompt = self.tokenizer.decode(tokens[:cont_idxs[0]])
             next_logit_tensor = self.get_next_token_logit_tensor(
@@ -214,7 +214,7 @@ class OpenAIChatAPIEvalWrapper(OpenAIEvalInterface):
                 output_logits = torch.cat([output_logits, next_logit_tensor])
             padding = torch.nn.functional.one_hot(
                 torch.full((seqlen - output_logits.shape[0],), padding_tok),
-                num_classes=self.tokenizer.vocab_size)
+                num_classes=len(self.tokenizer))
             output_logits = torch.cat([output_logits, padding])
             output_logits_batch.append(output_logits)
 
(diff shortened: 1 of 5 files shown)


llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
```diff
@@ -228,9 +228,14 @@ def process_result(self, completion: Optional['ChatCompletion']):
             tensors = []
             for t in self.tokenizer(
                     completion.choices[0].message.content)['input_ids']:
-                tensors.append(
-                    self.tokenizer.construct_logit_tensor(
-                        {self.tokenizer.decode([t]): 0.0}))
+
+                # Construct tensor of shape (vocab_size,) with logprobs for each token
```
aspfohl, line 232: From https://github.com/mosaicml/llm-foundry/blob/ac7835415cf6fa91664e513c7a471f2802cdc8a2/llmfoundry/tokenizers/tiktoken.py#L361

dakinggg: Should this fix just be in `tiktoken.py`? or why are you copying it over here?

dakinggg: nvm

dakinggg (commented):
Want to add a test that would've caught your issues?

aspfohl:
@dakinggg  Good call, caught an issue in interface.py that would have caused the same bug when I got to casual models.. 

Updated [8a7c929](https://github.com/mosaicml/llm-foundry/pull/914/commits/8a7c929c17d70ba54152fc90b0be7566bcf1d2b2)

dakinggg (approved):
Would be nice if @maxisawesome or @bmosaicml also took a glance

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
```diff
@@ -228,9 +228,15 @@ def process_result(self, completion: Optional['ChatCompletion']):
...
+                tokenizer_logprobs = {self.tokenizer.decode([t]): 0.0}
```
maxisawesome, line 233: Do we want to make this 1.0?

llmfoundry/models/inference_api_wrapper/openai_causal_lm.py
```diff
@@ -228,9 +228,15 @@ def process_result(self, completion: Optional['ChatCompletion']):
...
+                tensor = torch.tensor([min(tokenizer_logprobs.values()) - 1] *
```
maxisawesome, line 234: I know all this is taken from tiktoken.py but this logic doesn't make sense to me.
L233: set each log prob to 0 in a dict
L234/235: take minimum of all tokenizer_logprobs (we set them all to 0) and subtract 1 to get -1, duplicate that value (always -1?) to make a tensor of size len(tokenizer). 
L236-238: at tensor location for each token in tokenizer_logprobs, set probability to 0.0 instead of -1.

So we're not returning tru logprobs at all if I'm not mistaken?

maxisawesome: @bmosaicml

aspfohl: [Silliest refactor ever](https://github.com/mosaicml/llm-foundry/pull/914/commits/861a2bd66ee2cc1c79d1e38c11b32d3de7b5714b), but this should make this significantly clearly

bmosaicml: OpenAI only returns top 5 logprobs, but there's still `len(tokenizer) - 5 logprobs` that need to be mocked in order for the metric's math to be computed properly. n.b. recall that the metrics expect to be able to index into a tensor of size len(tokenizer) and inspect the logprob there

bmosaicml:
Great catch! I can only imagine how hard this bug was to track down. 🙃

bmosaicml (approved):

PR #915, opened 2024-01-27
Add the model license file for mlflow

dakinggg:
Log the license file as an artifact to mlflow runs.

Screenshot 2024-01-26 at 4 08 31 PM
Screenshot 2024-01-26 at 4 36 41 PM

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -280,11 +280,12 @@ class HuggingFaceCheckpointer(Callback):
                             **self.mlflow_logging_config,
                         )
 
-                        if os.path.exists(
-                                os.path.join(local_save_path, 'LICENSE.txt')):
+                        license_path = os.path.join(local_save_path,
+                                                    'LICENSE.txt')
+                        if os.path.exists(license_path):
                             mlflow_logger._mlflow_client.log_artifact(
                                 mlflow_logger._run_id,
-                                os.path.join(local_save_path, 'LICENSE.txt'),
+                                license_path,
                             )
 
                         mlflow_logger.register_model(
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -6,6 +6,7 @@ import copy
 import logging
 import math
 import os
+import re
 import tempfile
 from pathlib import Path
 from typing import Optional, Sequence, Union
@@ -26,6 +27,23 @@ from llmfoundry.utils.huggingface_hub_utils import \
     edit_files_for_hf_compatibility
 
 log = logging.getLogger(__name__)
+
+_LICENSE_FILE_PATTERN = re.compile(r'license(\.[a-z]+|$)', re.IGNORECASE)
+
+
+def _maybe_get_license_filename(local_dir: str) -> Optional[str]:
+    """Returns the name of the license file if it exists in the local_dir.
+
+    Note: This is intended to be consistent with the code in MLflow.
+    https://github.com/mlflow/mlflow/blob/5d13d6ec620a02de9a5e31201bf1becdb9722ea5/mlflow/transformers/__init__.py#L1152
+
+    If the license file does not exist, returns None.
+    """
+    try:
+        return next(file for file in os.listdir(local_dir)
+                    if _LICENSE_FILE_PATTERN.search(file))
+    except StopIteration:
+        return None
 
 
 class HuggingFaceCheckpointer(Callback):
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -50,7 +50,8 @@
 ]
 
 install_requires = [
-    'mosaicml[libcloud,wandb,mlflow,oci,gcs]>=0.17.2,<0.18',
+    'mosaicml[libcloud,wandb,oci,gcs]>=0.17.2,<0.18',
+    'mlflow>=2.10,<3',
```
mvpatel2000, line 54: does this need to be in requires? can we gate mlflow in the hf ckptr?

dakinggg: we could move it to optional, but 🤷‍♂️ theres lots of stuff in this list that could be optional. can jira to clean it up more generally if you want. would technically be a breaking change to move mlflow out of the required deps.

mvpatel2000: It isn't correctly in install_requires so it wouldn't be breaking change? not sure i follow.

mvpatel2000: It isn't correctly in install_requires so it wouldn't be breaking change? not sure i follow.

mvpatel2000 (approved):

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -279,6 +279,14 @@ def _save_checkpoint(self, state: State, logger: Logger):
                             path=local_save_path,
                             **self.mlflow_logging_config,
                         )
+
+                        if os.path.exists(
+                                os.path.join(local_save_path, 'LICENSE.txt')):
+                            mlflow_logger._mlflow_client.log_artifact(
+                                mlflow_logger._run_id,
+                                os.path.join(local_save_path, 'LICENSE.txt'),
+                            )
```
mvpatel2000, line 288: You know what would be sick here? a walrus operator

mvpatel2000:
```
if (license_path := os.path.join(local_save_path, 'LICENSE.txt')) and os.path.exists(license_path):
    mlflow_logger._mlflow_client.log_artifact(
        mlflow_logger._run_id,
        license_path,
    )

```

mvpatel2000: or I guess u could just move the path creation to a new line. That's probably better tbh

dakinggg: oh yeah, this should just be a variable.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -279,6 +279,15 @@ def _save_checkpoint(self, state: State, logger: Logger):
                             path=local_save_path,
                             **self.mlflow_logging_config,
                         )
+
+                        license_path = os.path.join(local_save_path,
+                                                    'LICENSE.txt')
```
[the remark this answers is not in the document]
irenedea: perhaps something like
```
_LICENSE_FILE_PATTERN = re.compile(r"license(\.[a-z]+|$)", re.IGNORECASE)
...
license_filename = next(file for file in os.listdir(local_save_path) if _LICENSE_FILE_PATTERN.search(file))
license_path = os.path.join(license_path, license_filename)

if os.path.isfile(license_path):
```
to be consistent w what mlflow does

setup.py
```diff
@@ -50,7 +50,8 @@
 ]
 
 install_requires = [
-    'mosaicml[libcloud,wandb,mlflow,oci,gcs]>=0.17.2,<0.18',
+    'mosaicml[libcloud,wandb,oci,gcs]>=0.17.2,<0.18',
```
irenedea, line 53: hmm so i guess the potential issue here is that we're out of sync with the mlflow version that composer requires?

irenedea: The proper thing to do is probably to update the mlflow version in composer. We are going to pin composer in ftapi, so that's possible without a release. wdyt?

dakinggg: No, I think this is correct. LLM Foundry is using a feature of MLflow that requires a tighter pin than the pin in composer. Composer has no reason to make its pin tighter, because it does not rely on this new feature. We could of course instead port the entire feature into the MLflowLogger, but I think it makes more sense here.

llmfoundry/callbacks/hf_checkpointer.py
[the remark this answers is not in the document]
dakinggg: Ah, I thought they told me they always copy it as `LICENSE.txt`, will check on this.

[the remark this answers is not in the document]
dakinggg: Update to be consistent with mlflow logic and manually tested MPT, codellama, and llama. Thanks for catching!

setup.py
irenedea: ah yeah you're right. I was thinking about a case where composer requires an incompatible pin, but this should be fine. And agree, it makes more sense here 🚀

irenedea (approved):
LGTM! Thanks!

PR #921, opened 2024-01-30
Adding deprecation warning for Flash Attention 1 and user warning against using Triton attention.

ShashankMosaicML:


llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -222,6 +223,20 @@ def _validate_config(self) -> None:
                 'attn_impl'] not in ['torch', 'triton']:
             raise NotImplementedError(
                 'prefix_lm only implemented with torch and triton attention.')
+
+        if self.attn_config['attn_impl'] == 'flash' and is_flash_v1_installed():
+            warnings.warn(
+                DeprecationWarning(
+                    'The support for Flash Attention v1 will soon be deprecated. Please upgrade to Flash Attention v2.4.2.'
```
vchiley, line 230:
```suggestion
                    'Support for Flash Attention v1 is deprecated. Please upgrade to Flash Attention >= v2.4.2.'
```

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -222,6 +223,20 @@ def _validate_config(self) -> None:
...
             raise NotImplementedError(
                 'prefix_lm only implemented with torch and triton attention.')
+
+        if self.attn_config['attn_impl'] == 'flash' and is_flash_v1_installed():
+            warnings.warn(
+                DeprecationWarning(
+                    'Support for Flash Attention v1 is deprecated. Please upgrade to Flash Attention v2.4.2.'
+                ))
+
+        if self.attn_config[
```
dakinggg, line 233: does prefix lm require triton?

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -222,6 +223,20 @@ def _validate_config(self) -> None:
                 'attn_impl'] not in ['torch', 'triton']:
             raise NotImplementedError(
                 'prefix_lm only implemented with torch and triton attention.')
+
+        if self.attn_config['attn_impl'] == 'flash' and is_flash_v1_installed():
```
dakinggg, line 227: add the instructions for how to install and what will change in future version. so, say, you can install with .[gpu-flash2], and after deprecation .[gpu] will be updated to install flash2

llmfoundry/models/mpt/configuration_mpt.py
ShashankMosaicML: Yes, we need triton for prefix lm because flash attention does not support prefix lm right now.

dakinggg: 😢

vchiley: 😢

llmfoundry/models/mpt/configuration_mpt.py
ShashankMosaicML: Added the instruction on how to install Flash 2 in the warning, and updated readme to say that after we remove support for FA v1, .[gpu] will be updated to install flash2.

dakinggg (commented):
Could you also update the training readme that explains how to use flash attention?

ShashankMosaicML:
> Could you also update the training readme that explains how to use flash attention?

Done. Please take a look.

vchiley (approved):

dakinggg (approved):

PR #922, opened 2024-01-30
Throw error when no EOS

KuuCi:
This PR aims to throw an error if no eos_token_id is detected when building tokenizer


tests/utils/test_builders.py
```diff
@@ -48,6 +48,27 @@ def test_tokenizer_builder(tokenizer_name: str, tokenizer_kwargs: dict):
         assert isinstance(tokenizer, PreTrainedTokenizerBase)
 
 
+@pytest.mark.parametrize('tokenizer_name,tokenizer_kwargs', [
```
irenedea, line 51: parameterization seems unnecessary with L64 and L61

tests/utils/test_builders.py
```diff
@@ -48,6 +48,27 @@ def test_tokenizer_builder(tokenizer_name: str, tokenizer_kwargs: dict):
...
+    ('bert-base-uncased', {
+        'model_max_length': 10,
+        'eos_token_id': None,
+    }),
+    ('bert-base-uncased', {
+        'model_max_length': 10
+    }),
+])
+def test_tokenizer_no_EOS(tokenizer_name: str, tokenizer_kwargs: dict):
+    tokenizer_kwargs = {
```
irenedea, line 61: not necessary to pass in model_max_length here, you can just do

```
build_tokenizer('bert-base-uncased', {})
```

tests/utils/test_builders.py
```diff
@@ -48,6 +48,27 @@ def test_tokenizer_builder(tokenizer_name: str, tokenizer_kwargs: dict):
...
+        'model_max_length': 10,
+    }
+    tokenizer_name = 'bert-base-uncased'
+
+    with pytest.raises(ValueError) as exc_info:
+        build_tokenizer(tokenizer_name, tokenizer_kwargs)
+    assert 'must have an eos_token_id.' in str(
```
irenedea, line 68: I think instead of asserting a specific message separately, you can pass an expected message to `pytest.raises`

irenedea (changes requested):
Thanks! Had some comments!

tests/utils/test_builders.py
```diff
@@ -48,6 +48,14 @@ def test_tokenizer_builder(tokenizer_name: str, tokenizer_kwargs: dict):
         assert isinstance(tokenizer, PreTrainedTokenizerBase)
 
 
+def test_tokenizer_no_EOS(tokenizer_name: str, tokenizer_kwargs: dict):
+    with pytest.raises(ValueError, match=r".*must have an eos_token_id.*"):
```
irenedea, line 52: nit: Can we match the strings exactly?

tests/utils/test_builders.py
```diff
@@ -48,6 +48,14 @@ def test_tokenizer_builder(tokenizer_name: str, tokenizer_kwargs: dict):
...
+        build_tokenizer('bert-base-uncased', {})
+    
+    with pytest.raises(ValueError, match=r".*must have an eos_token_id.*"):
+        build_tokenizer('bert-base-uncased', {'eos_token_id': None})
```
irenedea, line 56: looks like for `bert-base-uncased` eos_token_id is already None by default, is there a reason to do this?

llmfoundry/utils/builders.py
```diff
@@ -407,6 +407,10 @@ def build_tokenizer(
             int(1e30),
         )
 
+    if not hasattr(tokenizer, 'eos_token_id') or tokenizer.eos_token_id is None:
+        raise ValueError(
+            f'The tokenizer {tokenizer_name} must have an eos_token_id.')
+
```
irenedea, line 413: let's actually check for `eos_token` instead of `eos_token_id`.

`eos_token_id` uses `eos_token` to get an id, so `eos_token` is a better "source of truth"

irenedea (approved):
LGTM! thanks for the better checking 💪

PR #926, opened 2024-01-31
Bump composer version

aspfohl:
I don't have a reason to prefer 18.1 > 18.0, but maybe @mvpatel2000 @dakinggg you'd have opinions https://github.com/mosaicml/composer/releases/tag/v0.18.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ classifiers = [
 ]
 
 install_requires = [
-    'mosaicml[libcloud,wandb,oci,gcs]>=0.17.2,<0.18',
+    'mosaicml[libcloud,wandb,oci,gcs]>=0.18.1,<0.19',
     'mlflow>=2.10,<3',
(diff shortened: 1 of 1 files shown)


dakinggg:
@aspfohl it should be 0.18.1. Also can you please kick off a regression testing run?

setup.py
```diff
@@ -50,7 +50,7 @@
 ]
 
 install_requires = [
-    'mosaicml[libcloud,wandb,oci,gcs]>=0.17.2,<0.18',
+    'mosaicml[libcloud,wandb,oci,gcs]>=0.18.1,<0.19',
```
dakinggg, line 53: there are a few more places you need to bump the version in the extra deps

aspfohl:
will pause for now 👍

setup.py
aspfohl: yep, caught in the test failures 👍

PR #927, opened 2024-01-31
Add rich formatting to tracebacks

jjanezhang:
# Add rich formatting to tracebacks
Somehow lost our rich formatting for tracebacks. Added it to eval and train.

## Testing
image


dakinggg (approved):
Can you add to eval.py too please?

PR #931, opened 2024-02-01
Remove .ci folder and move FILE_HEADER

irenedea:
We don't use `.ci`, so we should remove the folder and move FILE_HEADER.


mvpatel2000 (approved):
Can we do this in all repos to stay consistent? eg composer/streaming

irenedea:
@mvpatel2000 Yup that's the plan!

dakinggg (approved):

PR #933, opened 2024-02-02
Add finetuning streaming dataset conversion

bigning:
Changes:
1. add a JSONL IFT dataset
2. modified scripts/data_prep/convert_finetuning_dataset.py, adding tokenizer args. If it's set, it will run the tokenizer, and save token ids, instead of raw text. 

In a separate PR, we will add an example yaml to use streaming dataset for finetuning. 

test:
python convert_finetuning_dataset.py --dataset "json" --splits "train"    --out_root "example_data" --data_files "./dolly_hhrlhf_train_small.jsonl" --tokenizer=gpt2 --skip-preprocessing --max_seq_len=512

image


karan6181:
You might want to run [linting](https://github.com/mosaicml/composer/blob/dev/CONTRIBUTING.md#prerequisites) locally and also pytest using `pytest -vv -s .` command to ensure those passes locally.

scripts/data_prep/convert_finetuning_dataset.py
```diff
@@ -170,12 +188,22 @@ def main(args: Namespace) -> None:
                 'include the "--skip-preprocessing" flag to avoid this error.'
             )
 
-    columns = ['prompt', 'response']
+    tokenizer = None
+    args.tokenizer_kwargs.update({'model_max_length': args.max_seq_len})
```
dakinggg, line 192: lets pull the `tokenizer_kwargs` out of args and define in a variable before updating, so that the `args` remains unmodified.

scripts/data_prep/convert_finetuning_dataset.py
```diff
@@ -38,6 +44,10 @@ def parse_args() -> Namespace:
                         default=None,
                         help='Name or import path of function used to preprocess (reformat) the dataset. ' +\
                              'See README for additional details.')
+    parser.add_argument('--data_files',
```
dakinggg, line 47: clarify docs that this is a 1:1 mapping between splits and file. Also could you test if there are any restrictions on file naming? I know sometimes HF requires that the split name match the filename and stuff like that.

scripts/data_prep/convert_finetuning_dataset.py
```diff
@@ -170,12 +188,22 @@ def main(args: Namespace) -> None:
...
+    if args.tokenizer:
+        tokenizer = build_tokenizer(args.tokenizer, args.tokenizer_kwargs)
+        columns = {'input_ids': 'bytes', 'labels': 'bytes'}
+    else:
+        columns = {'prompt': 'str', 'response': 'str'}
 
-    for split_name in args.splits:
+    for i, split_name in enumerate(args.splits):
+        data_file = None
+        if len(args.data_files) > 0:
```
dakinggg, line 201: lets do some up front validation that splits and data files have the same length and error if they don't

scripts/data_prep/dolly_hhrlhf_train.jsonl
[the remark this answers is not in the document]
bigning: changed to only including the first 1000 lines.

scripts/data_prep/convert_finetuning_dataset.py
bigning: > Also could you test if there are any restrictions on file naming?

For data_files it's fine, since you input the file name, not the directory name.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -228,6 +229,33 @@ def _tokenize_formatted_example(
         raise ValueError(f'Unknown conversation type {example_format=}')
 
 
+def is_valid_ift_example(pad_token_id: int, max_seq_len: int,
+                         example: Dict) -> bool:
+    """Check if it's an valid ift example.
```
dakinggg, line 234:
```suggestion
    """Check if the example is a valid ift example.
```

llmfoundry/data/finetuning/tasks.py
```diff
@@ -228,6 +229,33 @@ def _tokenize_formatted_example(
...
+
+    This functions does the following check:
+    a. Length of input_ids should less than max_seq_len
```
dakinggg, line 237:
```suggestion
    a. Length of input_ids should be less than max_seq_len
```

llmfoundry/data/finetuning/tasks.py
```diff
@@ -228,6 +229,33 @@ def _tokenize_formatted_example(
...
+    b. Both input_ids and labels should not be empty
+    c. Labels should has at least 1 non-padding token.
```
dakinggg, line 239:
```suggestion
    c. Labels should have at least 1 non-padding token.
```

scripts/data_prep/convert_finetuning_dataset.py
```diff
@@ -73,6 +89,17 @@ def parse_args() -> Namespace:
...
 
+    if parsed.tokenizer_kwargs is not None:
+        parsed.tokenizer_kwargs = json.loads(parsed.tokenizer_kwargs)
+    else:
+        parsed.tokenizer_kwargs = {}
+
+    if len(parsed.data_files) > 0 and len(parsed.data_files) != len(
+            parsed.splits):
+        raise ValueError(
+            f'If set data_files, data_files and splits must be 1:1 mapping. Got {len(parsed.data_files)=} while {len(parsed.splits)=}'
```
dakinggg, line 100:
```suggestion
            f'If data_files is set, data_files and splits must have the same length. Got {len(parsed.data_files)=} while {len(parsed.splits)=}'
```

dakinggg (approved):
LGTM, but can you please include evidence of manual testing in the PR description?

bigning:
> LGTM, but can you please include evidence of manual testing in the PR description?

Thanks very much for the review! Added the screenshot of the run. Please let me know if more information is needed.

dakinggg:
@bigning i'd suggest also just loading the dataset manually and checking that the first sample looks as you expect it to

PR #937, opened 2024-02-02
Tessa/callibration script

tbarton16:
Here is code we use to test our benchmark tasks by using a series of progressively more advanced models to see if the benchmarks effectively differentiate between them, and at which number of shots they performed best at.

* Select an independent variable and a series of models that correspond to the settings of that variable
* Select clusters 
* Edit the list of tasks in the `base_callibration.yaml` to reflect the ones you want to see
* Run the launcher script
* When everything is done, run the `analyze_output` notebook which collates the results from wandb


maxisawesome:
lgtm! I kinda hate checking in notebooks but I do think it's better than a script in this case.

maxisawesome (approved):

scripts/eval/callibration/base_callibration.yaml
```diff
@@ -0,0 +1,225 @@
...
+  gpus: 8
+  cluster:
+scheduling:
+  priority: low
+  preemptible: true
+image: mosaicml/llm-foundry:2.1.0_cu121-8e96a95
+integrations:
+- integration_type: git_repo
+  git_repo: mosaicml/llm-foundry
+  git_branch: main
```
dakinggg, line 12: should probably be pinned to a release.

scripts/eval/callibration/base_callibration.yaml
dakinggg, line 1: Lets move this to the `eval/yamls` folder

dakinggg (commented):
Given that this notebook/script is mostly for y'all, has lots of hardcoded stuff, etc, lets note in the README that the calibration scripts are experimental and subject to change at any time.

bmosaicml:
Would you mind adding the MCLI name of a test run you launched so I can go back and `describe run` and view logs later?

Additionally a screenshot of the resulting notebook would be good so that when I go back to this later I can confirm that I got the correct results?

bmosaicml (approved):
Great work! can you just address Daniel's comments as well as update the description as I requested?

Thx Tessa!!!

PR #940, opened 2024-02-05
Fixing the gen_attention_mask_in_length function to handle the case when sequence id is -1 due to attention masking

ShashankMosaicML:


rajammanabrolu (approved):
lgtm thanks for fixing this so quickly!

dakinggg (approved):
LGTM

PR #944, opened 2024-02-05
Add and use VersionedDeprecationWarning

irenedea:
We should include versions with our deprecation warnings so we can easily tell when we should officially remove something.

When adding a new VersionedDeprecationWarning, one should pass the most recent released foundry version pus two versions to the `remove_version` argument.


llmfoundry/utils/warnings.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+
+class VersionedDeprecationWarning(DeprecationWarning):
```
dakinggg, line 5: Should we upstream this to composer?

llmfoundry/utils/warnings.py
```diff
@@ -0,0 +1,27 @@
...
+    """A custom deprecation warning class that includes version information.
+
+    Attributes:
+        message (str): The deprecation message describing why the feature is deprecated.
+        after_version (str): The version after which the feature will be deprecated.
```
dakinggg, line 10: The wording here isn't totally clear to me. Should it be "Feature X is deprecated and will be removed after version Y"?

llmfoundry/utils/warnings.py
irenedea: hmm could but then we'd need to wait for a release to use it in foundry?

dakinggg: yeah you can also just do it in this pr and open a companion PR to composer

dakinggg (commented):
I think the flash attention v1 deprecation is the only one you need to keep and the others can be deleted now.

llmfoundry/utils/warnings.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+
+class VersionedDeprecationWarning(DeprecationWarning):
+    """A custom deprecation warning class that includes version information.
+
+    Attributes:
+        message (str): The deprecation message describing why the feature is deprecated.
+        after_version (str): The version after which the feature will be removed.
```
dakinggg, line 10: update docstring

llmfoundry/callbacks/generate_callback.py
```diff
@@ -20,10 +22,10 @@ def __init__(self, prompts: List[str], batch_log_interval: int,
                  **kwargs: Any):
 
         warnings.warn(
-            ('Accessing llmfoundry.callbacks.generate_callback.Generate '
-             'is deprecated and will be removed in a future release. '
-             'Please use composer.callbacks.Generate instead.'),
-            DeprecationWarning,
+            VersionedDeprecationWarning('Accessing llmfoundry.callbacks.generate_callback.Generate ' + \
+             'is deprecated and will be removed in a future release. Please use composer.callbacks.Generate instead.',
```
dakinggg, line 26:
```suggestion
             'is deprecated. Please use composer.callbacks.Generate instead.',
```

llmfoundry/data/packing.py
```diff
@@ -433,9 +435,11 @@ def profile(raw_batch_size: int) -> Tuple[Optional[float], Optional[float]]:
     import warnings
 
     warnings.warn(
-        DeprecationWarning(
+        VersionedDeprecationWarning(
             'Please use scripts/misc/profile_packing.py to profile packing.' +
-            'This script will be removed in later releases.'))
+            'This script will be removed in later releases.',
```
dakinggg, line 440:
```suggestion
```

scripts/train/train.py
```diff
@@ -219,8 +221,10 @@ def main(cfg: DictConfig) -> Trainer:
                                           default_value=None)
         if eval_gauntlet_config is not None:
             warnings.warn(
-                'Use of the key `model_gauntlet` is deprecated, please use the key `eval_gauntlet`',
-                DeprecationWarning)
+                VersionedDeprecationWarning(
+                    'Use of the key `model_gauntlet` is deprecated, please use the key `eval_gauntlet`',
```
dakinggg, line 225:
```suggestion
                    'Use of the key `model_gauntlet` is deprecated, please use the key `eval_gauntlet`.',
```

dakinggg (approved):
Just some tiny wording nits that don't really matter cause you're about to delete them anyway. Thanks!

llmfoundry/utils/warnings.py
irenedea, line 10:
```suggestion
        remove_version (str): The version after which the feature will be removed.
```

PR #945, opened 2024-02-05
add finutuning with streaming dataset example

bigning:
Adding an IFT finetuning yaml with streaming dataset `scripts/train/yamls/finetune/gpt2-arc-easy--cpu-streaming-dataset.yaml`

command to run the finutuning job:
under `scripts/train` path, run command:
`composer train.py yamls/finetune/gpt2-arc-easy--cpu-streaming-dataset.yaml`

image


.github/workflows/code-quality.yaml
```diff
@@ -4,10 +4,12 @@ on:
     branches:
     - main
     - release/**
+    - add_finetuning_streaming_dataset_conversion
```
dakinggg, line 7: leaving behind a comment to remove before merging

.github/workflows/code-quality.yaml
```diff
@@ -4,10 +4,13 @@ on:
     branches:
     - main
     - release/**
+    # todo: remove this before merging
+    - add_finetuning_streaming_dataset_conversion
```
dakinggg, line 8: remove

.github/workflows/code-quality.yaml
```diff
@@ -4,10 +4,13 @@ on:
...
   pull_request:
     branches:
     - main
     - release/**
+    - add_finetuning_streaming_dataset_conversion
```
dakinggg, line 13: remove

llmfoundry/data/finetuning/collator.py
```diff
@@ -106,7 +106,7 @@ def __init__(
 
     def __call__(self, examples: List[Dict[str,
                                            Any]]) -> Dict[str, torch.Tensor]:
-        for check_key in ['input_ids', 'labels', 'attention_mask']:
+        for check_key in ['input_ids', 'labels']:
```
dakinggg, line 109: whats going on here? was this just a mistake before to require this key? seems like yes

llmfoundry/data/finetuning/tasks.py
```diff
@@ -371,10 +373,19 @@ def __init__(self,
...
 
         self.tokenizer = tokenizer
+        self.max_seq_len = max_seq_len
 
     # How to process a sample
     def __getitem__(self, idx: int) -> Dict[str, Any]:
         sample = super().__getitem__(idx)
+        if 'input_ids' in sample:
+            # already tokenized data
+            sample['input_ids'] = np.frombuffer(
```
dakinggg, line 383: were the tolist calls necessary here? I notice that the equivalent code in `text_data.py` doesn't have the tolist

scripts/train/yamls/finetune/gpt2-arc-easy--cpu-streaming-dataset.yaml
```diff
@@ -0,0 +1,77 @@
...
+  name: gpt2
+  kwargs:
+    model_max_length: ${max_seq_len}
+
+# Dataloaders
+train_loader:
+  name: finetuning
+  dataset:
+    ############
+    remote: /Users/ning.wang/projects/llm-foundry/scripts/data_prep/example_data/
```
dakinggg, line 24: should copy the structure of e.g. `gpt-neo-125m.yaml` in the `pretrain` folder for how to specify reasonable defaults for local/remote. Should not reference your specific path.

llmfoundry/data/finetuning/collator.py
bigning: yeah, it looks it was a mistake. Both `self._process_and_batch_decoder_only` and `self._process_and_batch_encoder_decoder` will compute and add `attention_mask` to the sample. So I think it's not needed here.

llmfoundry/data/finetuning/tasks.py
bigning: Prior to this change, Line 389 returns a python list (not a numpy array or torch.Tensor). So in order to make it return the same type as before, I convert it to a list. 

The code in text_data.py returns a tensor by calling `torch.from_numpy()`, so that it doesn't need the tolist().

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -152,6 +152,7 @@ def build_finetuning_dataloader(cfg: DictConfig,
             sampling_method=cfg.dataset.get('sampling_method', 'balanced'),
             sampling_granularity=cfg.dataset.get('sampling_granularity', 1),
             batching_method=cfg.dataset.get('batching_method', 'random'),
+            max_seq_len=cfg.dataset.max_seq_len,
```
karan6181, line 155: Is `cfg.dataset.max_seq_len` a mandatory argument? What happens if `max_seq_len` is missing from `cfg.dataset`?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -371,10 +373,19 @@ def __init__(self,
...
+                sample['input_ids'],
+                dtype=np.int64)[:self.max_seq_len].tolist().copy()
```
karan6181, line 385: Wondering, what is the reason behind `np.int64`? why not `int32` or lower?

scripts/train/yamls/finetune/gpt2-arc-easy--cpu-streaming-dataset.yaml
```diff
@@ -0,0 +1,80 @@
+max_seq_len: 512
```
karan6181, line 1: Nit (filename): two double dashes `--` before `CPU.` Is that intentional?

karan6181 (commented):
Any unit tests which would test the streaming dataset sample using getitem and iterator?

llmfoundry/data/finetuning/dataloader.py
bigning: if it's missing, it will be using the default 2048. I added this because it's needed when convert to list from buffer in  `np.from_buffer(sample['input_ids'], dtype=np.int64)[:self.max_seq_len].tolist()`. I guess the `[:self.max_seq_len]` is not needed if the data already truncates at `max_seq_len` when generating the MDS file ?

llmfoundry/data/finetuning/tasks.py
bigning: Honestly, i don't know. I just copied from [here](https://github.com/mosaicml/llm-foundry/blob/105f76631d7f74edebc187f3427008ce4506ee85/llmfoundry/data/text_data.py#L170-L171) . 

I guess it's better to set the dtype explicitly when converting to `numpy.array` [here](https://github.com/mosaicml/llm-foundry/blob/105f76631d7f74edebc187f3427008ce4506ee85/scripts/data_prep/convert_finetuning_dataset.py#L258) , and make this two place use the same dtype.  Do you suggest to do that ?

scripts/train/yamls/finetune/gpt2-arc-easy--cpu-streaming-dataset.yaml
bigning: I just copied from [this file](https://github.com/mosaicml/llm-foundry/blob/105f76631d7f74edebc187f3427008ce4506ee85/scripts/train/finetune_example/gpt2-arc-easy--cpu.yaml), and  append `-streaming-dataset` to the filename. Let me just use one dash.

bigning:
> Any unit tests which would test the streaming dataset sample using getitem and iterator?

good question, the existing unit test doesn't cover tokenized streaming dataset. let me add one.

llmfoundry/data/finetuning/collator.py
dakinggg: 👍

llmfoundry/data/finetuning/dataloader.py
dakinggg: I think we should do it the way you implemented it. This matches the behavior of `text_data.py`, and allows you to run with data truncated to the requested max seq len.

llmfoundry/data/finetuning/tasks.py
dakinggg: ah, makes sense, thanks!

llmfoundry/data/finetuning/tasks.py
dakinggg: as discussed offline, lets make the change we discussed in this pr

tests/data/test_dataloader.py
```diff
@@ -49,8 +51,15 @@ def get_abs_data_path(data_local: str):
     return os.path.join(os.getcwd(), data_local)
 
 
-def build_mock_ft_streaming_dataset(data_path: str, split: str):
-    columns = {'prompt': 'str', 'response': 'str'}
+def build_mock_ft_streaming_dataset(
```
dakinggg, line 54: can you also test that the bytes version still works?

tests/data/test_dataloader.py
```diff
@@ -547,11 +572,6 @@ def test_finetuning_dataloader_streaming(tmp_path: pathlib.Path):
 
     cfg = om.create(cfg)
 
-    tokenizer = build_tokenizer(
-        tokenizer_name='gpt2',
-        tokenizer_kwargs={'model_max_length': max_seq_len},
-    )
-
     _ = build_finetuning_dataloader(cfg, tokenizer, 4)
```
dakinggg, line 575: should probably also make this test a little more thorough and test that the first batch in the dataloader looks how you expect

llmfoundry/data/finetuning/tasks.py
```diff
@@ -371,10 +373,27 @@ def __init__(self,
         )
 
         self.tokenizer = tokenizer
+        self.max_seq_len = max_seq_len
 
     # How to process a sample
     def __getitem__(self, idx: int) -> Dict[str, Any]:
         sample = super().__getitem__(idx)
+        if 'input_ids' in sample:
+            # already tokenized data
```
dakinggg, line 382:
```suggestion
            # Already tokenized data
```

dakinggg (approved):
LGTM! Left a couple comments to make the test more thorough, but feel free to merge once that is addressed!

PR #948, opened 2024-02-06
Bump llm-foundry version to 0.5.0

irenedea:
- [x] Bump version 0.4.0 to 0.5.0
- [x] Remove deprecations (depends on https://github.com/mosaicml/llm-foundry/pull/944)


mcli/mcli-1b-eval.yaml
```diff
@@ -1,7 +1,7 @@
 integrations:
 - integration_type: git_repo
   git_repo: mosaicml/llm-foundry
-  git_branch: v0.4.0
+  git_branch: v0.5.0
```
dakinggg, line 4: these tags shouldn't change until there is a tagged release for v0.5.0, off the of the `release/v0.5.0` branch. So these should be a step after the release I think?

irenedea: ah yes that's true. this wouldn't work until we do the full github release 🥲

dakinggg (approved):

PR #954, opened 2024-02-07
Adding curriculum learning callback (experimental)

snarayan21:
This PR adds the curriculum learning callback. This callback enables cleanly resuming from existing checkpoints by specifying the new dataset and passing in a `dataset_index` argument to the callback. Each new dataset is treated as a fresh epoch of data, and fast resumption using StreamingDataset is possible as well. 

This callback is dependent on changes to StreamingDataset's `load_state_dict` function to enable setting the dataset state dict multiple times correctly, which has now been merged! [PR Link](https://github.com/mosaicml/streaming/pull/593)

This callback also depends on the addition of `Event.BEFORE_LOAD` into composer, PR [here](https://github.com/mosaicml/composer/pull/2974), which has also been merged.

Testing is detailed below:

Test run 1 -- test resumption from existing checkpoints that did not use the CL Callback. Loss is discontinuous when starting from a new dataset (as expected) while LR schedule is continuous.
* Steps 0-49: No CL callback enabled, training on GitHub data.
* Steps 50-99: CL callback enabled with `dataset_index` 1 on C4. Training proceeds on the new dataset but LR schedule is continuous.
* Steps 100-149: CL callback enabled, resuming on C4, maintaining `dataset_index`. Training resumes correctly and deterministically.
* Steps 150-200: CL callback enabled with `dataset_index` 2 on Pile. Training finishes and LR schedule is still continuous.
Screenshot 2024-02-06 at 6 33 03 PM
Screenshot 2024-02-06 at 6 32 02 PM

Test run 2 -- test model training when using the CL callback from the very beginning. This was the same as Test run 1, except for the first segment from batches 0-49, the CL callback was enabled with `dataset_index` 0.  Loss is discontinuous when starting from a new dataset (as expected) while LR schedule is continuous.
Screenshot 2024-02-06 at 6 32 48 PM
Screenshot 2024-02-06 at 6 32 34 PM


llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+                f'CurriculumLearning callback can only be used with a train ',
+                f'dataloader of type DataLoader, but got {type(dataloader)}.')
+        dataset = dataloader.dataset
+        if not isinstance(dataset, StreamingDataset):
+            raise ValueError(
+                f'CurriculumLearning callback only supports StreamingDataset ',
+                f'because it requires loading and saving dataset state. ',
+                f'Instead, got a dataset of type {type(dataset)}')
+        # Save the current dataset state so we can restore it if needed.
+        self.current_dataset_state = dataset.state_dict(0, False)
```
mvpatel2000, line 57: I don't really understand what's going on here -- can you elaborate?

mvpatel2000 (commented):
Why is there a bunch of other stuff eg rotary embedding here?

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: We need to store dataset state like `num_canonical_nodes`, `initial_physical_nodes`, and `shuffle_seed` at init time because it gets overwritten by checkpoint loading. So if I'm resuming with a different dataset, the only time I'll have access to this info is at init time, and then I load it in in the `after_load` function

mvpatel2000: Why can't I fetch this on event init? Isn't it accessible the same way?

snarayan21: Sadly you can't, I tried it. The `State` obj in composer is created [here](https://github.com/mosaicml/composer/blob/12261d62335bab7e341735bad368a08eb3212cb7/composer/trainer/trainer.py#L1043), but the dataloader doesn't get set until [here](https://github.com/mosaicml/composer/blob/12261d62335bab7e341735bad368a08eb3212cb7/composer/trainer/trainer.py#L1300), which is after Event.INIT [here](https://github.com/mosaicml/composer/blob/12261d62335bab7e341735bad368a08eb3212cb7/composer/trainer/trainer.py#L1199).

snarayan21:
Whoops, idk where that rope stuff came from lol. fixed

llmfoundry/callbacks/curriculum_learning_callback.py
mvpatel2000: a little annoying. I suppose it's fine...

llmfoundry/callbacks/__init__.py
```diff
@@ -18,12 +20,7 @@
...
-    'MonolithicCheckpointSaver',
-    'GlobalLRScaling',
-    'LayerFreezing',
-    'ScheduledGarbageCollector',
-    'EvalGauntlet',
-    'HuggingFaceCheckpointer',
-    'AsyncEval',
+    'FDiffMetrics', 'MonolithicCheckpointSaver', 'GlobalLRScaling',
+    'LayerFreezing', 'ScheduledGarbageCollector', 'EvalGauntlet',
+    'HuggingFaceCheckpointer', 'AsyncEval', 'CurriculumLearning'
```
mvpatel2000, line 25:
```suggestion
    'HuggingFaceCheckpointer', 'AsyncEval', 'CurriculumLearning',
```
nit restore trailing comma

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: indeed

llmfoundry/callbacks/__init__.py
snarayan21: updated, linting changed formatting

llmfoundry/utils/builders.py
```diff
@@ -159,6 +161,7 @@ def build_callback(
     name: str,
     kwargs: Union[DictConfig, Dict[str, Any]],
     config: Any = None,
+    train_loader: Union[None, DataLoader, Iterable] = None,
```
mvpatel2000, line 164: I don't like this but I can't think of a cleaner way. @dakinggg  thoughts?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+class CurriculumLearning(Callback):
+    """Starts an epoch with a different dataset when resuming from a checkpoint.
+
+    This callback is currently experimental. The API may change in the future.
+
+    Args:
+        dataset_index (int): The index of the dataset currently being used.
+        current_dataset_config (Dict): The configuration of the dataset currently
+            being used.
+        dataset (StreamingDataset): The dataset currently being used.
```
mvpatel2000, line 29: Currently you pass in dataloader and not dataset

Should this be a dataset instead?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+        elif self.dataset_index == 0 and len(self.all_dataset_configs) == 0:
+            # Make sure to save our current dataset config if we are just starting training.
+            self.new_dataset_setup = True
+
+    def state_dict(self):
+        if self.new_dataset_setup and not self.dataset_config_appended:
+            # Append the new dataset config to the list of all dataset configs.
+            self.all_dataset_configs.append(self.current_dataset_config)
+            # Only append the dataset config once, not on every single checkpoint save.
+            self.dataset_config_appended = True
```
mvpatel2000, line 95: Can we get rid of the vars here and instead do directly in after_load?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+            train_loader,
+            DataLoader), 'CurriculumLearning callback requires a DataLoader.'
+        dataset = train_loader.dataset
+        assert isinstance(
+            dataset, StreamingDataset
+        ), 'CurriculumLearning callback requires a StreamingDataset.'
+        if self.saved_dataset_index < self.dataset_index:
+            # Ignore the dataset state that was read in from the checkpoint, and
+            # replace with the new dataset state. This preserves resumption info.
+            if self.current_dataset_state['epoch'] < 0:
```
mvpatel2000, line 76: How can epoch be < 0?

scripts/train/train.py
```diff
@@ -482,6 +474,15 @@ def main(cfg: DictConfig) -> Trainer:
         device_train_batch_size,
     )
 
+    # Callbacks
```
dakinggg, line 477: I'm assuming you had to move this below the dataloader creation. Given that the ordering now matters, please leave behind a comment.

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: Oh I meant for this to be dataloader. I'd rather check if the dataset is a StreamingDataset in the callback rather than in build_callback or in trainer.py because it's particular to this callback alone, and keeps the existing code cleaner

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Enable curriculum learning by resuming with a different dataset.
+
+This callback is currently experimental. The API may change in the future.
```
dakinggg, line 6:
```suggestion
This callback is currently experimental. The API may change without warning in the future.
```

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+from streaming import StreamingDataset
+from torch.utils.data import DataLoader
+
+log = logging.getLogger(__name__)
+
+
+class CurriculumLearning(Callback):
+    """Starts an epoch with a different dataset when resuming from a checkpoint.
+
+    This callback is currently experimental. The API may change in the future.
```
dakinggg, line 23:
```suggestion
    This callback is currently experimental. The API may change without warning in the future.
```

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+                 current_dataset_config: Dict):
+        self.dataset_index = dataset_index
+        self.saved_dataset_index = 0
+        self.all_dataset_configs = []
+        # The current dataset config is resolved and passed in train.py
+        self.current_dataset_config = current_dataset_config
+        self.new_dataset_setup = False
+        self.dataset_config_appended = False
+
+        # Must pass in dataset directly since it is not actually accessible at Event.INIT in
```
dakinggg, line 43: Can we lazy init this so that this callback doesn't have to take in the dataloader? I understand what you need is not accessible on `INIT`, but we can we just defer to whenever the first time it is accessible is?

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: When a streamingdataset is first initialized and we call `state_dict`, the epoch that gets returned is `-1` because we increment the epoch counter at the start of  `__iter__`, which has not yet been called. So to set the state dict correctly, we have to check if `epoch` is -1.

scripts/train/train.py
snarayan21: resolved!

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: done

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: done

llmfoundry/callbacks/curriculum_learning_callback.py
mvpatel2000: Can you extend comment to explain this scenario?

llmfoundry/callbacks/curriculum_learning_callback.py
dakinggg: Yeah, i'd prefer fixing the events in composer to support this new location that you need something than requiring the callback to take in the dataloader.

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: Well, because the dataloader is `Union[DataLoader, Iterable]` it won't always have the `dataset` attribute. So I'd rather just do the dataloader and dataset checks both inside `init` rather than separately

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: word yeah makes sense. I wanted to enable this with as minimal code changes as possible, but I can add that in

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: modified to do this

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: added!

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: modified the callback, assuming `Event.BEFORE_LOAD` exists

llmfoundry/utils/builders.py
snarayan21: addressed with `BEFORE_LOAD` event.

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: with the `BEFORE_LOAD` event, still doing type checks inside there.

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+class CurriculumLearning(Callback):
+    """Starts an epoch with a different dataset when resuming from a checkpoint.
+
+    This callback is currently experimental. The API may change without warning in the future.
+
+    Args:
+        dataset_index (int): The index of the dataset currently being used.
+        current_dataset_config (Dict): The configuration of the dataset currently
+            being used.
+        dataloader (Union[DataLoader, Iterable]): The training dataloader currently being used.
```
dakinggg, line 30: remove

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+                f'CurriculumLearning callback can only be used with a train ',
+                f'dataloader of type DataLoader, but got {type(train_loader)}.')
+        dataset = train_loader.dataset
+        if not isinstance(dataset, StreamingDataset):
+            raise ValueError(
+                f'CurriculumLearning callback only supports StreamingDataset ',
+                f'because it requires loading and saving dataset state. ',
+                f'Instead, got a dataset of type {type(dataset)}')
+        # Save the current dataset state so we can restore it if needed.
+        self.current_dataset_state = dataset.state_dict(0, False)
```
dakinggg, line 59: can you pass the args here via kwarg?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def after_load(self, state: State, logger: Logger):
+        del logger
+
+        # As saved_dataset_index is loaded from state_dict, this only run when
```
dakinggg, line 64:
```suggestion
        # As saved_dataset_index is loaded from state_dict, this only runs when
```

llmfoundry/utils/builders.py
dakinggg: remove

scripts/train/train.py
```diff
@@ -482,6 +474,17 @@ def main(cfg: DictConfig) -> Trainer:
...
```
dakinggg, line 477: This is no longer true and can be move back up, yeah?

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: done

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: done

snarayan21:
Tested the callback again with the `BEFORE_LOAD` event. Looks good.
Screenshot 2024-02-08 at 3 04 52 PM
Screenshot 2024-02-08 at 3 05 02 PM

mvpatel2000 (commented):
@snarayan21  does this need a stream release?

snarayan21:
I released streaming v0.7.4 yesterday, so with that and composer release, this PR should be ready to merge (barring comments ofc)

mvpatel2000:
> @mvpatel2000 I released streaming v0.7.4 yesterday, so with that and composer release, this PR should be ready to merge (barring comments ofc)

Can we bump min version or is that already done

snarayan21:
it was a streaming patch release, and foundry's `setup.py` doesn't need to be updated

dakinggg:
@snarayan21 but this callback will only work from streaming 0.7.4 onwards right? if so, the callback either needs to check the version, or we bump min version. i think we should just bump min version.

snarayan21:
Oh yeah true. I misunderstood. Bumped min versions for both streaming and composer

scripts/train/train.py
```diff
@@ -482,6 +480,8 @@ def main(cfg: DictConfig) -> Trainer:
         device_train_batch_size,
     )
 
+    use_async_eval = any(isinstance(c, AsyncEval) for c in callbacks)
+
```
mvpatel2000, line 484: why is this here?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -0,0 +1,105 @@
...
+            # Ignore the dataset state that was read in from the checkpoint, and
+            # replace with the new dataset state. This preserves resumption info.
+            if self.current_dataset_state['epoch'] < 0:
+                # Make sure the epoch in the loaded state dict is not negative.
+                # Since `__iter__` has not yet been called on the dataset, the
+                # epoch index in the dataset will still be -1. We need to ensure
+                # that we set the epoch correctly to 0 in this case.
+                self.current_dataset_state['epoch'] = 0
+            dataset.load_state_dict(  # type: ignore
+                self.current_dataset_state)
```
mvpatel2000, line 86: Why do we need to do this?

Isn't it always sample 0 ep 0?

scripts/train/train.py
dakinggg: should be moved back where it was, although doesn't really matter

snarayan21: moved it back

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: part of this mightve been wrapped up in an earlier discussion, but we need to correctly set the other dataset state fields like `initial_physical_nodes`, `shuffle_seed`, and `num_canonical_nodes` which may be different in the new dataset compared to the old one. Checkpoint loading will overwrite the initial dataset state with potentially different values for these fields, so we need to set the dataset state dict correctly.

snarayan21: And the only time we have access to the unaltered, new dataset state is at the `BEFORE_LOAD` event.

mvpatel2000 (approved):
LGTM!

PR #956, opened 2024-02-07
Make CodeEval respect device_eval_batch_size

josejg:
See [the corresponding PR in composer](https://github.com/mosaicml/composer/pull/2969). 

`num_beams` shouldn't have been included as part of the icl_cfg since it gets passed as `generations_per_sample` to the eval suite. To actually specify beams we can now do it thanks to @maxisawesome 's refactor via generation_kwargs


dakinggg:
@josejg does this need to wait for a release with the composer change? If so, could you please change to draft PR so we don't merge until that is through?

josejg:
I think we want to get this as soon as possible so it can be run as part of gauntlet v3

dakinggg:
Sure, but does it need to wait for a composer release or is it standalone?

llmfoundry/utils/builders.py
```diff
@@ -497,10 +497,10 @@ def _validate_cfg(icl_cfg: DictConfig):
             icl_cfg.batch_size = default_batch_size
         if 'pass_at_k' not in icl_cfg:
             icl_cfg.pass_at_k = 1
-        if 'num_beams' not in icl_cfg:
```
dakinggg, line 500: Should we error if num_beams is specified # outside of generation kwargs? I think yes because people will probably still have yamls with that

josejg: raised an error explaining the remapping

dakinggg (commented):
Can you ctrl+f for `num_beams`? I saw at least one other (commented out) in tasks.yaml

maxisawesome:
> Can you ctrl+f for `num_beams`? I saw at least one other (commented out) in tasks.yaml

I changed the commented ones, there's one in test that shouldn't change.

dakinggg (approved):

PR #957, opened 2024-02-07
Update TUTORIAL.md

sdonoso:
fix indentation problem

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -371,7 +371,6 @@ model:
       target_modules:
       - q_proj
       - k_proj
-      - 'Wqkv'
 ```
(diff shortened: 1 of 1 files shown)


TUTORIAL.md
```diff
@@ -371,8 +371,7 @@ model:
       target_modules:
       - q_proj
       - k_proj
-    target_modules:
-    - 'Wqkv'
+      - 'Wqkv'
```
dakinggg, line 374: oops, thanks for catching! However, `Wqkv` should be removed entirely as no model will have both `q_proj`/`k_proj` and `Wqkv` :)

dakinggg (approved):
Thank you!

PR #958, opened 2024-02-07
allow te to use meta device with deferred init

cli99:
Specifying the `meta` device (`init_device: meta`) triggers the deferring init (defer module parameter initialization until after FSDP sharding)  in transformer engine repo. This resolves the excessive memory usage problem when using FSDP + te FP8.  The fix goes with the NV te fix https://github.com/NVIDIA/TransformerEngine/pull/596.

### DIFF
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -522,8 +522,7 @@ class GroupedQueryAttention(nn.Module):
         fc_kwargs: dict[str, Any] = {
             'bias': bias,
         }
-        if fc_type != 'te':
-            fc_kwargs['device'] = device
+        fc_kwargs['device'] = device
         self.Wqkv = FC_CLASS_REGISTRY[fc_type](
             self.d_model,
             self.d_model + 2 * self.kv_n_heads * self.head_dim,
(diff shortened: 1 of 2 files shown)


j316chuck (approved):
Great find! Curious how you found this fix?

cli99:
@j316chuck yes, we identified the excessive memory problem and verified fix a while ago, with the help of the memory profiling.

mvpatel2000 (approved):

llmfoundry/models/layers/ffn.py
```diff
@@ -97,8 +97,8 @@ def __init__(
         self.fc_kwargs: dict[str, Any] = {
             'bias': bias,
         }
-        if fc_type != 'te':
-            self.fc_kwargs['device'] = device
+
+        self.fc_kwargs['device'] = device
```
vchiley, line 101: y not just add at line 99

llmfoundry/models/layers/attention.py
```diff
@@ -522,8 +522,7 @@ def __init__(
         fc_kwargs: dict[str, Any] = {
             'bias': bias,
         }
-        if fc_type != 'te':
-            fc_kwargs['device'] = device
+        fc_kwargs['device'] = device
```
vchiley, line 525: y not just add at line 524

PR #960, opened 2024-02-08
strengthened chat formatting validation

milocress:


llmfoundry/data/finetuning/tasks.py
```diff
@@ -109,37 +109,42 @@ def _is_empty_or_nonexistent(dirpath: str) -> bool:
 
 def _get_role_key(message: Dict[str, str]) -> str:
     role_keys = _ALLOWED_ROLE_KEYS.intersection(message.keys())
-    assert len(role_keys) == 1
+    assert len(
```
dakinggg, line 112: I missed in the initial PR I guess, but we use `raise` for errors that might actually happen and we want to communicate to the user, and `assert` is just for documenting invariants, satisfying pyright, or in tests. Could you update everywhere to `raise`?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -109,37 +109,42 @@ def _is_empty_or_nonexistent(dirpath: str) -> bool:
...
 def _get_content_key(message: Dict[str, str]) -> str:
     content_keys = _ALLOWED_CONTENT_KEYS.intersection(message.keys())
-    assert len(content_keys) == 1
+    assert len(content_keys
+              ) == 1, f'Expected 1 content key, but found {len(content_keys)}'
     content_key = list(content_keys)[0]
     return content_key
 
 
 def _get_message_key(example: ChatFormattedDict):
```
dakinggg, line 126: I think this doesn't cover the error I sent you, which results from the data not matching the expected types. Since this is data validation, lets not rely on the types being correct, and do some runtime checking to raise nice errors when data is malformatted

dakinggg (commented):
Also please add tests for the new error cases you are checking

dakinggg (commented):
Code looks good, will do final review once the tests are passing.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -79,7 +85,11 @@ def _get_example_type(example: Example) -> ExampleType:
     Raises:
         KeyError: If the example type is unknown.
     """
-    if 'messages' in example:
+    if not isinstance(example, Mapping):
+        raise TypeError(
```
dakinggg, line 89: update message

llmfoundry/data/finetuning/tasks.py
```diff
@@ -102,6 +112,72 @@ def _is_empty_or_nonexistent(dirpath: str) -> bool:
     return not os.path.isdir(dirpath) or len(os.listdir(dirpath)) == 0
 
 
+def _get_role_key(message: Dict[str, str]) -> str:
+    if not isinstance(message, Mapping):
```
dakinggg, line 116: this check now doesn't really make sense with the parameter typing. Should they just all be `Mapping`?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -118,7 +194,8 @@ def _slice_chat_formatted_example(
         ValueError: If the chat example has less than two messages or if the last message is not from the assistant.
         KeyError: If a message does not have a role or content.
     """
-    messages = example['messages']
+    _validate_chat_formatted_example(example)
```
irenedea, line 197: I think there's some validation below that is now unnecessary bc you're doing it here now

llmfoundry/data/finetuning/tasks.py
```diff
@@ -102,6 +112,72 @@ def _is_empty_or_nonexistent(dirpath: str) -> bool:
...
+        raise TypeError(
+            f'Expected message to be a mapping, but found {type(message)}')
+    role_keys = _ALLOWED_ROLE_KEYS.intersection(message.keys())
+    if len(role_keys) != 1:
+        raise ValueError(f'Expected 1 role key, but found {len(role_keys)}')
+    role_key = list(role_keys)[0]
+    return role_key
+
+
+def _get_content_key(message: Dict[str, str]) -> str:
```
irenedea, line 126: these get functions are pretty similar in their implementations, can we generalize?

irenedea: What about some kind of `safe_get` function to safely get from a mapping? This could also be used for prompt response checks

overly simplified pseudo code
```
def safe_get(dictionary: dict, allowed_keys: set, type: Any):
    ...
    if not isinstance(value, type):
        raise TypeError()
    return value


def validate_chat_formatted_data(example: dict):
    messages = safe_get(example, _ALLOWED_MESSAGES_KEYS, type=list) 
    for idx, message in enumerate(messages):
        role = safe_get(example, _ALLOWED_ROLE_KEYS, type=str)
        _ = safe_get(example, _ALLOWED_CONTENT_KEYS, type=str)
        
        if role not in _ALLOWED_ROLES:
            raise ValueError()
        
        if idx == len(messages) - 1:
            if role not in _ALLOWED_LAST_MESSAGE_ROLES:
                raise ValueError()

```

llmfoundry/data/finetuning/tasks.py
milocress: "dict[str, str]" is incompatible with "Mapping[str, str]"

pyright does not know about dicts being mappings. We need a strict static check but a looser runtime check.

dakinggg (approved):

PR #964, opened 2024-02-09
Make turbo an optional dependency

snarayan21:
Makes turbo an optional dependency and gates its imports. This only affects users of Lion 8b optimizer.


mvpatel2000 (approved):
Specify it's a target option in error?

snarayan21:
modified the error message to direct users to `pip install llm-foundry[turbo]`.

PR #965, opened 2024-02-09
Fix eval.py with lora

dakinggg:
Missed `eval.py` completely in the LoRA refactor. `eval.py` should work with lora weights from HF now.

### DIFF
--- a/tests/models/test_model.py
+++ b/tests/models/test_model.py
@@ -13,6 +13,7 @@ import torch
 import torch.nn as nn
 from accelerate import init_empty_weights
 from composer.core.precision import Precision, get_precision_context
+from composer.models.huggingface import maybe_get_underlying_model
 from composer.optim import DecoupledAdamW
 from composer.trainer.dist_strategy import prepare_fsdp_module
 from composer.utils import dist, get_device, reproducibility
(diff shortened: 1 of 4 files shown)


tests/models/test_model.py
```diff
@@ -499,8 +500,18 @@ def test_loss_fn():
...
-def test_opt_wrapping():
-    conf = {
+@pytest.mark.parametrize('peft_config', [
+    None,
+    {
+        'peft_type': 'LORA',
+        'task_type': 'CAUSAL_LM'
+    },
+])
+def test_opt_wrapping(peft_config: Optional[dict[str, str]]):
```
irenedea, line 510: what is opt wrapping?

irenedea (approved):

tests/models/test_model.py
dakinggg: fsdp wrapping, opt is a model

dakinggg:
Will wait for @danbider to double check this works for him

danbider (approved):
LGTM, tested in two successful runs with llama-2-7b + lora adapter trained on starcoder:
human_eval
```human-eval-llama-7b-peft-VZDizn```
gauntlet
```eval-llama-7b-peft-xqBzgQ```

PR #971, opened 2024-02-12
Replace `num_beams` with `generations_per_sample`

eitanturok:
In `composer`'s `build_icl_dataloader` function  ([here](https://github.com/eitanturok/composer/blob/39db8483790a8fab6b2f3a08162cf68ba5396c3b/composer/datasets/in_context_learning_evaluation.py#L1167)), we specify the number of generations per sample with the variable `generations_per_sample`.

But when we call `build_icl_dataloader` in `llm-foundry` [here](https://github.com/eitanturok/llm-foundry/blob/96c47079e0db9d3532515f4634662beeadd2f3b8/llmfoundry/utils/builders.py#L521), we specified the number of generations per sample with the variable `num_beams`. This was confusing because 
1. we use different variable names
2. the variable `num_beams` is also used as a parameter for beam search in inference.

To resolve this, I replaced all instances of `num_beams` with `generations_per_sample` in `llm-foundry`. If `generations_per_sample` is not specified in the config and `num_beams` *is* specified, we will use the value set by `num_beams` and raise a warning that `num_beams` is depreciated.

Cheers!

### DIFF
--- a/llmfoundry/utils/builders.py
+++ b/llmfoundry/utils/builders.py
@@ -505,7 +505,7 @@ def build_icl_evaluators(
                     'To specify the number of generations per sample please use '
                     f'generations_per_sample: {icl_cfg.num_beams}. To specify '
                     'the number of beams for beam search decoding please use '
-                    f'num_beams: {icl_cfg.num_beams} in generation_kwargs.')
+                    f'num_beams: {icl_cfg.num_beams} in generation_kwargs.'),
                     DeprecationWarning,
                 )
             else:
--- a/llmfoundry/utils/builders.py
+++ b/llmfoundry/utils/builders.py
@@ -497,8 +497,16 @@ def build_icl_evaluators(
             icl_cfg.batch_size = default_batch_size
         if 'pass_at_k' not in icl_cfg:
             icl_cfg.pass_at_k = 1
-        if 'num_beams' not in icl_cfg:
-            icl_cfg.num_beams = 20
+        if 'generations_per_sample' not in icl_cfg:
+            if 'num_beams' in icl_cfg:
+                icl_cfg.generations_per_sample = icl.num_beams
+                warnings.warn(
+                    ('num_beams is deprecated and will be removed in a future release.'
+                     f'Please use generations_per_sample: {icl.num_beams}'),
+                    DeprecationWarning,
+                )
+            else:
+                icl_cfg.generations_per_sample = 20
 
     for icl_cfg in icl_tasks_list:
         assert isinstance(icl_cfg, DictConfig)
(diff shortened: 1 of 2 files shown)


llmfoundry/utils/builders.py
```diff
@@ -497,8 +497,16 @@ def _validate_cfg(icl_cfg: DictConfig):
             icl_cfg.batch_size = default_batch_size
         if 'pass_at_k' not in icl_cfg:
             icl_cfg.pass_at_k = 1
-        if 'num_beams' not in icl_cfg:
-            icl_cfg.num_beams = 20
+        if 'generations_per_sample' not in icl_cfg:
+            if 'num_beams' in icl_cfg:
+                icl_cfg.generations_per_sample = icl.num_beams
```
maxisawesome, line 502: probably meant icl_cfg.num_beams?

llmfoundry/utils/builders.py
```diff
@@ -497,8 +497,16 @@ def _validate_cfg(icl_cfg: DictConfig):
...
+                warnings.warn(
+                    ('num_beams is deprecated and will be removed in a future release.'
+                     f'Please use generations_per_sample: {icl.num_beams}'),
+                    DeprecationWarning,
```
maxisawesome, line 506: We're deprecating num_beams entirely in favor of passing generation keyword args in w/ `generation_kwargs`, mind updating the warning to say so?

maxisawesome: (also icl.num_beams here)

llmfoundry/utils/builders.py
eitanturok: Thanks -- fixed

llmfoundry/utils/builders.py
eitanturok: lol thanks, fixed the icl.num_beams

maxisawesome: I think something more like this:
```
'num_beams' is deprecated and will be removed in a future release. '
 'To specify keyword arguments for `model.generate()`, please use '
 '`generation_kwargs` with your corresponding values.' 
```
The idea is we don't want them to use generations_per_sample or num_beams - we want them to put any and all generation_kwargs in that dictionary, like this:
```
label: winograd
dataset_uri: eval/local_data/language_understanding/winograd_wsc.jsonl # ADD YOUR OWN DATASET URI
num_fewshot: [0]
icl_task_type: schema
generation_kwargs:
  num_beams: 9001
  temperateure: .9
```
(these are not real params btw)

We should also have a deprecation warning for `generations_per_sample` even though we're literally introducing it right now XD

also these strings are throwing a syntax error lol

eitanturok: According to the PR [Make CodeEval respect device_eval_batch_size](https://github.com/mosaicml/composer/pull/2969), `generations_per_sample` does not go in `generation_kwargs` -- it is it's own parameter. Thus we do not want to put it in `generation_kwargs`.

eitanturok:
Overall, these changes are so that we use the same variable names in both `llmfoundry` and `composer`. 

Right now, the PR [Make CodeEval respect device_eval_batch_size](https://github.com/mosaicml/composer/pull/2969) uses the variable `generations_per_sample` to specify the number of generations per sample and so we change `num_beams`, which is how we specify the number of generations per sample in `llmfoundry`, to `generations_per_sample`.

josejg:
I believe this PR is redundant with the accompanying foundry PR of the composer you linked: https://github.com/mosaicml/llm-foundry/pull/956

I didn't raise a DeprecationWarning because I do think specifying `num_beams` was a bug and misnomer, and thus should not be supported any longer as it's a footgun.

dakinggg:
I'm going to close this PR, it is being covered in #956 . Let me know if you disagree.

PR #973, opened 2024-02-13
Reset fewshot_seed for icl tasks

maxisawesome:
The default was incorrectly getting set to `17` instead of `1234` as we desired. This PR changes that.


llmfoundry/utils/builders.py
```diff
@@ -499,6 +499,7 @@ def _validate_cfg(icl_cfg: DictConfig):
             icl_cfg.pass_at_k = 1
         if 'num_beams' not in icl_cfg:
             icl_cfg.num_beams = 20
+        ## NOTE: This is one possible location to set the default
```
maxisawesome, line 502: We set many other defaults here. More explanation below.

llmfoundry/utils/builders.py
```diff
@@ -533,13 +538,18 @@ def _validate_cfg(icl_cfg: DictConfig):
                 num_fewshot=num_fewshot,
                 prompt_string=icl_cfg.prompt_string,
                 example_delimiter=icl_cfg.example_delimiter,
+                hf_loading_vars=hf_loading_vars,
+                hf_parsing_map=hf_parsing_map,
                 continuation_delimiter=icl_cfg.continuation_delimiter,
                 question_prelimiter=icl_cfg.get('question_prelimiter', ''),
                 destination_path=destination_path,
+                ## NOTE: This is the other possible location to set the default
+                fewshot_random_seed=icl_cfg.get('fewshot_random_seed', 1234),
```
maxisawesome, line 547: We could set variables that are potentially not in icl_cfg with `icl_cfg.get(<parameter name>, None)` but this will explicitly pass `None` in as the value rather than letting the default value in `get_icl_task_dataloader` be used. Thus we use `icl_cfg.get('fewshot_random_seed', 1234)`, effectively using `1234` as the default.

maxisawesome:
incorrect branch

PR #974, opened 2024-02-13
Fix fewshot_random_seed default setting

maxisawesome:
Using `seed` as the default for `fewshot_random_seed` sets it to `17` by default instead of `1234` which breaks regression testing. This fixes that.

Also removes some human eval descriptions that snuck in with the previous PR that should not have been there.

### DIFF
--- a/llmfoundry/utils/builders.py
+++ b/llmfoundry/utils/builders.py
@@ -53,7 +53,6 @@ def build_evaluators(
     device_eval_batch_size: int,
     icl_seq_len: int,
     icl_subset_num_batches: Optional[int],
-    fewshot_random_seed: Optional[int] = 1234,
 ) -> Tuple[List[Evaluator], List[str], Optional[EvalGauntlet]]:
 
     evaluators = []
@@ -73,7 +72,6 @@ def build_evaluators(
             tokenizer,
             device_eval_batch_size,
             icl_seq_len,
-            fewshot_random_seed,
             icl_subset_num_batches,
         )
         evaluators.extend(icl_evaluators)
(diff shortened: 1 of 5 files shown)


llmfoundry/utils/builders.py
```diff
@@ -547,8 +543,8 @@ def _validate_cfg(icl_cfg: DictConfig):
                 continuation_delimiter=icl_cfg.continuation_delimiter,
                 question_prelimiter=icl_cfg.get('question_prelimiter', ''),
                 destination_path=destination_path,
-                fewshot_random_seed=icl_cfg.get('fewshot_random_seed',
-                                                fewshot_random_seed),
+                ## NOTE: This is the other possible location to set the default
+                fewshot_random_seed=icl_cfg.get('fewshot_random_seed', 1234),
```
maxisawesome, line 547: We could set variables that are potentially not in icl_cfg with icl_cfg.get(<parameter name>, None) but this will explicitly pass None in as the value rather than letting the default value in get_icl_task_dataloader be used. Thus we use icl_cfg.get('fewshot_random_seed', 1234), effectively using 1234 as the default.

llmfoundry/utils/builders.py
```diff
@@ -504,6 +499,7 @@ def _validate_cfg(icl_cfg: DictConfig):
             icl_cfg.pass_at_k = 1
         if 'num_beams' not in icl_cfg:
             icl_cfg.num_beams = 20
+        ## NOTE: This is one possible location to set the default
```
maxisawesome, line 502: Can set defaults here or below.

scripts/eval/yamls/hf_eval.yaml
```diff
@@ -37,11 +37,11 @@ models:
 device_eval_batch_size: 4
 
 # FSDP config for model sharding
-fsdp_config:
-  sharding_strategy: FULL_SHARD
-  mixed_precision: FULL
-  forward_prefetch: True
-  limit_all_gathers: True
+# fsdp_config:
```
dakinggg, line 40: I thought the previous change here was intentional, was it not?

dakinggg (approved):

scripts/eval/yamls/hf_eval.yaml
maxisawesome: It was intentional, this was a mistake. w/ mixed init, fsdp is required

PR #979, opened 2024-02-15
Torch 2.2 - Part 2

dakinggg:
This part adds torch 2.2 to the ci workflow (and removes a dependency that doesn't work with 3.11). If this goes ok, Part 3 will remove support for 2.1 and update the images in the readme.


mvpatel2000 (approved):

j316chuck:
@dakinggg we saw some issues with the torch 2.2 images in the regression tests just fyi. We might need to fix on composer side before foundry tests pass as well

PR #981, opened 2024-02-16
PyTorch 2.2 - Part 3

dakinggg:
This remove support torch 2.1. Also updates some yamls that had fallen out of date.

### DIFF
--- a/mcli/mcli-pretokenize-oci-upload.yaml
+++ b/mcli/mcli-pretokenize-oci-upload.yaml
@@ -1,5 +1,5 @@
 name: c4-2k-pre-tokenized
-image: mosaicml/pytorch:1.13.1_cu117-python3.10-ubuntu20.04
+image: mosaicml/llm-foundry:2.2.0_cu121_flash2-latest
 compute:
   gpus: 8  # Number of GPUs to use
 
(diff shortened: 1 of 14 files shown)


dakinggg:
@mvpatel2000 I'm gonna red button this because the 2.2 tests pass on main, its just that the 2.1 tests are failing here because i bumped torch version so its trying to reinstall and doing it badly I think. but please lmk if that seems right

mcli/mcli-pretokenize-oci-upload.yaml
```diff
@@ -1,5 +1,5 @@
 name: c4-2k-pre-tokenized
-image: mosaicml/pytorch:1.13.1_cu117-python3.10-ubuntu20.04
+image: mosaicml/llm-foundry:2.1.0_cu121_flash2-latest
```
mvpatel2000, line 2: why are images 2.1?

mvpatel2000 (approved):
Should we bump composer docker latest to torch 2.2?

mcli/mcli-pretokenize-oci-upload.yaml
dakinggg: because the example yamls are pinned to the 0.5.0 release of llm foundry. we can update when we go to 0.6.0

PR #983, opened 2024-02-20
Token accuracy metrics

dakinggg:
Adds a token accuracy metric. Tested with a full training run as well.


dakinggg:
@mvpatel2000 Yup! Shouldn't be any cost, and is a useful metric to look at with loss/perplexity

irenedea (approved):

samhavens:
So we will now automatically see token accuracy in wandb?

dakinggg:
@samhavens yes

PR #985, opened 2024-02-21
Add deeper support for multi-turn chats and loss-generating tokens in finetuning

alextrott16:
The main purpose of this PR is to support training on non-terminal responses in multi-round chats. This is achieved by tokenizing at the level of conversation "turns" and exposing some options for what turns are used as training targets (i.e. generate loss). This also adds support for treating prompt tokens as loss-generating.

The script for converting a finetuning dataset to streaming has also been updated (with some bug fixes).


llmfoundry/data/finetuning/collator.py
```diff
@@ -226,6 +268,8 @@ def _process_and_batch_decoder_only(
             return_tensors='pt',
         )
 
+        batch['sequence_id'] = batch['attention_mask'] - 1
```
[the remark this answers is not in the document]
alextrott16: Yep that's right. This is just to make sure that there is `sequence_id` in the case that no bin packing is used (so there's always `sequence_id`).

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -522,49 +567,57 @@ def _build_collate_fn(
                                 torch.logical_and(
                                     is_subseq, batch['attention_mask'][j] ==
                                     1)],
-                                             skip_special_tokens=False))
+                                             skip_special_tokens=False,
+                                             clean_up_tokenization_spaces=True))
```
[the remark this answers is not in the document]
alextrott16: Just makes printouts better? Look it's tokenizers idk.

scripts/data_prep/convert_finetuning_dataset.py
```diff
@@ -236,33 +239,39 @@ def main(args: Namespace) -> None:
...
                         examples_removed += 1
                         continue
 
-                    sample_to_write = {}
+                    sample_to_write = {'turns': []}
                     # convert to bytes
-                    for key in columns.keys():
-                        sample_to_write[key] = np.asarray(sample[key],
-                                                          dtype=np.uint32)
+                    for turn in sample['turns']:
```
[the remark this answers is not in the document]
alextrott16: Based on offline discussion, keeping the removal of np arrays because json encoding/decoding works well.

tests/data/test_dataloader.py
```diff
@@ -549,10 +549,8 @@ def test_finetuning_dataloader_custom_split_remote(split: str):
 
 @pytest.mark.parametrize('pretokenize', [True, False])
 @pytest.mark.parametrize('use_multiple_streams', [True, False])
-@pytest.mark.parametrize('use_bytes', [True, False])
```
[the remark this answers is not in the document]
alextrott16: The changes to the way streaming examples are formatted made this setting irrelevant

llmfoundry/data/finetuning/tasks.py
```diff
@@ -291,18 +350,21 @@ def is_valid_ift_example(pad_token_id: int, max_seq_len: int,
...
         bool: Indicator of whether the input example is valid
     """
-    less_than_max_seq_len = len(example['input_ids']) < max_seq_len
-    non_empty_input = len(example['input_ids']) > 0
-    non_empty_labels = len(example['labels']) > 0
-    non_padding_response = any(
-        token_id != pad_token_id for token_id in example['labels'])
-    return (less_than_max_seq_len and non_empty_input and non_empty_labels and
-            non_padding_response)
+    for turn in example['turns']:
```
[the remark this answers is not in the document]
alextrott16: Good catch. No I don't think I thought this through all the way. Will revisit.

tests/data/test_dataloader.py
```diff
@@ -79,23 +68,34 @@ def build_mock_ft_streaming_dataset(
...
                 sample = tokenize_formatted_example(sample, tokenizer=tokenizer)
-                sample_to_write = {}
-                for key in columns.keys():
-                    if use_bytes:
-                        sample_to_write[key] = np.asarray(sample[key]).tobytes()
-                    else:
-                        sample_to_write[key] = np.asarray(sample[key],
-                                                          dtype=np.uint32)
+                sample_to_write = {'turns': []}
+                # convert to bytes
```
dakinggg, line 83: there is no bytes here right? cause we're just doing json

tests/data/test_dataloader.py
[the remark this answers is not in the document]
dakinggg: I don't quite follow this. We still want to test the case where the dataset was written with bytes, since we support loading that for backwards compat, right?

tests/data/test_dataloader.py
alextrott16: Correct. Oversight from me. I'll remove that comment.

tests/data/test_dataloader.py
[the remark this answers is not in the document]
alextrott16: No, no, no. Maintaining backwards compatibility doesn't mean that you _also_ maintain test coverage of the previous functionality, duh.... wait, no, that doesn't sound right.... I should probably restore the test cases.

tests/data/test_dataloader.py
```diff
@@ -713,6 +809,72 @@ def test_malformed_data(
...
+
+    device_batch_size = 1
+    dataloader = build_finetuning_dataloader(cfg, tokenizer,
+                                             device_batch_size).dataloader
+
+    # We should be able to iterate through this dataset without crashing
+    for i, batch in enumerate(dataloader):
+        # This check here just makes sure that the labels sequence is all padding
+        # (except for an EOS at the end)
+        for subseq in range(int(batch['sequence_id'][0].max()) + 1):
```
dakinggg, line 866: this for loop is not making sense to me. can you explain more whats going on here?

llmfoundry/data/finetuning/collator.py
```diff
@@ -13,6 +13,168 @@
...
+def stitch_turns_decoder_only(
+        example_turns: list[dict[str, list[int]]],
+        target_prompts: str,
+        target_responses: str,
+        eos_token_id: Optional[int] = None,
+        validate: bool = False) -> tuple[list[int], list[int]]:
+    target_prompts = target_prompts.lower()
+    target_responses = target_responses.lower()
+
+    if validate:
```
dakinggg, line 113: why is validation optional?

llmfoundry/data/finetuning/collator.py
```diff
@@ -13,6 +13,168 @@
...
+        validate_target_settings(target_prompts,
+                                 target_responses,
+                                 decoder_only_format=True)
+
+    if target_prompts.startswith('length'):
```
dakinggg, line 118: .

alextrott16: Instructions unclear

tests/data/test_dataloader.py
alextrott16: Packing is used here, so it's going through each packed sequence in sequence 0. `subseq` is the sequence index, so it's just looping through all the `sequence_id` values (which equals the number of sub-sequences packed into this sequence). Basically, this for loop is _unpacking_ the packed sequence.

llmfoundry/data/finetuning/collator.py
dakinggg: lol woops. i meant to come back and think about how to make this not have a special case for how it gets looked up in the registry...but not gonna block this pr on that

tests/data/test_dataloader.py
dakinggg: ah and the `[0]` everywhere is because your device batch size is 1

llmfoundry/data/finetuning/collator.py
alextrott16: Just to add flexibility I guess. There are other validation checks that should catch invalid settings in the dataloader and in the convert-to-streaming script. So, if we did validation every time this function was run, it'd be redundant with those checks and waste time. But I figured there might be times this function would want to validate on its own, so I added it as an option. And I figure this will be useful if we eventually allow each sample to have its own target prompts/responses settings.

tests/data/test_dataloader.py
alextrott16: Correct

dakinggg:
Ok, last test @alextrott16, can you launch a training run on chat formatted data with one of the new loss settings and just make sure it trains properly?

dakinggg (approved):

PR #987, opened 2024-02-22
Add default for eval interval

irenedea:
Set a default for eval_interval to reflect the composer default and so that it does not need to be explicitly expressed in the input yaml.


scripts/train/train.py
```diff
@@ -248,7 +248,8 @@ def main(cfg: DictConfig) -> Trainer:
                                                must_exist=True)
     eval_interval: Union[int, str] = pop_config(cfg,
                                                 'eval_interval',
-                                                must_exist=True)
+                                                default_value=1,
```
j316chuck, line 251: Does this default to 'ba', if so 1 might be too high?

irenedea: I just wanted to match the composer default. I think `1` is interpreted as `1ep`. (And if eval_loader is not specified, then this has no effect)

dakinggg (approved):

PR #989, opened 2024-02-22
if condition in tie weights added

megha95:
Issue: `AutoModelForCausalLM.from_pretrained()` calls `tie_weights()` function of `MPTForCausalLM` which resets `lm_head` (or output embeddings vector) to None. This creates a bug for when output embeddings are not shared with input embeddings in MPT models.

This PR fixes this bug by adding an if condition on config.


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -826,7 +826,8 @@ def set_output_embeddings(
             self.transformer.set_input_embeddings(new_embeddings)
 
     def tie_weights(self) -> None:
-        self.lm_head = None
+        if getattr(self.config, "tie_word_embeddings", True):
```
dakinggg, line 829: Can you add a really simple test that just checks that `tie_weights` does as expected if config is true, and no op if config is false?

dakinggg (approved):

llmfoundry/models/mpt/modeling_mpt.py
megha95: looks there is a test already for this in `tests/models/test_model.py::test_generate_with_device_map`

dakinggg: I meant, an explicit test that just calls `tie_weights` and makes sure it behaves as we want in both cases (also CI is failing, need to run precommit)

megha95: done

PR #990, opened 2024-02-22
Patch test, lock mcli version

aspfohl:

PR #991, opened 2024-02-22
Bump gha timeouts

aspfohl:
Hit both of these in https://github.com/mosaicml/llm-foundry/pull/990 - I think this is just a symptom of the codebase getting larger, but probably could be optimized a little


dakinggg (approved):
Sad times, approving since no time to make faster rn

PR #992, opened 2024-02-23
log details to metadata for run analytics

angel-ruiz7:
This will log information via the `MosaicMLLogger` to place the following keys in a run's metadata for [analytics purposes](https://adb-2548836972759138.18.azuredatabricks.net/sql/dashboardsv3/01eec385b9a41e3aacdabb8c7952b91e?o=2548836972759138). The data to log includes

- `model_name`: `string`
- `script`: 'Training', 'Eval'
- `train_task_type`: `PRETRAIN`, `CONTINUED_PRETRAIN`, `INSTRUCTION_FINETUNE`
- `train_loader_name`: `string`
- `train_dataset_hf_name`: `string`
- `eval_loader_name`: `string`
- `eval_dataset_hf_name`: `string`
- `tokenizer_name`: `string`
- `n_heads`: `number`
- `d_model`:  `int`
- `callbacks`: `string[]`
- `train_loader_workers`: `int`
- `eval_loader_workers`: `int`
- `gauntlet_configured`: `boolean`
- `icl_configured`: `boolean`

## Screenshots
# Using the [`Quickstart`](https://github.com/mosaicml/llm-foundry?tab=readme-ov-file#quickstart) example
Screenshot 2024-03-22 at 4 39 18 PM

# Using the [`gpt2-small`](https://github.com/mosaicml/llm-foundry/blob/main/scripts/train/yamls/pretrain/gpt2-small.yaml) example
Screenshot 2024-03-22 at 4 45 16 PM


angel-ruiz7:
still need to add the subtype. how should we do this? currently, we manually look through the command to find the yaml or python file to classify the run. what would be the best approach to do this inside of `train.py`?

cc @aspfohl @irenedea

scripts/eval/eval.py
```diff
@@ -56,6 +56,15 @@ def load_model(model_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
     return composer_model
 
 
+def log_analytics_details(mosaicml_logger: MosaicMLLogger, model_name: str):
+    mosaicml_logger.log_metrics({
+        'uses_llmfoundry': True,
```
aspfohl, line 61: I actually don't think we need this directly - presence of any foundry metadata will be enough on the analytics side to know the run uses foundry

scripts/eval/eval.py
```diff
@@ -56,6 +56,15 @@ def load_model(model_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+        'model_name': model_name,
+        'llmfoundry_run_type': 'Eval',
```
aspfohl, line 63: Could all of the keys be prefixed with something like `mosaicml/llmfoundry/`? Any other suggestions here? 

Eg etas are logged as `mosaicml/time/remaining_estimate`

aspfohl, line 63: Nit: could these be all lowercase?

scripts/eval/eval.py
```diff
@@ -118,6 +127,7 @@ def evaluate_model(
                                None)
 
         if mosaicml_logger is not None:
+            log_analytics_details(mosaicml_logger, model_cfg.model_name)
```
aspfohl, line 130: Is model_cfg.model_name the best thing to use here @irenedea @dakinggg ? Any suggestions for getting this standardized?

aspfohl (commented):
More of an ask for foundry team: Is there anything else that would be useful for analytics?

Should we make it configurable to turn this on or off? Or is presence of MosaicMLLogger enough (users could always turn it off via MOSAICML_PLATFORM env var)

scripts/eval/eval.py
angel-ruiz7: added `llmfoundry/` as a prefix b/c the mosaicml logger already prefixes w/ `mosaicml`

scripts/eval/eval.py
irenedea: Ah sorry, this is confusing but model_name actually is a user-defined name (or "nickname", rather) for the model being evaluated. This allows us to distinguish the models that are being evaluated. See [here](https://github.com/mosaicml/llm-foundry/blob/main/scripts/eval/yamls/hf_eval.yaml#L11) .

The model name we actually want to use is going to be `model.pretrained_model_name_or_path` if `model.name == 'hf_causal_lm'`. What this means is that the user is evaluating a model from huggingface (hence `hf_causal_lm`) loaded by name or path given by `pretrained_model_name_or_path`. For example, this would be for mpt-7b:
```
  model:
    name: hf_causal_lm
    pretrained_model_name_or_path: mosaicml/mpt-7b
    init_device: mixed
    pretrained: true
```

A few more things I think would be useful to log:
- If `model.name` is `hf_causal_lm`, whether `model.pretrained` is true. This tells us if users are loading in the pretrained weights. Otherwise, they may be specifying `load_path` or randomly initializing (like training from scratch)
- Whether `load_path` is provided (see [here](https://github.com/mosaicml/llm-foundry/blob/main/scripts/eval/yamls/mpt_eval.yaml#L27), for example). This tells us if users are loading model weights from a composer checkpoint. 
- Whether `name == 'mpt_causal_lm'`. This would give us signal on whether a user is bringing their own custom mpt-defined architecture.

scripts/eval/eval.py
```diff
@@ -56,6 +56,45 @@ def load_model(model_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
 
+def log_analytics_details(mosaicml_logger: MosaicMLLogger,
+                          model_config: DictConfig, tokenizer_name: str,
+                          load_path: Union[str,
+                                           None], icl_tasks: Union[str,
+                                                                   ListConfig],
+                          eval_gauntlet_config: Optional[Union[str,
+                                                               DictConfig]]):
+    metrics = {
+        'llmfoundry/llmfoundry_run_type': 'eval',
```
aspfohl, line 67: `llmfoundry_run_type` -> `script` maybe?

scripts/eval/eval.py
```diff
@@ -56,6 +56,45 @@ def load_model(model_cfg: DictConfig, tokenizer: PreTrainedTokenizerBase,
...
+
+    if model_config.get('vocab_size', None) is not None:
+        metrics['llmfoundry/vocab_size'] = model_config.get('vocab_size')
+    if model_config.get('d_model', None) is not None:
+        metrics['llmfoundry/d_model'] = model_config.get('d_model')
+    if model_config.get('n_heads', None) is not None:
+        metrics['llmfoundry/n_heads'] = model_config.get('n_heads')
+
+    if load_path is not None:
+        metrics['llmfoundry/cloud_provider_loading'] = load_path.split(':')[0]
```
aspfohl, line 88: This would raise a index error for local paths - I would build a helper (can use for train.py too) that safely parses load_path / save_folder

Also nit: maybe loading -> data?

scripts/eval/eval.py
```diff
@@ -151,6 +191,10 @@ def evaluate_model(
 
     assert composer_model is not None
 
+    if mosaicml_logger is not None:
+        log_analytics_details(mosaicml_logger, model_cfg, tokenizer_name,
```
aspfohl, line 195: A complication here: we call evaluate_model multiple times - so this (namely model name) could be overwritten. 

I'd maybe suggest just skipping model_name (& tokenizer name) for eval and logging for analytics a single time, outside of evaluate_model

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+                                                  None], tokenizer_name: str,
+                          load_path: Union[str, None], save_folder: Union[str,
+                                                                          None],
+                          icl_tasks_config: Optional[Union[ListConfig, str]],
+                          eval_gauntlet: Optional[Union[DictConfig, str]]):
+    metrics = {
+        'llmfoundry/tokenizer_name':
+            tokenizer_name,
+        'llmfoundry/llmfoundry_run_type':
+            'training',
```
aspfohl, line 169:
```suggestion
        'llmfoundry/script': 'train',
```

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+    # TODO: what's a good name for this? ideally we want the keys to be the same for
+    # training and eval to make querying easier.
+    if load_path is not None:
+        metrics['llmfoundry/cloud_provider_loading'] = load_path.split(':')[0]
+    if save_folder is not None:
+        metrics['llmfoundry/cloud_provider_saving'] = save_folder.split(':')[0]
+
+    if train_loader_config.get('dataset').get('hf_name', None) is not None:
+        metrics['llmfoundry/train_dataset_hf_name'] = train_loader_config.get(
+            'dataset').get('hf_name')
```
aspfohl, line 199:
```suggestion
            'dataset', {}).get('hf_name')
```

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+    if train_loader_config.get('name') == 'finetuning':
+        metrics['llmfoundry/llmfoundry_run_subtype'] = 'IFT (finetuning)'
```
aspfohl, line 201: What if we make these consistent with "task types" (from finetuning)
```suggestion
        metrics['llmfoundry/train_task_type'] = 'INSTRUCTION_FINETUNE'
```
& CONTINUED_PRETRAIN + PRETRAIN

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+        metrics['llmfoundry/eval_loaders'] = []
+
+        if isinstance(eval_loader_config, ListConfig):
+            eval_loader_configs: ListConfig = eval_loader_config
+        else:
+            eval_loader_configs = ListConfig([eval_loader_config])
+
+        for loader_config in eval_loader_configs:
+            eval_loader_info = {}
+            eval_loader_info['eval_loader_name'] = loader_config.get('name')
```
aspfohl, line 218:
```suggestion
            eval_loader_info['name'] = loader_config.get('name')
```

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+            eval_loader_info['eval_loader_workers'] = loader_config.get(
+                'dataset').get('num_workers')
```
aspfohl, line 220:
```suggestion
            eval_loader_info['num_workers'] = loader_config.get(
                'dataset', {}).get('num_workers')
```

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+            if loader_config['dataset'].get('hf_name', None) is not None:
```
aspfohl, line 221: `dataset` isn't guaranteed here, right?

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+                eval_loader_info['eval_dataset_hf_name'] = loader_config.get(
+                    'dataset').get('hf_name')
```
aspfohl, line 223:
```suggestion
                eval_loader_info['dataset_hf_name'] = loader_config.get(
                    'dataset', {}).get('hf_name')
```

scripts/train/train.py
```diff
@@ -149,6 +150,97 @@ def build_composer_model(model_cfg: DictConfig,
...
+
+            # Log as a key-sorted JSON string, so that we can easily parse it in Spark / SQL
+            metrics['llmfoundry/eval_loaders'].append(
+                json.dumps(eval_loader_info, sort_keys=True))
+
+    # TODO: do we need error checking here?
+    if model_config['name'] == 'hf_casual_lm':
```
aspfohl, line 230: Question for @dakinggg @irenedea How do we get the other model names?

scripts/eval/eval.py
angel-ruiz7: I'm thinking about moving this out to `main`. The only thing I'm curious about here is how this affects [certain fields that rely on the `model_cfg` variable](https://github.com/mosaicml/llm-foundry/blob/angel/log-data-for-run-analytics/scripts/eval/eval.py#L60). Since we load multiple, should I throw this information into a list like we did with `llmfoundry/eval_loaders`?

scripts/eval/eval.py
angel-ruiz7: I _think_ that in this case, the `split()` function would just return the whole path for local paths. 
Screenshot 2024-03-11 at 1 03 43 PM

Something that I didn't think about though was the naming of these paths. In this case, should we just put `local` instead of the pathname?

scripts/eval/eval.py
angel-ruiz7: Implemented this solution tentatively. If you think it will be useful, I can add `model_name` and `tokenizer_name` back into metrics, but for now I removed them per the comment above. @aspfohl

scripts/eval/eval.py
```diff
@@ -287,6 +336,15 @@ def main(cfg: DictConfig) -> Tuple[List[Trainer], pd.DataFrame]:
     models_df = None
     composite_scores = None
     trainers = []
+
+    mosaicml_logger = find_mosaicml_logger([
+        build_logger(name, logger_cfg)
```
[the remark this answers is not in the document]
angel-ruiz7: @aspfohl mentioned that since `evaluate_model` is called twice, we're likely to overwrite some values with our current logging paradigms. I think that logging everything at once is good from an analytics perspective because we can see everything before all of the calls to `evaluate_model` are ran (although i'm not sure how long this usually takes?).

the reason why i built the loggers again on the outside is that we [pop/update the `loggers_cfg` metadata every time `evaluate_model` is called](https://github.com/mosaicml/llm-foundry/blob/angel/log-data-for-run-analytics/scripts/eval/eval.py#L164). (naive question) why do we do this, and how should we go about this?

I have two ideas: 
1. move all analytics logging into `main`. for `loggers_cfg`, can we pop/update the metadata once as well? If not, we can pass in a function (or `loggers_cfg`) to perform these operations in `evaluate_model`. **this is what I will do right now**
2. if logging for every model is better, we can log analytics for each one, but rename them or place them in a larger list so that we don't have naming conflicts/overwrites in our metadata. For example, we can create a list in metadata, `llmfoundry/models`, and update this list with model data every time `evaluate_model` is called.

Thoughts? @dakinggg @irenedea @aspfohl

llmfoundry/utils/logging_utils.py
```diff
@@ -2,6 +2,29 @@
...
+    return next(
+        (logger for logger in loggers if isinstance(logger, MosaicMLLogger)),
+        None)
+
+
+def get_cloud_provider_from_path(path: str) -> str:
+    """Gets the cloud provider from the a given string path.
+
+    If we see a ':', we know that the service provider is present in the URI.
+    Otherwise, we assume that the model is local.
```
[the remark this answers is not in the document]
angel-ruiz7: > https://github.com/mosaicml/composer/blob/dev/composer/utils/file_helpers.py#L327

gotta love copilot 🤠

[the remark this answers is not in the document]
angel-ruiz7: @irenedea @dakinggg -- can we just grab the providers from `save_folder` / `load_path` in composer? it looks like the same paths from foundry are passed into there, but I want to make sure this is correct.

In this case, I think it may be better to log `save_folder` / `load_path` from composer, rather than foundry. That way we can make sure to grab these providers for non-foundry runs as well. How do y'all feel about this?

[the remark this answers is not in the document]
dakinggg: Yeah those can be grabbed in composer

scripts/eval/eval.py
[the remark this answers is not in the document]
irenedea: Oh i see, yeah that makes sense. Let's just build all the loggers once and then pass the built loggers to evaluate_model

llmfoundry/utils/mosaicmllogger_utils.py
```diff
@@ -0,0 +1,166 @@
...
+import json
+from typing import Any, Dict, List, Optional, Union
+
+from composer.loggers import MosaicMLLogger
+from composer.loggers.logger_destination import LoggerDestination
+from composer.utils.file_helpers import parse_uri
+from omegaconf import DictConfig, ListConfig
+
+
+def find_mosaicml_logger(
```
irenedea, line 12: add a docstring for this

llmfoundry/utils/logging_utils.py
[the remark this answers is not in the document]
irenedea: Yes, I think that's fine to log save_folder and load_path from composer.

irenedea (commented):
Can you include evidence of this working in the PR description? Some manual tests and screenshots would be good.

llmfoundry/utils/__init__.py
```diff
@@ -13,23 +13,18 @@
...
-    'update_batch_size_info',
-    'log_config',
-    'pop_config',
+    'build_callback', 'build_logger', 'build_algorithm', 'build_optimizer',
+    'build_scheduler', 'build_icl_evaluators', 'build_tokenizer',
+    'calculate_batch_size_info', 'convert_and_save_ft_weights',
+    'download_from_http_fileserver', 'download_from_hf_hub',
+    'get_hf_tokenizer_from_composer_state_dict', 'update_batch_size_info',
+    'log_config', 'pop_config', 'find_mosaicml_logger', 'log_eval_analytics',
+    'log_train_analytics'
```
irenedea, line 29:
```suggestion
    'log_train_analytics',
```

add a trailing comma so the auto format is nice

irenedea (approved):
🚀 LGTM! Just one super super tiny formatting comment :) Thanks Angel! Will be great to have more logging and data 😄

PR #1001, opened 2024-02-29
add magic filename for sharded state dicts

milocress:


scripts/train/train.py
```diff
@@ -262,10 +262,15 @@ def main(cfg: DictConfig) -> Trainer:
...
-                                           must_exist=False,
-                                           default_value='latest-rank{rank}.pt')
+    is_state_dict_sharded: bool = cfg.get('fsdp_config',
+                                          {}).get('state_dict_type',
+                                                  'sharded') == 'sharded'
+    save_latest_filename: str = pop_config(
+        cfg,
+        'save_latest_filename',
+        must_exist=False,
+        default_value='latest-rank{rank}_sharded_magic_keyword.pt'
```
dakinggg, line 272:
```suggestion
        default_value='latest-sharded-rank{rank}'
```

scripts/train/train.py
```diff
@@ -262,10 +262,15 @@ def main(cfg: DictConfig) -> Trainer:
                                             'save_folder',
                                             must_exist=False,
                                             default_value=None)
-    save_latest_filename: str = pop_config(cfg,
-                                           'save_latest_filename',
-                                           must_exist=False,
-                                           default_value='latest-rank{rank}.pt')
+    is_state_dict_sharded: bool = cfg.get('fsdp_config',
```
dakinggg, line 265: the fsdp config is already popped off above, you should use that

dakinggg: this doesn't look resolved?

scripts/train/train.py
```diff
@@ -262,10 +262,14 @@ def main(cfg: DictConfig) -> Trainer:
                                             'save_folder',
                                             must_exist=False,
                                             default_value=None)
-    save_latest_filename: str = pop_config(cfg,
-                                           'save_latest_filename',
-                                           must_exist=False,
-                                           default_value='latest-rank{rank}.pt')
+    is_state_dict_sharded: bool = (fsdp_config.get('state_dict_type', 'sharded')
```
dakinggg, line 265: isn't the default `full`?

scripts/train/train.py
```diff
@@ -262,10 +262,14 @@ def main(cfg: DictConfig) -> Trainer:
...
+                                   == 'sharded') if fsdp_config else True
```
dakinggg, line 266: if there is no fsdp, the checkpoint is not sharded

scripts/train/train.py
milocress: no we're changing the default to sharded I think

dakinggg: ah no, the intention is not to change the default in llm foundry, just in the finetuning api. sorry, was not clear about this

dakinggg (approved):
LGTM! Could you do a manual test that (1) if you set sharded state dict, you get a sharded file name, and (2) if you don't, you get a full file name?

Feel free to merge after doing that

PR #1003, opened 2024-03-02
Fix extra BOS token in front of response for some tokenizers

dakinggg:
Some tokenizers, like llama and mistral, automatically add a BOS token to the start of the input to the tokenizer. This is fine and expected. However, we did not realize that when you call `tokenizer(text='hello', text_target='goodbye')`, it _also_ automatically adds a BOS token to the `labels`. This is unexpected, and would require a user to suffix all of their prompts with BOS after finetuning. So we automatically strip BOS from the labels now for tokenizers that have this behavior.

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -231,11 +231,7 @@ def _tokenize_chat_formatted_example(
         TokenizedExample: The tokenized example.
     """
     prompt, response = _slice_chat_formatted_example(example, tokenizer)
-    return _tokenize_with_bos_removal(
-        tokenizer=tokenizer,
-        text=prompt,
-        text_target=response,
-    )
+    return tokenizer(text=prompt, text_target=response)
 
 
 def _tokenize_prompt_response_formatted_example(
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -246,7 +246,15 @@ def _tokenize_prompt_response_formatted_example(
             f'Unable to tokenize example because {response_key} was not a string. {example=}'
         )
 
-    return tokenizer(text=prompt, text_target=response)
+    tokenized_sample = tokenizer(text=prompt, text_target=response)
+
+    # Remove the BOS token from the start of the labels if it was automatically added
+    if hasattr(tokenizer, 'add_bos_token') and tokenizer.add_bos_token:
+        if tokenizer.bos_token_id is not None and tokenized_sample['labels'][
+                0] == tokenizer.bos_token_id:
+            tokenized_sample['labels'] = tokenized_sample['labels'][1:]
+
+    return tokenized_sample
 
 
 def tokenize_formatted_example(
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -195,6 +195,29 @@ def _slice_chat_formatted_example(
     return prompt, response
 
 
+def _tokenize_with_bos_removal(tokenizer: PreTrainedTokenizerBase, text: str,
+                               text_target: str) -> TokenizedExample:
+    """Tokenizes the prompt and response using the provided tokenizer.
+
+    Args:
+        tokenizer (PreTrainedTokenizerBase): The tokenizer to use for tokenization.
+        prompt (str): The prompt to tokenize.
+        response (str): The response to tokenize.
+
+    Returns:
+        TokenizedExample: The tokenized example.
+    """
+    tokenized_sample = tokenizer(text=text, text_target=text_target)
+
+    # Remove the BOS token from the start of the labels if it was automatically added
+    if hasattr(tokenizer, 'add_bos_token') and tokenizer.add_bos_token:
+        if tokenizer.bos_token_id is not None and tokenized_sample['labels'][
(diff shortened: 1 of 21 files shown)


tests/data/test_template_tokenization.py
```diff
@@ -163,3 +163,23 @@ def test_tokenize_instruct_example_well_formed():
...
+    # This tokenizer automatically adds bos tokens
+    tokenizer = transformers.AutoTokenizer.from_pretrained(
+        'mistralai/Mixtral-8x7B-v0.1')
+
+    example = {'prompt': 'prompt', 'response': 'response'}
+
+    tokenized_example = tokenize_formatted_example(example, tokenizer)
+
+    assert len(tokenized_example['labels']) == 1
+    assert tokenized_example['labels'][0] != '<s>'
```
irenedea, line 178: `tokenized_example['labels']` produces a list of int. How about comparing against tokenizer.bos_token_id

```suggestion
    assert len(tokenized_example['labels']) == 1
    assert tokenized_example['labels'][0] != tokenizer.bos_token_id
    assert len(tokenized_example['input_ids'] == 2
    assert tokenized_example['input_ids'][0] == tokenizer.bos_token_id
```

llmfoundry/data/finetuning/tasks.py
```diff
@@ -246,7 +246,15 @@ def _tokenize_prompt_response_formatted_example(
             f'Unable to tokenize example because {response_key} was not a string. {example=}'
         )
 
-    return tokenizer(text=prompt, text_target=response)
+    tokenized_sample = tokenizer(text=prompt, text_target=response)
+
+    # Remove the BOS token from the start of the labels if it was automatically added
+    if hasattr(tokenizer, 'add_bos_token') and tokenizer.add_bos_token:
```
irenedea, line 252: we will also need to do this above on L211. pull out into a shared helper fn?

tests/data/test_template_tokenization.py
```diff
@@ -163,3 +163,23 @@ def test_tokenize_instruct_example_well_formed():
             tokenized_example = tokenize_formatted_example(example, tokenizer)
             assert 'input_ids' in tokenized_example
             assert 'labels' in tokenized_example
+
+
+def test_tokenize_no_labels_bos():
+    # This tokenizer automatically adds bos tokens
+    tokenizer = transformers.AutoTokenizer.from_pretrained(
+        'mistralai/Mixtral-8x7B-v0.1')
+
```
irenedea, line 172:
```suggestion
     assert tokenizer.add_bos_token == True
```

tests/data/test_template_tokenization.py
```diff
@@ -163,3 +163,23 @@ def test_tokenize_instruct_example_well_formed():
...
+    assert len(tokenized_example['labels']) == 1
+    assert tokenized_example['labels'][0] != '<s>'
+
+    # This tokenizer does not have the add_bos_token attribute
+    tokenizer = transformers.AutoTokenizer.from_pretrained('mosaicml/mpt-7b')
+
+    tokenized_example = tokenize_formatted_example(example, tokenizer)
+
+    assert len(tokenized_example['labels']) == 1
```
irenedea, line 185:
```suggestion
    assert len(tokenized_example['labels']) == 1
    assert tokenized_example['labels'][0] != tokenizer.bos_token_id
    assert len(tokenized_example['input_ids']) == 1
    assert tokenized_example['input_ids'][0] != tokenizer.bos_token_id
```

irenedea (approved):
Thanks for fixing! Just need to make sure it's done for chat tokenizer as well.

tests/data/test_template_tokenization.py
dakinggg: oops, thanks

llmfoundry/data/finetuning/tasks.py
dakinggg: ah good catch, thanks. will fix

alextrott16 (commented):
So, not to be that guy, but I'm not sure that I agree with this behavior. It also conflicts with some of the incoming changes in the [multi-turn chat PR](https://github.com/mosaicml/llm-foundry/pull/985/files/b3e19d1a577aec2ae1202e04e326bfbea7c37552#top).

I am less opinionated about what should happen in the prompt-response formatted case, but when applying the chat template I think the best course of action is to use `tokenize(..., add_special_tokens=False)`. The reason is that the convention is to have the chat template add the special tokens itself into the templated string before tokenizing. Internally in the HF code, when using `apply_chat_template` with `tokenize=True`, you'll see that the method manually turns off adding special tokens when tokenizing because they have already been baked into the templated string. My additions in the linked PR reflect that convention, so I think our bases will be covered in that case.

When it comes to the prompt-response format (where you never use `apply_chat_template`), I think the correct thing to do is less clear. For one, starting the `labels` tokens with BOS is actually correct in the encoder-decoder case. For the decoder-only case, having the first generated token be BOS is really not that bad, IMHO. And it should not change the way that prompts need to be formatted at inference time.

I think the thing that matters most is that our tokenization here reflects the conventions we expect to be applied at inference time.

dakinggg:
Discussed offline, deferring the chat changes to https://github.com/mosaicml/llm-foundry/pull/985

alextrott16 (approved):
LGTM

PR #1006, opened 2024-03-02
Fix evaluators actually pulling eval metrics

mvpatel2000:
As title. Previously evaluators were pulling train metrics.


dakinggg (commented):
manual and unit test please

dakinggg (approved):

PR #1012, opened 2024-03-05
adding stopping time for current dataset in CurriculumLearning callback

ShashankMosaicML:


llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -94,6 +100,15 @@ def after_load(self, state: State, logger: Logger):
...
             self.all_dataset_configs.append(self.current_dataset_config)
 
+    def batch_start(self, state: State, logger: Logger) -> None:
+        del logger
+        if (self.current_dataset_max_duration is not None) and (
+                state.timestamp > self.current_dataset_max_duration):
+            state.stop_training()
+            log.debug(
+                f'Training stopped for the current dataset in curriculum learning as the training has reached current_dataset_max_duration = {self.current_dataset_max_duration}.'
+            )
```
vchiley, line 110:
```suggestion
    def before_dataloader(self, state: State, logger: Logger) -> None:
        del logger
        if (self.current_dataset_max_duration is not None) and (
                state.timestamp > self.current_dataset_max_duration):
            log.debug(
                f'Training stopped for the current dataset in curriculum learning as the training has reached current_dataset_max_duration = {self.current_dataset_max_duration}.'
            )
            state.stop_training()
```

seems like this should be triggered at `BEFORE_DATALOADER` (or we could move it to `BATCH_CHECKPOINT`)

also just to be safe, I wouldn't have anything happen after `state.stop_training()`

PR #1014, opened 2024-03-06
Custom Exceptions for Mosaic Logger

jjanezhang:
Creates custom exceptions to label all user solvable errors in the finetuning run path. Logs exceptions in these areas using the MosaicML logger.

Manually tested by running ft yaml with missing train split.
image

Dependent on: https://github.com/mosaicml/composer/pull/3089


llmfoundry/data/finetuning/tasks.py
```diff
@@ -44,13 +44,20 @@ def preprocessing_fn(example: Dict) -> Dict[str, str]:
...
+from composer.loggers.mosaicml_logger import (MOSAICML_ACCESS_TOKEN_ENV_VAR,
+                                              MOSAICML_PLATFORM_ENV_VAR)
 from composer.utils import dist
 from streaming import Stream, StreamingDataset
 from transformers import PreTrainedTokenizerBase
 
 from llmfoundry.utils.logging_utils import SpecificWarningFilter
 
 log = logging.getLogger(__name__)
+if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower(
```
dakinggg, line 57: can you just try/except in train.py?

dakinggg: oh you did, i assume this one can be removed?

jjanezhang:
I could move the new conversation chat tests to the `test_teamplate_tokenization.py` file if that's more relevant?

llmfoundry/data/finetuning/tasks.py
```diff
@@ -774,9 +771,12 @@ def build_from_hf(
                 if not all(
                         Path(f).suffix in SUPPORTED_EXTENSIONS
                         for f in dataset_files):
-                    raise ValueError(
-                        f'Dataset at local path {dataset_name} contains invalid file types. '
-                        + f'Allowed file types are: {SUPPORTED_EXTENSIONS}')
+                    raise InvalidFileExtensionError(dataset_name,
+                                                    SUPPORTED_EXTENSIONS)
+
+            if split is None:
```
dakinggg, line 777: any particular reason you added this? The split none check is higher up in the stack (dataloader.py)

scripts/train/train.py
```diff
@@ -499,11 +499,16 @@ def main(cfg: DictConfig) -> Trainer:
 
     # Dataloaders
     log.info('Building train loader...')
-    train_loader = build_dataloader(
-        train_loader_config,
-        tokenizer,
-        device_train_batch_size,
-    )
+    try:
```
dakinggg, line 502: you probably want to wrap the eval loader too

tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import List
+
+import pytest
+
+
+def generate_exclusive_test_params(param_names: List[str]):
```
dakinggg, line 9: can you docstring this? its a bit hard to read

llmfoundry/utils/exceptions.py
```diff
@@ -0,0 +1,215 @@
...
+        super().__init__(message)
+
+
+class NotEnoughDatasetSamplesError(ValueError):
+    """Error thrown when there is not enough data to train a model."""
+
+    def __init__(self, dataset_name: str, split: str,
+                 dataloader_batch_size: int, world_size: int,
+                 full_dataset_size: int, minimum_dataset_size: int) -> None:
+        self.dataset_name = dataset_name
```
dakinggg, line 25: do these need to be set as attrs for some reason? same throughout

llmfoundry/utils/exceptions.py
```diff
@@ -0,0 +1,215 @@
...
+## Tasks exceptions
+class UnknownConversationTypeError(KeyError):
+    """Error thrown when an unknown conversation type is used in a task."""
+
+    def __init__(self, example: Mapping) -> None:
+        message = f'Unknown conversation type {example=}'
+        super().__init__(message)
+
+
+class TooManyKeysInExampleError(ValueError):
```
dakinggg, line 50: seems like we should probably say what the keys in the offending example are?

dakinggg (commented):
lgtm, would like a second pair of eyes just because its touching so many places cc @irenedea

llmfoundry/utils/exceptions.py
jjanezhang: yes, will use them to construct the user facing messages

dakinggg: Meaning when you intercept with excepthook? Cause here you just pass it along to ValueError

jjanezhang: When we catch the exception with the mosaic logger, we can pull the attributes out and also save them in the run metadata.

dakinggg: ahh ok. got it.

llmfoundry/data/finetuning/tasks.py
jjanezhang: I was running a manual test with a run that was missing an HF split and was finding that in `_build_collate_fn` before the split check later in the ft dataloader, we end up calling `build_from_hf` which would throw the missing split exception.

dakinggg: hm, I don't think is quite the correct condition. If you look where its checked in the dataloader, split is only required if its a remote jsonl dataset, not necessarily required if its an actual hf dataset. at least that is the idea. also the `_build_collate_fn` should go through the same `build_finetuning_dataloader` code, so should hit the other error. How was it failing?

jjanezhang: I get a 
```
│ /usr/lib/python3/dist-packages/llmfoundry/data/finetuning/tasks.py:793 in    │
│ build_from_hf                                                                │
│                                                                              │
│   790 │   │   │   detected_cpus_with_margin = detected_cpu_count - 8         │
│   791 │   │   │   num_cpus_to_use = max(1, detected_cpus_with_margin)        │
│   792 │   │   │                                                              │
│ ❱ 793 │   │   │   columns_to_remove = list(dataset[0].keys())                │
│   794 │   │   │   tokenized_dataset = dataset.map(                           │
│   795 │   │   │   │   dataset_mapper,                                        │
│   796 │   │   │   │   batched=False,                                         │
│                                                                              │
│ /usr/lib/python3/dist-packages/datasets/dataset_dict.py:80 in __getitem__    │
│                                                                              │
│     77 │   │   │   │   split for split in (Split.TRAIN, Split.TEST, Split.VA │
│     78 │   │   │   ]                                                         │
│     79 │   │   │   suggested_split = available_suggested_splits[0] if availa │
│ ❱   80 │   │   │   raise KeyError(                                           │
│     81 │   │   │   │   f"Invalid key: {k}. Please first select a split. For  │
│     82 │   │   │   │   f"`my_dataset_dictionary['{suggested_split}'][{k}]`.  │
│     83 │   │   │   │   f"Available splits: {sorted(self)}"                   │
╰──────────────────────────────────────────────────────────────────────────────╯
KeyError: "Invalid key: 0. Please first select a split. For example:
`my_dataset_dictionary['train'][0]`. Available splits: ['test', 'train']"
``` 

when I use the `mosaicml/dolly_hhrlhf` as my hf dataset name with no split which isn't caught by our code?

dakinggg: I believe that if the dataset _has_ splits, you must specify one, but I think you can have a dataset without splits. I'm actually not 100% sure about this. Maybe its ok to just force split here. I can't find an example dataset that doesn't have a split right now, so maybe I'm just wrong about that. @alextrott16 do you know?

scripts/train/train.py
```diff
@@ -510,11 +510,16 @@ def main(cfg: DictConfig) -> Trainer:
...
```
irenedea, line 513: Instead of doing this try-except, wdyt about using sys.excepthook with @dakinggg's new custom init code extension: https://github.com/mosaicml/llm-foundry/blob/main/scripts/train/train.py#L163

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -446,14 +456,18 @@ def _args_str(original_args: Namespace) -> str:
...
+                            concat_tokens=args.concat_tokens,
+                            eos_text=args.eos_text,
+                            bos_text=args.bos_text,
+                            no_wrap=args.no_wrap,
+                            compression=args.compression,
+                            processes=args.processes,
+                            reprocess=args.reprocess,
+                            args_str=_args_str(args))
+    except Exception as e:
+        mosaicml_logger.log_exception(e)
```
irenedea, line 472: this will error if mosaicml_logger is not set above

scripts/train/train.py
irenedea: I realize @dakinggg's init code extension point won't work with the data prep scripts. Is it still possible to use sys.excepthook directly instead of try-except?

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -20,8 +23,16 @@
...
                                               merge_shard_groups)
+from llmfoundry.utils.exceptions import (InputFolderMissingDataError,
+                                         OutputFolderNotEmptyError)
 
 log = logging.getLogger(__name__)
+mosaicml_logger = None
+if os.environ.get(MOSAICML_PLATFORM_ENV_VAR, 'false').lower(
+) == 'true' and os.environ.get(MOSAICML_ACCESS_TOKEN_ENV_VAR):
+    # Adds mosaicml logger to composer if the run was sent from Mosaic platform, access token is set
+    mosaicml_logger = MosaicMLLogger()
```
irenedea, line 34: can we move this to under `if __name__ == '__main__':`, doesn't seem used elsewhere in the file so no need for it to be a global var

scripts/data_prep/convert_text_to_mds.py
irenedea: should wrap in a conditional check to make sure mosaicml_logger is not None

scripts/data_prep/convert_text_to_mds.py
irenedea: similar for the convert_delta_to_json file

irenedea: Since this logic is repeated in several places, can we use a helper function instead?

scripts/train/train.py
jjanezhang: I'm planning to do try-except as a temp solution (~next month or so). Another option would be to create an excepthook override script in foundry and run it at the start of `train.py` + all the data prep scripts. 

I am working on a cloud change to inject excepthook override for all python interpreters used during a run, and then we can remove these.

irenedea: Gotcha, OK with leaving the try-excepts as is then. Thanks for explaining!

llmfoundry/utils/exceptions.py
```diff
@@ -0,0 +1,215 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Custom exceptions for the LLMFoundry."""
```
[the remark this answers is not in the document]
jjanezhang: I thought about doing more general errors, something like this:
```
class FinetuneDataFormatError(Exception):
    def __init__(self, message: str, attributes: List[Any], errorType: enum) -> None:
        self.finetuneErrorType = errorType
        self.attributes = attributes
        super().__init__(message)

raise FinetuneDataFormatError(f'Expected 2 keys in message, but found {len(keys)}', [keys], IncorrectMessageKeyQuantityError)
```
so that in mcloud, we would get the finetuneErrorType and attributes and map it to "Expected 2 keys in each chat message (role and content), but found keys ${attributes.keys} in at least one message in your train dataset."

I chose against this because I found that supporting the message we want to raise in foundry and the required data we need for the mcloud user message construction is harder to maintain if it is distributed in multiple files throughout the codebase 

and we don't want to hard code the user message in foundry because it's awk to show those open source and it's harder to make changes for quick release

llmfoundry/data/finetuning/tasks.py
jjanezhang: Moved the check up in the dataloader and made the split not option in `build_from_hf`!

llmfoundry/utils/exceptions.py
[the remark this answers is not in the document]
irenedea: sgtm!

irenedea (approved):

llmfoundry/data/finetuning/tasks.py
```diff
@@ -94,13 +107,11 @@ def _get_example_type(example: Example) -> ExampleType:
...
-    elif any([
-            pr in example
-            for pr in _ALLOWED_PROMPT_KEYS.union(_ALLOWED_RESPONSE_KEYS)
-    ]):
+    elif any(p in example for p in _ALLOWED_PROMPT_KEYS) and any(
+            r in example for r in _ALLOWED_RESPONSE_KEYS):
         return 'prompt_response'
     else:
-        raise KeyError(f'Unknown conversation type {example=}')
+        raise UnknownConversationTypeError(example)
```
irenedea, line 114: Change to UnknownExampleTypeError

irenedea: make sure to change the error message along with it, thanks!

PR #1018, opened 2024-03-08
Bump transformers to 4.38.2

dakinggg:
The substantive change here is to fix rope, as hf changed the api...again


ShashankMosaicML (commented):
Do you think it would be a good idea to make tests configurable with transformers version? Are there tests currently that are configured by transformers version?

This is because now that we have different code paths for different transformers versions, we would like to ensure everything works for all the versions.

dakinggg:
I don't really want to set up tests for each transformer version as that will increase our test load quite a bit, and LLM Foundry actually pins transformers versions _beyond_ the ones that are special cased here. It is true that we are not actually testing the BC, apart from that I ran manually the `test_flash_torch_triton.py` tests on 4.37 and 4.38. I think this is sufficient for now, and we can revisit in the future if needed.

ShashankMosaicML (approved):

mvpatel2000 (approved):

PR #1023, opened 2024-03-11
Bump version to 0.6.0

dakinggg:
- [ ] Run regression tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,11 +99,11 @@ extra_deps['tensorboard'] = [
     'mosaicml[tensorboard]>=0.20.1,<0.21',
 ]
 
+# Flash 2 group kept for backwards compatibility
 extra_deps['gpu-flash2'] = [
     'flash-attn==2.5.0',
 ]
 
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -98,16 +99,13 @@
...
-    'flash-attn==1.0.9',
-    # PyPI does not support direct dependencies, so we remove this line before uploading from PyPI
-    'xentropy-cuda-lib@git+https://github.com/HazyResearch/flash-attention.git@v1.0.9#subdirectory=csrc/xentropy',
-]
-
 extra_deps['gpu-flash2'] = [
     'flash-attn==2.5.0',
 ]
 
+# Removing support for flash attention 1 without breaking the installation commands
```
mvpatel2000, line 106: This should instead be a comment on flash2 saying here for backwards compatibility reasons imo.

mvpatel2000 (approved):

setup.py
dakinggg: fair

irenedea (approved):

PR #1036, opened 2024-03-14
Fix llama attention patch

dakinggg:
Missed in the latest release (its deprecated so probably won't do a patch, but fixing on main anyway)

Tests pass locally: `10 passed, 12 deselected, 12 warnings in 7.35s`
Full train run comparison:
Screenshot 2024-03-14 at 4 53 24 PM


irenedea (approved):
👍

PR #1038, opened 2024-03-15
Adds a decorator for experimental features

dakinggg:
Also adds the decorator to a couple places

### DIFF
--- a/llmfoundry/utils/warnings.py
+++ b/llmfoundry/utils/warnings.py
@@ -1,6 +1,7 @@
 # Copyright 2024 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
-
+from typing import Any, Callable
+import warnings
 
 class VersionedDeprecationWarning(DeprecationWarning):
     """A custom deprecation warning class that includes version information.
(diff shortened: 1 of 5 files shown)


llmfoundry/utils/warnings.py
```diff
@@ -38,3 +40,16 @@ def __init__(self, feature_name: str) -> None:
         super().__init__(
             f'{feature_name} is experimental and may change with future versions.'
         )
+
+
+# Decorator version of experimental warning
+def experimental(feature_name: str):
```
irenedea, line 46: nit: add doc string

irenedea: wdyt about making feature_name optional and defaulting to `func.__name__`?

irenedea (approved):
nice! had a nit and a small suggestion

llmfoundry/utils/warnings.py
dakinggg: will add docstring

dakinggg: on the default, i'd actually prefer to force the writer of the code to think about what the message will actually be, rather than default. not super strong opinion tho

irenedea: In theory, I agree with that, but in practice it seems like we'll just be writing the names of classes again and the classes _should_ just be named well. Would like to not have this extra room for misspellings haha

dakinggg: well so far there are 3, and one of those i changed the string from the name to make more clear

irenedea: 66% uses the default then 😄 ? tbh, don't feel strongly here, OK to leave as is

PR #1039, opened 2024-03-15
Check the user provided eos / bos token id against the tokenizer eos / bos token id

ShashankMosaicML:
Currently, we need to specify the eos or bos token id in the dataset config for sequence id masking to work. This PR adds code to check if the provided eos and bos token ids match the token ids in the tokenizer. If there is a mismatch, an error will be raised. The error can be suppressed through the flags `override_eos_token_id_mismatch_error` and `override_bos_token_id_mismatch_error`

Verification runs: 
```
125m-incorrect-eos-token-id-override-B1AOYT
125m-incorrect-eos-token-id-5DXPCI 
125m-correct-eos-token-id-Ge3qzT 
```

```
125m-incorrect-bos-token-id-override-mxBMzA
125m-incorrect-bos-token-id-EeBgKH
125m-correct-bos-token-id-I97jYG
```


vchiley:
Should we also pull out `'bos_token_id'`?

ShashankMosaicML:
> Should we also pull out `'bos_token_id'`?

`ConcatenatedSequenceCollatorWrapper` throws an error if we set both the `eos_token_id ` and `bos_token_id`. That is why I only set `eos_token_id` here. I can add a flag that dictates which of these to automatically pick. But then, we will have to explicitly set one of those.

samhavens:
> @samhavens @codestar12 do you think `eos_token_id` or `bos_token_id` is more important and/or widely used? (probably `eos` but wanted to check)

Llama and t5 models both expect both eos and bos, OPT I think is BOS only?

llmfoundry/data/text_data.py
```diff
@@ -256,6 +259,29 @@ def build_text_dataloader(
...
+            tokenizer,
+            'eos_token_id') and eos_token_id != tokenizer.eos_token_id:
+        if cfg.dataset.pop('override_eos_token_id_mismatch_error', False):
+            log.warning(
+                f'Provided eos_token_id={eos_token_id} does not match tokenizer.eos_token_id={tokenizer.eos_token_id}'
+            )
+        else:
+            raise ValueError(
+                f'Provided eos_token_id={eos_token_id} does not match tokenizer.eos_token_id={tokenizer.eos_token_id}.  To override this error, set the override_eos_token_id_mismatch_error flag to True in the dataset config section of the YAML.'
+            )
```
vchiley, line 272:
```suggestion
        eos_mismatch_str = f'Provided {eos_token_id=} does not match {tokenizer.eos_token_id=}.'
        if cfg.dataset.pop('override_eos_token_id_mismatch_error', False):
            log.warning(eos_mismatch_str)
        else:
            raise ValueError(
                eos_mismatch_str + ' To override this error, set the override_eos_token_id_mismatch_error flag to True in the dataset config section of the YAML.'
            )
```

avoid repeated code

llmfoundry/data/text_data.py
```diff
@@ -256,6 +259,29 @@ def build_text_dataloader(
...
+            tokenizer,
+            'bos_token_id') and bos_token_id != tokenizer.bos_token_id:
+        if cfg.dataset.pop('override_bos_token_id_mismatch_error', False):
+            log.warning(
+                f'Provided bos_token_id={bos_token_id} does not match tokenizer.bos_token_id={tokenizer.bos_token_id}'
+            )
+        else:
+            raise ValueError(
+                f'Provided bos_token_id={bos_token_id} does not match tokenizer.bos_token_id={tokenizer.bos_token_id}. To override this error, set the override_bos_token_id_mismatch_error flag to True in the dataset config section of the YAML.'
+            )
```
vchiley, line 284:
```suggestion
        bos_mismatch_str = f'Provided bos_token_id={bos_token_id} does not match tokenizer.bos_token_id={tokenizer.bos_token_id}.'
        if cfg.dataset.pop('override_bos_token_id_mismatch_error', False):
            log.warning(bos_mismatch_str)
        else:
            raise ValueError(
                bos_mismatch_str + ' To override this error, set the override_bos_token_id_mismatch_error flag to True in the dataset config section of the YAML.'
            )
```

avoid repeated code

vchiley (commented):
Should we also warn/error if tokenizer does have `eos_token_id` or `bos_token_id` but user does not set it?

vchiley (approved):

PR #1043, opened 2024-03-20
LLM Foundry CLI (just registry)

dakinggg:
Adds an LLM Foundry CLI, with two commands for registry discoverability
Screenshot 2024-03-20 at 1 38 26 AM
Screenshot 2024-03-20 at 1 38 47 AM


llmfoundry/cli/registry_cli.py
```diff
@@ -0,0 +1,73 @@
...
+
+import typer
+from rich.console import Console
+from rich.table import Table
+
+console = Console()
+from typing import Optional
+
+from llmfoundry import registry
+from llmfoundry.utils import registry_utils
```
irenedea, line 12: nit
```suggestion
from llmfoundry.utils.registry_utils import TypedRegistry
```

llmfoundry/cli/registry_cli.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+import typer
+from rich.console import Console
+from rich.table import Table
+
+console = Console()
```
irenedea, line 8: nit move after imports

irenedea (approved):

llmfoundry/cli/cli.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+import typer
+
+from llmfoundry.cli import registry_cli
+
+app = typer.Typer(pretty_exceptions_show_locals=False)
```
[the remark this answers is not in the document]
dakinggg: don't really have a strong opinion. there are a few easy to use python cli tools (typer, click, fire). Composer just uses argparse and only has one command, so is a bit different. We can always bail to a different CLI tool if needed, but typer is very popular and I'm familiar with it.

mvpatel2000 (approved):
Approving to unblock

PR #1047, opened 2024-03-21
Deprecate attention patching for llama

dakinggg:
Now that flash attention 2 is integrated into transformers, we don't need to monkeypatch llama.

Also changes our versioned deprecation warning to a userwarning so that it shows up in the logs.

See `llama-patch-triton-after-6-MYjOo9` for a run that sets attention patch type, and changes to flash attention 2 and emits the warning.


irenedea (approved):
thanks!

PR #1050, opened 2024-03-21
Triton RMSNorm

josejg:
This PR adds support for FlashAttention's Triton implementation of [RMSnorm](https://github.com/Dao-AILab/flash-attention/blob/36587c01cb4390de0a590b2131e3fcc4859ba09c/flash_attn/ops/triton/layer_norm.py#L893) (commonly found in LLaMa-like models).

The flag is enabled by setting `parameters.model.norm_type` to  `triton_rmsnorm` .

Under a LLaMa-2-7b-like workload (displayed below) using this implementation produces a lift of ~1400 tok/s/gpu while preserving the loss. The improvement is so large because the current RMSnorm implementation is done in vanilla PyTorch (see [here](https://github.com/mosaicml/llm-foundry/blob/3348b596d2b9375f50bb1fc46be2b3cff20ae849/llmfoundry/models/layers/norm.py#L56)).

image


llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            hidden_size = normalized_shape
+        elif isinstance(normalized_shape, list):
+            hidden_size = math.prod(normalized_shape)
+        elif isinstance(normalized_shape, torch.Size):
+            hidden_size = math.prod(normalized_shape)
+
+        factory_kwargs = {'device': device, 'dtype': dtype}
+        super().__init__()
+        self.eps = eps
+        if dropout_p > 0.0:
```
dakinggg, line 135: the others don't support dropout here, do we need to make the signature different from the others?

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            self.drop = torch.nn.Dropout(dropout_p)
+        else:
+            self.drop = None
+        self.weight = torch.nn.Parameter(
```
dakinggg, line 139: the others have a flag for whether to use weight, should that exist here too?

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            torch.empty(hidden_size, **factory_kwargs))
+        self.register_parameter('bias', None)
+        torch.nn.init.ones_(self.weight)
```
dakinggg, line 142: the init shouldn't be in the constructor, right?

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+        try:
+            from flash_attn.ops.triton.layer_norm import rms_norm_fn
+            self.rms_norm_fn = rms_norm_fn
+        except ImportError:
+            raise ImportError(
+                'triton_rms_norm requires Flash Attention to be installed. ' +
+                'Please `pip install llm-foundry[gpu-flash2]`.')
```
dakinggg, line 150: this should just say to install fa, because the model code can end up isolated from the foundry code

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+    def forward(self,
```
dakinggg, line 152: i don't think forward should take anything other than `x`, same as the other norms?

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        device: Optional[torch.device] = None,
+    ):
+        if isinstance(normalized_shape, int):
+            hidden_size = normalized_shape
+        elif isinstance(normalized_shape, list):
+            hidden_size = math.prod(normalized_shape)
+        elif isinstance(normalized_shape, torch.Size):
+            hidden_size = math.prod(normalized_shape)
+
+        factory_kwargs = {'device': device, 'dtype': dtype}
```
dakinggg, line 132: these are only used ones, just specify them directly in the param constructor?

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+
+    def __init__(
+        self,
+        normalized_shape: Union[int, List[int], torch.Size],
+        eps: float = 1e-5,
+        dropout_p: float = 0.0,
+        dtype: Optional[torch.dtype] = None,
+        device: Optional[torch.device] = None,
+    ):
+        if isinstance(normalized_shape, int):
```
dakinggg, line 125: how come the others don't need this if/else stuff?

llmfoundry/models/layers/norm.py
josejg: made default 0

llmfoundry/models/layers/norm.py
josejg: Not an option in the FA backend

llmfoundry/models/layers/norm.py
josejg: refactored

llmfoundry/models/layers/norm.py
josejg: 👍

llmfoundry/models/layers/norm.py
josejg: agreed, making them defaults

llmfoundry/models/layers/norm.py
josejg: fixed

llmfoundry/models/layers/norm.py
josejg: FA expects flat tensor since they only consider the NLP mode of LayerNorm, whereas torch's is more generic

tests/models/test_model.py
```diff
@@ -599,6 +599,9 @@ def test_lora_id():
 def test_mpt_creation(norm_type: str, no_bias: bool, tie_word_embeddings: bool,
                       expansion_ratio: Union[int, float], ffn_hidden_size: int,
                       ffn_act_fn: dict):
+    if norm_type == 'triton_rms_norm' and not is_flash_v2_installed():
+        pytest.skip(f'norm_type=triton_rms_norm requires flash Attention to be installed')
```
dakinggg, line 603:
```suggestion
        pytest.skip(f'norm_type=triton_rms_norm requires flash attention to be installed')
```

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,55 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        elif isinstance(normalized_shape, list):
+            hidden_size = math.prod(normalized_shape)
+        elif isinstance(normalized_shape, torch.Size):
+            hidden_size = math.prod(normalized_shape)
+
+        self.weight = torch.nn.Parameter(
+            torch.ones(hidden_size, device=device, dtype=dtype))
+        self.register_parameter('bias', None)
+
+        try:
```
dakinggg, line 139: nit: move try/except to the top of the constructor and attr setting out of it, so its just the import that is try/excepted

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,55 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        if isinstance(normalized_shape, int):
+            hidden_size = normalized_shape
+        elif isinstance(normalized_shape, list):
+            hidden_size = math.prod(normalized_shape)
+        elif isinstance(normalized_shape, torch.Size):
+            hidden_size = math.prod(normalized_shape)
+
+        self.weight = torch.nn.Parameter(
+            torch.ones(hidden_size, device=device, dtype=dtype))
+        self.register_parameter('bias', None)
```
dakinggg, line 137: needed?

llmfoundry/models/layers/norm.py
```diff
@@ -111,9 +112,55 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        except ImportError:
+            raise ImportError(
+                'triton_rms_norm requires Flash Attention to be installed. ' +
+                'Please pip install flash-attn.')
+
+    def forward(self, x: torch.Tensor):
+        return self.rms_norm_fn(
+            x,
+            self.weight,
+            self.bias,
```
dakinggg, line 151: where does self.bias come from?

dakinggg (approved):
LGTM, can you add a simple test that checks numerical equivalence to the base rmsnorm impl?

vchiley (approved):

PR #1052, opened 2024-03-22
Metrics registry

dakinggg:
Adds registry for the metrics

Same loss before and after
Screenshot 2024-03-22 at 3 57 08 PM


llmfoundry/metrics/__init__.py
```diff
@@ -1,8 +1,80 @@
...
+    'masked_accuracy',
+]
+
+DEFAULT_ENC_DEC_METRICS = [
+    'language_cross_entropy',
+    'masked_accuracy',
+]
+
+
+def build_metric(name: str, kwargs: Dict[str, Any]) -> Metric:
```
dakinggg, line 60: remove dupe

llmfoundry/metrics/__init__.py
```diff
@@ -1,8 +1,80 @@
...
+    """Builds a metric from the registry."""
+    return builder(name=name,
+                   registry=metrics,
+                   partial_function=True,
+                   pre_validation_function=Metric,
+                   post_validation_function=None,
+                   kwargs=kwargs)
+
 
 __all__ = [
```
dakinggg, line 70: add to all

llmfoundry/utils/builders.py
```diff
@@ -196,6 +198,16 @@ def build_algorithm(name: str, kwargs: Dict[str, Any]) -> Algorithm:
                                    kwargs=kwargs)
 
 
+def build_metric(name: str, kwargs: Dict[str, Any]) -> Metric:
```
irenedea, line 201: can we make this kwargs optional or be **kwargs? Don't like that we need to pass in `{}` everytime

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -123,25 +118,17 @@ def __init__(self, om_model_config: DictConfig,
...
+        use_train_metrics = om_model_config.get('use_train_metrics', True)
         train_metrics = [
-            LanguageCrossEntropy(),
-            LanguagePerplexity(),
-            TokenAccuracy()
-        ]
+            build_metric(metric, {})
+            for metric in DEFAULT_CAUSAL_LM_TRAIN_METRICS +
+            om_model_config.get('additional_train_metrics', [])
+        ] if use_train_metrics else []
```
irenedea, line 126: suggest
```
train_metric_names = DEFAULT_CAUSAL_LM_TRAIN_METRICS + om_model_config.get('additional_train_metrics',[])
train_metrics = [build_metric(metric, {}) for metric in train_metric_names] if use_train_metrics else []
```

and similarly for eval_metrics

irenedea (approved):

llmfoundry/utils/builders.py
dakinggg: I didn't want to do **, cause then you can't have kwargs to your constructor that match other args in the function, but it can be optional, sure

PR #1054, opened 2024-03-23
Add chat schema example for mlflow

dakinggg:
Screenshot 2024-03-22 at 5 53 19 PM
Screenshot 2024-03-22 at 5 53 23 PM


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -129,39 +129,30 @@ def __init__(
...
-                ColSpec('double', 'temperature', optional=True),
-                ColSpec('integer', 'max_tokens', optional=True),
-                ColSpec('string', 'stop', optional=True),
-                ColSpec('integer', 'candidate_count', optional=True)
-            ])
-
-            output_schema = Schema([ColSpec('string', 'predictions')])
-
-            default_signature = ModelSignature(inputs=input_schema,
-                                               outputs=output_schema)
```
irenedea, line 157: why don't we need this anymore?

irenedea (approved):

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -129,39 +129,30 @@ def __init__(
...
+            mlflow_logging_config['metadata'] = passed_metadata
+            mlflow_logging_config.setdefault('task', 'llm/v1/completions')
 
             default_input_example = {
                 'prompt': np.array(['What is Machine Learning?'])
             }
+            is_chat = mlflow_logging_config['task'].endswith(
```
nancyhung, line 142:
```suggestion
            is_chat = mlflow_logging_config['task'].contains(
```

nancyhung (approved):
lgtm

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: actually probably fine to do this

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: they told me that when using `llm/v1/{completions|chat}`, the input schema and signature would be added automatically, so no need to hardcode here, which would be more brittle.

PR #1057, opened 2024-03-24
Models registry

dakinggg:
Round of manual tests:

- [x] MPT pretrain before and after (`registry-condition-mpt-1-YoWX7j` vs `registry-baseline-mpt-1-gTfymO`)
- [x] llama finetune before and after (`registry-condition-llama-1-qxchzh` vs `registry-baseline-llama-1-eYIlHX`)
- [x] multi model eval before and after (`registry-condition-eval-1-dXlla3` vs `registry-baseline-eval-1-9Csev8`)

PR #1060, opened 2024-03-25
Remove under construction from registry

dakinggg:
Merge after #1057


mvpatel2000 (approved):
Approving to unblock

README.md
```diff
@@ -255,9 +255,96 @@ export HUGGING_FACE_HUB_TOKEN=your-auth-token
 
 and uncomment the line containing `--hf_repo_for_upload ...` in the above call to `inference/convert_composer_to_hf.py`.
 
-# :construction: UNDER CONSTRUCTION: Registry
+# Registry
+
+Some components of LLM Foundry are registrable. This means that you can register options for these components, and then use them in your yaml config, without forking the library.
```
irenedea, line 260: how about something like:

```suggestion
You can use the registry to customize your workflows without forking the library. Some components of LLM Foundry are registrable, such as models, loggers, and callbacks. This means that you can register new options for these components, and then use them in your yaml config.
```

Wanted to mention examples of components (loggers, models, callbacks) and explain that its for customization.

irenedea (approved):

PR #1071, opened 2024-03-27
Expose more exception attributes

jjanezhang:
## Expose more exception attributes

- Exposing more exception attributes so I can use them to construct better messages for FT runs.


irenedea (approved):
LGTM

PR #1075, opened 2024-03-29
Add expandeable segments flag

dakinggg:


mvpatel2000 (commented):
Why should env vars be in yamls its a bit weird imo 🤷

scripts/train/train.py
```diff
@@ -130,11 +130,19 @@ def main(cfg: DictConfig) -> Trainer:
...
     max_split_size_mb: Optional[int] = cfg.pop('max_split_size_mb', None)
     if max_split_size_mb is not None:
-        os.environ[
-            'PYTORCH_CUDA_ALLOC_CONF'] = f'max_split_size_mb:{max_split_size_mb}'
+        cuda_alloc_conf.append(f'max_split_size_mb:{max_split_size_mb}')
+
+    # Expandeable segments
+    expandeable_segments: bool = cfg.pop('expandeable_segments', False)
+    if expandeable_segments:
+        cuda_alloc_conf.append('expandeable_segments:True')
```
mvpatel2000, line 142:
```suggestion
    if cfg.pop('expandeable_segments', False):
        cuda_alloc_conf.append('expandeable_segments:True')
```

scripts/train/train.py
```diff
@@ -130,11 +130,19 @@ def main(cfg: DictConfig) -> Trainer:
     # Create copy of config for logging
     logged_cfg: DictConfig = copy.deepcopy(cfg)
 
+    cuda_alloc_conf = []
     # Get max split size mb
     max_split_size_mb: Optional[int] = cfg.pop('max_split_size_mb', None)
     if max_split_size_mb is not None:
-        os.environ[
-            'PYTORCH_CUDA_ALLOC_CONF'] = f'max_split_size_mb:{max_split_size_mb}'
+        cuda_alloc_conf.append(f'max_split_size_mb:{max_split_size_mb}')
```
mvpatel2000, line 137: Should we deprecate?

mvpatel2000 (approved):

scripts/train/train.py
dakinggg: I don't see any indication in torch this is deprecated, do you?

PR #1078, opened 2024-03-30
Background mlflow model registration

irenedea:
Spawn a new process that handles the mlflow model registration (which uploads the model to UC). This way, global rank 0 isn't stuck on with a long operation with variable speeds, which could lead to nccl timeouts. This also allows other operations to happen concurrently such as composer checkpoint saving, etc.

At the end of training, we wait for all child processes to finish and clean up temporary directories. 

### Manual Test
`test-hf-mlflow-background-register-ib6LgH`
`test-hf-mlflow-background-register-llama70b-Lhda8z `

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -76,19 +76,21 @@ def _maybe_get_license_filename(
 
 
 def _register_model_with_run_id_multiprocess(mlflow_logger: MLFlowLogger,
-                                             logging_level: int, model_uri: str,
-                                             name: str,
+                                             composer_logging_level: int,
+                                             model_uri: str, name: str,
                                              await_creation_for: int):
     """Function for calling MLFlowLogger.register_model_with_run_id from a.
 
     spawned child process.
     """
     # Setup logging for child process. This ensures that any logs from composer are surfaced.
-    logging.basicConfig(
-        format=
-        f'%(asctime)s: rank{dist.get_global_rank()}[%(process)d][%(threadName)s]: %(levelname)s: %(name)s: %(message)s'
-    )
-    logging.getLogger('composer').setLevel(logging_level)
+    if composer_logging_level > 0:
+        # If logging_level is 0, then the composer logger was unset.
+        logging.basicConfig(
+            format=
+            f'%(asctime)s: rank{dist.get_global_rank()}[%(process)d][%(threadName)s]: %(levelname)s: %(name)s: %(message)s'
+        )
+        logging.getLogger('composer').setLevel(composer_logging_level)
 
     # Register model.
     mlflow_logger.register_model_with_run_id(
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -228,7 +228,7 @@ class HuggingFaceCheckpointer(Callback):
         elif event == Event.FIT_END:
             # Wait for all child processes spawned by the callback to finish.
             while not self._all_child_processes_done():
-                time.sleep(30)
+                time.sleep(2)
 
     def _is_last_batch(self, state: State):
         elapsed_duration = state.get_elapsed_duration()
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -82,9 +82,9 @@ def _register_model_with_run_id_multiprocess(
     name: str,
     await_creation_for: int,
 ):
-    """Function for calling MLFlowLogger.register_model_with_run_id from a.
+    """Call MLFlowLogger.register_model_with_run_id.
 
-    spawned child process.
+    Used mainly to register from a child process.
     """
     # Setup logging for child process. This ensures that any logs from composer are surfaced.
     if composer_logging_level > 0:
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -78,11 +78,18 @@ def _register_model_with_run_id_multiprocess(mlflow_logger: MLFlowLogger,
                                              logging_level: int, model_uri: str,
                                              name: str,
                                              await_creation_for: int):
+    """Function for calling MLFlowLogger.register_model_with_run_id from a.
+
+    spawned child process.
+    """
+    # Setup logging for child process. This ensures that any logs from composer are surfaced.
     logging.basicConfig(
         format=
         f'%(asctime)s: rank{dist.get_global_rank()}[%(process)d][%(threadName)s]: %(levelname)s: %(name)s: %(message)s'
     )
     logging.getLogger('composer').setLevel(logging_level)
+
+    # Register model.
     mlflow_logger.register_model_with_run_id(
         model_uri=model_uri, name=name, await_creation_for=await_creation_for)
 
@@ -411,7 +418,7 @@ class HuggingFaceCheckpointer(Callback):
                                 os.path.join(local_save_path, license_filename),
                             )
 
-                        # Register the model to mlflow in a child process.
+                        # Spawn a new process to register the model.
                         process = SpawnProcess(
                             target=_register_model_with_run_id_multiprocess,
                             kwargs={
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -254,8 +254,6 @@ class HuggingFaceCheckpointer(Callback):
         return False
 
     def _all_child_processes_done(self) -> bool:
-        if len(self.child_processes) == 0:
-            return True
         not_done = any(process.is_alive() for process in self.child_processes)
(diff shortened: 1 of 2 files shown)


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -235,158 +276,175 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
-                composer_model = state.model.module
-                original_model: PreTrainedModel = state.model.module.model
-                state_dict_model = state.model.module.model
-                original_tokenizer = state.model.module.tokenizer
-            elif isinstance(state.model.model, FSDP):
-                composer_model = state.model
-                original_model: PreTrainedModel = state.model.model.module
-                state_dict_model = state.model.model
-                original_tokenizer = state.model.tokenizer
+
```
irenedea, line 279: a lot of these code "changes" below are because:
* we now create the temp directory not using a context, so the indentation has changed. This is because we need to keep these files alive for `_register_model_with_run_id_multiprocess`, so now we manually clean up the temp directory at the end of the run along with waiting for the processes to finish. 
* I've added an extra `dist.barrier()` between HF checkpoint upload and mlflow registry logic.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -218,6 +253,12 @@ def _is_last_batch(self, state: State):
 
         return False
 
+    def _all_child_processes_done(self) -> bool:
```
dakinggg, line 256: we should have a timeout on this func in case something goes wrong. rn this upload can't more than hour anyway, so maybe set it to that?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -72,6 +75,26 @@ def _maybe_get_license_filename(
         return None
 
 
+def _register_model_with_run_id_multiprocess(mlflow_logger: MLFlowLogger,
+                                             logging_level: int, model_uri: str,
+                                             name: str,
+                                             await_creation_for: int):
+    """Function for calling MLFlowLogger.register_model_with_run_id from a.
```
dakinggg, line 82:
```suggestion
    """Call MLFlowLogger.register_model_with_run_id
```

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -72,6 +75,26 @@ def _maybe_get_license_filename(
...
+
+    spawned child process.
```
dakinggg, line 84:
```suggestion
    Used mainly to register from a child process.
```

llmfoundry/callbacks/hf_checkpointer.py
irenedea: sounds good, i'll add to the `elif event == Event.FIT_END:` block

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -235,158 +276,175 @@ def _save_checkpoint(self, state: State, logger: Logger):
...
+                            'name':
+                                self.mlflow_registered_model_name,
+                            'await_creation_for':
+                                3600,
+                        })
+                    process.start()
+                    self.child_processes.append(process)
+
+                    # Save the temporary directory to be cleaned up later.
+                    if use_temp_dir:
```
dakinggg, line 444: does this need to be done down here for some reason? Or can we make more clear by doing up at the top where `use_temp_dir` is set

irenedea: It's done here because we only need to save the temp dir for later clean up at when we register to mlflow. I can do it at the top of the if statement though, but I thought it was clearer here because it happens after we spawn a process.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -72,6 +75,26 @@ def _maybe_get_license_filename(
...
+def _register_model_with_run_id_multiprocess(mlflow_logger: MLFlowLogger,
+                                             logging_level: int, model_uri: str,
+                                             name: str,
+                                             await_creation_for: int):
+    """Function for calling MLFlowLogger.register_model_with_run_id from a.
+
+    spawned child process.
+    """
+    # Setup logging for child process. This ensures that any logs from composer are surfaced.
+    logging.basicConfig(
```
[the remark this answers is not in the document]
irenedea: hmm... good point. There's no super nice way to check if logging has been set up previously.
Can condition on something like https://stackoverflow.com/questions/53129716/how-to-check-if-a-logger-exists . Wdyt?

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: ok either way

llmfoundry/callbacks/hf_checkpointer.py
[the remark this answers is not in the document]
dakinggg: do we actually need composer logging in the subprocess? like the logs from mlflow all still show up? or no?

llmfoundry/callbacks/hf_checkpointer.py
irenedea: I'll keep as is then. Will do a refactor later to make the logic more clear.

llmfoundry/callbacks/hf_checkpointer.py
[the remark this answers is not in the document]
irenedea: with logging:
```
2024-04-02 15:42:41,021: rank0[8477][MainThread]: INFO: composer.loggers.mlflow_logger: Registered model mpt125test already exists. Creating a new version of this model...
Uploading artifacts:   0%|                                                                                                                                                                     | 0/14 [00:00<?, ?it/s]2024/04/02 15:42:48 INFO mlflow.store.artifact.cloud_artifact_repo: The progress bar can be disabled by setting the environment variable MLFLOW_ENABLE_ARTIFACTS_PROGRESS_BAR to false
2024-04-02 15:42:49,204: rank0[8477][ThreadPoolExecutor-11_0]: WARNING: urllib3.connectionpool: Connection pool is full, discarding connection: s3.amazonaws.com. Connection pool size: 10
Uploading artifacts:  43%|███████████████████████████████████████████████████████████████████▎                                                                                         | 6/14 [00:00<00:04,  1.89it/s]2024-04-02 15:42:49,263: rank0[8477][ThreadPoolExecutor-26_0]: WARNING: urllib3.connectionpool: Connection pool is full, discarding connection: s3.amazonaws.com. Connection pool size: 10
Uploading artifacts:  64%|████████████████████████████████████████████████████████████████████████████████████████████████████▉                                                        | 9/14 [00:00<00:02,  1.89it/s]2024-04-02 15:42:49,353: rank0[8477][ThreadPoolExecutor-35_0]: WARNING: urllib3.connectionpool: Connection pool is full, discarding connection: s3.amazonaws.com. Connection pool size: 10
2024-04-02 15:42:49,709: rank0[8477][ThreadPoolExecutor-38_0]: WARNING: urllib3.connectionpool: Connection pool is full, discarding connection: s3.amazonaws.com. Connection pool size: 10
Uploading artifacts: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 14/14 [00:01<00:00,  7.45it/s]
2024-04-02 15:42:52,941: rank0[8477][MainThread]: INFO: composer.loggers.mlflow_logger: Successfully created model version 2 for model main.irenedea.mpt125test
```
without logging:
```
Uploading artifacts:   0%|                                                                                                                                                                     | 0/14 [00:00<?, ?it/s]2024/04/02 15:45:28 INFO mlflow.store.artifact.cloud_artifact_repo: The progress bar can be disabled by setting the environment variable MLFLOW_ENABLE_ARTIFACTS_PROGRESS_BAR to false
Uploading artifacts: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 14/14 [00:02<00:00,  6.99it/s]
```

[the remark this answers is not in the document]
irenedea: Okay, I think if the logging level is 0, then the logger was unset, so we can avoid setting up logging in that case. Pushed the code, lmk if you think this is good

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -202,6 +234,21 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
...
                     '5GB')
+        elif event == Event.FIT_END:
+            # Wait for all child processes spawned by the callback to finish.
+            timeout = 3600
+            wait_start = time.time()
+            while not self._all_child_processes_done():
+                wait_time = time.time() - wait_start
+                if wait_time > timeout:
+                    raise TimeoutError(
+                        f'Waited {wait_time} seconds for child processes to complete. Exceed timeout of {timeout} seconds.'
```
dakinggg, line 245:
```suggestion
                        f'Waited {wait_time} seconds for child processes to complete. Exceeded timeout of {timeout} seconds.'
```

dakinggg (approved):
LGTM!

PR #1079, opened 2024-04-01
Doing the loss reduction in foundry instead of in the loss functions.

ShashankMosaicML:
This PR gives us more flexibility to reduce losses in foundry in custom ways.

We see that the changes do not affect the MFU or the convergence for 125M and 7B models.

Screenshot 2024-04-01 at 4 58 49 PM

Screenshot 2024-04-02 at 9 17 43 AM

Memory consumption is also similar:

Screenshot 2024-04-02 at 9 21 09 AM
Screenshot 2024-04-02 at 9 21 38 AM

### DIFF
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -1017,9 +1017,9 @@ class ComposerMPTCausalLM(HuggingFaceModel):
 
     def loss(self, outputs: CausalLMOutputWithPast,
              batch: Mapping) -> torch.Tensor:
-        targets = self.get_targets(batch).view(-1)
+        targets = self.get_targets(batch)
         losses = self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
-                              targets)
+                              targets.view(-1))
 
         if torch.all(targets == -100):
             raise ValueError(
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -1021,7 +1021,7 @@ class ComposerMPTCausalLM(HuggingFaceModel):
         losses = self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
                               targets.view(-1))
 
-        if torch.all(targets == -100):
+        if torch.all(targets == self.loss_fn.ignore_index):
             loss = losses.sum()
         else:
             loss = losses.sum() / (targets != self.loss_fn.ignore_index).sum()
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -984,7 +984,8 @@ class ComposerMPTCausalLM(HuggingFaceModel):
                 from flash_attn.losses.cross_entropy import \
                     CrossEntropyLoss as FusedCrossEntropyLoss
 
-                self.loss_fn = FusedCrossEntropyLoss(ignore_index=-100)
+                self.loss_fn = FusedCrossEntropyLoss(ignore_index=-100,
+                                                     reduction='none')
             except:
                 raise ValueError(
                     'Fused Cross Entropy is not installed. Either (1) have a CUDA-compatible GPU '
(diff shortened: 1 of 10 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1015,9 +1017,18 @@ def forward(self, batch: MutableMapping) -> CausalLMOutputWithPast:
...
-        return self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
-                            targets.view(-1))
+        targets = self.get_targets(batch).view(-1)
+        losses = self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
+                              targets)
+
+        if torch.all(targets == -100):
+            raise ValueError(
+                'All targets are -100. This might be due to all tokens in a sequence being padding tokens.'
+            )
```
vchiley, line 1027:
```suggestion
            loss = losses.sum()
```

Should we return 0 in this case?
maybe log a warning, but Error doesn't make sense.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1015,9 +1017,18 @@ def forward(self, batch: MutableMapping) -> CausalLMOutputWithPast:
 
     def loss(self, outputs: CausalLMOutputWithPast,
              batch: Mapping) -> torch.Tensor:
-        targets = self.get_targets(batch)
-        return self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
-                            targets.view(-1))
+        targets = self.get_targets(batch).view(-1)
```
vchiley, line 1020: The only place targets is used is in `(targets != self.loss_fn.ignore_index).sum()`.

Do we need to view it as -1 for that?

(I'm a bit scared of destroying batch dim unless its NEEDED)

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: This was done to preserve functionality. The code before this change would result in a NaN being returned as loss, due to dividing by 0. A better way to capture that would be raise an error. That is why I raise this error.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: There is only one other place the targets is being used in this function, and that also uses `.view(1)`

Screenshot 2024-04-01 at 5 12 21 PM

dakinggg: @vchiley CE loss expects the flattened targets, right?

vchiley: Just trying to future proof the code.
I'd argue that we should pass it into the loss fn as `targets.view(1)` but don't destroy the batch info unless needed.
(what if in the future someone wants the batch info for something else)

vchiley: i guess ppl can undo this change in the future, I'm just not sure what the point of including it now is 🤷‍♂️

ShashankMosaicML: Makes sense, let me update the PR.

vchiley (approved):

vchiley (approved):

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -984,7 +984,8 @@ def __init__(
                 from flash_attn.losses.cross_entropy import \
                     CrossEntropyLoss as FusedCrossEntropyLoss
 
-                self.loss_fn = FusedCrossEntropyLoss(ignore_index=-100)
+                self.loss_fn = FusedCrossEntropyLoss(ignore_index=-100,
```
vchiley, line 987: Outside the scope of this pr but should we make expose `ignore_index` in the config (default `-100`)?
cc @dakinggg

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1016,8 +1018,15 @@ def forward(self, batch: MutableMapping) -> CausalLMOutputWithPast:
...
         targets = self.get_targets(batch)
-        return self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
-                            targets.view(-1))
+        losses = self.loss_fn(outputs.logits.view(-1, outputs.logits.size(-1)),
+                              targets.view(-1))
+
+        if torch.all(targets == self.loss_fn.ignore_index):
+            loss = losses.sum()
+        else:
+            loss = losses.sum() / (targets != self.loss_fn.ignore_index).sum()
```
vchiley, line 1027: Outside the scope of this PR, but should we enable expose loss fn `reduction` in the config (default `'mean'`)?

llmfoundry/models/mpt/modeling_mpt.py
dakinggg: I don't think so. Everyone uses -100, not sure why anyone would want to change this

vchiley: resolving cuz outside the scope of this PR

llmfoundry/models/mpt/modeling_mpt.py
vchiley: resolving cuz outside the scope of this PR

PR #1080, opened 2024-04-02
Norms registry

dakinggg:
Starting the layers part of the registry with just the norms so we can review the overall setup of the layers registry in a smaller PR.


irenedea (approved):
lgtm!

llmfoundry/layers_registry.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2024 MosaicML LLM Foundry authors
```
irenedea, line 1: just curious why is this a top level file in `llmfoundry`?

dakinggg: resolved offline, may revisit in the future

mvpatel2000 (approved):

PR #1083, opened 2024-04-03
Fixes for streaming and auto packing

dakinggg:
Failed multi stream run before: `multi-stream-7-qGYQwc`
Succeeded multi stream run after: `multi-stream-16-bsEVkl`

Failed single stream before: `multi-stream-shashank-1-Cppzyo`
Succeeded multi stream run after: `multi-stream-shashank-10-ELpDau`

### DIFF
--- a/llmfoundry/data/packing.py
+++ b/llmfoundry/data/packing.py
@@ -389,17 +389,18 @@ def profile_packing(
     dataloader_cfg.persistent_workers = False
 
     # If streaming dataset, use a temporary local folder for profiling
+    local_rank_zero = dist.get_global_rank() - dist.get_local_rank()
     if dataloader_cfg.dataset.get('remote') is not None:
         tmp_path_to_broadcast = tempfile.TemporaryDirectory().name
         gathered_paths = dist.all_gather_object(tmp_path_to_broadcast)
-        tmp_path = gathered_paths[0]
+        tmp_path = gathered_paths[local_rank_zero]
         dataloader_cfg.dataset.local = tmp_path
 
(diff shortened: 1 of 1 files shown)


llmfoundry/data/packing.py
```diff
@@ -389,7 +390,17 @@ def profile_packing(
...
+        gathered_paths = dist.all_gather_object(tmp_path_to_broadcast)
+        tmp_path = gathered_paths[0]
+        dataloader_cfg.dataset.local = tmp_path
+
+    if dataloader_cfg.dataset.get('streams') is not None:
+        for stream_config in dataloader_cfg.dataset.streams.values():
+            tmp_path_to_broadcast = tempfile.TemporaryDirectory().name
+            gathered_paths = dist.all_gather_object(tmp_path_to_broadcast)
+            tmp_path = gathered_paths[0]
+            stream_config.local = tmp_path
```
mvpatel2000, line 403: This will take global rank 0 path and fail on multi-node. You want global rank minus local rank indexed into gathered_paths

dakinggg: wait actually this works (although not the way I intended) because streaming does not require that your local path already exist.

dakinggg: multi node run that works with the current code: `multi-stream-18-Ah40Ce`.

mvpatel2000: @knighton is so good he predicts user errors before they happen and prevents it

We should still use local path

dakinggg: fixed, and test run: `multi-stream-19-EM0XH8`

PR #1095, opened 2024-04-04
FFN layer registry

dakinggg:
Before and after
Screenshot 2024-04-04 at 7 14 24 PM
Screenshot 2024-04-04 at 7 15 23 PM


llmfoundry/registry.py
```diff
@@ -119,4 +120,5 @@
     'models',
     'metrics',
     'dataloaders',
+    'norms',
```
dakinggg, line 123: add

tests/test_registry.py
```diff
@@ -30,6 +30,7 @@ def test_expected_registries_exist():
         'dataloaders',
         'metrics',
         'models',
+        'norms',
```
dakinggg, line 33: add

mvpatel2000 (commented):
will review after tests pass

dakinggg:
And with the MB merge:

Screenshot 2024-04-11 at 8 24 30 PM
Screenshot 2024-04-11 at 8 24 40 PM

@irenedea @mvpatel2000 this one is ready

dakinggg:
jk, also failed a test

dakinggg:
ok @mvpatel2000 tests passing now

mvpatel2000 (commented):
why do the megablocks ones have to be separate?

@vchiley to review the mb build stuff

dakinggg:
@mvpatel2000 because there are a bunch of places in the code that we do something only for the mb layers (e.g. the mb args as an attr on the main class.

llmfoundry/models/layers/ffn.py
```diff
@@ -200,131 +225,225 @@ def build_ffn(
...
+                                                ffn_hidden_size),
+        moe_num_experts=moe_num_experts,
+        moe_top_k=moe_top_k,
+        mlp_type=mlp_type,
+        bias=bias,
+        moe_jitter_eps=moe_jitter_eps,
+        activation_fn=resolve_ffn_act_fn(ffn_act_fn),
+        moe_normalize_expert_weights=moe_normalize_expert_weights,
+        uniform_expert_assignment=uniform_expert_assignment,
+        device=device,  # pyright: ignore[reportGeneralTypeIssues]
```
mvpatel2000, line 277: Why ignore? I think we can remove this 🤔

mvpatel2000: hm nvm I think @bigning has a PR for this

mvpatel2000 (approved):

llmfoundry/models/layers/ffn.py
dakinggg: Ah ok, I'll try. Was just a copy paste from existing code.

dakinggg: ok fixed

PR #1103, opened 2024-04-09
Updating the streaming version in setup.py

ShashankMosaicML:


vchiley:
Should we adde `allow_unsafe_types=False, replication=None` args [here](https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/data/text_data.py#L153)?

ShashankMosaicML:
> allow_unsafe_types=False, replication=None

Done. PTAL.

snarayan21 (commented):
Can you add the `allow_unsafe_types` and `replication` args as arguments to the StreamingTextDataset and StreamingFinetuningDataset init functions as well? What's the rationale for passing in fixed values? If I want to set `replication` from my yaml, for example, this isn't possible right

llmfoundry/data/text_data.py
```diff
@@ -151,6 +151,8 @@ def __init__(self,
             sampling_method=sampling_method,
             sampling_granularity=sampling_granularity,
             batching_method=batching_method,
+            allow_unsafe_types=False,
+            replication=None,
```
vchiley, line 155: I'd error on the side of passing them in as args at line 111 then setting the default there.

ShashankMosaicML: Done. PTAL.

llmfoundry/data/finetuning/tasks.py
vchiley, line 1: we have made a habit of copy pasta-ing args doc str from streaming into the datasets here (since they inherit from streaming datasets)
please copy pasta the doc str for the new args

llmfoundry/data/text_data.py
vchiley, line 1: we have made a habit of copy pasta-ing args doc str from streaming into the datasets here (since they inherit from streaming datasets)
please copy pasta the doc str for the new args

vchiley (changes requested):
sorry nit

llmfoundry/data/finetuning/tasks.py
ShashankMosaicML: Done. Ptal.

llmfoundry/data/text_data.py
ShashankMosaicML: Done. Ptal.

vchiley (approved):

karan6181 (approved):
LGTM. Thank You!

PR #1106, opened 2024-04-10
Add remote code option to allow execution of DBRX tokenizer

b-chu:
Without trust_remote_code, FT API cannot load the local DBRX tokenizer during convert_text_to_mds


irenedea (approved):
LGTM! thank you

PR #1107, opened 2024-04-10
Fix overwriting FP8 act ckpt flag in the train script

cli99:
The current activation checkpointing with FP8 TE requires using the non-reentrant implementation from TE. The  fsdp_config for activation checkpointing with TE FP8 is
```
  fsdp_config:
    activation_checkpointing: true
    activation_checkpointing_reentrant: false
    activation_cpu_offload: false
    te_checkpoint_wrapper: true
```
The activation_checkpointing_reentrant param in yaml get overwritten in the [train.py](https://github.com/mosaicml/llm-foundry-private/blob/main/scripts/train/train.py#L78-L90) script due to legacy reasons. Need to rewrite the check.


mvpatel2000 (approved):
Thanks! Sorry for the bad merge

PR #1108, opened 2024-04-12
Dbrx finetune yaml requires save folder specified to enable autoresume

mvpatel2000:
Dbrx finetune yaml requires save folder specified to enable autoresume. As title. Otherwise yaml errors out of box


dakinggg (approved):
Can you get the lora one too?

PR #1115, opened 2024-04-15
Add FreebaseQA to tasks and gauntlet

moeiniamir:
Added [FreebaseQA](https://aclanthology.org/N19-1028/) which is used in several recent papers [1](https://link.springer.com/chapter/10.1007/978-3-031-44216-2_28) [2](https://arxiv.org/abs/2403.09712) [3](https://openreview.net/forum?id=B9klVS7Ddk). Tested on [Vicuna-7B](https://huggingface.co/lmsys/vicuna-7b-v1.3) and obtained similar results to [3](https://openreview.net/forum?id=B9klVS7Ddk). Feel free to cherry-pick if we only need the JSON file itself.


bmosaicml:
Hi @moeiniamir this dataset is very cool. Can you explain the MARLIN section a bit more?

moeiniamir:
> Hi @moeiniamir this dataset is very cool. Can you explain the MARLIN section a bit more?

I mistakenly submitted the pull request on my main branch. I just monkey-patched [Marlin](https://github.com/IST-DASLab/marlin) for internal use since it requires replacing the linear layers of the model, and I found no clean way to include that in the framework. Should I make a separate branch until the commit before that, or is it doable on your side?

PR #1119, opened 2024-04-18
Mlflow datasets

KuuCi:
Add dataset tracking in MLFlow

https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/974778159721961?o=3360802220363900&searchFilter=&orderByKey=attributes.start_time&orderByAsc=false&startTime=ALL&lifecycleFilter=Active&modelVersionFilter=All+Runs&datasetsFilter=W10%3D


llmfoundry/utils/config_utils.py
```diff
@@ -7,7 +7,8 @@
 import warnings
 from typing import Any, Dict, Literal, Mapping, Optional, Tuple, Union
 
-from composer.utils import dist
+import mlflow
```
dakinggg, line 10: This should be conditionally imported, as mlflow is an optional dependency

llmfoundry/utils/config_utils.py
```diff
@@ -184,3 +185,85 @@ def log_config(cfg: DictConfig) -> None:
...
+                                 {}).get('dataset', {}).get('local', None)
+            split = cfg.get(f'{data_split}_loader',
+                            {}).get('dataset', {}).get('split', None)
+            data_paths.add(('local', local_path, data_split))
+
+    return data_paths
+
+
+def log_dataset_uri(cfg: DictConfig) -> mlflow.data.meta_dataset.MetaDataset:
+    """Extracts dataset information from the provided configuration."""
```
dakinggg, line 238:
```suggestion
    """Logs dataset tracking information to MLflow."""
```

llmfoundry/utils/config_utils.py
```diff
@@ -184,3 +185,85 @@ def log_config(cfg: DictConfig) -> None:
...
+    for dataset_type, path, split in data_paths:
+        source_class = dataset_source_mapping.get(dataset_type)
+
+        if source_class:
+            if dataset_type == 'delta_table':
+                source = source_class(delta_table_name=path)
+            elif dataset_type == 'hf' or dataset_type == 'uc_volume':
+                source = source_class(path=path)
+            else:
+                source = source_class(url=path)
```
dakinggg, line 261: how come this is `url=` but the fallback case is `uri=`?

llmfoundry/utils/config_utils.py
```diff
@@ -184,3 +185,85 @@ def log_config(cfg: DictConfig) -> None:
...
+        else:
+            log.info(
+                f'{dataset_type} unknown, defaulting to http dataset source')
+            source = mlflow.data.http_dataset_source.HTTPDatasetSource(uri=path)
+
+        mlflow.log_input(
+            mlflow.data.meta_dataset.MetaDataset(
+                source, name=f'{dataset_type} | {split}'))
```
dakinggg, line 269: I think the name should just be the split? Not a super strong opinion though. Like, the type is shown in the UI by MLflow.

llmfoundry/utils/config_utils.py
```diff
@@ -184,3 +185,85 @@ def log_config(cfg: DictConfig) -> None:
...
+def parse_source_dataset(cfg: DictConfig):
+    """Parse a run config for dataset information."""
+    data_paths = set()
+
+    for data_split in ['train', 'eval']:
+        split = cfg.get(f'{data_split}_loader', {}).get('dataset',
+                                                        {}).get('split', None)
+        source_dataset_path = cfg.get(f'source_dataset_{data_split}', {})
+
+        # check for Delta table
```
dakinggg, line 200:
```suggestion
        # Check for Delta table
```

and throughout

llmfoundry/utils/config_utils.py
```diff
@@ -184,3 +185,85 @@ def log_config(cfg: DictConfig) -> None:
...
+        elif source_dataset_path and source_dataset_path.startswith('/Volumes'):
+            data_paths.add(('uc_volume', source_dataset_path, data_split))
+        # check for HF path
+        elif cfg.get(f'{data_split}_loader', {}).get('dataset',
+                                                     {}).get('hf_name'):
+            hf_path = cfg.get(f'{data_split}_loader', {}).get('dataset',
+                                                              {}).get('hf_name')
+            backend, _, _ = parse_uri(hf_path)
+            if backend:
+                hf_path = f'{hf_path.rstrip("/")}/{split}/' if split else hf_path
```
dakinggg, line 213: This one shouldn't end in a `/` since it will have _some_ file extension (although we don't know what extension here, and that seems fine for now).

dakinggg (commented):
Looking good!

llmfoundry/utils/config_utils.py
KuuCi: whoops

llmfoundry/utils/config_utils.py
KuuCi: I'm down

llmfoundry/utils/config_utils.py
```diff
@@ -184,3 +185,85 @@ def log_config(cfg: DictConfig) -> None:
...
+    for data_split in ['train', 'eval']:
+        split = cfg.get(f'{data_split}_loader', {}).get('dataset',
+                                                        {}).get('split', None)
+        source_dataset_path = cfg.get(f'source_dataset_{data_split}', {})
+
+        # check for Delta table
+        if source_dataset_path and len(source_dataset_path.split('.')) >= 3:
+            data_paths.add(('delta_table', source_dataset_path, data_split))
+        # check for UC volume
+        elif source_dataset_path and source_dataset_path.startswith('/Volumes'):
```
dakinggg, line 204: shouldn't the UC check be for dbfs:/? clearly what you did worked, so what am i missing?

llmfoundry/utils/config_utils.py
```diff
@@ -184,3 +185,85 @@ def log_config(cfg: DictConfig) -> None:
             raise e
         if mlflow.active_run():
             mlflow.log_params(params=om.to_container(cfg, resolve=True))
+            log_dataset_uri(cfg)
+
+
+def parse_source_dataset(cfg: DictConfig):
```
dakinggg, line 191: Also please add unit tests for this function

llmfoundry/utils/config_utils.py
KuuCi: https://github.com/mlflow/mlflow/blob/e39dc8ca3438624a6ff1ea46bd9adf83e8355002/mlflow/data/uc_volume_dataset_source.py#L10

They just set it up that way in mlflow

dakinggg: but isn't this checking against what we pass in? And we pass in a path starting with dbfs:/

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,92 @@ def log_config(cfg: DictConfig) -> None:
...
+    """Parse a run config for dataset information."""
+    data_paths = set()
+
+    for data_split in ['train', 'eval']:
+        split = cfg.get(f'{data_split}_loader', {}).get('dataset',
```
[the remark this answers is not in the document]
irenedea: seems like there's some repeated chained `get`s throughout this fn like `cfg.get(f'source_dataset_{data_split}', {})` and `cfg.get(f'source_dataset_{data_split}', {}).get('dataset...`

Can you use some shared variables here to make the code more concise and readable?

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +185,92 @@ def log_config(cfg: DictConfig) -> None:
...
+        source_dataset_path = cfg.get(f'source_dataset_{data_split}', {})
+
+        # Check for Delta table
+        if source_dataset_path and len(source_dataset_path.split('.')) >= 3:
+            data_paths.add(('delta_table', source_dataset_path, data_split))
+        # Check for UC volume
+        elif source_dataset_path and source_dataset_path.startswith('/Volumes'):
+            data_paths.add(('uc_volume', source_dataset_path, data_split))
+        # Check for HF path
+        elif cfg.get(f'{data_split}_loader', {}).get('dataset',
```
irenedea, line 209: The better way to do this check would be to check `'hf_name' in dataset` rather than `dataset.get('hf_name')`. Using non boolean values as conditionals can sometimes be error prone; if there's a way to have a pure boolean check that's preferred.

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +185,92 @@ def log_config(cfg: DictConfig) -> None:
...
+                                                     {}).get('hf_name'):
+            hf_path = cfg.get(f'{data_split}_loader', {}).get('dataset',
+                                                              {}).get('hf_name')
+            backend, _, _ = parse_uri(hf_path)
+            if backend:
+                hf_path = f'{hf_path.rstrip("/")}/{split}' if split else hf_path
+                data_paths.add((backend, hf_path, data_split))
+            else:
+                data_paths.add(('hf', hf_path, data_split))
```
irenedea, line 218: we can also pass a local path to hf_name like
```
train_dataloader:
    dataset:
        hf_name: /my/local/path
```

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +185,92 @@ def log_config(cfg: DictConfig) -> None:
...
+        elif cfg.get(f'{data_split}_loader', {}).get('dataset',
+                                                     {}).get('remote', None):
+            remote_path = cfg.get(f'{data_split}_loader',
+                                  {}).get('dataset', {}).get('remote', None)
+            backend, _, _ = parse_uri(remote_path)
+            remote_path = f'{remote_path.rstrip("/")}/{split}/' if split else remote_path
+            data_paths.add((backend, remote_path, data_split))
+        # check for local path
+        elif cfg.get(f'{data_split}_loader', {}).get('dataset',
+                                                     {}).get('local', None):
```
irenedea, line 229: For streaming datasets, the dataset path is passed via `remote`, which can also be a local path. `local` is like a local path to save the dataset if you don't want to automatically save to /tmp

irenedea: We might need to do a similar metadata thing that you did with the source_datasets for streaming datasets. i'll dm you an example.

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +185,92 @@ def log_config(cfg: DictConfig) -> None:
...
+            local_path = cfg.get(f'{data_split}_loader',
+                                 {}).get('dataset', {}).get('local', None)
+            split = cfg.get(f'{data_split}_loader',
+                            {}).get('dataset', {}).get('split', None)
+            data_paths.add(('local', local_path, data_split))
+
+    return data_paths
+
+
+def log_dataset_uri(cfg: DictConfig) -> Optional[mlflow.data.meta_dataset.MetaDataset]:
```
irenedea: prefix with `_` because it's a private helper

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +185,92 @@ def log_config(cfg: DictConfig) -> None:
...
+        'oci': mlflow.data.http_dataset_source.HTTPDatasetSource,
+        'https': mlflow.data.http_dataset_source.HTTPDatasetSource,
+        'hf': mlflow.data.huggingface_dataset_source.HuggingFaceDatasetSource,
+        'delta_table': mlflow.data.delta_dataset_source.DeltaDatasetSource,
+        'uc_volume': mlflow.data.uc_volume_dataset_source.UCVolumeDatasetSource,
+        'local': mlflow.data.http_dataset_source.HTTPDatasetSource,
+    }
+
+    for dataset_type, path, split in data_paths:
+        source_class = dataset_source_mapping.get(dataset_type)
```
irenedea, line 258: super nit: use syntactic sugar for dictionary gets
```suggestion
        source_class = dataset_source_mapping[dataset_type]
```

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +185,92 @@ def log_config(cfg: DictConfig) -> None:
...
+                            {}).get('dataset', {}).get('split', None)
+            data_paths.add(('local', local_path, data_split))
+
+    return data_paths
+
+
+def log_dataset_uri(cfg: DictConfig) -> Optional[mlflow.data.meta_dataset.MetaDataset]:
+    """Logs dataset tracking information to MLflow."""
+    if mlflow is None:
+        log.warning('MLflow is not installed. Skipping dataset logging.')
```
irenedea, line 242: we probably should error here? We should never call this function if mlflow isn't installed.

llmfoundry/utils/config_utils.py
irenedea: I'm guessing pyright won't be very happy about this? I think there's precedent in foundry/composer. Check out where `peft` is conditionally imported.

llmfoundry/utils/config_utils.py
KuuCi: should be .get here because key isn't guaranteed to be in dataset_source_mapping right?

llmfoundry/utils/config_utils.py
KuuCi: This should be checked here:
https://github.com/mosaicml/llm-foundry/blob/cbf0c30ec5974eb5b1491319e0fe74cfa1e3d2fe/llmfoundry/utils/config_utils.py#L231
With remote path taking higher precedent than local

KuuCi: probably dont need new source_datasets here, just check if it's a local path in the elif (se if backend if present)

tests/utils/test_mlflow_logging.py
```diff
@@ -0,0 +1,100 @@
...
+def test__parse_source_dataset_delta_table():
+    cfg = create_config(source_dataset_train='db.schema.train_table',
+                        source_dataset_eval='db.schema.eval_table')
+    expected = {('delta_table', 'db.schema.train_table', 'train'),
+                ('delta_table', 'db.schema.eval_table', 'eval')}
+    assert _parse_source_dataset(cfg) == expected
+
+
+def test__parse_source_dataset_uc_volume():
+    cfg = create_config(source_dataset_train='/Volumes/train_data',
```
dakinggg, line 30: should be dbfs prefix?

tests/utils/test_mlflow_logging.py
```diff
@@ -0,0 +1,100 @@
...
+        }})
+    expected = {('local', '/local/train_dataset', 'train'),
+                ('local', '/local/eval_dataset', 'eval')}
+    assert _parse_source_dataset(cfg) == expected
+
+
+@pytest.mark.usefixtures('mock_mlflow_classes')
+def test__log_dataset_uri_all_sources():
+    cfg = create_config(
+        train_loader={'dataset': {
```
dakinggg, line 79: should make the loader configs at match the sources

tests/utils/test_mlflow_logging.py
```diff
@@ -0,0 +1,100 @@
...
+        eval_loader={'dataset': {
+            'hf_name': 'huggingface/eval_dataset'
+        }},
+        source_dataset_train='db.schema.train_table',
+        source_dataset_eval='/Volumes/eval_data')
+
+    with patch('mlflow.data.meta_dataset.MetaDataset'):
+        with patch('mlflow.log_input') as mock_log_input:
+            _log_dataset_uri(cfg)
+            assert mock_log_input.call_count == 2
```
dakinggg, line 91: can also assert here that the calls were made with the args you expected

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,102 @@ def log_config(cfg: DictConfig) -> None:
...
         if mlflow.active_run():
             mlflow.log_params(params=om.to_container(cfg, resolve=True))
+            _log_dataset_uri(cfg)
+
+
+def _parse_source_dataset(cfg: DictConfig) -> Set[Tuple[str, str, str]]:
+    """Parse a run config for dataset information.
+
+    Given a config dictionary, parse through it to determine what the datasource
+    should be catagorized as. Possible data sources are Delta Tables, UC Volumes,
```
dakinggg, line 197:
```suggestion
    should be categorized as. Possible data sources are Delta Tables, UC Volumes,
```

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,102 @@ def log_config(cfg: DictConfig) -> None:
...
+    Args:
+        cfg (DictConfig): A config dictionary of a run
+
+    Returns:
+        Set[Tuple[str]]: A set of tuples formatted as (data type, path, split)
+    """
+    data_paths = set()
+
+    for data_split in ['train', 'eval']:
+        data_set = cfg.get(f'{data_split}_loader', {}).get('dataset', {})
```
dakinggg, line 209: eval loader could be a list, probably want to handle that case (https://github.com/mosaicml/llm-foundry/pull/603)

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,102 @@ def log_config(cfg: DictConfig) -> None:
...
+        cfg_split = data_set.get('split', None)
+        source_dataset_path = cfg.get(f'source_dataset_{data_split}', {})
```
dakinggg, line 211: default here should probably be `None` not empty dict?

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,102 @@ def log_config(cfg: DictConfig) -> None:
...
+        elif 'remote' in data_set:
+            remote_path = data_set['remote']
+            backend, _, _ = parse_uri(remote_path)
+            if backend:
+                remote_path = f'{remote_path.rstrip("/")}/{cfg_split}/' if cfg_split else remote_path
+                data_paths.add((backend, remote_path, data_split))
+            else:
+                data_paths.add(('local', remote_path, data_split))
+        else:
+            raise KeyError('DataSource Not Found')
```
dakinggg, line 240: lets not error here. a bug in this code doesn't need to take down the entire job. lets just log a warning

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,102 @@ def log_config(cfg: DictConfig) -> None:
...
+    """
+    if mlflow is None:
+        raise ImportError('MLflow is not installed. Skipping dataset logging.')
+
+    # Figure out which data source to use
+    data_paths = _parse_source_dataset(cfg)
+
+    dataset_source_mapping = {
+        's3': mlflow.data.http_dataset_source.HTTPDatasetSource,
+        'oci': mlflow.data.http_dataset_source.HTTPDatasetSource,
```
dakinggg, line 259: might as well add `gs` and `azure` here (https://github.com/mosaicml/composer/blob/4e0068a086291b0a35e9c7ce9fe8ce71e3d961be/composer/utils/file_helpers.py#L386-L427)

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,102 @@ def log_config(cfg: DictConfig) -> None:
...
+        'uc_volume': mlflow.data.uc_volume_dataset_source.UCVolumeDatasetSource,
+        'local': mlflow.data.http_dataset_source.HTTPDatasetSource,
+    }
+
+    # Map data source types to their respective MLFlow DataSource.
+    for dataset_type, path, split in data_paths:
+
+        if dataset_type in dataset_source_mapping:
+            source_class = dataset_source_mapping[dataset_type]
+            if dataset_type == 'delta_table':
```
dakinggg, line 272: instead of this if/else, can make the dictionary mapping above tuples, where the second item in the tuple is the path arg key name, and then just use that here

llmfoundry/utils/config_utils.py
```diff
@@ -5,15 +5,20 @@
...
 
-from composer.utils import dist
+from composer.utils import dist, parse_uri
 from omegaconf import DictConfig, ListConfig
 from omegaconf import OmegaConf as om
 
 from llmfoundry.layers_registry import ffns_with_megablocks
 from llmfoundry.models.utils import init_empty_weights
 
+try:
```
dakinggg, line 17: mlflow as actually not an optional dependency so you don't need this try/except. my bad

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
             wandb.config.update(om.to_container(cfg, resolve=True))
 
     if 'mlflow' in cfg.get('loggers', {}):
-        try:
-            import mlflow
-        except ImportError as e:
-            raise e
+        if not mlflow:
```
dakinggg, line 186: same, not needed

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+
+def _process_data_source(source_dataset_path: Optional[str],
+                         data_set: Dict[str, str], cfg_split: Optional[str],
+                         true_split: str, data_paths: List[Tuple[str, str,
+                                                                 str]]):
+    """Add a data source by mutating data_paths.
+
+    Given various dataset attributes, attempt to determine what type of dataset is being added, and parse
+    the dataset accordingly.
+
```
dakinggg, line 239: Missing Args: header and indentation

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+    source_dataset_path (Optional[str]): The source dataset in a cfg metadata
```
dakinggg, line 240:
```suggestion
    source_dataset_path (Optional[str]): The source dataset in cfg metadata
```

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+        eval_split = eval_data_set.get('split', None)
+        eval_source_path = cfg.get('source_dataset_eval', None)
+        _process_data_source(eval_source_path, eval_data_set, eval_split,
+                             'eval', data_paths)
+
+    return data_paths
+
+
+def _process_data_source(source_dataset_path: Optional[str],
+                         data_set: Dict[str, str], cfg_split: Optional[str],
```
dakinggg, line 232: `dataset` is one word (throughout)

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+                                                                 str]]):
+    """Add a data source by mutating data_paths.
+
+    Given various dataset attributes, attempt to determine what type of dataset is being added, and parse
+    the dataset accordingly.
+
+    source_dataset_path (Optional[str]): The source dataset in a cfg metadata
+    data_set (Dict[str, str]): The dataset from cfg
+    cfg_split (str): The split listed for the dataset in cfg
+    true_split (str): The split of the dataset to be added
```
dakinggg, line 243:
```suggestion
    true_split (str): The split of the dataset to be added (i.e. train or eval)
```

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+    data_paths (List[Tuple[str, str, str]]): A list of tuples formatted as (data type, path, split)
+    """
+    # Check for Delta table
+    if source_dataset_path and len(source_dataset_path.split('.')) >= 3:
```
dakinggg, line 247: delta table will have exactly three dot separated components as far as I know, so >= would be unknown source I think?

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+        data_paths.append(
+            ('uc_volume', source_dataset_path[len('dbfs:'):], true_split))
+    # Check for HF path
+    elif 'hf_name' in data_set:
+        hf_path = data_set['hf_name']
+        backend, _, _ = parse_uri(hf_path)
+        if backend:
+            hf_path = f'{hf_path.rstrip("/")}/{cfg_split}' if cfg_split else hf_path
+            data_paths.append((backend, hf_path, true_split))
+        elif hf_path.startswith('/'):
```
dakinggg, line 260: local path doesn't have to start with `/`, can we check `os.path.exists` here instead?

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+            data_paths.append(('local', hf_path, true_split))
+        else:
+            data_paths.append(('hf', hf_path, true_split))
+    # check for remote path
```
dakinggg, line 264:
```suggestion
    # Check for remote path
```

llmfoundry/utils/config_utils.py
```diff
@@ -178,9 +183,136 @@ def log_config(cfg: DictConfig) -> None:
...
+        log.warning('DataSource Not Found.')
+
+
+def _log_dataset_uri(cfg: DictConfig) -> None:
+    """Logs dataset tracking information to MLflow.
+
+    Args:
+        cfg (DictConfig): A config dictionary of a run
+    """
+    if mlflow is None:
```
dakinggg, line 283: same

llmfoundry/utils/config_utils.py
KuuCi: Should just be == 3 then

llmfoundry/utils/config_utils.py
```diff
@@ -177,10 +179,132 @@ def log_config(cfg: DictConfig) -> None:
...
+    # Check for UC volume
+    elif source_dataset_path and source_dataset_path.startswith('dbfs:'):
+        data_paths.append(
+            ('uc_volume', source_dataset_path[len('dbfs:'):], true_split))
+    # Check for HF path
+    elif 'hf_name' in dataset:
+        hf_path = dataset['hf_name']
+        backend, _, _ = parse_uri(hf_path)
+        if backend:
+            hf_path = f'{hf_path.rstrip("/")}/{cfg_split}' if cfg_split else hf_path
```
dakinggg, line 253: better to use `os.path.join` here

llmfoundry/utils/config_utils.py
```diff
@@ -177,10 +179,132 @@ def log_config(cfg: DictConfig) -> None:
...
+        elif os.path.exists(hf_path):
+            data_paths.append(('local', hf_path, true_split))
+        else:
+            data_paths.append(('hf', hf_path, true_split))
+    # Check for remote path
+    elif 'remote' in dataset:
+        remote_path = dataset['remote']
+        backend, _, _ = parse_uri(remote_path)
+        if backend:
+            remote_path = f'{remote_path.rstrip("/")}/{cfg_split}/' if cfg_split else remote_path
```
dakinggg, line 264: same, `os.path.join`

dakinggg (approved):
lgtm

PR #1121, opened 2024-04-19
Add option for subclasses to convert model and tokenizer in hf checkpointer

dakinggg:
This will allow subclasses of the hf checkpointer to apply an arbitrary transformation to the model and tokenizer before saving.


mvpatel2000 (commented):
PR descriptions are your friend. Please write one and then I will review :P

irenedea (approved):
LGTM!

mvpatel2000 (approved):

PR #1125, opened 2024-04-19
Fix HF checkpointer + mlflow bugs

dakinggg:
Fixes an edge case where an mlflow model would not be registered if the hf save interval was exactly the same as the total duration in batches and the max duration was specified in epochs.

Manual tests for chat and not chat (and deployed and checked that input example still works), both test the edge case:
- `dur-edge-case-chat-4-mztOoR`
- `dur-edge-case-pr-3-u5qIfu`

Note, there is now a scary exception in the mlflow model saving logs, but it is harmless. The deployment and input example still works.


setup.py
```diff
@@ -54,7 +54,8 @@
 
 install_requires = [
     'mosaicml[libcloud,wandb,oci,gcs]>=0.21.3,<0.22',
-    'mlflow>=2.10,<2.12',
+    'mlflow>=2.12.1,<2.13',
+    'pandas',
```
dakinggg, line 58:
```suggestion
```

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -260,6 +258,13 @@ def _is_last_batch(self, state: State):
             return True
 
         assert state.max_duration is not None  # for pyright
+
```
irenedea, line 261: Can you add a comment to describe the case that this handles?

irenedea (approved):

mvpatel2000 (approved):
Approving to unblock

PR #1126, opened 2024-04-20
Change main to a dev version

dakinggg:
Following https://github.com/mosaicml/composer/pull/3188, add the dev tag to source versions


mvpatel2000 (approved):
should write up release instructions so people know to change before release

PR #1127, opened 2024-04-20
Permit 2.3 nightlies

dakinggg:
Now that we have more frequent releases, I think we can allow for torch nightly versions on main. Releases should tighten the pin to released versions of torch.


mvpatel2000 (commented):
2.3 comes out in a week, imo we should just wait and not do this. The risk is high of breaking given imminent release

dakinggg:
Fair enough

PR #1128, opened 2024-04-20
Clean up the publicly exported API

dakinggg:
Fixes up a bunch of __all__s


llmfoundry/eval/datasets/__init__.py
```diff
@@ -8,14 +8,13 @@
...
-                                            stop_sequences_criteria, strip_data,
-                                            tokenizer_needs_prefix_space,
-                                            trim_context)
+
+# isort: off
+from llmfoundry.eval.datasets.utils import (
+    MultiTokenEOSCriteria, convert_tokens_to_tensors, get_continuation_span,
+    get_fewshot_sample_idxs, make_padded_input, stop_sequences_criteria,
+    strip_data, tokenizer_needs_prefix_space, trim_context)
+# isort: on
```
mvpatel2000, line 17: why?

mvpatel2000 (approved):
Can we please add the trailng commas thing? @b-chu has instructions (reisort)

llmfoundry/eval/datasets/__init__.py
dakinggg: because isort and yapf disagreed

dakinggg:
@mvpatel2000 Brian said it wasn't as simple as just changing the pyproject and would take a few hours, so I am not going to include that in this PR

@b-chu Not sure if there is a way to automate the alls. Some of the choices of what to reexport are more subjective, but if there is a good way to automate, I'm very open to it.

mvpatel2000:
> @mvpatel2000 Brian said it wasn't as simple as just changing the pyproject and would take a few hours, so I am not going to include that in this PR

Can we prioritize doing this soon? it has frequently come up in PR reviews

PR #1130, opened 2024-04-22
Add llama3 instructions to llama2 yaml

b-chu:
Adds instructions for what to change in the yaml to have it work for Llama 3 models


dakinggg (commented):
Instead of inline like this, can you add a readme section somewhere that is like "How to finetune a new Hugging Face model". Should say something like:
(1) Go to this yaml
(2) Change these fields to point to your new model
(3) Open a GitHub issue if it doesn't work. Most new causal LMs should work out of the box.

PR #1133, opened 2024-04-24
First initialize dist with gloo

dakinggg:
There seems to be an issue with the first nccl barrier, where the timeout does not properly apply. It does apply correctly to a gloo barrier though. So we add a gloo barrier at the start to make sure all ranks are alive, and then proceed with the normal process group intialization.

Run with this that runs successfully: `gloo-baseline-1-cAcvEB`
Run with this change and a rank sleeping in the gloo section (does timeout): `gloo-messed-up-no-gloo-2-CZw9ru` (on https://github.com/mosaicml/llm-foundry/pull/1133/commits/a7616ea5a110faf6c0fa5595b460123af8437ec1)
Run with this change and a rank sleeping in the nccl section (does not timeout): `gloo-messed-up-5-HB6gOH` (on https://github.com/mosaicml/llm-foundry/pull/1133/commits/ac258b93997a189e5e3157397e5d1ccab367ab30)

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -114,6 +114,22 @@ def validate_config(cfg: DictConfig):
                 f'MoEs with expert parallelism (moe_world_size {moe_world_size} > 1) require `use_orig_params=True`.'
             )
 
+def _initialize_gloo_and_nccl(dist_timeout: Union[int, float]):
+    # First, initialize with a gloo process group and test a barrier
+    log.debug('Initializing dist with cpu...')
+    dist.initialize_dist('cpu', timeout=dist_timeout)
+    log.debug('Testing barrier with cpu...')
+    dist.barrier()
+    log.debug('Barrier test passed with cpu. Destroying process group...')
+    tdist.destroy_process_group()
+    log.debug('Process group destroyed.')
+
+    # Now, initialize with the correct device
+    log.debug('Initializing dist with device...')
+    dist.initialize_dist(get_device(None), timeout=dist_timeout)
+    log.debug('Testing barrier with device...')
+    dist.barrier()
+    log.debug('Barrier test passed with device.')
 
 def main(cfg: DictConfig) -> Trainer:
     # Run user provided code if specified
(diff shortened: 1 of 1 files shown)


scripts/train/train.py
```diff
@@ -170,7 +171,38 @@ def main(cfg: DictConfig) -> Trainer:
...
+    log.debug('Barrier test passed with cpu. Destroying process group...')
+    tdist.destroy_process_group()
+    log.debug('Process group destroyed.')
+
+    # Now, initialize with the correct device
+    log.debug('Initializing dist with device...')
     dist.initialize_dist(get_device(None), timeout=dist_timeout)
+    log.debug('Testing barrier with device...')
+    dist.barrier()
+    log.debug('Barrier test passed with device.')
```
irenedea, line 205: Can you throw this into a helper function?

scripts/train/train.py
```diff
@@ -170,7 +171,38 @@ def main(cfg: DictConfig) -> Trainer:
                                                  'dist_timeout',
                                                  must_exist=False,
                                                  default_value=600.0)
+    python_log_level: Optional[str] = pop_config(cfg,
+                                                 'python_log_level',
+                                                 must_exist=False,
+                                                 default_value='debug')
+    # set logging level
+    if python_log_level is not None:
```
mvpatel2000, line 179:
```suggestion
    # Set logging level
    if python_log_level is not None:
```

scripts/train/train.py
```diff
@@ -170,7 +171,38 @@ def main(cfg: DictConfig) -> Trainer:
...
+        logging.getLogger(__name__).setLevel(
+            python_log_level.upper())  # Train script
+
+    # First, initialize with a gloo process group and test a barrier
+    log.debug('Initializing dist with cpu...')
+    dist.initialize_dist('cpu', timeout=dist_timeout)
+    log.debug('Testing barrier with cpu...')
+    dist.barrier()
+    log.debug('Barrier test passed with cpu. Destroying process group...')
+    tdist.destroy_process_group()
```
mvpatel2000, line 197:
```suggestion
    torch.distributed.destroy_process_group()
```
I prefer not abbreviating and importing in a non-standard way

scripts/train/train.py
```diff
@@ -170,7 +171,38 @@ def main(cfg: DictConfig) -> Trainer:
...
+            # 2022-06-29 11:22:26,152: rank0[822018][MainThread]: INFO: Message here
+            format=
+            f'%(asctime)s: rank{dist.get_global_rank()}[%(process)d][%(threadName)s]: %(levelname)s: %(name)s: %(message)s'
+        )
+        logging.getLogger('llmfoundry').setLevel(
+            python_log_level.upper())  # Foundry module
+        logging.getLogger(__name__).setLevel(
+            python_log_level.upper())  # Train script
+
+    # First, initialize with a gloo process group and test a barrier
```
mvpatel2000, line 191: Can you leave a longer explanation here from PR description to code?

scripts/train/train.py
irenedea, line 197: nit just use the full name, i think it's more clear. i thought tdist was a typo initially haha
```suggestion
    torch.distributed.destroy_process_group()
```

irenedea (approved):
minor comments, thank you!

mvpatel2000:
This might break on >300 GPUs per @sandeep-krishnamurthy-db's point on GLOO's limitations. Can you doubel check max size for GLOO?

scripts/train/train.py
dakinggg: done

scripts/train/train.py
dakinggg: done

mvpatel2000 (approved):
CC: @bigning this is roughly what im thinking for control plane except we wouldn't destroy it and would name it something else

PR #1134, opened 2024-04-24
Fix saving of generation_config for Llama-3

eldarkurtic:
The existing version of `HuggingFaceCheckpointer` does not respect model's `generation_config.json` and during saving it initializes it from the `config.json`. In the case of Llama-3 model, there is a discrepancy with `eos_token_id` which is set to  `128001` in `config.json` but to `[128001, 128009]` in `generation_config.json`. This means that Llama-3 models saved with `HuggingFaceCheckpointer` have `"eos_token_id": 128001` instead of `"eos_token_id": [128001, 128009],`.  

This creates problems when we want to use Llama-3 models produced with llm-foundry as they will most likely always generate text until the max number of tokens is exhausted instead of stopping at `128009` token.


dakinggg (approved):
Thanks for the fix! Please run `pre-commit` and then we can merge

PR #1136, opened 2024-04-25
Spawn a timeout process

irenedea:


llmfoundry/data/finetuning/tasks.py
```diff
@@ -809,12 +826,15 @@ def dataset_mapper(example: Dict):
                 desc='Tokenizing dataset',
             )
 
-            filtered_dataset = tokenized_dataset.filter(
-                partial(is_valid_ift_example, max_seq_len, target_prompts,
-                        target_responses, decoder_only_format),
-                num_proc=num_cpus_to_use,
-                desc='Filtering out long prompts',
-            )
+            filter_timeout = 600
```
dakinggg, line 829: can we make this configurable with a default? Also would suggest longer than 10 minutes, especially if it is not configurable. maybe 30

irenedea: Updated

dakinggg (approved):

irenedea:
Closing, we've root caused the hang issue. It was because dbrx tokenizer originally used the tiktoken tokenizer class, which was not compatible with huggingface dataset caching. This resulted in the dataset being reprocessed for every rank, which we've seen lead to deadlocks. Updating to the latest dbrx tokenizer, which uses the HuggingFace GPT2Tokenizer, solves this hang. https://huggingface.co/databricks/dbrx-base/blob/main/tokenizer_config.json

PR #1144, opened 2024-04-26
fix DatasetConstants.splints default value to protect dictionary overwriting

ivan-kud:
The values of `DatasetConstants.splints` dictionary of "c4" dataset overwrites the values for the "The Pile" splits dictionary. This is because the dict type of `DatasetConstants.splints` default value is mutable, so `default_factory` should be used.

Also there are typos for `train_small` amount of raw samples.

And I correct type hinting for `DataSplitConstants.raw_samples`. The `None` value is checked below in the code.

### DIFF
--- a/scripts/data_prep/convert_dataset_hf.py
+++ b/scripts/data_prep/convert_dataset_hf.py
@@ -89,7 +89,7 @@ def parse_args() -> Namespace:
 class DataSplitConstants:
     hf_split: str
     folder_split: str
-    raw_samples: Union[int, None]
+    raw_samples: Optional[int]
     truncated_samples: Union[int, None]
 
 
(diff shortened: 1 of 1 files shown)


scripts/data_prep/convert_dataset_hf.py
```diff
@@ -89,15 +89,15 @@ def parse_args() -> Namespace:
 class DataSplitConstants:
     hf_split: str
     folder_split: str
-    raw_samples: int
+    raw_samples: Union[int, None]
```
dakinggg, line 92:
```suggestion
    raw_samples: Optional[int]
```

dakinggg (approved):
Thanks!

PR #1154, opened 2024-04-30
Only create remote_ud on global rank 0 in HF checkpointer

irenedea:
Manual Test
`ift-mpt-7b-4nipk7-3uX2Rb`


irenedea:
Doesn't work because the rud uses dist operations to wait for all the uploads to finish, so this results in nccl timeouts. All ranks need to have a rud to go through the dist operations.

PR #1157, opened 2024-05-01
Move sentencepiece import

aspfohl:
This moves sentencepiece import down to where it's used. If you do not have it installed and want to load a `.model` file, an import error will still be raised

### DIFF
--- a/llmfoundry/utils/checkpoint_conversion_helpers.py
+++ b/llmfoundry/utils/checkpoint_conversion_helpers.py
@@ -91,7 +91,7 @@ def get_hf_tokenizer_from_composer_state_dict(
                     import sentencepiece as spm
                 except ImportError as e:
                     raise ImportError(
-                        'To load SentencePiece model, you need to install `sentencepiece`.'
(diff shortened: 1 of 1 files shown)


llmfoundry/utils/checkpoint_conversion_helpers.py
```diff
@@ -88,6 +87,13 @@ def get_hf_tokenizer_from_composer_state_dict(
                 with open(tokenizer_file_path, 'w') as _tmp_file:
                     _tmp_file.write(saved_content['content'])
             elif saved_content['file_extension'] == '.model':
+                try:
+                    import sentencepiece as spm
+                except ImportError as e:
+                    raise ImportError(
+                        'To load SentencePiece model, you need to install `sentencepiece`.'
```
dakinggg, line 94:
```suggestion
                        'Your tokenizer uses `sentencepiece`. Please install `sentencepiece` to load it.'
```

dakinggg (approved):

PR #1159, opened 2024-05-01
retry loop on chunked encoding error

milocress:


bigning (commented):
maybe use composer retry with backoff,  jitter and logging  https://github.com/mosaicml/composer/blob/5eddaf3e26e9930488cf329796eab3614d0dbfa2/composer/utils/retrying.py#L39 ?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -390,7 +397,9 @@ def _download_remote_hf_dataset(remote_path: str, split: str) -> str:
             finetune_dir, f'.node_{dist.get_node_rank()}_local_rank0_completed')
         if dist.get_local_rank() == 0:
             try:
-                get_file(path=name, destination=destination, overwrite=True)
+                retry(num_attempts=num_retries, exc_class=ChunkedEncodingError)(
```
bigning, line 400: not a blocker, but just fyi, In checkpoint downloading with OCI, we've seen different kinds of request exception. Let's see if there any other exception in this case.

bigning (approved):

llmfoundry/data/finetuning/dataloader.py
milocress: Sounds good, will add other network errors here

irenedea (changes requested):
I thought we already have retry logic in composer for OCI issues. `get_file` is imported from composer, should we be adding it there? @bigning

bigning:
> I thought we already have retry logic in composer for OCI issues. `get_file` is imported from composer, should we be adding it there? @bigning

@irenedea  I actually tried that https://github.com/mosaicml/composer/pull/3163, but it regressed the auto-resume which uses `get_file`, since auto-resume will try to download file no matter file exists or not. Adding retry to `get_file` slows down auto-resume

PR #1161, opened 2024-05-01
Add line splitting and other linting

b-chu:
Adds line splitting and trailing commas similar to https://github.com/mosaicml/composer/pull/3078

### DIFF
--- a/llmfoundry/__init__.py
+++ b/llmfoundry/__init__.py
@@ -12,20 +12,37 @@ import logging
 from llmfoundry.utils.logging_utils import SpecificWarningFilter
 
 # Filter out Hugging Face warning for not using a pinned revision of the model
-hf_dynamic_modules_logger = logging.getLogger(
-    'transformers.dynamic_module_utils')
+logger = logging.getLogger('transformers.dynamic_module_utils')
 new_files_warning_filter = SpecificWarningFilter(
-    'A new version of the following files was downloaded from')
+    'A new version of the following files was downloaded from',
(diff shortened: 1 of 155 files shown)


llmfoundry/__init__.py
```diff
@@ -13,19 +13,38 @@
...
 hf_dynamic_modules_logger = logging.getLogger(
-    'transformers.dynamic_module_utils')
+    'transformers.dynamic_module_utils',
+)
 new_files_warning_filter = SpecificWarningFilter(
-    'A new version of the following files was downloaded from')
+    'A new version of the following files was downloaded from',
+)
 
 hf_dynamic_modules_logger.addFilter(new_files_warning_filter)
```
mvpatel2000, line 22: can we just call it logger and make it 1 line instead of indenting?

mvpatel2000 (commented):
@b-chu should we raise min line length to be 100? it seems smaller than composer? or maybe im dumb.

I think looks good to me, would like daniel's pass too

llmfoundry/__init__.py
b-chu: Done

b-chu:
Yes foundry is 80 but composer is 120. There are many other style inconsistencies between the two repos :) I'm going to say that's out of the scope of this PR and we should discuss how/if we want to reconcile.

mvpatel2000 (approved):

dakinggg (approved):

PR #1169, opened 2024-05-03
Fix import and mocking

dakinggg:
A couple of small fixes related to composer dev version


llmfoundry/eval/datasets/in_context_learning_evaluation.py
```diff
@@ -26,6 +25,7 @@
     get_continuation_span,
     get_fewshot_sample_idxs,
     make_padded_input,
+    stop_sequences_criteria,
```
milocress, line 28: what does this do?

milocress (approved):

llmfoundry/eval/datasets/in_context_learning_evaluation.py
dakinggg: moves the import from composer to llmfoundry because the function was moved from composer to llmfoundry

PR #1179, opened 2024-05-07
log eval dataset misconfiguration

milocress:
# Make Eval Dataset Misconfiguration Errors Visible through Mosaic Logger

Wraps eval dataset creation with a mosaic logger try/catch.

## In Train context

`mpt-125m-chinchilla-regression-6OSoWo` (the log trace indicates that the error is caught and the context attribute is set)

```
[rank7]: │   623 │   │   if mosaicml_logger is not None:                                │
[rank7]: │   624 │   │   │   e.context = 'TrainContext'                                 │
[rank7]: │   625 │   │   │   mosaicml_logger.log_exception(e)                           │
[rank7]: │ ❱ 626 │   │   raise e 
```

## In Eval context

`mpt-125m-chinchilla-regression-6OSoWo` (the log trace indicates that the error is caught and the context attribute is set)

```
[rank0]: │   656 │   │   │   if mosaicml_logger is not None:                            │
[rank0]: │   657 │   │   │   │   e.context = 'EvalContext'                              │
[rank0]: │   658 │   │   │   │   mosaicml_logger.log_exception(e)                       │
[rank0]: │ ❱ 659 │   │   │   raise e  
```


dakinggg (commented):
Can you add a run name for a manual test?

llmfoundry/utils/exceptions.py
```diff
@@ -35,9 +35,17 @@
 ALLOWED_PROMPT_KEYS = {'prompt'}
 ALLOWED_MESSAGES_KEYS = {'messages'}
 
+ErrorContext = Union[Literal['TrainContext'], Literal['EvalContext']]
```
jjanezhang, line 38: Thoughts on calling this generally ModelTrainingError with the property "location"  = "train dataloader" || "eval dataloader"

and then we could extend this to catching more errors with metadata in the future!

jjanezhang (commented):
Looking great, had a comment on renaming so we can use this for many things in the future. 

Flagging that the dataset type won't be added in the `convert_delta_to_json.py` and `convert_text_to_mds.py`. I think we would have to parse the datapath to get the train split to wrap the error if you'd like to take a stab! Also happy to do it :) thanks so much Milo!

milocress:
> Flagging that the dataset type won't be added in the `convert_delta_to_json.py` and `convert_text_to_mds.py`. I think we would have to parse the datapath to get the train split to wrap the error if you'd like to take a stab! Also happy to do it :) thanks so much Milo!

I am happy to do this, but let's split it into a separate PR so this one stays small.

llmfoundry/utils/exceptions.py
milocress: I think `ModelTrainingError` might not be general enough because we'd want this in our eval script too, and maybe also in our data prep scripts. I was thinking initially we could drop the "Training" and call it a `ModelError`, but that's even more misleading because it implies the issue is with the model and not the train/eval dataloaders. I ended up sticking with ContextualError, because it seemed like the most general way to describe an error that could happen anywhere in the codebase and contained a context, but changed the `context` attribute to `location` to make the situation clearer.

PR #1193, opened 2024-05-09
fix eval

milocress:
### Eval
`eval-7b-hf-dIE2JY`

### Train
`ift-mpt-7b-lrhex4-Ykz4Tq`

---
if both of these runs run without error, you're good to merge IMO


dakinggg (approved):
LGTM, but please wait for the manual tests to pass before merging.

PR #1194, opened 2024-05-10
Log exception on inactivity callback

jjanezhang:
## Log exception on inactivity callback

Logs an exception on timeout so we can write a run event.

## Testing
Irene tested a run `test-1b-5tv00z` that timed out and I confirmed that we wrote the run timeout exception to metadata:
Screenshot 2024-05-10 at 2 24 15 PM


llmfoundry/callbacks/inactivity_callback.py
```diff
@@ -0,0 +1,56 @@
...
+        mosaicml_logger.log_exception(RunTimeoutError(timeout=timeout))
+    os.kill(os.getpid(), signal.SIGINT)
+
+
+class InactivityCallback(Callback):
+
+    def __init__(
+        self,
+        timeout: int = 1800,
+        mosaicml_logger: Optional[MosaicMLLogger] = None,
```
dakinggg, line 32: instead of passing via constructor, can you acquire the mosaic logger on `Event.INIT`? It should be on `state.callbacks` at that point. That allows the callback construction to not need a special case.

llmfoundry/callbacks/inactivity_callback.py
```diff
@@ -0,0 +1,56 @@
...
+from composer.loggers import MosaicMLLogger
+
+from llmfoundry.utils.exceptions import RunTimeoutError
+
+log = logging.getLogger(__name__)
+
+
+def _timeout(timeout: int, mosaicml_logger: Optional[MosaicMLLogger] = None):
+    log.error(
+        f'Timeout after no Trainer events were triggered for {timeout} seconds.',
```
dakinggg, line 20:
```suggestion
        f'Timeout after {timeout} seconds of inactivity after fit_end.',
```
(since the callback only does fit end for now, can update later if we add more timeout locations)

scripts/train/train.py
```diff
@@ -365,6 +365,11 @@ def main(cfg: DictConfig) -> Trainer:
     callback_configs = train_cfg.callbacks or {}
 
     # Callbacks
+    if callback_configs is not None:
```
dakinggg, line 368: remove, after making the constructor change

llmfoundry/callbacks/inactivity_callback.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    if mosaicml_logger is not None:
+        mosaicml_logger.log_exception(RunTimeoutError(timeout=timeout))
+    os.kill(os.getpid(), signal.SIGINT)
+
+
+class InactivityCallback(Callback):
```
dakinggg, line 27: Lets rename this to `RunTimeoutCallback` (matches the exception)

llmfoundry/callbacks/__init__.py
```diff
@@ -19,10 +19,13 @@
 from llmfoundry.callbacks.eval_output_logging_callback import EvalOutputLogging
 from llmfoundry.callbacks.fdiff_callback import FDiffMetrics
 from llmfoundry.callbacks.hf_checkpointer import HuggingFaceCheckpointer
-from llmfoundry.callbacks.log_mbmoe_tok_per_expert_callback import \
-    MegaBlocksMoE_TokPerExpert
-from llmfoundry.callbacks.monolithic_ckpt_callback import \
-    MonolithicCheckpointSaver
+from llmfoundry.callbacks.inactivity_callback import RunTimeoutCallback
```
dakinggg, line 22: can you change the filename too?

llmfoundry/callbacks/__init__.py
```diff
@@ -47,6 +50,7 @@
 callbacks.register('oom_observer', func=OOMObserver)
 callbacks.register('eval_output_logging', func=EvalOutputLogging)
 callbacks.register('mbmoe_tok_per_expert', func=MegaBlocksMoE_TokPerExpert)
+callbacks.register('inactivity', func=RunTimeoutCallback)
```
dakinggg, line 53: and the name its registered under

dakinggg (approved):

PR #1199, opened 2024-05-14
Allow EOS token for finetuning

jimwu6:
This is needed to allow the finetuning dataset to be constructed correctly.

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -390,6 +390,7 @@ def _validate_config(
         'allow_pad_trimming',
         'seq_parallel_replication',
         'auto_packing_replication',
+        'eos_token_id',
     }
     if not set(kwargs.keys()).issubset(allowed_additional_kwargs):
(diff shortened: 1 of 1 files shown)


dakinggg:
Where do you see this needed? I'm pretty sure finetuning just uses the eos from the tokenizer.

milocress:
> Where do you see this needed? I'm pretty sure finetuning just uses the eos from the tokenizer.

It looks like it's one of the things `**`ed into the superclass, I think there are some cases where omitting this causes an error. eg.

```
[rank2]: ValueError: sequence_id is a required argument when MPT is configured with attn_uses_sequence_id=True and the model is in train mode.
```

milocress (approved):

dakinggg:
@milocress that should only be for pretraining style. finetuning style handles packing and sequence id on its own. e.g. https://github.com/mosaicml/llm-foundry/blob/fb9a2259e880b0baa3d3523ff42def9ea6c29ce3/llmfoundry/data/packing.py#L155

PR #1201, opened 2024-05-14
Make `fc_type` a dict to pass fc kwargs through

snarayan21:
Some custom FC layers will need custom kwargs. This PR enables that by changing `fc_type` from `str` to `Union[str, Dict]`, and converting it to dict thereafter.

Default configs have also been moved to their own file, keeping those defaults easily accessible.

We will likely seek to deprecate fc_type down the line and replace it with fc_config instead.


llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -133,7 +137,8 @@ def __init__(
                 init_nonlinearity (str): The nonlinearity to use for parameter initialization with kaiming initialization schemes.
                 ---
                 See llmfoundry.models.utils.param_init_fns.py for info on other param init config options
-            fc_type (str): choose fc layer implementation. Options: torch and te. te layers support fp8 when using H100 GPUs.
+            fc_type (str | Dict): choose fc layer implementation. Options: torch and te. te layers support fp8 when using H100 GPUs. Can
```
dakinggg, line 140:
```suggestion
            fc_type (str | Dict): Choose fc layer implementation. Options: torch and te. te layers support fp8 when using H100 GPUs. Can
```

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -163,7 +168,12 @@ def __init__(
         self.init_config = init_config if init_config is not None else copy.deepcopy(
             init_config_defaults,
         )
-        self.fc_type = fc_type
+        # fc_type can be a string or a dictionary
```
dakinggg, line 171:
```suggestion
```
comment doesn't really add anything imo

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -163,7 +168,12 @@ def __init__(
...
+        if isinstance(fc_type, dict):
```
dakinggg, line 172: Would rewrite as
```
if isinstance(fc_type, str):
    fc_type = {'name': fc_type}
self.fc_type = fc_type
```

llmfoundry/models/layers/ffn.py
```diff
@@ -139,24 +140,27 @@ def __init__(
...
-            'bias': bias,
-        }
 
-        self.fc_kwargs['device'] = device
+        if fc_type is None:
+            self.fc_type = copy.deepcopy(fc_type_defaults)
+            self.fc_type['bias'] = bias
+            self.fc_type['device'] = device
+        else:
+            self.fc_type = fc_type
```
dakinggg, line 149: don't we still need to set bias and device in this case?

llmfoundry/models/layers/ffn.py
```diff
@@ -139,24 +140,27 @@ def __init__(
...
+        self.fc_type_name = self.fc_type['name']
```
dakinggg, line 150: any reason you don't pop the name off here so you can avoid the filter in the builder?

llmfoundry/models/mpt/configuration_mpt.py
snarayan21: changed

llmfoundry/models/mpt/configuration_mpt.py
snarayan21: done

llmfoundry/models/layers/ffn.py
snarayan21: hmm...I was assuming that MPTBlock init would pass the `fc_type` dict to both ffn and attention, and I modified MPTBlock init to make sure to add in `bias` and `device` keys to the dict. However, that also means that instantiating the FFN or Attention alone (that is, not by way of MPTBlock init) would not add these keys to the dict...

Related to this, I want to keep `fc_type` as a dict in MPTBlock, ffn, and attention, but I need a good way to set the default value for this dict since the `fc_type` arg will default to `None`. Importing the `fc_type_defaults` dict from `configuration_mpt.py` gives me a circular import error though. Is there a different place where we put constants that I can move the default dicts to, or should I just create a new file?

snarayan21: I'll just implement what I'm talking about above and get feedback lol

llmfoundry/models/layers/ffn.py
snarayan21: If we pop off the name from the dict in layer 1's FFN, then layer 2's FFN will not have the `name` key since it's been popped off already and the dict is shared between the two. So if I want to pop the name off, I'd have to `copy.deepcopy` the `fc_type` dict every time in both attention and ffn. Unless there's a different way?

dakinggg: yeah i guess the way you have it is fine

llmfoundry/models/layers/ffn.py
```diff
@@ -139,24 +140,27 @@ def __init__(
...
             ffn_hidden_size,
         )
-        self.fc_kwargs: dict[str, Any] = {
-            'bias': bias,
-        }
 
-        self.fc_kwargs['device'] = device
+        # Usually, fc_type dict should be passed in through MPTBlock's __init__ function.
+        if fc_type is None:
+            fc_type = fc_type_defaults
```
dakinggg, line 146: should be a deepcopy of the defaults right? to avoid the same issue of accidentally modifying the defaults dict for everyone

dakinggg: applies throughout

snarayan21: yeah i guess we want to avoid a situation where someone accidentally modifies the dict and it ripples throughout the model...

snarayan21: fixed

dakinggg (approved):

PR #1209, opened 2024-05-15
Add fc to HF export

dakinggg:
The HF export was missing the fc.py file and the associated register, so would result in an error when trying to load the HF exported model. This PR both fixes the issue, and adds a unit test that would've caught it, which requires clearing the registry for the unit test to simulate a fresh environment.

The move of `MultiTokenEOSCriteria` is just because Composer dev has removed that class, and it is now in foundry.

### DIFF
--- a/tests/a_scripts/inference/test_convert_composer_to_hf.py
+++ b/tests/a_scripts/inference/test_convert_composer_to_hf.py
@@ -1410,7 +1410,8 @@ def test_mptmoe_huggingface_conversion_callback(
 
 
 def test_mpt_convert_simple(
-    monkeypatch: pytest.MonkeyPatch, tmp_path: pathlib.Path
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: pathlib.Path,
 ):
     delete_transformers_cache()
 
--- a/tests/a_scripts/inference/test_convert_composer_to_hf.py
+++ b/tests/a_scripts/inference/test_convert_composer_to_hf.py
@@ -1442,9 +1442,9 @@ def test_mpt_convert_simple(
         cfg=model_cfg,
     )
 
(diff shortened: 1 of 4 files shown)


irenedea (approved):

snarayan21 (approved):
lgtm, thanks for the description too

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -1407,6 +1409,53 @@ def test_mptmoe_huggingface_conversion_callback(
...
+    )
+
+    original_model.model.save_pretrained(tmp_path)
+
+    edit_files_for_hf_compatibility(tmp_path)
+
+    monkeypatch.setattr(catalogue, 'REGISTRY', {})
+
+    _ = transformers.AutoModelForCausalLM.from_pretrained(
+        tmp_path,
```
milocress, line 1452:
```suggestion
        str(tmp_path),
```

milocress (approved):

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -1407,6 +1409,53 @@ def test_mptmoe_huggingface_conversion_callback(
...
+
+    original_model = build_composer_model(
+        name='mpt_causal_lm',
+        tokenizer=None,
+        cfg=model_cfg,
+    )
+
+    original_model.model.save_pretrained(tmp_path)
+
+    edit_files_for_hf_compatibility(tmp_path)
```
milocress, line 1447:
```suggestion
    edit_files_for_hf_compatibility(str(tmp_path))
```

milocress (approved):

PR #1210, opened 2024-05-15
Added torch_dmoe defaults, bug fixes for 2D inputs

snarayan21:
Users are hitting issues where changing their `ffn_type` from `mb_dmoe` to `torch_dmoe` results in errors with their config. This is because `torch_dmoe` does not have any defaults set. This PR ensures that `torch_dmoe` has the same defaults as `mb_dmoe`, according to the [Arguments dataclass](https://github.com/databricks/megablocks/blob/main/megablocks/layers/arguments.py) in the Megablocks repo. Fixed typing in some places as well. Added a test to make sure that `mb_dmoe` and `torch_dmoe` are the same in fwd and bwd for these defauly values.

There was also a bug for some inputs (top k = 1) where the outputs of the `max` function were 1 dimensional instead of 2. This bug has also been addressed.

There's a separate bug with Megablocks where, if both the `hidden_size` and `ffn_hidden_size` are both 128, there's an unintended bug -- this is why `hidden_size` has been bumped to 256. With values of `hidden_size` and `ffn_hidden_size` not divisible by 128, there's an unhelpful error thrown. Will open a separate PR on megablocks side to address this.

### DIFF
--- a/llmfoundry/models/layers/dmoe.py
+++ b/llmfoundry/models/layers/dmoe.py
@@ -1,9 +1,11 @@
 # Copyright 2024 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Callable, Optional
+from functools import partial
+from typing import Callable, Optional, Union
 
 import torch
+import torch.nn.functional as F
 
 __all__ = [
     'dMoE',
@@ -12,6 +14,8 @@ __all__ = [
     'GLU',
     'DroplessMLP',
 ]
+
+DEFAULT_ACTIVATION_FN = partial(F.gelu, approximate='tanh')
 
 
 # Add option to route tokens uniformly across experts. We use
@@ -36,8 +40,8 @@ class LearnedRouter(torch.nn.Module):
         hidden_size: int,
         moe_num_experts: int,
         moe_top_k: int,
-        moe_jitter_eps: float,
-        moe_normalize_expert_weights: bool,
+        moe_jitter_eps: Optional[float],
+        moe_normalize_expert_weights: Optional[Union[int, float]],
         uniform_expert_assignment: bool,
         device: Optional[torch.device],
     ) -> None:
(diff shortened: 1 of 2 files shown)


llmfoundry/models/layers/dmoe.py
```diff
@@ -66,11 +72,12 @@ def jitter(self, x: torch.Tensor) -> torch.Tensor:
...
         return low + noise * (high - low)
 
-    def _top_k(self, scores: torch.Tensor) -> torch.Tensor:
+    def _top_k(self, scores: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
         if self.moe_top_k == 1:
-            return scores.max(
+            values, indices = scores.max(
                 dim=-1,
             )  # pyright: ignore[reportGeneralTypeIssues]
+            return values.unsqueeze(-1), indices.unsqueeze(-1)
```
mvpatel2000, line 80: should you always unsqueeze?

llmfoundry/models/layers/dmoe.py
```diff
@@ -66,11 +72,12 @@ def jitter(self, x: torch.Tensor) -> torch.Tensor:
         )
         return low + noise * (high - low)
 
-    def _top_k(self, scores: torch.Tensor) -> torch.Tensor:
+    def _top_k(self, scores: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
         if self.moe_top_k == 1:
-            return scores.max(
+            values, indices = scores.max(
                 dim=-1,
             )  # pyright: ignore[reportGeneralTypeIssues]
```
mvpatel2000, line 79: why is this pyright necessary?

mvpatel2000:
LGTM but want @bigning 's signoff

llmfoundry/models/layers/dmoe.py
snarayan21: that was there before

llmfoundry/models/layers/dmoe.py
snarayan21: Yes, since the input to `_topk`, the `scores` matrix, is always 2D ([link](https://github.com/mosaicml/llm-foundry/blob/8274c6cf0e756816c7dc303370b68c7bb1663b27/llmfoundry/models/layers/dmoe.py#L84)). So the output of max gives 1D arrays for both `values` and `indices` as topk is 1. We want them to be 2D arrays

llmfoundry/models/layers/dmoe.py
snarayan21: removed it and the one below, fixed return type

bigning (approved):
LGTM, thanks!

PR #1222, opened 2024-05-20
add error when chat template fails

milocress:
# Add Error when Chat Template Fails
Creates the error type `ChatTemplateError`, which encodes the chat template, the error, and the sample that triggered it.

### DIFF
--- a/llmfoundry/utils/exceptions.py
+++ b/llmfoundry/utils/exceptions.py
@@ -116,6 +116,21 @@ class ConsecutiveRepeatedChatRolesError(ValueError, ContextualError):
         super().__init__(message)
 
 
+class ChatTemplateError(ValueError, ContextualError):
+    """Error thrown when a chat template fails to process a sample."""
+
+    def __init__(
+        self,
+        template: str,
+        sample: Dict[str, Any],
+        inner_message: str,
+    ) -> None:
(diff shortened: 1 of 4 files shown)


llmfoundry/utils/exceptions.py
```diff
@@ -116,6 +116,21 @@ def __init__(self, repeated_role: str) -> None:
...
+
+    def __init__(
+        self,
+        template: str,
+        sample: List[Dict[str, Any]],
+        inner_message: str,
+    ) -> None:
+        self.template = template
+        self.sample = sample
+        message = f'Failed to process sample {sample} with template {template}. {inner_message}'
```
dakinggg, line 130: I think putting the template itself into the error is maybe a bit too much text?

dakinggg (approved):

llmfoundry/utils/exceptions.py
milocress: Our template and the huggingface one are about one-liners. I was thinking including the template could make the error message more actionable. For example the Mistral error says something about repeated roles when the actual issue is no system prompts being allowed, which is fixable by including the mistral template.

dakinggg: hmm ok, all good

PR #1225, opened 2024-05-22
Add user error superclass

milocress:
Adds a `UserError` superclass to distinguish network/internal errors from user-actionable errors.


jjanezhang (approved):
⛷️

jjanezhang:
ooc: why this and not adding `is_user_error` as a field to the contextual error and default true?

PR #1226, opened 2024-05-22
Loss v len callback

ShashankMosaicML:
Adds the loss v len callback that logs loss vs context length for each iteration into tables. Using those tables we can plot graphs such as the following:

Screenshot 2024-05-22 at 2 58 54 PM

### DIFF
--- a/llmfoundry/callbacks/loss_perp_v_len_callback.py
+++ b/llmfoundry/callbacks/loss_perp_v_len_callback.py
@@ -2,14 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import warnings
-from typing import Dict, Mapping, Optional, Tuple, Union
+from typing import Any, Dict, Mapping, Optional, Tuple
 
 import torch
 from composer.core import Callback, State
 from composer.loggers import Logger, MLFlowLogger
 from composer.utils import dist
-from flash_attn.losses.cross_entropy import \
-    CrossEntropyLoss as FusedCrossEntropyLoss
 from torchmetrics import Metric
 
 from llmfoundry.models.mpt import ComposerMPTCausalLM
@@ -200,7 +198,7 @@ class LossPerpVLen(Metric):
         labels: torch.Tensor,
         logits: torch.Tensor,
         sequence_id: Optional[torch.Tensor],
-        loss_fn: Union[torch.nn.CrossEntropyLoss, FusedCrossEntropyLoss],
+        loss_fn: Any,
     ) -> None:
         """Updates the internal state with results from a new batch.
 
@@ -208,7 +206,7 @@ class LossPerpVLen(Metric):
             labels (torch.Tensor): A Tensor of ground-truth values to compare against.
             logits (torch.Tensor): A Tensor of labels.
             sequence_id (torch.Tensor | None): The sequence ids for tokens.
-            loss_fn (torch.nn.CrossEntropyLoss | flash_attn.losses.cross_entropy.CrossEntropyLoss): The cross entropy loss to use.
+            loss_fn (Any): The cross entropy loss to use.
         """
         valid_labels_mask = torch.where(
             labels != self.ignore_index,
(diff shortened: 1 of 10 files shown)


llmfoundry/callbacks/loss_perp_v_len_callback.py
```diff
@@ -0,0 +1,351 @@
...
+    def __init__(
+        self,
+        ignore_index: int,
+        dist_sync_on_step: bool = False,
+    ):
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+
+        self.needs_batch = True
+        self.ignore_index = ignore_index
+        self.loss_fn = torch.nn.CrossEntropyLoss(
```
vchiley, line 168: Is it possible to use the model loss fn instead of torch.nn.CrossEntropyLoss
does this make senes?

ShashankMosaicML: Done.

llmfoundry/callbacks/loss_perp_v_len_callback.py
```diff
@@ -0,0 +1,349 @@
...
+    Args:
+        log_batch_interval (int): The interval for logging.
+        compute_batch_interval (int): The interval for computing the metric.
+        ignore_index (int): Specifies a target value that is ignored for computing loss.
+    """
+
+    def __init__(
+        self,
+        log_batch_interval: int,
+        compute_batch_interval: int,
```
vchiley, line 31: Should we add a comment here saying that eventually we should consider making this into one arg?
(once that other issue gets resolved)

llmfoundry/callbacks/loss_perp_v_len_callback.py
```diff
@@ -0,0 +1,349 @@
...
+
+from llmfoundry.models.mpt import ComposerMPTCausalLM
+
+__all__ = [
+    'LossPerpVsContextLengthLogger',
+]
+
+
+class LossPerpVsContextLengthLogger(Callback):
+    """Logs the average loss and perplexity for every context length.
```
vchiley, line 20: Should we note that this only works with MLflow? (Is that true or am I misremembering something?)

llmfoundry/callbacks/loss_perp_v_len_callback.py
```diff
@@ -0,0 +1,349 @@
...
+from torchmetrics import Metric
+
+from llmfoundry.models.mpt import ComposerMPTCausalLM
+
+__all__ = [
+    'LossPerpVsContextLengthLogger',
+]
+
+
+class LossPerpVsContextLengthLogger(Callback):
```
dakinggg, line 19: @experimental_class

llmfoundry/callbacks/loss_perp_v_len_callback.py
```diff
@@ -0,0 +1,349 @@
...
+                'log_batch_interval is shorter than the compute_batch_interval for LossPerpVsContextLengthLogger.',
+            )
+        self.log_batch_interval = log_batch_interval
+        self.compute_batch_interval = compute_batch_interval
+        self.ignore_index = ignore_index
+        self.metric_dict = {}
+        self.loss_perp_v_len = LossPerpVLen(ignore_index)
+
+    def after_backward(self, state: State, logger: Logger) -> None:
+        if not isinstance(state.model, ComposerMPTCausalLM):
```
dakinggg, line 45: this check should happen on the `init` event so that it fails fast

llmfoundry/callbacks/loss_perp_v_len_callback.py
ShashankMosaicML: Done.

llmfoundry/callbacks/loss_perp_v_len_callback.py
ShashankMosaicML: Done.

llmfoundry/callbacks/loss_perp_v_len_callback.py
ShashankMosaicML: Done.

llmfoundry/callbacks/loss_perp_v_len_callback.py
ShashankMosaicML: Done.

dakinggg (approved):

PR #1228, opened 2024-05-22
Quick patch to check that Dataset Keys contain non-None Values

KuuCi:
A bug arose where user set `remote` to be None, which led the parse_uri to attempt to parse None, which throws an error. This change checks for values so that None vales will never hit parse_uri


KuuCi:
working version

dakinggg (approved):

PR #1234, opened 2024-05-23
Move MLFlow dataset outside of log_config 

KuuCi:
Finetuning product now defaults log_config to false, meaning that datasets will no longer show up for fine tuning runs. This attempts to fix that by making dataset logging independent of config logging

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -55,7 +55,7 @@ from llmfoundry.utils.config_utils import (
     TRAIN_CONFIG_KEYS,
     TrainConfig,
     log_config,
-    log_log_dataset_uri,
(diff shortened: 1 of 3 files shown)


scripts/train/train.py
```diff
@@ -530,6 +531,7 @@ def main(cfg: DictConfig) -> Trainer:
     if train_cfg.log_config:
         log.info('Logging config')
         log_config(logged_cfg)
+    log_dataset_uri(train_cfg)
```
dakinggg, line 534:
```suggestion
    log_dataset_uri(logged_cfg)
```

dakinggg (approved):

PR #1238, opened 2024-05-24
Add retries to downloads in convert_text_to_mds.py

irenedea:
Add retries to downloads in convert_text_to_mds.py.

Test run: `test-mds-conversion-TQ4LB3` confirmed still works


dakinggg (approved):
just add quick manual test that the download script still works

irenedea:
@dakinggg Done, added to PR description

PR #1240, opened 2024-05-25
Chunk file reads and tokenization for text to mds conversion

irenedea:
Read files and tokenize in 1MB chunks.

Addresses two issues:
1. Tokenization is significantly slower on long sequences.
2. Loading extremely large files into memory at once can cause OOMs

# Manual Tests
## Convert files of specific sizes
With changes:
`test-mds-conversion-500mb-rZDKAm` Took 8 minutes
`test-mds-conversion-baseline-5gb-HRj4NE` (ignore the fact that baseline is in the name lol) Took 1.5 hours

Without changes:
`test-mds-conversion-baseline-500mb-QJy3gj` Hanging-- stopped after 2 days

## Training 1 epoch with sec filings small dataset
Confirmed that the token counts are the same.

With changes:
`test-mds-conversion-mpt-7b-TRKKfd` 

Without changes:
`test-mds-conversion-baseline-mpt-7b-XTf6c2` 

## Training 1 epoch with 5.5MB file
Confirmed that the token counts are the same.

With changes:
`test-mds-conversion-mpt-7b-5mb-UY6HtO`

Without changes:
`test-mds-conversion-baseline-mpt-7b-5mb-71friv`

### DIFF
--- a/scripts/data_prep/convert_text_to_mds.py
+++ b/scripts/data_prep/convert_text_to_mds.py
@@ -94,7 +94,7 @@ class ConcatTokensFromFilesDataset(AbstractConcatTokensDataset):
                 # Add the EOS token to the buffer to separate files.
                 buffer += self.eos_tokens
 
-        # Finish up the last of the tokens.
+        # Yield any remaining samples of size max_length.
         while len(buffer) >= self.max_length:
             concat_sample = buffer[:self.max_length]
             buffer = buffer[self.max_length:] if self.should_wrap else []
--- a/scripts/data_prep/convert_text_to_mds.py
+++ b/scripts/data_prep/convert_text_to_mds.py
@@ -94,6 +94,12 @@ class ConcatTokensFromFilesDataset(AbstractConcatTokensDataset):
                 # Add the EOS token to the buffer to separate files.
                 buffer += self.eos_tokens
 
+        # Finish up the last of the tokens.
+        while len(buffer) >= self.max_length:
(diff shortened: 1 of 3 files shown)


mvpatel2000 (approved):
Why does it doing it in chunks make it faster?

irenedea:
@mvpatel2000 Tokenizing large strings at once is really slow. HF tokenizers are typically optimized to shorter strings: https://github.com/huggingface/transformers/issues/25873#issuecomment-1701727782

scripts/data_prep/convert_text_to_mds.py
```diff
@@ -37,6 +39,68 @@
...
+                    if not first_chunk:
+                        if iids[0] == self.tokenizer.bos_token_id:
+                            iids = iids[1:]
+
+                    # Add the tokens to the buffer
+                    buffer += iids
+                    while len(buffer) >= self.max_length:
+                        concat_sample = buffer[:self.max_length]
+                        buffer = buffer[self.
+                                        max_length:] if self.should_wrap else []
```
XiaohanZhangCMU, line 89: curious where is should_wrap compute/defined?

irenedea: in the abstract parent class: https://github.com/mosaicml/llm-foundry/pull/1240/files#diff-f700c538c3149c18f8d2cdfa1f0a23aa3ac22157962387de842a168c46cfd44bR58

PR #1241, opened 2024-05-27
Make HF conversion automatically add missing imports

dakinggg:
This PR removes the need for manually importing everything in the modeling file, by automatically including the missing imports needed for HF.


mvpatel2000 (approved):
Very nice!

PR #1244, opened 2024-05-30
Fix the error message thrown from dataloader

shitaoli-db:
Previously the dataloader will throw the error message that contain duplicated fileName + extension in the `Could not find a file with any of the supported extensions:`. This PR patched a simple fix to remove the unnecessary concat in the error message string.


dakinggg (commented):
Could you please provide a manual test run showing that this fix works?

PR #1245, opened 2024-05-31
[MCLOUD-4623] Add more detailed exception when user has uppercase in their example case but could potentially match the exampe type

shitaoli-db:
As title, we want to show user a better exception, i.e. to give lowercase keys, when the user example contain upper case keys like "Prompt" that could potentially be fixed by using lower cases.

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -35,7 +35,7 @@ import importlib
 import logging
 import os
 import warnings
-from collections.abc import Mapping
+from collections.abc import KeysView, Mapping
 from functools import partial
 from pathlib import Path
 from typing import (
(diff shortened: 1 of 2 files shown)


shitaoli-db:
> This PR seems a bit overkill to me. It adds code _just_ for detecting one possible you may have messed up your data (out of the many possible ways), when the error message would already tell you fairly clearly what was wrong. If this is an error we've seen a handful of times, maybe we can just update the `UnknownExampleTypeError` to mention that you may want to check your casing as the keys are case sensitive.

Yeah, I was thinking the same at first place, I think another idea is no change in the LLM foundry but rather to only change on https://github.com/databricks-mosaic/mcloud/pull/4083 for user facing error. @jjanezhang  WDYT on this?

shitaoli-db:
Closed for now since we agreeed that the change should be on mcloud side.

PR #1246, opened 2024-05-31
Wrap `FileNotFound` exceptions in the finetuning dataloader and `convert_text_to_mds`

angel-ruiz7:
Creates two exceptions:
-  `DatasetMissingFileError` --> Tells the user that a dataset file could not be found during a failure in the finetuning dataloader where `split` / `path` were not defined properly
- `DatasetInvalidFolderError` --> Tells the user that an `input_folder` is invalid and while trying to list objects in `convert_text_to_mds.py`

mapi exceptions implemented in https://github.com/databricks-mosaic/mcloud/pull/4088

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -25,6 +25,7 @@ from llmfoundry.data.packing import BinPackCollator, auto_packing_ratio
 from llmfoundry.data.text_data import build_streams
 from llmfoundry.utils.config_utils import to_dict_container
 from llmfoundry.utils.exceptions import (
+    DatasetMissingFileError,
     MissingHuggingFaceURLSplitError,
     NotEnoughDatasetSamplesError,
 )
@@ -541,14 +542,7 @@ def _download_remote_hf_dataset(remote_path: str, split: str) -> str:
                 get_file(path=name, destination=destination, overwrite=True)
             except FileNotFoundError as e:
                 if extension == SUPPORTED_EXTENSIONS[-1]:
-                    files_searched = [
-                        f'{name}/{split}{ext}' for ext in SUPPORTED_EXTENSIONS
-                    ]
-                    raise FileNotFoundError(
-                        f'Could not find a file with any of ' + \
-                        f'the supported extensions: {SUPPORTED_EXTENSIONS}\n' + \
-                        f'at {files_searched}',
-                    ) from e
+                    raise DatasetMissingFileError(file_name=f"name/{split}") from e
                 else:
                     log.debug(
                         f'Could not find {name}, looking for another extension',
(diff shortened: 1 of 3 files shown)


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -683,6 +686,8 @@ def build_collate_fn(
         'timeout': 0,
     })
 
+    mosaicml_logger = maybe_create_mosaicml_logger()
```
angel-ruiz7, line 689: idk if this is necessary, given that this exception was already logged somewhere. is this caught and logged somewhere else (similar to `convert_text_to_mds`)?

@jjanezhang

irenedea:
@jjanezhang @angel-ruiz7 This is awesome! I was actually just seeing a lot of FileNotFound errors last week. Are there any plans to handle FileNotFound errors that are surfaced when downloading checkpoints? I saw some errors last week due to a bad custom weights path.

llmfoundry/utils/exceptions.py
```diff
@@ -354,3 +356,26 @@ class RunTimeoutError(InternalError):
     def __init__(self, timeout: int) -> None:
         message = f'Run timed out after {timeout} seconds.'
         super().__init__(message, timeout=timeout)
+
+
+class DatasetMissingFileError(UserError):
```
snarayan21, line 361: Is this error supposed to cover both cases where a shard file is missing AND the index.json metadata file is missing? From the PR description, it sounds like this error is only for a missing/badly specified index.json file. So maybe it's more appropriate to call this `DatasetMissingIndexError` instead

llmfoundry/utils/exceptions.py
```diff
@@ -354,3 +356,26 @@ class RunTimeoutError(InternalError):
...
+    """Error thrown when a dataset cannot find a file."""
+
+    def __init__(
+        self,
+        file_name: str,
+        supported_extensions: Any,  # Run into type error when using List[str]
+    ) -> None:
+        message = "Could not find the file '{file_name}' with any of the supported extensions: "
+        message += ', '.join(supported_extensions) + '.'
+        message += ' Please check your train / eval data and try again.'
```
snarayan21, line 371:
```suggestion
        message += ' Please check your train / eval dataset configuration and try again.'
```
Not sure how this will be surfaced or if users will have access to this info, but this is more specific at least

snarayan21 (commented):
some feedback, nothing crazy

llmfoundry/utils/exceptions.py
```diff
@@ -354,3 +356,26 @@ class RunTimeoutError(InternalError):
...
         super().__init__(message, timeout=timeout)
+
+
+class DatasetMissingFileError(UserError):
+    """Error thrown when a dataset cannot find a file."""
+
+    def __init__(
+        self,
+        file_name: str,
+        supported_extensions: Any,  # Run into type error when using List[str]
```
irenedea, line 367: what's the type error that you get? We do use `List[str]` in other exceptions, see here https://github.com/mosaicml/llm-foundry/blob/9fbfccc832fa5d365d0b5a5bc83218d2ff1623cd/llmfoundry/utils/exceptions.py#L227 for example. Would be great if we can use `List[str]` here.

irenedea:
@angel-ruiz7 Are you still working on this?

PR #1252, opened 2024-06-05
Add registry for ICL datasets

sanjari-orb:
Purpose of PR: Create a registry for ICL eval dataset types. This will allow users to create custom in-context learning datasets and add them to the registry to run custom ICL evaluations during training.

### DIFF
--- a/llmfoundry/eval/datasets/in_context_learning_evaluation.py
+++ b/llmfoundry/eval/datasets/in_context_learning_evaluation.py
@@ -116,11 +116,11 @@ class InContextLearningDataset(Dataset):
         max_seq_len: int,
         pad_tok_id: int,
         num_fewshot: int,
-        fewshot_random_seed: int,
-        prompt_string: str,
-        example_delimiter: str,
-        continuation_delimiter: str,
         destination_path: str,
+        fewshot_random_seed: int = 1234,
+        prompt_string: str = '',
+        example_delimiter: str = '\n',
+        continuation_delimiter: str = ' ',
         prelimiter: str = '',
         context_key: str = 'context',
         answer_key: str = 'answer',
@@ -136,8 +136,6 @@ class InContextLearningDataset(Dataset):
         static_keys: Optional[List] = None,
         list_keys: Optional[List] = None,
         tensor_keys: Optional[List] = None,
-        *args: Any,
-        **kwargs: Any,
     ):
         self.tokenizer = tokenizer
         self.prefix_space = tokenizer_needs_prefix_space(self.tokenizer)
@@ -194,6 +192,17 @@ class InContextLearningDataset(Dataset):
         return batch['input_ids'].shape[0]
 
     def get_effective_batch_size(self, batch_size: int) -> int:
+        r"""Returns effective batch size computed for given ICL task.
+
+        The effective batch size may not be equal to the configured evaluation
+        batch size because for certain ICL tasks, >1 prompts can get created
+        for every input query depending on the number of choices/continuations.
+        This requires the effective batch size to be reduced to prevent larger batches than expected during eval. For example,
+        check InContextLearningMultipleChoiceTaskDataset.
+
+        Args:
+            batch_size (int): Original batch size configured for ICL evaluations
+        """
         return batch_size
 
     def update_generation_kwargs(self, generation_kwargs: Dict) -> None:
@@ -562,11 +571,11 @@ class InContextLearningGenerationTaskWithAnswersDataset(
         max_seq_len: int,
         pad_tok_id: int,
         num_fewshot: int,
-        fewshot_random_seed: int,
-        prompt_string: str,
-        example_delimiter: str,
-        continuation_delimiter: str,
         destination_path: str,
+        fewshot_random_seed: int = 1234,
+        prompt_string: str = '',
+        example_delimiter: str = '\n',
+        continuation_delimiter: str = ' ',
         prelimiter: str = '',
         context_key: str = 'context',
         answer_key: str = 'answer',
@@ -580,8 +589,6 @@ class InContextLearningGenerationTaskWithAnswersDataset(
         cot_delimiter: str = '',
         early_stopping_criteria: Optional[List[str]] = None,
         do_normalization: bool = True,
-        *args: Any,
-        **kwargs: Any,
     ):
         if tokenizer.eos_token_id is None:
             raise ValueError(
@@ -626,8 +633,6 @@ class InContextLearningGenerationTaskWithAnswersDataset(
             static_keys=static_keys,
             list_keys=list_keys,
             tensor_keys=tensor_keys,
-            *args,
-            **kwargs,
         )
         # NOTE: set these after init call because they take class vars
         self.early_stopping_criteria = early_stopping_criteria
@@ -781,9 +786,7 @@ class InContextLearningGenerationTaskWithAnswersDataset(
 
     def split_batch(self, batch: Any,
                     microbatch_size: Union[int, float]) -> Sequence[Any]:
-        """Handling for certain specialty columns that must be split into.
-
-        batches in different formats.
+        """Split batch handling for special columns.
 
         Args:
             batch (Dict): Batch of data
@@ -820,6 +823,7 @@ class InContextLearningGenerationTaskWithAnswersDataset(
                         for idx in range(num_chunks)]
         return batched_list
 
+
 class InContextLearningLMTaskDataset(InContextLearningDataset):
     """A dataset that constructs batches for in-context learning language.
 
@@ -840,11 +844,11 @@ class InContextLearningLMTaskDataset(InContextLearningDataset):
         max_seq_len: int,
         pad_tok_id: int,
         num_fewshot: int,
-        fewshot_random_seed: int,
-        prompt_string: str,
-        example_delimiter: str,
-        continuation_delimiter: str,
         destination_path: str,
+        fewshot_random_seed: int = 1234,
+        prompt_string: str = '',
+        example_delimiter: str = '\n',
+        continuation_delimiter: str = ' ',
         prelimiter: str = '',
         context_key: str = 'context',
         strip_dataset: bool = True,
@@ -855,8 +859,6 @@ class InContextLearningLMTaskDataset(InContextLearningDataset):
         generation_kwargs: Optional[Dict] = None,
         static_keys: Optional[List] = None,
         list_keys: Optional[List] = None,
-        *args: Any,
-        **kwargs: Any,
     ):
         super().__init__(
             dataset_uri=dataset_uri,
@@ -898,8 +900,6 @@ class InContextLearningLMTaskDataset(InContextLearningDataset):
                 'labels': 'context',
             },
             padding_side='right',
-            *args,
-            **kwargs,
         )
 
 
@@ -936,12 +936,14 @@ class InContextLearningMultipleChoiceTaskDataset(InContextLearningDataset):
         max_seq_len: int,
         pad_tok_id: int,
         num_fewshot: int,
-        fewshot_random_seed: int,
-        prompt_string: str,
-        example_delimiter: str,
-        continuation_delimiter: str,
         destination_path: str,
+        fewshot_random_seed: int = 1234,
+        prompt_string: str = '',
+        example_delimiter: str = '\n',
+        continuation_delimiter: str = ' ',
         prelimiter: str = '',
+        context_key: str = 'query',
+        tensor_keys: Optional[List] = None,
         answer_key: str = 'answer',
         strip_dataset: bool = True,
         tokenize_labels: bool = True,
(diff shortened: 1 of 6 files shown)


dakinggg:
Hi @sanjari-orb could you please add a PR description describing the change? Thank you!

llmfoundry/registry.py
```diff
@@ -206,6 +206,17 @@
     description=_metrics_description,
 )
 
+_icl_dataloaders_description = (
```
dakinggg, line 209: I guess this is actually datasets that are registered so should probably be called icl_datasets?

llmfoundry/eval/datasets/in_context_learning_evaluation.py
```diff
@@ -134,6 +136,8 @@ def __init__(
         static_keys: Optional[List] = None,
         list_keys: Optional[List] = None,
         tensor_keys: Optional[List] = None,
+        *args: Any,
```
dakinggg, line 139: I think that if you adjust `build_icl_dataloader` to take in arbitrary kwargs and then pass those along directly to the registry constructor, you don't need to add arbitrary support for kwargs here, right?

dakinggg: I think this will require a couple other changes too. I think we should push all of the config defaults to the actual class constructors instead of having them in `build_icl_dataloader` and `get_icl_task_dataloader`. If `get_icl_task_dataloader` just takes arbitrary kwargs containing the optional args, then I believe you should be able to not have any special casing for different class types. Required args for all ICL classes can be named, and optional args that only some subclasses have can be just kwargs that get passed along. Let me know what you think!

sanjari-orb:
Hi @dakinggg Sorry this was still a draft because I was still trying to get it to work. But thanks for the comments, I'll take them into account and update the PR soon.

dakinggg:
No worries, thanks for the contribution!

llmfoundry/eval/datasets/in_context_learning_evaluation.py
sanjari-orb: Yes I agree. The `super().__init__` calls in every specific ICL Dataset's implementation will become verbose, but I think that is a better design than not keeping the fields as named arguments.

sanjari-orb:
Hi @dakinggg could you point me to the steps to run the unit tests locally?

dakinggg:
Please see the makefile here (https://github.com/mosaicml/llm-foundry/blob/main/Makefile). Sorry there aren't better instructions!

# CPU tests
make test

# Multi CPU tests
make test-dist

# Single GPU tests
make test-gpu

# Multi GPU tests
make test-dist-gpu

llmfoundry/eval/datasets/in_context_learning_evaluation.py
```diff
@@ -1323,108 +1513,46 @@ def build_icl_dataloader(
...
-            early_stopping_criteria=early_stopping_criteria,
-            do_normalization=do_normalization,
-            generation_kwargs=generation_kwargs,
-        )
-        effective_batchsize = batch_size
-    else:
-        raise Exception(f'Unrecognized ICL task type: {icl_task_type}')
-
+    name = icl_task_type
+    kwargs: Dict[str, Any] = {
```
dakinggg, line 1517: I think we can get rid of this hardcoding of the kwargs, and remove the *args, **kwargs from the Dataset classes, by:
(1) Move the defaulters [here](https://github.com/mosaicml/llm-foundry/blob/5571101a50804406ef0fe23e7ea6795b3c4a1bcb/llmfoundry/utils/builders.py#L548-L563) into the Dataset class constructors themselves, so they that aren't hardcoded in some other place.
(2) just pass kwargs, all the way through until the construction from the registry

Does that make sense?

llmfoundry/registry.py
```diff
@@ -206,6 +206,17 @@
     description=_metrics_description,
 )
 
+_icl_datasets_description = (
+    'The ICL dataloaders registry is used to register an torch.utils.data.Dataset class which can be used for ICL tasks.'
```
dakinggg, line 210:
```suggestion
    'The ICL datasets registry is used to register an torch.utils.data.Dataset class which can be used for ICL tasks.'
```

llmfoundry/registry.py
```diff
@@ -206,6 +206,17 @@
...
+)
+icl_datasets = create_registry(
+    'llmfoundry',
+    'icl_datasets',
+    generic_type=Type[Dataset],
```
dakinggg, line 215: For your customization did you want to make this a generic `Dataset`? Otherwise I would suggest having it be `Type[InContextLearningDataset]`

dakinggg (commented):
Thanks! Had a couple comments. Also, once the tests are passing, I'll run this PR through our regression tests to check and make sure we aren't accidentally changing any evals.

llmfoundry/eval/datasets/in_context_learning_evaluation.py
```diff
@@ -519,46 +526,12 @@ def collate_fn(self, data: List[Dict[str, Any]]) -> Dict[str, Any]:
...
-        batched_list = [{k: v[idx]
-                         for k, v in chunked.items()}
-                        for idx in range(num_chunks)]
-        return batched_list
+    def split_batch(
+        self, 
+        batch: Any,
+        microbatch_size: Union[int, float]
+    ) -> Sequence[Any]:
+        return _default_split_batch(batch, microbatch_size)
```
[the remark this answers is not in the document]
sanjari-orb: This was the resulting behavior of the previous code https://github.com/mosaicml/llm-foundry/blob/5571101a50804406ef0fe23e7ea6795b3c4a1bcb/llmfoundry/eval/datasets/in_context_learning_evaluation.py#L1408-L1417

So I moved the default split_batch function to `InContextLearningGenerationTaskWithAnswersDataset` since it was only being used there and retained the `split_batch` implementation of `InContextLearningMultipleChoiceTaskDataset`. 

Setting split_batch=None internally gets translated to `default_split_batch` here: https://github.com/mosaicml/composer/blob/7fbac5f7f3144ee6c0f7731b125cd308546bc237/composer/core/data_spec.py#L193

llmfoundry/registry.py
sanjari-orb: I do have to make my own variant of `InContextLearningDataset` which can read from parquet files and is not limited to json file inputs, but yeah I can just inherit from `InContextLearningDataset`.

llmfoundry/eval/datasets/in_context_learning_evaluation.py
sanjari-orb: Points 1,2 make sense to me. But will we not need `*args, **kwargs` in the Dataset classes to catch any extra parameters which get sent from the registry constructor, since the arguments vary slightly for every icl dataset type.

sanjari-orb: Actually might be alright

dakinggg: Yeah, my thought is that the only reason the *args/**kwargs exists is because the defaults are set independent of dataset type, and so some of those defaults are passed to constructors that don't need them. If the defaults move into the constructors, each dataset should only be receiving args that it actually uses.

llmfoundry/registry.py
dakinggg: Yeah, I think inheriting makes sense here. We want each registry to have a defined interface, and this one is that it matches the `InContextLearningDataset` interface.

llmfoundry/eval/datasets/in_context_learning_evaluation.py
sanjari-orb: So are you suggesting constructing the kwargs for this call https://github.com/mosaicml/llm-foundry/blob/5571101a50804406ef0fe23e7ea6795b3c4a1bcb/llmfoundry/utils/builders.py#L601 so that based on the dataset type we choose which fields to include (for eg, cot_delimiter will get included only in a particular dataset). 

With kwargs in the dataset constructors, I can use a single call like:
```
            icl_constructor_kwargs: Dict[str, Any] = {
                'dataset_uri': icl_cfg['dataset_uri'],
                'tokenizer': tokenizer,
                'max_seq_len': icl_cfg['max_seq_len'],
                'pad_tok_id': pad_tok_id,
                'num_fewshot': num_fewshot,
                'prompt_string': icl_cfg['prompt_string'],
                'example_delimiter': icl_cfg['example_delimiter'],
                'continuation_delimiter': icl_cfg['continuation_delimiter'],
                'question_prelimiter': icl_cfg.get('question_prelimiter', ''),
                'fewshot_random_seed': icl_cfg['fewshot_random_seed'],
                'cot_delimiter': icl_cfg.get('cot_delimiter', ''),
                'early_stopping_criteria': early_stopping_criteria,
                'do_normalization': icl_cfg.get('do_normalization', True),
                'generation_kwargs': icl_cfg.get('generation_kwargs', {}),
            }

            dataloaders = get_icl_task_dataloader(
                icl_task_type=icl_cfg['icl_task_type'],
                dataset_uri=icl_cfg['dataset_uri'],
                tokenizer=tokenizer,
                batch_size=icl_cfg['batch_size'],
                hf_loading_vars=hf_loading_vars,
                hf_parsing_map=hf_parsing_map,
                has_categories=icl_cfg.get('has_categories', False),
                kwargs=icl_constructor_kwargs
            )
```

dakinggg: The problem with kwargs in the dataset constructors that get ignored is that it can silently ignore typos/code errors. I am suggesting removing all of this code (https://github.com/mosaicml/llm-foundry/blob/5571101a50804406ef0fe23e7ea6795b3c4a1bcb/llmfoundry/utils/builders.py#L548-L569), so that there are no dataset specific args that are getting passed into other constructors that don't use them. These defaults can be moved into the relevant dataset constructor. Then icl_cfg can be passed along to [here](https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/eval/datasets/in_context_learning_evaluation.py#L1511) as arbitrary kwargs before being **ed into the registried constructor. Things that are not in `icl_cfg` (e.g. tokenizer) can be passed directly by named arg since all the classes accept these in their constructor.

Happy to help with implementation too if I am not being clear. Feel free to message me on slack as well.

llmfoundry/registry.py
sanjari-orb: One problem I am running into is that the since the `InContextLearningDataset` class is defined in the same place as the builder methods (`build_icl_dataloader`), using `InContextLearningDataset` creates a circular import between this and the builder method. Do you think the refactoring can be taken up as a follow up?

dakinggg: Yeah, I'm ok with that. Please note in a comment about this (and the registry description)

llmfoundry/eval/datasets/in_context_learning_evaluation.py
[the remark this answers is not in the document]
dakinggg: ahh ok, I follow you now. Makes sense, thanks for explaining!

sanjari-orb:
@dakinggg Is there a linter I can use to fix the code quality checks?

dakinggg:
@sanjari-orb yeah, running `pre-commit` should do it

llmfoundry/utils/builders.py
```diff
@@ -589,38 +579,37 @@ def _validate_cfg(icl_cfg: Dict[str, Any]):
...
-                batch_size=icl_cfg['batch_size'],
-                max_seq_len=icl_cfg['max_seq_len'],
-                pad_tok_id=pad_tok_id,
-                num_fewshot=num_fewshot,
-                prompt_string=icl_cfg['prompt_string'],
-                example_delimiter=icl_cfg['example_delimiter'],
+                icl_task_type=icl_cfg['icl_task_type'],
+                dataset_uri=icl_cfg['dataset_uri'],
+                tokenizer=tokenizer,
+                batch_size=icl_cfg.get('batch_size', default_batch_size),
```
dakinggg, line 607: I think the default batch size is already set in _validate_cfg? can remove in one of the places

llmfoundry/eval/datasets/in_context_learning_evaluation.py
```diff
@@ -1292,25 +1474,13 @@ def tokenize_example(
 def build_icl_dataloader(
     icl_task_type: str,
     dataset_uri: str,
-    tokenizer: transformers.PreTrainedTokenizerBase,
+    tokenizer: Union[transformers.PreTrainedTokenizer,
```
dakinggg, line 1477: Any particular reason you changed the type here? Both of those classes inherit from the base class, so I've used the base type for tokenizer typing.

dakinggg (approved):
Successful regression test run name: `llm-foundry-regression-tests-runner-ActFp3`

LGTM, thank you for the contribution!

llmfoundry/eval/datasets/in_context_learning_evaluation.py
sanjari-orb: Probably an artifact of some merge error by me, will revert it back

sanjari-orb:
@dakinggg I'm not sure why the PR GPU tests failed: https://github.com/mosaicml/llm-foundry/actions/runs/9514022193/job/26225300100?pr=1252. Could you take a look?

PR #1256, opened 2024-06-05
Add curriculum learning callback

b-chu:
# Curriculum learning callback
## Requirements
- Requires StreamingDataset
- Each datamix must have a duration and train_loader
- Duration must be positive and defined in terms of epochs or tokens
- Duration units must match max_duration
- The length of the schedule must be equal to max_duration
- The part of the schedule that has already been trained on cannot be changed in future resumption runs
## Features
- Defines schedules by specifying datamixes in terms of training duration
- Supports epochs and tokens for each iteration
- Enables single run curriculum learning
- Swaps the entire dataloader at the start of each iteration
## Other
- Refactor build_tokenizer to avoid circular dependencies
- Refactor process_init_dist to avoid circular dependencies

# Manual tests
## Matches old callback behavior
image

## Resumes correctly in the middle of the schedule
image

## Resumes correctly when new datamix added to schedule
image

## Resumes correctly when callback added after initial training run
image

# API
## Old API:
```
train_loader:
  <some params>
callbacks:
  curriculum_learning:
    dataset_index: 0
```
Start a new run
```
train_loader:
  <some params>
callbacks:
  curriculum_learning:
    dataset_index: 1
```
Start a new run
```
train_loader:
  <some params>
callbacks:
  curriculum_learning:
    dataset_index: 2
```

## New API:
```
train_loader:
  <dataloader parameters>
callback:
  curriculum_learning:
  - duration: <number>tok
    train_loader:  # matches top level train_loader
      <dataloader parameters>
  - duration: <number>tok
    train_loader:
      <dataloader parameters>
  - duration: <number>tok
    train_loader:
      <dataloader parameters>
```

### DIFF
--- a/llmfoundry/callbacks/curriculum_learning_callback.py
+++ b/llmfoundry/callbacks/curriculum_learning_callback.py
@@ -7,47 +7,247 @@ This callback is currently experimental. The API may change without warning in
 the future.
 """
 
+import copy
 import logging
-from typing import Any, Dict
+from typing import Any
 
-from composer.core import State
-from composer.loggers import Logger
+from composer import DataSpec
+from composer.core import State, Time, TimeUnit, ensure_time
+from composer.loggers import Logger, MosaicMLLogger
 from streaming import StreamingDataset
+from streaming.base.util import clean_stale_shared_memory
 from torch.utils.data import DataLoader
 
 from llmfoundry.interfaces import CallbackWithConfig
-from llmfoundry.utils.warnings import experimental_class
+from llmfoundry.utils.exceptions import (
+    BaseContextualError,
+    TrainDataLoaderLocation,
+)
+from llmfoundry.utils.mosaicml_logger_utils import no_override_excepthook
 
 log = logging.getLogger(__name__)
 
 __all__ = ['CurriculumLearning']
 
 
-@experimental_class('CurriculumLearning callback')
 class CurriculumLearning(CallbackWithConfig):
     """Starts an epoch with a different dataset when resuming from a checkpoint.
 
+    Example schedule:
+    [
+        {
+            'duration': <number>tok,
+            'train_loader': <dataloader parameters>, # matches top level train_loader
+        },
+        {
+            'duration': <number>tok,
+            'train_loader': <dataloader parameters>,
+        },
+        {
+            'duration': <number>tok,
+            'train_loader': <dataloader parameters>,
+        ],
+    ]
+
     Args:
-        train_config (Dict): The configuration of the dataset currently
+        train_config (dict): The configuration of the dataset currently
             being used. Note that this is the full train config and must
-            contain the 'train_loader' key.
-        dataset_index (int): The index of the dataset currently being used.
+            contain the 'train_loader', 'device_train_batch_size', and
+            'tokenizer' keys.
+        schedule (list[dict[str, Any]]): The list of datamixes to use and their
+            durations. Duration units must match max_duration and be in terms of
+            a TimeUnit that is supported by Iteration. The duration values must
+            be positive. There must be at least one datamix in the schedule. The
+            first datamix in the schedule must match the train_loader in the
+            train_config. On resumption, previously trained on datamixes and
+            durations cannot be changed. The duration of the current datamix
+            must be greater than the saved timestamp. The dataset must be a
+            StreamingDataset.
     """
 
-    def __init__(self, train_config: Dict, dataset_index: int):
-        self.dataset_index = dataset_index
-        self.saved_dataset_index = 0
-        self.all_dataset_configs = []
-        self.current_dataset_state = {}
-        # The current dataset config is resolved and passed in train.py
-        self.current_dataset_config = train_config['train_loader']
+    def __init__(
+        self,
+        train_config: dict[str, Any],
+        schedule: list[dict[str, Any]],
+    ):
+        # Ensure all duration units are in epochs or tokens and values are positive
+        self._schedule = schedule
+        if len(self._schedule) == 0:
+            raise ValueError('The schedule must have at least one datamix.')
+        for index, datamix in enumerate(self._schedule):
+            self._validate_datamix(datamix)
+
+            if (
+                index == 0 and
+                train_config['train_loader'] != datamix['train_loader']
+            ):
+                raise ValueError((
+                    'The first datamix in the schedule must match the '
+                    'train_loader in the train_config.'
+                ))
+
+        self._schedule_index = 0
+        self.device_train_batch_size = train_config['device_train_batch_size']
+        self.tokenizer = None
+
+    def init(self, state: State, logger: Logger):
+        del logger  # unused
+
+        if not hasattr(state.model, 'tokenizer'):
+            raise ValueError('state.model must have a tokenizer attribute.')
+        self.tokenizer = state.model.tokenizer
 
     def before_load(self, state: State, logger: Logger):
-        del logger
+        del logger  # unused
 
-        # Save the current dataset state so we can restore it correctly
-        # if we are resuming with a new dataset.
-        train_loader = state.train_dataloader
+        # Ensure all duration units are the same as max_duration
+        datamix_units = [datamix['duration'].unit for datamix in self._schedule]
+        assert state.max_duration is not None, 'max_duration should have beeen set.'
+        if any(state.max_duration.unit != unit for unit in datamix_units):
+            raise ValueError((
+                f'All durations in the schedule must have the same units as '
+                f'the max_duration. Expected {state.max_duration.unit}, but '
+                f'got {datamix_units}.'
+            ))
+
+        # Ensure schedule duration is equal to max_duration
+        schedule_duration = Time(0, state.max_duration.unit)
+        for datamix in self._schedule:
+            assert isinstance(datamix['duration'], Time)
+            schedule_duration += datamix['duration']
+        if schedule_duration != state.max_duration:
+            raise ValueError((
+                'The sum of all durations in the schedule must be equal to the '
+                'max_duration.'
+            ))
+
+        self._validate_dataloader(state.train_dataloader)
+
+    def after_load(self, state: State, logger: Logger):
+        del logger  # unused
+
+        self._validate_dataloader(state.train_dataloader)
+
+        # If checkpoint was saved before iteration was incremented, we need to increment it now
+        if ((
+            self._schedule[self._schedule_index]['duration'].unit
+            == TimeUnit.TOKEN and state.timestamp.token_in_iteration >=
+            self._schedule[self._schedule_index]['duration'].value
+        ) or (
+            self._schedule[self._schedule_index]['duration'].unit
+            == TimeUnit.EPOCH and state.timestamp.epoch_in_iteration >=
+            self._schedule[self._schedule_index]['duration'].value
+        )):
+            log.warning((
+                'The CurriculumLearning callback has detected that the previous run did not correctly '
+                'increment the iteration.'
+            ))
+            self._schedule_index += 1
+            state.timestamp = state.timestamp.to_next_iteration()
+
+    def iteration_start(self, state: State, logger: Logger):
+        # Swap the dataset if starting a new iteration that's not the original datamix
+        if self._schedule_index > 0:
+            # TODO: trainer._train_data_spec should be updated whenever the dataloader is updated
+            # Dataloaders with the same prefix access the same shared memory
+            # which is stale
+            clean_stale_shared_memory()
+            datamix = copy.deepcopy(self._schedule[self._schedule_index])
+            data_spec = self._build_train_loader(
+                train_loader_config=datamix['train_loader'],
+                logger=logger,
+            )
+            state.set_dataloader(
+                dataloader=data_spec.dataloader,
+                dataloader_label='train',
+            )
+            state.train_dataloader = state.dataloader
+            self._validate_dataloader(state.train_dataloader)
+
+        # Set the length of the new iteration
+        state._iteration_length = self._schedule[self._schedule_index
+                                                ]['duration']
+
+    def iteration_end(self, state: State, logger: Logger):
+        del state, logger  # unused
+
+        self._schedule_index += 1
+
+    def state_dict(self):
+        return {
+            'schedule': self._schedule,
+            'schedule_index': self._schedule_index,
+        }
+
+    def load_state_dict(self, state: dict[str, Any]):
+        self._schedule_index = state['schedule_index']
+
+        # Ensure that the schedule has not changed on previously trained datamixes
+        for idx in range(state['schedule_index']):
+            if self._schedule[idx] != state['schedule'][idx]:
+                raise ValueError((
+                    f'Previous datamixes must stay the same across ',
+                    f'resumptions. Expected {state["schedule"][idx]} but got ',
+                    f'{self._schedule[idx]}',
+                ))
+
+        # Ensure that the datamix has not changed on the current datamix
+        current_loader = self._schedule[self._schedule_index]['train_loader']
+        saved_loader = state['schedule'][self._schedule_index]['train_loader']
+        if current_loader != saved_loader:
+            raise ValueError((
+                f'The current datamix must stay the same across resumptions. ',
+                f'Expected {saved_loader} but got {current_loader}',
+            ))
+
+        # Ensure that the current datamix duration is greater than timestamp
+        duration = self._schedule[self._schedule_index]['duration']
+        if duration.unit != TimeUnit.TOKEN and duration.unit != TimeUnit.EPOCH:
+            raise ValueError((
+                f'Duration must be in terms of tokens or epochs, but got ',
+                f'{duration.unit}.',
+            ))
+        if ((
+            duration.unit == TimeUnit.TOKEN and
+            duration > state['timestamp'].token_in_iteration
+        ) or (
+            duration.unit == TimeUnit.EPOCH and
+            duration > state['timestamp'].epoch_in_iteration
+        )):
+            raise ValueError((
+                'The duration of the current datamix must be less or equal to '
+                'than the saved timestamp.'
+            ))
+
+    def _build_train_loader(
+        self,
+        train_loader_config: dict[str, Any],
+        logger: Logger,
+    ) -> DataSpec:
+        from llmfoundry.data.dataloader import build_dataloader
+
+        # Copied from scripts/train/train.py
+        log.info(
+            f'Building train loader in CurriculumLearning callback for dataset {self._schedule_index}',
+        )
+        assert self.tokenizer is not None
+        try:
+            return build_dataloader(
+                train_loader_config,
+                self.tokenizer,
+                self.device_train_batch_size,
+            )
+        except BaseContextualError as e:
+            for destination in logger.destinations:
+                if (
+                    isinstance(destination, MosaicMLLogger) and
+                    no_override_excepthook()
+                ):
+                    e.location = TrainDataLoaderLocation
+                    destination.log_exception(e)
+            raise e
+
+    def _validate_dataloader(self, train_loader: Any):
         # Check if we are using a DataLoader and StreamingDataset
         if not isinstance(train_loader, DataLoader):
             raise ValueError(
@@ -61,54 +261,23 @@ class CurriculumLearning(CallbackWithConfig):
                 f'because it requires loading and saving dataset state. ',
                 f'Instead, got a dataset of type {type(dataset)}',
             )
-        assert isinstance(dataset, StreamingDataset)
-        # Save the current dataset state so we can restore it if needed.
-        self.current_dataset_state = dataset.state_dict(  # type: ignore
-            num_samples=0, from_beginning=False)
 
-    def after_load(self, state: State, logger: Logger):
-        del logger
+    def _validate_datamix(self, datamix: dict[str, Any]):
+        if 'duration' not in datamix:
+            raise ValueError('Each datamix must have a duration.')
+        datamix['duration'] = ensure_time(
+            datamix['duration'],
+            TimeUnit.EPOCH,
+        )
+        if datamix['duration'].value <= 0:
+            raise ValueError('The duration must be positive.')
+        if (
+            datamix['duration'].unit != TimeUnit.EPOCH and
+            datamix['duration'].unit != TimeUnit.TOKEN
+        ):
+            raise ValueError(
+                'Schedules can only be defined in terms of epochs or tokens.',
+            )
 
-        # As saved_dataset_index is loaded from state_dict, this only runs when
-        # a user explicitly increments the dataset_index and not on any other
-        # resumption, including autoresume.
-        train_loader = state._train_dataloader
-        assert isinstance(
-            train_loader,
-            DataLoader,
-        ), 'CurriculumLearning callback requires a DataLoader.'
-        dataset = train_loader.dataset
-        assert isinstance(
-            dataset,
-            StreamingDataset,
-        ), 'CurriculumLearning callback requires a StreamingDataset.'
-        if self.saved_dataset_index < self.dataset_index:
-            # Ignore the dataset state that was read in from the checkpoint, and
-            # replace with the new dataset state. This preserves resumption info.
-            if self.current_dataset_state['epoch'] < 0:
-                # Make sure the epoch in the loaded state dict is not negative.
-                # Since `__iter__` has not yet been called on the dataset, the
-                # epoch index in the dataset will still be -1. We need to ensure
-                # that we set the epoch correctly to 0 in this case.
-                self.current_dataset_state['epoch'] = 0
-            dataset.load_state_dict(  # type: ignore
-                self.current_dataset_state)
-            # Start a new epoch since we are using a new dataset.
-            # This will also reset the sample_in_epoch written to checkpoint,
-            # making sure that subsequent resumptions proceed correctly.
-            state.timestamp = state.timestamp.to_next_epoch()
-            # Append the new dataset config to the list of all dataset configs.
-            self.all_dataset_configs.append(self.current_dataset_config)
-        elif self.dataset_index == 0 and len(self.all_dataset_configs) == 0:
-            # Make sure to track our current dataset config if we are just starting training.
-            self.all_dataset_configs.append(self.current_dataset_config)
-
-    def state_dict(self):
-        return {
-            'dataset_index': self.dataset_index,
-            'all_dataset_configs': self.all_dataset_configs,
-        }
-
-    def load_state_dict(self, state: Dict[str, Any]):
-        self.saved_dataset_index = state.get('dataset_index', 0)
-        self.all_dataset_configs = state.get('all_dataset_configs', [])
+        if 'train_loader' not in datamix:
+            raise ValueError('Each datamix must have a train_loader.')
--- a/tests/callbacks/test_curriculum_learning_callback.py
+++ b/tests/callbacks/test_curriculum_learning_callback.py
@@ -1,14 +1,283 @@
 # Copyright 2024 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
+from contextlib import nullcontext
+from typing import Any, Callable, Optional
+from unittest.mock import MagicMock
+
+import pytest
+from composer.core import State
+from composer.core.time import Time, TimeUnit
+from composer.devices import DeviceCPU
+from composer.loggers import Logger
+from omegaconf import OmegaConf as om
+from torch.utils.data import DataLoader
+
+from llmfoundry.data.text_data import StreamingTextDataset
 from llmfoundry.utils.builders import build_callback
 
 
-def test_curriculum_learning_callback_builds():
-    kwargs = {'dataset_index': 0}
+@pytest.mark.parametrize(
+    'datamix,duration',
+    [
+        (None, '1ep'),
+        ({
+            'dataset': 'some_dataset',
+        }, '1ep'),
+        (None, '10tok'),
+        (None, ''),
+        ({}, '1ep'),
+    ],
+)
+def test_curriculum_learning_callback_init(
+    datamix: Optional[dict[str, Any]],
+    duration: str,
+    tiny_ft_dataloader_cfg: dict[str, Any],
+):
+    test_cfg = _get_test_cfg()
+    test_cfg['train_loader'] = tiny_ft_dataloader_cfg
+    train_loader = test_cfg['train_loader'] if datamix is None else datamix
+    kwargs = {
+        'schedule': [{
+            'duration': duration,
+            'train_loader': train_loader,
+        }, {
+            'duration': '2ep',
+            'train_loader': {},
+        }],
+    }
+    if duration == '':
+        del kwargs['schedule'][0]['duration']
+    if datamix is not None and len(datamix) == 0:
+        del kwargs['schedule'][0]['train_loader']
+
+    context = nullcontext()
+    if datamix is not None or duration == '':
+        context = pytest.raises(ValueError)
+    with context:
+        callback = build_callback(
+            'curriculum_learning',
+            kwargs=kwargs,
+            train_config=test_cfg,
+        )
+        assert callback is not None
+
+
+@pytest.mark.parametrize('duration', ['1ep', '10tok', '2ep'])
+def test_curriculum_learning_callback_before_load(
+    duration: str,
+    build_tiny_mpt: Callable,
+):
+    model = build_tiny_mpt(loss_fn='torch_crossentropy')
+    state = State(
+        model=model,
+        rank_zero_seed=0,
+        run_name='test_state',
+        device=DeviceCPU(),
+    )
+    state.max_duration = '3ep'
+    dl_mock = MagicMock(spec=DataLoader)
+    dl_mock.dataset = MagicMock(spec=StreamingTextDataset)
+    state.train_dataloader = dl_mock
+    logger = Logger(state)
+
+    test_cfg = _get_test_cfg()
+    kwargs = {
+        'schedule': [{
+            'duration': duration,
+            'train_loader': test_cfg['train_loader'],
+        }, {
+            'duration': '2ep',
+            'train_loader': test_cfg['train_loader'],
+        }],
+    }
+
     callback = build_callback(
         'curriculum_learning',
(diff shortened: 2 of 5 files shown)


snarayan21:
@b-chu about the new API, couple questions:
```
train_loader:
  <some params>
callbacks:
  curriculum_learning:
    duration: 5000000tok
    schedule:
    - duration: 5000000tok
      train_loader:
        <some params>
    - duration: 5000000tok
      train_loader:
        <some params>
```
1. so I still have to specify `train_loader` as a top-level entry?
2. the first `duration` specified is for the top-level `train_loader`?

mvpatel2000 (commented):
This needs a composer release first, right?

b-chu:
Yes, this needs a composer release. I'll rerun cicd after that release and before merging. Yes, train_loader is specified still and curriculum_learning.duration is its duration. We discussed offline with data team and they'll try the callback later when doing a longer training run. I think there's slight discrepancies in rng when running on interactive, but comparing to a run with no CL callback, the new callback matches the loss exactly while the old callback is slightly different. Also when comparing two different datasets/splits the loss is much greater than the plots above.
image

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,189 @@ class CurriculumLearning(CallbackWithConfig):
...
-    def __init__(self, train_config: Dict, dataset_index: int):
-        self.dataset_index = dataset_index
-        self.saved_dataset_index = 0
-        self.all_dataset_configs = []
-        self.current_dataset_state = {}
-        # The current dataset config is resolved and passed in train.py
-        self.current_dataset_config = train_config['train_loader']
+    def __init__(
+        self,
+        train_config: dict[str, Any],
```
mvpatel2000, line 45: is this still necessary? since now the callback stores the entire state of all phases

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,189 @@ class CurriculumLearning(CallbackWithConfig):
...
+        duration: Union[str, int, Time],
+        schedule: list[dict[str, Any]],
+    ):
+        from llmfoundry.utils.builders import build_tokenizer
+        from llmfoundry.utils.config_utils import calculate_batch_size_info
+        non_positive_error = ValueError('The duration must be positive.')
+        unit_error = ValueError(
+            'Schedules can only be defined in terms of epochs or tokens.',
```
mvpatel2000, line 53: imo copy pasting twice makes code more legible

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,189 @@ class CurriculumLearning(CallbackWithConfig):
...
-        # The current dataset config is resolved and passed in train.py
-        self.current_dataset_config = train_config['train_loader']
+    def __init__(
+        self,
+        train_config: dict[str, Any],
+        duration: Union[str, int, Time],
+        schedule: list[dict[str, Any]],
+    ):
+        from llmfoundry.utils.builders import build_tokenizer
+        from llmfoundry.utils.config_utils import calculate_batch_size_info
```
mvpatel2000, line 50: why does this have to be in-line?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,189 @@ class CurriculumLearning(CallbackWithConfig):
...
+        non_positive_error = ValueError('The duration must be positive.')
+        unit_error = ValueError(
+            'Schedules can only be defined in terms of epochs or tokens.',
+        )
+
+        # Ensure all duration values are positive
+        # Ensure all duration units are in epochs or  tokens
```
mvpatel2000, line 57:
```suggestion
        # Ensure all duration units are in epochs or  tokens
```
```suggestion
        # Ensure all duration values are positive and in epochs or tokens
```

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,189 @@ class CurriculumLearning(CallbackWithConfig):
...
+        self.device_train_batch_size, _, _ = calculate_batch_size_info(
+            train_config['global_train_batch_size'],
+            train_config['device_train_microbatch_size'],
+            data_replication_degree=1,
+        )
+
+        # Copied from scripts/train/train.py
+        tokenizer_name = train_config['tokenizer']['name']
+        tokenizer_kwargs = train_config['tokenizer'].get('kwargs', {})
+        self.tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
```
mvpatel2000, line 90: building tokenizer seems bad? shouldnt we just accept it as arg in init?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,189 @@ class CurriculumLearning(CallbackWithConfig):
...
+        for datamix in self._schedule:
+            if datamix['duration'].unit != state.max_duration.unit:
+                units_match = False
+        if not units_match:
+            raise ValueError((
+                'All durations in the schedule must have the same units as '
+                'the max_duration.'
+            ))
+
+        # Ensure schedule duration is greater than max_duration
```
mvpatel2000, line 109: why greater and not = ?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,189 @@ class CurriculumLearning(CallbackWithConfig):
...
+            'schedule_index': self._schedule_index,
+        }
+
+    def load_state_dict(self, state: dict[str, Any]):
+        # Ensure that the schedule has not changed on already trained datamixes
+        assert self._duration == state['duration']
+        for idx in range(state['schedule_index'] + 1):
+            assert self._schedule[idx] == state['schedule'][idx]
+
+        self._schedule_index = state['schedule_index']
```
mvpatel2000, line 199: this breaks if you load old version of callback, right?

mvpatel2000 (commented):
first pass, some high level questions

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: If you have it as an arg in init, then that means the user needs to provide the tokenizer in the callback section of the yaml and you still need to build the tokenizer. There's not way to pass the tokenizer into the callback outside of init.

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: This is needed to get the tokenizer config to build the tokenizer (which is explained more in the comment below).

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: There's circular dependencies that this resolves without having uglier package trees

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Changed

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Outdated

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Yes, but the previous version is experimental so I'm fine breaking BC. I don't mind raising an error instead of the generic "dataset_index not found in state" though

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,23 +7,29 @@
...
 import logging
-from typing import Any, Dict
+from typing import Any, Union
 
-from composer.core import State
-from composer.loggers import Logger
+from composer import DataSpec
+from composer.core import State, Time, TimeUnit, ensure_time
+from composer.loggers import Logger, MosaicMLLogger
+from composer.trainer.trainer import _get_initial_device_train_microbatch_size
```
mvpatel2000, line 17: should we make this fn public?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
+            state.set_dataloader(
+                dataloader=data_spec.dataloader,
+                dataloader_label='train',
+            )
+            # state.train_dataloader = state.dataloader # FIXME
+            state.device_train_microbatch_size = _get_initial_device_train_microbatch_size(
+                state.device_train_microbatch_size,
+                state.auto_microbatching,
+                state.train_dataloader,
+            )
```
mvpatel2000, line 155: why do we need this? why is it a fn of dataloader?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
+            datamix = copy.deepcopy(self._schedule[self._schedule_index])
+            data_spec = self._build_train_loader(
+                train_loader_config=datamix['train_loader'],
+                logger=logger,
+            )
+            state.set_dataloader(
+                dataloader=data_spec.dataloader,
+                dataloader_label='train',
+            )
+            # state.train_dataloader = state.dataloader # FIXME
```
mvpatel2000, line 150: FIXME?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
-        self.current_dataset_state = {}
-        # The current dataset config is resolved and passed in train.py
-        self.current_dataset_config = train_config['train_loader']
+    def __init__(
+        self,
+        train_config: dict[str, Any],
+        schedule: list[dict[str, Any]],
+    ):
+        from llmfoundry.utils.builders import build_tokenizer
+        from llmfoundry.utils.config_utils import calculate_batch_size_info
```
mvpatel2000, line 49: Plz put a comment why inline import is needed?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
+
+        # Ensure all duration values are positive
+        # Ensure all duration units are in epochs or tokens
```
mvpatel2000, line 52:
```suggestion
        # Ensure all duration values are positive and are in epochs or tokens
```

llmfoundry/callbacks/curriculum_learning_callback.py
milocress, line 150: What's with this

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
+        self._schedule = schedule
+        if len(self._schedule) == 0:
+            raise ValueError('The schedule must have at least one datamix.')
+        for index, datamix in enumerate(self._schedule):
+            self._validate_data_mix(datamix)
```
mvpatel2000, line 57: This fn is called `_validate_datamix` in the code

milocress, line 57: > this function doesn't seem to exist - @mvpatel2000

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
+
+            if (
+                index == 0 and
+                train_config['train_loader'] != datamix['train_loader']
+            ):
+                raise ValueError((
+                    'The first datamix in the schedule must match the '
+                    'train_loader in the train_config.'
+                ))
```
mvpatel2000, line 66: so you have to duplicate the datamix in the yaml...?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
+        self.device_train_batch_size, _, _ = calculate_batch_size_info(
+            train_config['global_train_batch_size'],
+            train_config['device_train_microbatch_size'],
+            data_replication_degree=1,
+        )
+
+        # Copied from scripts/train/train.py
+        tokenizer_name = train_config['tokenizer']['name']
+        tokenizer_kwargs = train_config['tokenizer'].get('kwargs', {})
+        self.tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
```
mvpatel2000, line 80: Why is this needed on init? Can we not pull this on `Event.INIT`?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -34,20 +40,188 @@ class CurriculumLearning(CallbackWithConfig):
...
+                    'train_loader in the train_config.'
+                ))
+
+        self._schedule_index = 0
+
+        # Copied from llmfoundry/utils/config_utils.py
+        self.device_train_batch_size, _, _ = calculate_batch_size_info(
+            train_config['global_train_batch_size'],
+            train_config['device_train_microbatch_size'],
+            data_replication_degree=1,
```
mvpatel2000, line 74: why is this needed on `__init__` instead of being pulled on `Event.INIT`?

mvpatel2000 (commented):
First pass, will rereview once questions are answered

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Removed

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Removed

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Yes, the first datamix is a duplicate of the one in train_loader. This was the outcome of the decision with the motivation being that having a lone duration in the callback is less intuitive than having the datamix duplicated

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: You can do this during Event.INIT but then you'd need to save the train_config in the callback during init. It seems easier to do this during init

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Same reasoning as above

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,223 @@
...
+                'increment the iteration.'
+            ))
+            self._schedule_index += 1
+            state.timestamp = state.timestamp.to_next_iteration()
+
+    def iteration_start(self, state: State, logger: Logger):
+        # Reset and initialize state train dataloader
+        log.warning(
+            'trainer._train_data_spec should be updated whenever the dataloader is updated',
+        )
```
mvpatel2000, line 143: Unideal to emit a warning every iteration start. This should first check if we actualyl update dataloader... and even then I dont think this should be warning, maybe info?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,223 @@
...
+
+        # Swap the dataset if starting a new iteration that's not the original datamix
+        if self._schedule_index > 0:
+            clean_stale_shared_memory()
```
mvpatel2000, line 147: why is this needed? can we leave a comment?

llmfoundry/callbacks/curriculum_learning_callback.py
mvpatel2000: Why is it easier during __init__?

mvpatel2000 (commented):
First pass, mostly lgtm but some minor comments.

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: You get passed the train_config during init. If you want to do it during Event.init then you need to save the train_config as a class attribute in init then access it during Event.init to create more class attributes. It seems better to just create it in init and not have to save train_config.

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+        # If checkpoint was saved before iteration was incremented, we need to increment it now
+        if ((
+            self._schedule[self._schedule_index]['duration'].unit
+            == TimeUnit.TOKEN and state.timestamp.token_in_iteration
+            >= self._schedule[self._schedule_index]['duration'].value
+        ) or (
+            self._schedule[self._schedule_index]['duration'].unit
+            == TimeUnit.EPOCH and state.timestamp.epoch_in_iteration
+            >= self._schedule[self._schedule_index]['duration'].value
+        )):
```
mvpatel2000, line 131: why is this only token or epoch? why not batch?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+            'schedule': self._schedule,
+            'schedule_index': self._schedule_index,
+        }
+
+    def load_state_dict(self, state: dict[str, Any]):
+        self._schedule_index = state['schedule_index']
+
+        # Ensure that the schedule has not changed on previously trained datamixes
+        for idx in range(state['schedule_index']):
+            assert self._schedule[idx] == state['schedule'][idx]
```
mvpatel2000, line 182: can we instead do if + runtime error giving a helpful msg to user?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+
+        # Ensure that the datamix has not changed on the current datamix
+        current_loader = self._schedule[self._schedule_index]['train_loader']
+        saved_loader = state['schedule'][self._schedule_index]['train_loader']
+        assert current_loader == saved_loader
```
mvpatel2000, line 187: same comment on error

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+
+        # Ensure that the current datamix duration is greater than timestamp
+        duration = self._schedule[self._schedule_index]['duration']
+        assert isinstance(duration.unit, (TimeUnit.TOKEN, TimeUnit.EPOCH))
```
mvpatel2000, line 191: doesn't support batches or samples? same comment on error

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+            ))
+
+        self._validate_dataloader(state.train_dataloader)
+
+    def after_load(self, state: State, logger: Logger):
+        del logger
+
+        self._validate_dataloader(state.train_dataloader)
+
+        # If checkpoint was saved before iteration was incremented, we need to increment it now
```
mvpatel2000, line 122: how would this be possible?

mvpatel2000 (commented):
second pass, lgtm besides a few minor comments. requiring review from @milocress

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Design decision

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: You save a checkpoint at EPOCH_CHECKPOINT but the program crashes before ITERATION_END

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Design decision

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: I don't think this is critical to application logic since behavior will be exactly the same with and without assert
https://google.github.io/styleguide/pyguide.html#244-decision

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Same as above

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Same as above for error

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
-        # if we are resuming with a new dataset.
-        train_loader = state.train_dataloader
+        # Ensure all duration units are the same as max_duration
+        units_match = True
+        assert state.max_duration is not None, 'max_duration should have beeen set.'
+        for datamix in self._schedule:
+            if datamix['duration'].unit != state.max_duration.unit:
+                units_match = False
+        if not units_match:
+            raise ValueError((
```
dakinggg, line 99: would be nicer to gather all the units and put the max duration unit and all the datamix units in the error so that you don't have to iterate multiple times with each data mix

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+        if schedule_duration != state.max_duration:
+            raise ValueError((
+                'The sum of all durations in the schedule must be equal to the '
+                'max_duration.'
+            ))
+
+        self._validate_dataloader(state.train_dataloader)
+
+    def after_load(self, state: State, logger: Logger):
+        del logger
```
milocress, line 118: why are we deleting the logger?

tests/callbacks/test_curriculum_learning_callback.py
```diff
@@ -1,14 +1,28 @@
 # Copyright 2024 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
+from omegaconf import OmegaConf as om
+
 from llmfoundry.utils.builders import build_callback
 
 
 def test_curriculum_learning_callback_builds():
```
dakinggg, line 9: given that we are making it no longer experimental, its complete functionality should be unit tested

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+    def iteration_end(self, state: State, logger: Logger):
+        del state, logger  # unused
+
+        self._schedule_index += 1
+
+    def state_dict(self):
+        return {
+            'schedule': self._schedule,
+            'schedule_index': self._schedule_index,
+        }
```
milocress, line 175: Have you tested resumption with this `state_dict` method?

milocress (approved):
LGTM, just a couple questions

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+            ):
+                raise ValueError((
+                    'The first datamix in the schedule must match the '
+                    'train_loader in the train_config.'
+                ))
+
+        self._schedule_index = 0
+
+        # Copied from llmfoundry/utils/config_utils.py
+        self.device_train_batch_size, _, _ = calculate_batch_size_info(
```
dakinggg, line 78: shouldn't this just the batch size info from the config itself?

llmfoundry/callbacks/curriculum_learning_callback.py
dakinggg: this should be an error not an assert (https://github.com/mosaicml/composer/blob/dev/STYLE_GUIDE.md#4-use-of-assert). behavior is not unchanged without the assert, it would silently do something bad (or crash somewhere else, not certain)

dakinggg: as with the other places

llmfoundry/callbacks/curriculum_learning_callback.py
dakinggg: https://github.com/mosaicml/llm-foundry/blob/4b1fecb4f0a5a59f6b86fb8aa7bec58e759ef172/llmfoundry/utils/config_utils.py#L103

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: unused, added comment

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: That's only in the "nice" config which we don't have access to

b-chu: We have to recalculate since we only have access to the yaml config

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
dakinggg: I think that was a mistake..can you fix that? having this logic in two places is not good for maintainability and also makes it not extendable

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,225 @@
...
+            ))
+            self._schedule_index += 1
+            state.timestamp = state.timestamp.to_next_iteration()
+
+    def iteration_start(self, state: State, logger: Logger):
+
+        # Swap the dataset if starting a new iteration that's not the original datamix
+        if self._schedule_index > 0:
+            # Reset and initialize state train dataloader
+            log.info(
```
dakinggg, line 144: agree with mihir, this should not emit every iteration. also maybe im missing something, but what useful information is this communicating?  its just like a comment about the internals of the callback

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: The callback doesn't use previous indices in the schedule so it wouldn't change its behavior. If the user changes the previous datamixes, the callback resumes on the current datamix from the checkpoint and ignores everything previously in the schedule.

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
dakinggg: discussed offline, the current config should have the info needed here

b-chu: Done

tests/callbacks/test_curriculum_learning_callback.py
b-chu: Done

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,228 @@
...
+                train_config['train_loader'] != datamix['train_loader']
+            ):
+                raise ValueError((
+                    'The first datamix in the schedule must match the '
+                    'train_loader in the train_config.'
+                ))
+
+        self._schedule_index = 0
+
+        # Copied from llmfoundry/utils/config_utils.py
```
dakinggg, line 77:
```suggestion
```

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,228 @@
...
+        self.device_train_batch_size = train_config['device_train_batch_size']
+
+        # Copied from scripts/train/train.py
```
dakinggg, line 80: the tokenizer should be extracted from state.model wherever it is needed, no?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,228 @@
...
+                'The sum of all durations in the schedule must be equal to the '
+                'max_duration.'
+            ))
+
+        self._validate_dataloader(state.train_dataloader)
+
+    def after_load(self, state: State, logger: Logger):
+        del logger  # unused
+
+        self._validate_dataloader(state.train_dataloader)
```
dakinggg, line 114: is there a reason we validated dataloader both before and after load?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,228 @@
...
+            self._schedule[self._schedule_index]['duration'].unit
+            == TimeUnit.TOKEN and state.timestamp.token_in_iteration >=
+            self._schedule[self._schedule_index]['duration'].value
+        ) or (
+            self._schedule[self._schedule_index]['duration'].unit
+            == TimeUnit.EPOCH and state.timestamp.epoch_in_iteration >=
+            self._schedule[self._schedule_index]['duration'].value
+        )):
+            log.error((
+                'The CurriculumLearning callback has detected that the previous run did not correctly '
```
dakinggg, line 127: This shouldn't be an error right? cause the code here is self healing. just a warning? maybe im misunderstanding.

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,228 @@
...
+            'tokenizer' keys.
+        schedule (list[dict[str, Any]]): The list of datamixes to use and their
+            durations. Duration units must match max_duration and be in terms of
+            a TimeUnit that is supported by Iteration. The duration values must
+            be positive. There must be at least one datamix in the schedule. The
+            first datamix in the schedule must match the train_loader in the
+            train_config. On resumption, previously trained on datamixes and
+            durations cannot be changed. The duration of the current datamix
+            must be greater than the saved timestamp. The dataset must be a
+            StreamingDataset.
```
dakinggg, line 48: please include an example schedule config in the docstring.

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,228 @@
...
+                'The CurriculumLearning callback has detected that the previous run did not correctly '
+                'increment the iteration.'
+            ))
+            self._schedule_index += 1
+            state.timestamp = state.timestamp.to_next_iteration()
+
+    def iteration_start(self, state: State, logger: Logger):
+        # Swap the dataset if starting a new iteration that's not the original datamix
+        if self._schedule_index > 0:
+            # TODO: trainer._train_data_spec should be updated whenever the dataloader is updated
```
dakinggg, line 136: I don't quite understand this todo, could you jira with more details?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -7,47 +7,228 @@
...
+        )
+        try:
+            return build_dataloader(
+                train_loader_config,
+                self.tokenizer,
+                self.device_train_batch_size,
+            )
+        except BaseContextualError as e:
+            for destination in logger.destinations:
+                if isinstance(destination, MosaicMLLogger):
```
dakinggg, line 226: please make sure to merge with https://github.com/mosaicml/llm-foundry/pull/1301 if it goes in first

dakinggg (commented):
will approve after comments are addressed, overall lgtm

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done.

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done.

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done.

llmfoundry/callbacks/curriculum_learning_callback.py
dakinggg: #1301 is in, unresolving

dakinggg (approved):

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Merged.

PR #1267, opened 2024-06-07
Revert to older TE version

mvpatel2000:
Revert to older TE version. We're seeing some issues with flash-attn mismatch.


mvpatel2000:
Force merging after approval from @dakinggg

PR #1277, opened 2024-06-13
Fix packing + streaming + resumption

dakinggg:
There was a long standing bug when resuming a streaming, finetuning dataset with packing enabled. The issue would only occur if you resumed _not_ on an epoch boundary. The issue would be fairly significant, in that resumption would resume at an earlier sample than it should have, so data would be unexpectedly repeated. The root cause of the issue is that, when packing, the num samples recorded for streaming resumption needed to be multiplied by the packing ratio.

Manual test, with PR, resumption is deterministic:
Screenshot 2024-06-14 at 2 42 55 AM

Before this PR (note, loss is lower upon resumption because its repeating samples, and then it reverts to the same general loss curve once its through the repeats):
Screenshot 2024-06-14 at 11 24 59 AM

The above plots were with `num_workers=0`, when setting `num_workers=8`, we discovered a second deterministic resumption bug, where the wrong batch size was passed to streaming. This has now also been fixed, and here is a manual test with `num_workers=8`.

Screenshot 2024-06-14 at 2 33 17 PM

Note: There will still be some non determinism if `max_leftover_bins_to_keep` is not set to zero, because we don't save the reservoir. This will just be slight non determinism, as opposed to the current bug which would repeat samples.

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -683,7 +683,8 @@ class StreamingFinetuningDataset(StreamingDataset):
             num_samples = self.packing_ratio * num_samples
 
         return super().state_dict(
-            num_samples=num_samples, from_beginning=from_beginning
+            num_samples=num_samples,
+            from_beginning=from_beginning,
         )
 
 
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -592,6 +592,7 @@ class StreamingFinetuningDataset(StreamingDataset):
         max_seq_len: int = 2048,
         allow_unsafe_types: bool = False,
         replication: Optional[int] = None,
+        packing_ratio: Optional[float] = None,
         **kwargs: Any,
     ):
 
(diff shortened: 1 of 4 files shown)


llmfoundry/data/finetuning/tasks.py
```diff
@@ -644,9 +645,11 @@ def __init__(
 
         self.tokenizer = tokenizer
         self.max_seq_len = max_seq_len
+        self.packing_ratio = packing_ratio
 
     # How to process a sample
     def __getitem__(self, idx: int) -> Dict[str, Any]:
+        print(f'idx: {idx}')
```
snarayan21, line 652: remove?

dakinggg, line 652: Will remove before merging

snarayan21 (approved):
thanks for addressing this! minor nit but otherwise lgtm

llmfoundry/data/finetuning/tasks.py
snarayan21: oop just saw this

mvpatel2000 (approved):
Thanks for hunting this down!

PR #1278, opened 2024-06-13
Make expandable segments on by default

b-chu:
We almost always use expandable segments to help with memory issues


snarayan21 (approved):
dope

mvpatel2000 (approved):

PR #1280, opened 2024-06-13
Fix TE HF checkpoint saving

j316chuck:
# Description 
Fixes HF Checkpoint callback for TransformerEngine FP8 saving. This PR ensures we serialize the io.BytesIO extra_state tensors as regular tensors in`save_pretrained` so the code does not error. 

# Tests
- Added unit test, skipped on A100 GPU ✔️ 
- Added unit test, manually ran on H100 GPU ✅ 
```
tests/a_scripts/inference/test_convert_composer_to_hf.py::test_huggingface_conversion_callback[1ba-1ba-1ba-1-1-amp_fp8-full-mpt-True-None]
  /usr/lib/python3/dist-packages/transformer_engine/pytorch/module/base.py:394: UserWarning: The given buffer is not writable, and PyTorch does not support non-writable tensors. This means you can write to the underlying (supposedly non-writable) buffer using the tensor. You may want to copy the buffer to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at ../torch/csrc/utils/tensor_new.cpp:1524.)
    state_serialized = torch.frombuffer(pickle.dumps(state), dtype=torch.uint8)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
===================================================================================== 25 passed, 11 skipped, 1621 deselected, 266 warnings in 92.95s (0:01:32) ======================================================================================
Waiting up to 30 seconds for all training processes to terminate. Press Ctrl-C to exit immediately.
```
- Before: `failed-hf-checkpointer-fp8-llama3-8b-metamath-4ep-KOTaOP` 🔴 
- After: `success-hf-checkpointer-fp8-llama3-8b-metamath-4ep-yxNFTK` ✅ 

# Issues
Closes https://databricks.atlassian.net/browse/RGENAI-255

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -37,6 +37,12 @@ from llmfoundry.models.utils import init_empty_weights
 from llmfoundry.utils.huggingface_hub_utils import \
     edit_files_for_hf_compatibility
 
+try:
+    import transformer_engine.pytorch as te
+    is_te_imported = True
+except ModuleNotFoundError:
+    is_te_imported = False
+
 log = logging.getLogger(__name__)
 
 __all__ = ['HuggingFaceCheckpointer']
@@ -486,10 +492,18 @@ class HuggingFaceCheckpointer(Callback):
             )
 
             log.debug('Saving Hugging Face checkpoint to disk')
-            new_model_instance.save_pretrained(temp_save_dir)
-            if original_tokenizer is not None:
-                assert isinstance(original_tokenizer, PreTrainedTokenizerBase)
-                original_tokenizer.save_pretrained(temp_save_dir)
+            # This context manager casts the TE extra state in io.BytesIO format to tensor format
+            # Needed for proper hf ckpt saving.
+            context_manager = te.onnx_export(
+                True
+            ) if is_te_imported else contextlib.nullcontext()
+            with context_manager:
+                new_model_instance.save_pretrained(temp_save_dir)
+                if original_tokenizer is not None:
+                    assert isinstance(
+                        original_tokenizer, PreTrainedTokenizerBase
+                    )
+                    original_tokenizer.save_pretrained(temp_save_dir)
 
             # Only need to edit files for MPT because it has custom code
             if original_model.config.model_type == 'mpt':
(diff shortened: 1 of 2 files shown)


mvpatel2000 (commented):
Can i load a TE ckpt into a non TEd model?

LGTM but also @dakinggg  wdyt

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -486,10 +492,19 @@ def dtensor_to_tensor_hook(
             )
 
             log.debug('Saving Hugging Face checkpoint to disk')
-            new_model_instance.save_pretrained(temp_save_dir)
-            if original_tokenizer is not None:
-                assert isinstance(original_tokenizer, PreTrainedTokenizerBase)
-                original_tokenizer.save_pretrained(temp_save_dir)
+            # This context manager casts the TE extra state in io.BytesIO format to tensor format
+            # Needed for proper hf ckpt saving.
+            context_manager = te.onnx_export(
```
dakinggg, line 497: is there a more granular check we can do to decide whether to wrap in onnx export context manager? I'd prefer to only wrap if the model _actually_ needs it. not just if te is importable.

j316chuck: added precision state check

j316chuck:
@mvpatel2000 I don't know let me try. But curious the use case in which you would do that?

dakinggg (commented):
will review fully once CI passes

mvpatel2000 (commented):
LGTM but same comment on waiting for CI/CD topass

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -486,10 +492,20 @@ def dtensor_to_tensor_hook(
...
+                True,
+            ) if is_te_imported and state.precision == Precision.AMP_FP8 else contextlib.nullcontext(
+            )
+            with context_manager:
+                new_model_instance.save_pretrained(temp_save_dir)
+                if original_tokenizer is not None:
```
dakinggg, line 503: the tokenizer doesn't need to be in the context manager right?

dakinggg (approved):

PR #1281, opened 2024-06-14
Adds CI for torch 2.3.1

dakinggg:
Adds CI for torch 2.3.1. After this, there will be another PR to bump the minimum foundry version and remove support for torch 2.3.0


mvpatel2000 (approved):
should we remove the other one?

PR #1283, opened 2024-06-14
Add optional logging of text output to EvalOutputLogging

sjawhar:
Adds the ability to log text output using `EvalOutputLogging` in non-metrics/ICL use cases. This appears as a new `outputs` key in the logged dictionary, which is set to `stats.outpus` and de-tokenized when the model is a `HuggingFaceModel` and the dataset has a tokenizer.


llmfoundry/callbacks/eval_output_logging_callback.py
```diff
@@ -48,7 +48,7 @@ def eval_batch_end(self, state: State, logger: Logger) -> None:
                                       )
 
         # If batch mode is not generate, outputs will be logits
-        if state.batch['mode'] == 'generate':
+        if state.batch.get('mode') == 'generate':
```
sjawhar, line 51: Should the default assumption be that it is indeed "generate"?

sjawhar: Nope! Tested myself and no good.

mvpatel2000:
@maxisawesome can you please take a look? 

CC: @dakinggg

sjawhar:
image
YES!! It's finally working! The logging, that is. Not my fine-tuning, that's still garbage. But I can finally start debugging!

llmfoundry/callbacks/eval_output_logging_callback.py
```diff
@@ -47,28 +47,31 @@ def eval_batch_end(self, state: State, logger: Logger) -> None:
...
             # Outputs are already detokenized
             logging_dict['outputs'] = state.outputs
+        elif isinstance(state.outputs, torch.Tensor):
+            # If batch mode is not generate, outputs will be logits
+            logging_dict['outputs'] = state.outputs.argmax(dim=-1)
 
         input_ids = state.batch['input_ids']
         logged_input = []
         assert state.dataloader is not None
+        dataset = state.dataloader.dataset  # pyright: ignore[reportGeneralTypeIssues]
```
maxisawesome, line 61: This makes things much more legible but also doubles the dataset in memory 😔. Most eval datasets are small and it wouldn't matter much, but we have run into problems when people using surprisingly large eval datasets before, so I don't think this is a good idea in the end.

maxisawesome:
Looks great other than the one comment I left! Thanks for fixing this for non-generative evals.

llmfoundry/callbacks/eval_output_logging_callback.py
sjawhar: Is this a gc issue? Isn't the variable already defined in the `for input_list in` loop?

mvpatel2000: Discussed offline, resolving -- should not result in a copy since it stores a pointer.

mvpatel2000:
@sjawhar if you can fix lint + unit tests (the CPU ones) that would be awesome! then we can merge

tests/callbacks/test_eval_output_logging_callback.py
```diff
@@ -193,23 +205,27 @@ def test_eval_output_logging_lm(
...
     # We use the same data in each batch
-    assert json.loads(in_memory_logger.tables[f'lm_acc_step_0'])['data'] == [
+    assert logged_data['data'] == [
         [
             'The dog is',
             ' furry',
             ' furry',
             1,
             'InContextLearningLMAccuracy',
+            RegexMatcher(r' dog is furry(\[PAD\])+I'),
```
sjawhar, line 228: So I'm not sure you'll like this approach :sweat_smile: 

I actually was surprised to find `output` already in the logged data, since it's not in `eval_output_logger`. Is that from the ICL metric/eval itself? It makes the new `outputs` field look superfluous (though also possibly not, since `outputs` is like a `raw_output`). This pattern matching here to pass the test also feels a bit unprincipled.

sjawhar:
Anything else I can do to help get this merged?

dakinggg:
@sjawhar sorry about that, will take a look this week!

tests/callbacks/test_eval_output_logging_callback.py
dakinggg: as for the outputs already being there, thats cause of https://github.com/mosaicml/llm-foundry/blob/21c9e0a58a68b2ddb7cc7bea40a11c9add85d2c9/llmfoundry/callbacks/eval_output_logging_callback.py#L53 right?

sjawhar: Yes, I expected `outputs` (with an "s"), but not `output` (not with an "s"). Don't know where that's coming from.

llmfoundry/callbacks/eval_output_logging_callback.py
```diff
@@ -47,28 +47,34 @@ def eval_batch_end(self, state: State, logger: Logger) -> None:
                                           state.metric_outputs,
                                       )
 
-        # If batch mode is not generate, outputs will be logits
-        if state.batch['mode'] == 'generate':
+        if state.batch.get('mode') == 'generate':
             # Outputs are already detokenized
             logging_dict['outputs'] = state.outputs
+        elif isinstance(state.outputs, torch.Tensor):
```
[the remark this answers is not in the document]
sjawhar: I gave it a go

tests/callbacks/test_eval_output_logging_callback.py
dakinggg: oh i see what you mean. Yeah, the logging dict starts off as a deep copy of whatever the metric outputs were (https://github.com/mosaicml/llm-foundry/blob/88511f780c7437e89cdaf96a1771aac41c2a64a5/llmfoundry/callbacks/eval_output_logging_callback.py#L45-L48)

llmfoundry/callbacks/eval_output_logging_callback.py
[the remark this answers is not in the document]
dakinggg: possibly forgot to push this change?

dakinggg (approved):
LGTM, please update the PR title and description to reflect the changes in this PR. Thank you!

sjawhar:
> LGTM, please update the PR title and description to reflect the changes in this PR. Thank you!

Done

dakinggg:
@sjawhar looks like the tests failed with the recent change

llmfoundry/callbacks/eval_output_logging_callback.py
```diff
@@ -24,51 +25,85 @@ class EvalOutputLogging(Callback):
...
+        **kwargs: Any,
+    ):
         super().__init__(self, *args, **kwargs)
         self.log_tokens = log_tokens
         self.columns = None
         self.name = None
         self.rows = []
+        self.log_output_text = log_output_text
+
+    def init(self, state: State, logger: Logger) -> None:
```
dakinggg, line 42: to fix the test I think you just need to call `init` with some mock inputs and then it should be fine

sjawhar: Done

PR #1287, opened 2024-06-14
Add missing dependency group

dakinggg:
The Composer `mlflow` dependency group contains more than just mlflow, so we need to install the whole dep group.


snarayan21 (commented):
should we create a separate optional dep group in composer that contains only what foundry needs? or does foundry need all of the packages under `'mlflow'`?

dakinggg:
@snarayan21 foundry needs the whole dep group

snarayan21 (approved):
lgtm then

mvpatel2000 (approved):

PR #1292, opened 2024-06-20
Removing logging exception through update run metadata

jjanezhang:
We are moving to catching exceptions through the python excepthook override 7/8/24. We can remove logging to metadata then.


scripts/train/train.py
```diff
@@ -427,9 +425,7 @@ def main(cfg: DictConfig) -> Trainer:
             if eval_gauntlet_callback is not None:
                 callbacks.append(eval_gauntlet_callback)
         except BaseContextualError as e:
-            if mosaicml_logger is not None:
-                e.location = EvalDataLoaderLocation
-                mosaicml_logger.log_exception(e)
+            e.location = EvalDataLoaderLocation
```
mvpatel2000, line 428: why does location need to be set if the logger is being migrated off of? should we be removing this fn from logger in composer?

jjanezhang: The location attribute is on the exception class helps us tell the user if their data validation failed for their eval dataset or their train dataset. We still need this attribute set even when we are catching exceptions without the logger.

mvpatel2000 (approved):
LGTM

mvpatel2000 (approved):
LGTM

jjanezhang:
> Member

removed the logger from a couple more places, requested review again

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -667,18 +663,10 @@ def fetch_DT(args: Namespace) -> None:
...
-
-    try:
-        w = WorkspaceClient()
-        args.DATABRICKS_HOST = w.config.host
-        args.DATABRICKS_TOKEN = w.config.token
-
-        tik = time.time()
-        fetch_DT(args)
-        log.info(f'Elapsed time {time.time() - tik}')
+    w = WorkspaceClient()
```
mvpatel2000, line 666: does this import need to be gated?

mvpatel2000 (approved):

llmfoundry/callbacks/run_timeout_callback.py
```diff
@@ -8,19 +8,18 @@
...
 
-def _timeout(timeout: int, mosaicml_logger: Optional[MosaicMLLogger] = None):
+def _timeout(timeout: int):
     log.error(f'Timeout after {timeout} seconds of inactivity after fit_end.',)
-    if mosaicml_logger is not None and no_override_excepthook():
-        mosaicml_logger.log_exception(RunTimeoutError(timeout=timeout))
-    os.kill(os.getpid(), signal.SIGINT)
+    try:
+        raise RunTimeoutError(timeout=timeout)
+    except RunTimeoutError:
```
irenedea, line 21: this seems like a noop? Can we remove this?

irenedea (approved):
LGTM! minor comment

scripts/data_prep/convert_delta_to_json.py
jjanezhang: I don't think so, we aren't currently in gating it in the script.

PR #1296, opened 2024-06-21
Fix grad accum typing

dakinggg:
Fixes the typing on the config. It wasn't causing any errors because its just for logging, but it was getting incorrectly cast to a string.


snarayan21 (approved):
noice

PR #1298, opened 2024-06-21
Allow passing in lbl_process_group directly

dakinggg:
Allows `ffn_config['lbl_process_group']` to be direct constructed instead of just using string mappings.

### DIFF
--- a/tests/models/utils/test_config_moe_args.py
+++ b/tests/models/utils/test_config_moe_args.py
@@ -1,15 +1,24 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
 from typing import Any
 
 import pytest
-
 from torch import distributed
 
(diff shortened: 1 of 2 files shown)


tests/models/utils/test_config_moe_args.py
```diff
@@ -0,0 +1,45 @@
...
+
+from llmfoundry.models.utils.config_moe_args import (
+    config_megablocks_moe_args,
+    get_megablocks_device_mesh,
+)
+
+
+@pytest.mark.gpu
+@pytest.mark.world_size(2)
+def test_config_megablocks_moe_args_pg():
```
snarayan21, line 17: Could you add a test for the ValueError? otherwise lgtm

snarayan21 (commented):
one test, nice otherwise

mvpatel2000 (approved):

snarayan21 (approved):

PR #1299, opened 2024-06-21
Allows interweaving of arbitrary kinds of 'attention' layers, like sliding window, reuse prev layer kv cache etc.

ShashankMosaicML:
This allows for overriding default block configs for certain layers. This should contain two sub configs: order and overrides. order specifies the order of different kinds of layers (default refers to a layer that does not apply any overrides). For each kind of layer, specify the overrides in the overrides config. For example, to specify this model (https://research.character.ai/optimizing-inference/) , the following config will be needed:
```
model:
    ...
    (usual model configs)
    ...
    block_overrides:
        order:
        - name: default
        - order:
          - name: sliding_window_layer
          - name: sliding_window_layer_reuse
          - name: sliding_window_layer
          - repeat: 2
            name: sliding_window_layer_reuse
          - name: reuse_kv_layer
          repeat: 2
        overrides:
            sliding_window_layer:
                attn_config:
                    sliding_window_size: 1024
            sliding_window_layer_reuse:
                attn_config:
                    sliding_window_size: 1024
                    reuse_kv_layer_idx: -1 # Relative index of the layer whose kv cache to reuse
            reuse_kv_layer:
                attn_config:
                    reuse_kv_layer_idx: -6 # Relative index of the layer whose kv cache to reuse
```

Also prints the following log summarizing the network:

```
INFO: llmfoundry.models.mpt.modeling_mpt: The following is a summary of overrides per layer.
  idx  name                        overrides
-----  --------------------------  ----------------------------------------------------------
    0  default                     []
    1  sliding_window_layer        [{'sliding_window_size': 1024}]
    2  sliding_window_layer_reuse  [{'sliding_window_size': 1024}, {'reuse_kv_layer_idx': 1}]
    3  sliding_window_layer        [{'sliding_window_size': 1024}]
    4  sliding_window_layer_reuse  [{'sliding_window_size': 1024}, {'reuse_kv_layer_idx': 3}]
    5  sliding_window_layer_reuse  [{'sliding_window_size': 1024}, {'reuse_kv_layer_idx': 3}]
    6  reuse_kv_layer              [{'reuse_kv_layer_idx': 0}]
    7  sliding_window_layer        [{'sliding_window_size': 1024}]
    8  sliding_window_layer_reuse  [{'sliding_window_size': 1024}, {'reuse_kv_layer_idx': 7}]
    9  sliding_window_layer        [{'sliding_window_size': 1024}]
   10  sliding_window_layer_reuse  [{'sliding_window_size': 1024}, {'reuse_kv_layer_idx': 9}]
   11  sliding_window_layer_reuse  [{'sliding_window_size': 1024}, {'reuse_kv_layer_idx': 9}]
   12  reuse_kv_layer              [{'reuse_kv_layer_idx': 0}]
```
Note that the table above prints the absolute layer index for `reuse_kv_layer_idx`


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1232,6 +1396,12 @@ def flops_per_batch(self, batch: Mapping):
         # that the dataset has been constructed without padding. Additionally, we
         # assume the backward pass is approximately 2x the forward pass
 
+        if self.model.config.block_overrides is not None:
```
vchiley, line 1399: Can you add a jira ticket to update this?
(and also I'm pretty sure we never updated this for swa... 😬 )

ShashankMosaicML: Done.

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -145,6 +177,35 @@ def __init__(
             init_config_defaults,
         )
 
+        if 'reuse_kv_layer_idx' in self.attn_config and self.attn_config[
+            'attn_impl'] == 'torch':
+            raise NotImplementedError(
+                'reusing kv cache from a previous layer is not implemented for torch attention.',
+            )
+        if block_overrides is not None:
+            warnings.warn(
```
dakinggg, line 186: You can use https://github.com/mosaicml/llm-foundry/blob/21c9e0a58a68b2ddb7cc7bea40a11c9add85d2c9/llmfoundry/utils/warnings.py#L41 as the warning class

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -145,6 +177,35 @@ def __init__(
             init_config_defaults,
         )
 
+        if 'reuse_kv_layer_idx' in self.attn_config and self.attn_config[
+            'attn_impl'] == 'torch':
+            raise NotImplementedError(
+                'reusing kv cache from a previous layer is not implemented for torch attention.',
+            )
+        if block_overrides is not None:
```
dakinggg, line 185: suggest a helper function on this class for validating the overrides

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -440,6 +442,10 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
             nn.ModuleList: The list of Transformer blocks.
         """
         block_args = self.extract_block_args(config.to_dict())
+        self.kv_cache_layers = None
+
+        if config.block_overrides is not None:
+            return self._construct_blocks_with_overrides(config, block_args)
 
         return nn.ModuleList([
```
dakinggg, line 450: should probably make this block init a helper function so that you can call it inside the `construct_blocks_with_overrides`

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+        self,
+        config: MPTConfig,
+        block_args: Dict[str, Any],
+    ) -> nn.ModuleList:
+        if config.block_overrides is None:
+            raise ValueError(
+                'config.block_overrides should not be None when calling _construct_blocks_with_overrides.',
+            )
+        modules_order_expanded = {}
+        for type in 'start', 'repeating_pattern', 'end':
```
dakinggg, line 467: i didn't know you could this...but maybe wrap in a list?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+            modules_order_expanded[type] = []
+            if type in config.block_overrides:
```
dakinggg, line 469: can combine these lines to `for block in config.block_overrides.get(type, [])`

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+            if (
+                config.n_layers - (start_len + end_len)
+            ) % repeating_pattern_len != 0:
+                raise ValueError(
+                    'Number of layers should be divisible by the specified custom modules order.',
+                )
+            modules_order_expanded[
+                'repeating_pattern'
+            ] = modules_order_expanded['repeating_pattern'] * (
+                (config.n_layers -
```
dakinggg, line 493: pull this calculation out into a variable

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+    def _construct_blocks_with_overrides(
+        self,
+        config: MPTConfig,
+        block_args: Dict[str, Any],
+    ) -> nn.ModuleList:
+        if config.block_overrides is None:
+            raise ValueError(
+                'config.block_overrides should not be None when calling _construct_blocks_with_overrides.',
+            )
+        modules_order_expanded = {}
```
dakinggg, line 466: I think `get_modules_order` or something can be pulled out into a helper function and individually tested

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+                                             ] + modules_order_expanded['end']
+
+        module_list = []
+        layer_description_list = []
+        if len(model_modules_order_expanded) != config.n_layers:
+            raise ValueError(
+                f'The specified block overrides do not match the number of layers: {len(model_modules_order_expanded)} vs {config.n_layers}.',
+            )
+
+        for i in range(config.n_layers):
```
dakinggg, line 508: suggest helper function for this that just constructs the list of block args, and then construct the blocks after that whole list of configs has been created

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+    def _override_block_args(
+        self,
+        block_args: Dict[str, Any],
+        override_config: Dict[str, Any],
+        allowed_block_overrides: set,
+    ) -> Dict[str, Any]:
+        new_block_args = override_config | block_args
+        common_keys = override_config.keys() & block_args.keys()
+        for k in common_keys:
+            if type(override_config[k]) != type(block_args[k]):
```
dakinggg, line 577: is this true? seems like you could pretty easily want to (e.g.) override an int with a float or something

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+    ) -> List[dict[str, str]]:
+        overrides_list = []
+        for k, v in override_config.items():
+            if isinstance(v, dict):
+                overrides_list.extend(self._get_overrides_for_logging(v))
+            else:
+                overrides_list.append({k: v})
+        return overrides_list
+
+    def _override_block_args(
```
dakinggg, line 568: would be good to add some unit tests for this function

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+        self,
+        block_args: Dict[str, Any],
+        override_config: Dict[str, Any],
+        allowed_block_overrides: set,
+    ) -> Dict[str, Any]:
+        new_block_args = override_config | block_args
```
dakinggg, line 574: would it be simpler to make a shallow copy of `block_args` here and then just iterate over the keys in `override_config` (keeping the recursive bit)?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -448,6 +454,142 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+
+        for i in range(config.n_layers):
+            module_name = model_modules_order_expanded[i]
+            override_config = {}
+            if module_name != 'default':
+                override_config = copy.deepcopy(
+                    config.block_overrides['overrides'][module_name],
+                )
+                override_attn_config = override_config.get('attn_config', None)
+                if override_attn_config is not None and 'reuse_kv_layer_idx' in override_attn_config:
```
ShashankMosaicML, line 516: Add warning if layers sharing kv cache have different attn types or sliding window attentions.

tests/models/test_model.py
```diff
@@ -2617,3 +2622,169 @@ def test_head_dim_8_flash_mqa_attn(batch_size: int = 2):
...
+    'start',
+    [[], [{
+        'name': 'default',
+        'repeat': 1,
+    }, {
+        'name': 'layer_s',
+        'repeat': 2,
+    }]],
+)
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
ShashankMosaicML: Done.

tests/models/layers/test_flash_torch.py
```diff
@@ -537,3 +537,224 @@ def test_grouped_query_invalid_heads():
...
+    }],
+)
+def test_reuse_prev_layer_kv_cache(
+    pos_emb_config: dict,
+    device: str = 'cuda',
+):
+    """Checks reusing previous layer's kv cache."""
+    alibi = pos_emb_config['alibi']
+    rope = pos_emb_config['rope']
+    if alibi and not (
```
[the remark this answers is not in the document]
ShashankMosaicML: Done.

tests/models/layers/test_flash_torch.py
```diff
@@ -537,3 +537,224 @@ def test_grouped_query_invalid_heads():
...
+    if rope and (pos_emb_config['rope_impl']
+                 == 'dail') and (not is_flash_v2_installed()):
+        pytest.skip('dail implementation of rope requires flash attention 2.')
+
+    if (not is_flash_v2_installed(v2_version='v2.1.2')):
+        pytest.skip(
+            'Using sequence id with flash attention requires flash attention v2.1.2 or higher.',
+        )
+
+    cfg = om.create({
```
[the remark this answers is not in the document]
ShashankMosaicML: Done.

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -117,6 +118,37 @@ def __init__(
                 also be a dictionary that specifies the fc layer name and any kwargs for the fc layer.
             tie_word_embeddings (bool): Whether to tie the input embedding and output layers.
             use_pad_tok_in_ffn (bool): Whether to forward the pad token in the feedforward networks.
+            block_overrides: The allows for overriding default block configs for certain layers. This should contain two sub configs: order and overrides. order specifies the order of different kinds of layers (default refers to a layer that does not apply any overrides). For each kind of layer, specify the overrides in the overrides config.
```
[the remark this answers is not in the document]
ShashankMosaicML: Done.

llmfoundry/models/mpt/configuration_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Discussed offline, added test.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Discussed offline. We can remove this if we want to allow such an override in the future.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

tests/models/test_model.py
```diff
@@ -2617,3 +2622,169 @@ def test_head_dim_8_flash_mqa_attn(batch_size: int = 2):
...
+
+    batch = gen_random_batch(batch_size, test_cfg)
+
+    assert batch['input_ids'].shape == torch.Size([
+        batch_size,
+        test_cfg.max_seq_len,
+    ])
+    model.train()
+    with get_precision_context(test_cfg.precision):
+        outputs = model(batch)
```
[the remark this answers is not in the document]
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -440,14 +442,187 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+    def _get_override_block_args_list(
+        self,
+        config: MPTConfig,
+        block_args: Dict[str, Any],
+    ) -> List[Dict[str, Any]]:
+        if config.block_overrides is None:
+            raise ValueError(
+                'config.block_overrides should not be None when calling _get_override_block_args_list.',
+            )
+        repeat = config.block_overrides[
```
dakinggg, line 471: should just `config.block_overrides.get('repeat', 1)

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -440,14 +442,187 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+                ):
+                    MPTModel._validate_reuse_kv_layer_config(
+                        block_overrides=config.block_overrides,
+                        model_modules_order_expanded=
+                        model_modules_order_expanded,
+                        b_idx=b_idx,
+                        override_config=override_config,
+                        reuse_kv_layer_idx_dict=reuse_kv_layer_idx_dict,
+                    )
+                    if self.kv_cache_layers is None:
```
dakinggg, line 504: should this just be initialized to a set to start with?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -440,14 +442,187 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+                        self.kv_cache_layers = set()
+                    self.kv_cache_layers.add(
+                        override_config['attn_config']['reuse_kv_layer_idx'],
+                    )
+            layer_description_list.append([
+                b_idx,
+                module_name,
+                self._get_overrides_for_logging(override_config),
```
dakinggg, line 512: why do we want the flatten helper? wouldn't you want to see the actual structure of the overrides?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -724,7 +901,21 @@ def forward(
                 attention_mask,
             )
 
+        layer_kv_cache_dict = {}
         for b_idx, block in enumerate(self.blocks):
+            attn_block = block.attn if hasattr(
```
dakinggg, line 906: this should check the config.fuse_norm_attn_norm key instead of checking for the attr

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -440,14 +442,187 @@ def construct_blocks(self, config: MPTConfig) -> nn.ModuleList:
...
+            override_config = {}
+            if module_name != 'default':
+                override_config = copy.deepcopy(
+                    config.block_overrides['overrides'][module_name],
+                )
+                if 'reuse_kv_layer_idx' in override_config.get(
+                    'attn_config',
+                    {},
+                ):
+                    MPTModel._validate_reuse_kv_layer_config(
```
dakinggg, line 496: I think this would be simpler and easier to read if you validated this once after constructing the whole list instead of inside the construction loop. and its more like `resolve_reuse_kv_layer_config` than `validate`

dakinggg (commented):
mostly lgtm, are you done testing it?

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Without the flatten, we will see a nested dictionary which might be difficult to read. I don't have a strong opinion though. I can remove the flattening now, and bring it back later if we find it difficult to read.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Updated the name to `resolve_reuse_kv_layer_idx`. 

This for loop's main purpose is to resolve overrides and add then them to the list of overrides. If we create a list of overrides first and then resolve them, it would essentially mean that we add a for loop that is essentially identical to the one currently there. The only other thing that this for loop does is that it logs the overrides into a table.

ShashankMosaicML:
> mostly lgtm, are you done testing it?

Yes, we have finished testing this. Everything seems fine.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -741,9 +923,15 @@ def forward(
                 output_attentions=bool(output_attentions),
                 alibi_slopes=alibi_slopes,
                 flash_attn_padding_info=flash_attn_padding_info,
+                prev_layer_key_value=prev_layer_key_value,
```
dakinggg, line 926: minor suggestion to avoid breaking downstream code. only pass this arg along if it is non none. that way downstream code will only need the arg if the feature is trying to be used.

dakinggg (approved):

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

PR #1300, opened 2024-06-22
Add `all` transforms to train script

dakinggg:
In https://github.com/mosaicml/llm-foundry/pull/1290, I added a unit test for the `all` transform, but forgot to add it to `train.py`

Loss curve before and after, also verified the batch size calculations are done the same.
Screenshot 2024-06-22 at 1 52 11 PM

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -50,7 +50,6 @@ from llmfoundry.utils.config_utils import (
     make_dataclass_and_log_config,
     pop_config,
     process_init_device,
-    update_batch_size_info,
 )
 from llmfoundry.utils.exceptions import (
     BaseContextualError,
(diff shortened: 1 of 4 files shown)


scripts/train/train.py
```diff
@@ -197,7 +196,7 @@ def main(cfg: DictConfig) -> Trainer:
         cfg,
         TrainConfig,
         TRAIN_CONFIG_KEYS,
-        transforms=[update_batch_size_info],
+        transforms='all',
```
snarayan21, line 199: for my own understanding, this will apply all transforms (according [to here](https://github.com/dakinggg/llm-foundry/blob/c16e94d740482f2d83424bf35de17714812e5d4f/llmfoundry/utils/config_utils.py#L253)) -- suppose we add a new transform that's only applicable for certain cases, would we still want to have `all` in this case? Would it be safer to only specify the transforms we currently want to use as a list in `train.py`, and only modify that list as we see fit down the line?

snarayan21 (commented):
question

scripts/train/train.py
snarayan21: resolved offline

snarayan21 (approved):
noice

PR #1301, opened 2024-06-23
Ignore mosaicml logger for exception if excephook is active

jjanezhang:
## Ignore mosaicml logger for exception if excephook is active

Before we remove the logger all together (re: https://github.com/mosaicml/llm-foundry/pull/1292) we want to first test catching exceptions with a flag. This prevents us from logging an exception twice. I will run tests that turn the logger off in favor of overriding the excepthook to make sure the intended before still holds for finetuning.

Ran a test on dbx-staging to confirm that the flag works. We will have to update the foundry image used for finetuning to incorporate this change.
image

### DIFF
--- a/llmfoundry/utils/mosaicml_logger_utils.py
+++ b/llmfoundry/utils/mosaicml_logger_utils.py
@@ -37,6 +37,13 @@ def maybe_create_mosaicml_logger() -> Optional[MosaicMLLogger]:
         return MosaicMLLogger()
 
 
+def no_override_excepthook() -> bool:
+    return os.environ.get(
+        'OVERRIDE_EXCEPTHOOK',
+        'false',
+    ).lower() != 'true'
+
+
 def find_mosaicml_logger(
     loggers: List[LoggerDestination],
 ) -> Optional[MosaicMLLogger]:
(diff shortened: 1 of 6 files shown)


dakinggg (commented):
there are more places to do this, can you grep foundry for .log_exception and replace all of them (unless there was a reason you only did some)?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -676,6 +676,9 @@ def fetch_DT(args: Namespace) -> None:
         log.info(f'Elapsed time {time.time() - tik}')
 
     except Exception as e:
-        if mosaicml_logger is not None:
+        if mosaicml_logger is not None and os.environ.get(
```
dakinggg, line 679: can we add a helper function somewhere for the environ check? otherwise its a pretty easy condition to make a mistake on.

jjanezhang: This is reminding me that we need some end to end tests for custom exceptions, will work with Vincent to add those

llmfoundry/utils/mosaicml_logger_utils.py
```diff
@@ -37,6 +37,13 @@ def maybe_create_mosaicml_logger() -> Optional[MosaicMLLogger]:
         return MosaicMLLogger()
 
 
+def no_override_excepthook() -> bool:
```
dakinggg, line 40: nit: docstring please

dakinggg (approved):

PR #1302, opened 2024-06-24
Add Retries to run_query

KuuCi:
Fine-tuning has seen a couple of failures regarding Spark timing out, resulting in complete run failure. Just adding retries here to buffer and hopefully reduce that down.

https://databricks.atlassian.net/browse/MCLOUD-4793

Here is a passing run:
test-uc-mlflow-66pcq4

### DIFF
--- a/scripts/data_prep/convert_delta_to_json.py
+++ b/scripts/data_prep/convert_delta_to_json.py
@@ -377,8 +377,9 @@ def fetch(
                 cursor,
                 sparkSession,
             )
-            nrows = [row.asDict() for row in ans
-                    ][0].popitem()[1]  # pyright: ignore
+            nrows = [
+                row.asDict() for row in ans  # pyright: ignore
+            ][0].popitem()[1]
             log.info(f'total_rows = {nrows}')
             break
         except Exception as e:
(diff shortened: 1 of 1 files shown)


mvpatel2000 (commented):
should we make it a decorator or helper fn?

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -368,36 +369,52 @@ def fetch(
         dbsql (databricks.sql.connect): dbsql session
     """
     cursor = dbsql.cursor() if dbsql is not None else None
+    for row_retry in range(MAX_RETRY):
```
XiaohanZhangCMU, line 372: Can we reuse Composer's [retry logic] here (https://github.com/mosaicml/composer/blob/dev/composer/utils/retrying.py)? It has better backoff and jitter etc. (you may need to refactor the functions here a bit to use the decorator).

XiaohanZhangCMU (changes requested):

KuuCi:
Passing Run: test-uc-mlflow-WyKSz4

mvpatel2000 (approved):

PR #1303, opened 2024-06-24
Bumping mlflow version to include buffering

JackZ-db:
The MlFlow version was previously capped at 2.14, leading to llm-foundry regression tests generating abnormal mlflow traffic. Updated to the correct version of MlFlow, at 'mlflow>=2.14.1,<3.0'.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ classifiers = [
 
 install_requires = [
     'mosaicml[libcloud,wandb,oci,gcs,mlflow]>=0.23.4,<0.24',
-    'mlflow>=2.14.1,<3.0',
+    'mlflow>=2.14.1,<2.15',
     'accelerate>=0.25,<0.26',  # for HF inference `device_map`
     'transformers>=4.40,<4.41',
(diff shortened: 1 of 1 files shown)


chenmoneygithub (approved):
thank you!

setup.py
```diff
@@ -55,7 +55,7 @@
 
 install_requires = [
     'mosaicml[libcloud,wandb,oci,gcs,mlflow]>=0.23.4,<0.24',
-    'mlflow>=2.13.2,<2.14',
+    'mlflow>=2.14.1,<3.0',
```
mvpatel2000, line 58:
```suggestion
```

mvpatel2000 (commented):
Can we just remove given we use mlflow target w composer? @dakinggg @irenedea

setup.py
JackZ-db: yep, curious too, thanks Mihir!

dakinggg: please stick withj minor version cap

dakinggg (approved):

PR #1304, opened 2024-06-26
Avoid circular import in hf checkpointer

dakinggg:
If you extend LLM Foundry modeling/config code, the hf checkpointer may create a circular import in its attempt to automatically add all the missing imports to the files for hf. This PR simply excludes the dangerous circular import by preventing either of the entrypoint files (the modeling file or the config file), from including an import to the modeling file. The modeling file is expected to include a legitimate import of the config file, but the config file should not import the modeling file as this would be circular.


irenedea (approved):

snarayan21 (approved):
noice

PR #1308, opened 2024-06-27
External library usage interface

moeiniamir:
A similar feature to the [library usage of lm-eval-harness](https://github.com/EleutherAI/lm-evaluation-harness/blob/main/docs/interface.md) using a simple registry add-on.


dakinggg (commented):
Hi, the functions included in the library now here (https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/command_utils/eval.py) should cover this usage. Does that work for you?

PR #1313, opened 2024-06-27
Error if metadata matches existing keys

dakinggg:
We flatten the metadata so that it is easier to parse out of mlflow/run metadata. Because of this, a user's metadata key could overwrite the real key in the config that gets logged. This PR fixes that by simple erroring and asking the user to rename their metadata key. It also adds `flatten_metadata` to allow not flattening the metadata.

Original run that overwrites the key: `metadata-with-error-4-G9ps3T`
Run with duplicate key that now errors: `metadata-unflat-1-ltIBGL`
Run without flattening metadata: `metadata-unflat-2-VoxpxS`

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -320,15 +320,18 @@ def main(cfg: DictConfig) -> Trainer:
             loggers.append(mosaicml_logger)
 
     if train_cfg.metadata is not None:
-        # Flatten the metadata for logging
-        logged_cfg.pop('metadata', None)
-        common_keys = set(logged_cfg.keys()) & set(train_cfg.metadata.keys())
-        if common_keys:
-            raise ValueError(
-                f'Keys {common_keys} are already present in the config. Please rename them in metadata.',
-            )
+        # Optionally flatten the metadata for logging
+        if train_cfg.flatten_metadata:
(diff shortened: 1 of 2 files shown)


josejg:
Could we prefix `metadata_` in the metadata keys for clarity ?

josejg:
looks good to me, but I'll give it a try just in case

scripts/train/train.py
```diff
@@ -320,9 +320,21 @@ def main(cfg: DictConfig) -> Trainer:
...
-        # Flatten the metadata for logging
-        logged_cfg.pop('metadata', None)
-        logged_cfg.update(train_cfg.metadata, merge=True)
+        # Optionally flatten the metadata for logging
+        if train_cfg.flatten_metadata:
+            logged_cfg.pop('metadata', None)
+            common_keys = set(
+                logged_cfg.keys(),
+            ) & set(train_cfg.metadata.keys())
+            if common_keys:
```
irenedea, line 329: nit: to be a little more precise

```suggestion
            if len(common_keys) > 0:
```

irenedea (approved):

irenedea (approved):

PR #1318, opened 2024-06-29
Refactor hf checkpointer for config transformations

irenedea:
Subclasses can now easily extend the config transformations by overriding the `transform_config` method.


dakinggg (approved):
lgtm, simple unit test for the helper method since its a public method now would be good, but not blocking

PR #1320, opened 2024-06-29
Extra serverless

XiaohanZhangCMU:


setup.py
```diff
@@ -123,6 +123,11 @@
     'grouped-gemm==0.1.4',
 ]
 
+extra_deps['serverless'] = {
```
dakinggg, line 126: lets call it `databricks-serverless`?

dakinggg (approved):

setup.py
mvpatel2000: +1

setup.py
```diff
@@ -123,6 +123,11 @@
...
+    dep for key, deps in extra_deps.items() for dep in deps
+    if 'gpu' not in key and 'megablocks' not in key and
+    'databricks-connect' not in dep
```
mvpatel2000, line 129: so this skips `databricks-connect`?

XiaohanZhangCMU: Yes.

PR #1332, opened 2024-07-02
Currently multi-gpu generate does not work with hf.generate for hf checkpoints. This PR fixes that.

ShashankMosaicML:
Multi gpu generation using hf.generate with device map = 'auto' does pipeline parallelism and moves different modules to different gpus. This results in input tensors to certain operations being on different gpus than other inputs to that operation, which results in an error. This PR moves the tensors to match the other tensors. This should not slow down training because during training all of these tensor movements should be no-ops.

### DIFF
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -266,11 +266,11 @@ def flash_attn_fn(
 
     batch_size, seqlen = query.shape[:2]
 
-    indices_q = flash_attn_padding_info['indices_q']
-    indices_k = flash_attn_padding_info['indices_k']
-    indices_v = flash_attn_padding_info['indices_v']
-    cu_seqlens_q = flash_attn_padding_info['cu_seqlens_q']
-    cu_seqlens_k = flash_attn_padding_info['cu_seqlens_k']
+    indices_q = flash_attn_padding_info['indices_q'].to(query.device)
+    indices_k = flash_attn_padding_info['indices_k'].to(key.device)
+    indices_v = flash_attn_padding_info['indices_v'].to(value.device)
+    cu_seqlens_q = flash_attn_padding_info['cu_seqlens_q'].to(query.device)
+    cu_seqlens_k = flash_attn_padding_info['cu_seqlens_k'].to(key.device)
     max_seqlen_q = flash_attn_padding_info['max_seqlen_q']
     max_seqlen_k = flash_attn_padding_info['max_seqlen_k']
 
@@ -667,6 +667,8 @@ class GroupedQueryAttention(nn.Module):
             else:
                 (cos, sin) = rotary_emb(x=value, seq_len=seq_len)
             if is_transformers_version_gte('4.38'):
+                cos = cos.to(query.device)
+                sin = sin.to(query.device)
                 query, key = apply_rotary_pos_emb(
                     q=query,
                     k=key,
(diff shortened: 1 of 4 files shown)


llmfoundry/models/layers/blocks.py
```diff
@@ -198,7 +198,7 @@ def forward(
                 m = self.norm_2(x)
 
         n = self.apply_ffn(attention_mask, m)
-        x = x + self.resid_ffn_dropout(n)
+        x = x + self.resid_ffn_dropout(n).to(x.device)
```
vchiley, line 201: the fact that this one is needed is scary...

llmfoundry/models/layers/dmoe.py
```diff
@@ -279,7 +279,8 @@ def forward(
             topk_list = topk_idx.tolist()
 
             expert_tokens = x[None, token_list].reshape(-1, hidden_size)
-            mlp_output = self.mlp(expert_tokens, expert_idx)
+            mlp_output = self.mlp(expert_tokens,
+                                  expert_idx).to(expert_weights.device)
```
vchiley, line 283: same comment
> the fact that this one is needed is scary...

llmfoundry/models/layers/blocks.py
ShashankMosaicML: Discussed offline. Modified the code accordingly.

llmfoundry/models/layers/dmoe.py
ShashankMosaicML: Discussed offline. Modified the code accordingly.

llmfoundry/models/layers/blocks.py
snarayan21: why is this needed actually?

snarayan21 (commented):
would like a bit of clarity, thanks for submitting

llmfoundry/models/layers/attention.py
```diff
@@ -266,11 +266,11 @@ def flash_attn_fn(
 
     batch_size, seqlen = query.shape[:2]
 
-    indices_q = flash_attn_padding_info['indices_q']
-    indices_k = flash_attn_padding_info['indices_k']
-    indices_v = flash_attn_padding_info['indices_v']
-    cu_seqlens_q = flash_attn_padding_info['cu_seqlens_q']
-    cu_seqlens_k = flash_attn_padding_info['cu_seqlens_k']
+    indices_q = flash_attn_padding_info['indices_q'].to(query.device)
```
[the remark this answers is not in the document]
ShashankMosaicML: Done: https://github.com/mosaicml/llm-foundry/pull/1332#issue-2386827204

llmfoundry/models/layers/dmoe.py
```diff
@@ -280,9 +280,11 @@ def forward(
 
             expert_tokens = x[None, token_list].reshape(-1, hidden_size)
             mlp_output = self.mlp(expert_tokens, expert_idx)
+            expert_weights = expert_weights.to(mlp_output.device)
             expert_out = mlp_output * expert_weights[token_list, topk_list,
                                                      None]
-
+            out = out.to(mlp_output.device)
```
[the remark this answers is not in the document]
ShashankMosaicML: Added comments explaining. Also discussed offline.

vchiley (approved):

dakinggg (approved):
approving to unblock

PR #1334, opened 2024-07-02
Adding a child class of hf's rotary embedding to make hf generate work on multiple gpus.

ShashankMosaicML:
Multi gpu generation using hf.generate with device map = 'auto' does pipeline parallelism and moves different modules to different gpus. We see that for some reason, rope embedding's `inv_freq` buffer does not get moved to the right device. In this PR we explicitly move it to the correct device. This should not slow down training because during training this tensor movement should be a no-op.

Related to this: https://github.com/mosaicml/llm-foundry/pull/1332

### DIFF
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -109,7 +109,7 @@ def gen_rotary_embedding(
         )
     elif rope_impl == 'hf':
         if rope_hf_config['type'] == 'no_scaling':
-            return HFRotaryEmbedding(
+            return HFRotaryEmbeddingMP(
                 rope_head_dim,
                 max_position_embeddings=max_seq_len,
                 base=rope_theta,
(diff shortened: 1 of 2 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -306,6 +306,20 @@ def apply_sequence_id(
     return attn_bias
 
 
+class HFRotaryEmbeddingMP(HFRotaryEmbedding):
```
dakinggg, line 309:
```suggestion
class HFRotaryEmbeddingFoundry(HFRotaryEmbedding):
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -306,6 +306,20 @@ def apply_sequence_id(
...
+    @torch.no_grad()
+    def forward(
+        self,
+        x: torch.Tensor,
+        position_ids: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        # In this subclass, we move `inv_freq` to same device as position_ids. This operation should be a no-op during training.
+        # This is done to fix pipeline parallel generation using hf.generate. Please see this comment for details: https://github.com/mosaicml/llm-foundry/pull/1334#issue-2387337525
+        self.inv_freq = self.inv_freq.to(position_ids.device)
+        return super().forward(x, position_ids)
```
dakinggg, line 320: please pass args by keyword

dakinggg (approved):

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Done.

PR #1335, opened 2024-07-03
Add a config arg to just save an hf checkpoint

dakinggg:
We have an offline conversion script from composer to hf checkpoint format, but it only works for monolithic checkpoints, and has some other components hardcoded. We are also working on improved utilities for working with checkpoints in composer, but as a stopgap, we are introducing an arg here that will run train.py, but just call the hf checkpointer save and then exit.

Test run: `just-hf-4-nWWD05` (and confirmed the checkpoint shows up in object store)

### DIFF
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -161,7 +161,7 @@ class TrainConfig:
     load_strict_model_weights: bool = True
     load_ignore_keys: Optional[List[str]] = None
     save_ignore_keys: Optional[List[str]] = None
-    just_hf_checkpoint: bool = False
+    only_hf_checkpoint: bool = False
 
     # Dataloader
     device_train_microbatch_size: Union[str, int, float] = 'auto'
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -161,6 +161,7 @@ class TrainConfig:
     load_strict_model_weights: bool = True
     load_ignore_keys: Optional[List[str]] = None
     save_ignore_keys: Optional[List[str]] = None
+    just_hf_checkpoint: bool = False
 
     # Dataloader
     device_train_microbatch_size: Union[str, int, float] = 'auto'
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -534,7 +534,7 @@ def main(cfg: DictConfig) -> Trainer:
         ]
         if len(hf_checkpointer_callbacks) == 0:
             raise ValueError(
-                'No HuggingFaceCheckpointer callback found, but just_hf_checkpoint was set to True.',
+                'No HuggingFaceCheckpointer callback found, but just_hf_checkpoint was set to True. Please add a HuggingFaceCheckpointer.',
             )
 
         hf_checkpointer_callback = hf_checkpointer_callbacks[0]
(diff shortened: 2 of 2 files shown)


scripts/train/train.py
```diff
@@ -527,6 +527,20 @@ def main(cfg: DictConfig) -> Trainer:
...
     )
 
+    # Optionally just save an HF checkpoint
+    if train_cfg.just_hf_checkpoint:
+        hf_checkpointer_callbacks = [
+            c for c in callbacks if isinstance(c, HuggingFaceCheckpointer)
+        ]
+        if len(hf_checkpointer_callbacks) == 0:
+            raise ValueError(
+                'No HuggingFaceCheckpointer callback found, but just_hf_checkpoint was set to True.',
```
mvpatel2000, line 537:
```suggestion
                'No HuggingFaceCheckpointer callback found, but just_hf_checkpoint was set to True. Please add a HuggingFaceCheckpointer.',
```

llmfoundry/utils/config_utils.py
```diff
@@ -161,6 +161,7 @@ class TrainConfig:
     load_strict_model_weights: bool = True
     load_ignore_keys: Optional[List[str]] = None
     save_ignore_keys: Optional[List[str]] = None
+    just_hf_checkpoint: bool = False
```
mvpatel2000, line 164:
```suggestion
    only_hf_checkpoint: bool = False
```
More consistent with naming elsewhere

dakinggg: done

mvpatel2000 (dismissed):

scripts/train/train.py
```diff
@@ -527,6 +527,20 @@ def main(cfg: DictConfig) -> Trainer:
...
+    # Optionally just save an HF checkpoint
+    if train_cfg.only_hf_checkpoint:
+        hf_checkpointer_callbacks = [
+            c for c in callbacks if isinstance(c, HuggingFaceCheckpointer)
+        ]
+        if len(hf_checkpointer_callbacks) == 0:
+            raise ValueError(
+                'No HuggingFaceCheckpointer callback found, but only_hf_checkpoint was set to True. Please add a HuggingFaceCheckpointer.',
+            )
+
```
mvpatel2000, line 539: should raise error or warning if > 1 hf checkpointer

dakinggg: added warning

mvpatel2000 (approved):

scripts/train/train.py
```diff
@@ -527,6 +527,24 @@ def main(cfg: DictConfig) -> Trainer:
...
+            c for c in callbacks if isinstance(c, HuggingFaceCheckpointer)
+        ]
+        if len(hf_checkpointer_callbacks) == 0:
+            raise ValueError(
+                'No HuggingFaceCheckpointer callback found, but only_hf_checkpoint was set to True. Please add a HuggingFaceCheckpointer.',
+            )
+        if len(hf_checkpointer_callbacks) > 1:
+            warnings.warn(
+                'Multiple HuggingFaceCheckpointer callbacks found. Only the first one will be used.',
+            )
```
mvpatel2000, line 542: i would just raise an error

snarayan21 (approved):
dope

scripts/train/train.py
dakinggg: sure

PR #1337, opened 2024-07-03
Add CLI for train.py

KuuCi:
This PR allows users to call `composer llm-foundry train {YAML_PATH} {ARGS}` while maintaining correctness with `composer llm-foundry/train.py {PATH} {ARGS}`. The motivation is for DLE where we want to make the CLI much more intuitive in the docker images

Testing:
test-cli-cSn2Rb runs:
`composer -c -n 8 llmfoundry train /mnt/config/parameters.yaml || (echo "Command failed - killing python" && pkill python && exit 1)`

test-cli-qsRHEI runs:
`composer -c llmfoundry train /mnt/config/parameters.yaml || (echo "Command failed - killing python" && pkill python && exit 1)  `

test-cli-vGpXcw runs:
`composer train/train.py /mnt/config/parameters.yaml || (echo "Command failed - killing python" && pkill python && exit 1)`

Here is the MLflow experiement folder indicating all three runs act the same:
https://dbc-04ac0685-8857.staging.cloud.databricks.com/ml/experiments/3707544126254710?o=3360802220363900&searchFilter=&orderByKey=attributes.start_time&orderByAsc=false&startTime=ALL&lifecycleFilter=Active&modelVersionFilter=All+Runs&datasetsFilter=W10%3D


b-chu:
This seems like a breaking change

mvpatel2000 (dismissed):
Deleting `scripts/train/train.py` is a breaking change.

KuuCi:
We aren't deleting `scripts/train/train.py`, `scripts/train/train.py` is just calling `train/train.py` now

b-chu:
Ah, thanks for pointing that out. I'll give a more detailed review later

KuuCi:
will update to match scripts/train/train.py merges after first pass

KuuCi:
manual test runs updated

llmfoundry/train/train.py
```diff
@@ -0,0 +1,580 @@
...
+    # Disable resolving environment variables through omegaconf.
+    om.clear_resolver('oc.env')
+
+    # Load yaml and cli arguments.
+    with open(yaml_path) as f:
+        yaml_cfg = om.load(f)
+    cli_cfg = om.from_cli(args_list)
+    cfg = om.merge(yaml_cfg, cli_cfg)
+    assert isinstance(cfg, DictConfig)
+    train(cfg)
```
irenedea, line 580: remove? This doesn't need to be a script right?

scripts/train/train.py
```diff
@@ -575,4 +19,4 @@ def main(cfg: DictConfig) -> Trainer:
     cli_cfg = om.from_cli(args_list)
     cfg = om.merge(yaml_cfg, cli_cfg)
     assert isinstance(cfg, DictConfig)
```
irenedea, line 21: Is the plan to deprecate this script? Shall we add a deprecation warning?

scripts/train/train.py
```diff
@@ -575,4 +19,4 @@ def main(cfg: DictConfig) -> Trainer:
...
-    main(cfg)
+    train(cfg)
```
irenedea, line 22: could we have a helper function called `train_from_yaml` so we can share the code between here and the code in cli.py?

llmfoundry/train/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
```
irenedea, line 2: Can you do something like this: https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/utils/__init__.py#L70 

So that we can expose `train` here?

irenedea: That way the imports can be `from llmfoundry.train import train`

llmfoundry/train/train.py
KuuCi: Doesn't need to be in the script

scripts/train/train.py
KuuCi: I dont think we have current plan to deprecate, might be a follow up for daniel when he comes back

llmfoundry/cli/cli.py
```diff
@@ -1,12 +1,29 @@
 # Copyright 2024 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
+from typing import List, Optional
```
b-chu, line 4: Don't need List, use list

llmfoundry/cli/cli.py
```diff
@@ -1,12 +1,29 @@
...
 
+
+@app.command(name='train')
+def train(
+    yaml_path: str = typer.Argument(
+        ...,
+        help='Path to the YAML configuration file',
+    ),  # type: ignore
+    args_list: Optional[List[str]] = typer.
+    Argument(None, help='Additional command line arguments'),  # type: ignore
```
b-chu, line 22: Why type ignore?

llmfoundry/cli/cli.py
```diff
@@ -1,12 +1,29 @@
...
 app = typer.Typer(pretty_exceptions_show_locals=False)
 app.add_typer(registry_cli.app, name='registry')
 
+
+@app.command(name='train')
+def train(
+    yaml_path: str = typer.Argument(
+        ...,
+        help='Path to the YAML configuration file',
+    ),  # type: ignore
```
b-chu, line 20: Why type ignore?

tests/a_scripts/train/test_train.py
```diff
@@ -11,12 +11,12 @@
 from omegaconf import DictConfig, ListConfig
 from omegaconf import OmegaConf as om
 
+from llmfoundry.train import TrainConfig  # noqa: E402
```
b-chu, line 14: What is E402?

KuuCi: Not sure, was there when I started, all I did was change from scripts.train to llmfoundry.train

llmfoundry/cli/cli.py
KuuCi: Function calls and mutable objects not allowed within parameter default value expression

llmfoundry/cli/cli.py
KuuCi: Function calls and mutable objects not allowed within parameter default value expression

b-chu (approved):
Thanks!

PR #1338, opened 2024-07-03
Avoid HF race condition

dakinggg:
With the latest transformers version, it seems that there is a new race condition in the hf model code, so we apply the same fix as before, of running local rank 0 first to initialize the cache.

Before the transformers bump 6/6 runs succeeded. After the transformers bump 2/5 runs succeeded (6th failed for an unrelated reason). After this PR, 6/6 succeeded.

Transformers issue here: https://github.com/huggingface/transformers/issues/27421


irenedea (approved):
thank you!

mvpatel2000 (approved):

PR #1343, opened 2024-07-08
[MCLOUD-4910] Escape UC names during data prep

naren-loganathan:
[JIRA](https://databricks.atlassian.net/browse/MCLOUD-4910). This PR aims to get past an issue that occurs when specifying a delta table for training data that has a valid name with hyphens/special characters.

We now escape the table names with backticks.

Testing: `test_format_tablename`

Manual testing: When submitting fine-tuning runs with custom YAMLs pointing to this branch, the logs indicate that delta tables are read without `[INVALID_IDENTIFIER]` errors - even when hyphenated.

NOTE: We always apply escaping (`` `catalog`.`schema`.`tablename` ``). We do not expect the user to escape their own input. We do not handle cases where UC object names contain backticks.

### DIFF
--- a/scripts/data_prep/convert_delta_to_json.py
+++ b/scripts/data_prep/convert_delta_to_json.py
@@ -375,7 +375,7 @@ def get_total_rows(
     sparkSession: Optional[SparkSession],
 ):
     ans = run_query(
-        f'SELECT COUNT(*) FROM {format_tablename(tablename)}',
+        f'SELECT COUNT(*) FROM {tablename}',
         method,
         cursor,
         sparkSession,
@@ -393,7 +393,7 @@ def get_columns_info(
     sparkSession: Optional[SparkSession],
 ):
     ans = run_query(
-        f'SHOW COLUMNS IN {format_tablename(tablename)}',
+        f'SHOW COLUMNS IN {tablename}',
         method,
         cursor,
         sparkSession,
@@ -452,7 +452,7 @@ def fetch(
 
     if method == 'dbconnect' and sparkSession is not None:
         log.info(f'{processes=}')
-        df = sparkSession.table(format_tablename(tablename))
+        df = sparkSession.table(tablename)
 
         # Running the query and collecting the data as arrow or json.
         signed, _, _ = df.collect_cf('arrow')  # pyright: ignore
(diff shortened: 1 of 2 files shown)


XiaohanZhangCMU (changes requested):
Some general comments:

1. I thought it is easier just directly format args.delta_table_name in the main function, because you don't need to change any of the rest code.  
2. Can you add a test case where the tablename or schema name already comes in with the right format which has hyphen and escape? 
3. Can you reformat schema as well? I assume it follows the same reformatting rules.

naren-loganathan:
Thanks! Addressing these comments:

1. Fixed and verified manually with a custom YAML!
2. It seems like users cannot provide their own escaped UC paths due to validation checks being performed elsewhere. We will not handle names with backticks/escaped names in general. 
3. The expected delta table name is of the form `catalog.schema.tablename`. We escape the catalog, schema and tablename with backticks (i.e. we now do `catalog.schema.tablename` -> ``` `catalog`.`schema`.`tablename` ```).

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -283,6 +285,24 @@ def download(
 def download_starargs(args: Tuple) -> None:
     return download(*args)
 
+def format_tablename(tablename: str) -> str:
```
jjanezhang, line 288:
```suggestion
def format_tablename(table_name: str) -> str:
```

scripts/data_prep/convert_delta_to_json.py
```diff
@@ -610,7 +631,7 @@ def fetch_DT(args: Namespace) -> None:
     parser.add_argument(
         '--delta_table_name',
         required=True,
-        type=str,
+        type=format_tablename,
```
jjanezhang, line 634: do you like this better than keeping this as string and then running format_table_name in fetch DT?

jjanezhang (approved):
small nits! otherwise lgtm

XiaohanZhangCMU (approved):
LGTM. Please test the script a few times with actual tables (w/o hyphens). 

One example is (from a table in staging dogfood):
python scripts/data_prep/convert_delta_to_json.py --delta_table_name main.streaming.random_large_table --json_output_path /tmp/test_json1 --use_serverless True -- xxxxx

scripts/data_prep/convert_delta_to_json.py
naren-loganathan: Now running this within `fetch_DT` to make things clearer.

naren-loganathan:
Was able to get through data prep with `main.streaming.random_large_table`.

PR #1344, opened 2024-07-09
Log all extraneous_keys in one go for ease of development

josejg:
In the spirit of productivity, I think it'd be good for foundry to specify all extraneous keys it found in the config in a single error statement as otherwise you need to keep launching the run only to keep finding that a extraneous key needs to be removed at a time.


eracah (approved):
LGTM I hate guess and check for these types of things

mvpatel2000 (approved):
nice!

j316chuck:
This PR Is helping me debug too? Can we get it in!

PR #1347, opened 2024-07-11
Add fp32 to the set of valid inputs to attention layer

j316chuck:
# Description
Add fp32 to the set of valid inputs for attention layer. 

**Note**: 
- flash attn is incompatible with torch.float32 🔴 
- torch attn is compatible with torch.float32 ✅ 

# Tests: 

- Before:  `full-eval-fp32-train-fp8-llama3-8b-metamath-4ep-ObqlFj`  🔴 
   - `mixed_precision: full` 
- After: `torch-attn-full-eval-fp32-train-fp8-metamath-4ep-pmGJKN` ✅ 
   - `mixed_precision: full`

### DIFF
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -212,7 +212,7 @@ def check_valid_inputs(
     valid_dtypes: Optional[list[torch.dtype]] = None,
 ):
     if valid_dtypes is None:
-        valid_dtypes = [torch.float16, torch.bfloat16]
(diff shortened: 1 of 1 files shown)


mvpatel2000 (approved):
LGTM, can we unit test?

j316chuck:
Tested manually, there is no unit test for this.

dakinggg:
@j316chuck I don't think this is correct. Flash attention does not support fp32 (unless that changed recently?)

PR #1356, opened 2024-07-13
Add flag for saving only composer checkpoint

irenedea:
test: `ift-meta-llama-3-8b-bjyg3w-KRJ2jI`

Similar to https://github.com/mosaicml/llm-foundry/pull/1335 but for composer checkpoints.


snarayan21 (approved):
To clarify, this is when you pass in an HF model and want to save a composer checkpoint right? The other PR was for when I have a composer checkpoint and want to save an HF checkpoint.

LGTM, would like @dakinggg to stamp as well before merge

dakinggg (approved):

PR #1369, opened 2024-07-17
use default eval interval from composer

milocress:
https://github.com/mosaicml/composer/blob/da18d9e4513bc1629849158f5a16108350532adf/composer/trainer/trainer.py#L1042C1-L1043C1

Uses the default eval interval from composer rather than forcing eval_interval to be a required argument


b-chu (approved):
Thanks!

irenedea:
@milocress still working on this?

mvpatel2000 (approved):

PR #1376, opened 2024-07-19
Transform model

irenedea:


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -426,11 +435,15 @@ def dtensor_to_tensor_hook(
...
-                                tensor = tensor.cpu()
+                                tensor = tensor.to(dtype=self.dtype, device=torch.device('cpu'))
                             state_dict[fqn] = tensor
+                        else:
+                            state_dict[fqn] = None
+                        del tensor
+                    
                 if dist.get_global_rank() != 0:
                     for fqn in dtensor_fqns:
                         del state_dict[fqn]
```
bigning, line 449: can you try just state_dict = {} for non-0 rank?

bigning, line 449: I think we want to do gc before returning, right?

PR #1377, opened 2024-07-19
Allow flash attention up to 3

dakinggg:
Allows flash attention version up to 3 to make the docker image/install version correspondence less brittle (i.e. if we bump the flash attention version in the docker image to 2.6, the foundry install won't instantly break)


bigning (approved):
thanks!

vchiley (approved):
👍

PR #1380, opened 2024-07-22
HF Checkpoint OOM fix with earlier dtype conversion

snarayan21:
Porting over @irenedea's fix for HF Checkpoint OOMs from [this branch](https://github.com/mosaicml/llm-foundry/compare/main...irenedea:llm-foundry:transform-model)

Note: moved the `del tensor` call inside the `if` statement because otherwise we delete `tensor` every time (bad if `state_dict[fqn]` happens not to be a tensor or dtensor. could be some scalar for weight scales, for example.

Run OOMed without this change: running
Run successfully completed with this change: running


snarayan21:
Closing since this is a duplicate of #1376

PR #1381, opened 2024-07-22
Get a shared file system safe signal file name

dakinggg:
Use the previously added utils to get a signal file name that is safe on shared file systems. Previously the name would clash.

Tested simple train and load (don't have a multinode shared fs to actually test that part on):
train: dist-utils-train-1-UExzEF
load: dist-utils-load-2-Uoq35b

See also https://github.com/mosaicml/composer/pull/3485


mvpatel2000 (approved):

janEbert:
Hey, what was the reason to revert the commit? :)
Doesn't the reversion basically re-introduce the error that was explained in #1253?

dakinggg:
Hi, yes we reverted due to a bug in the released version of the helper function (did not work on systems where distributed was not initialized). That bug has since been fixed, and this commit should be reapplied. Thanks for the reminder!

PR #1382, opened 2024-07-22
Adds the convert_examples_ckpt from scripts to CLI

KuuCi:


dakinggg:
We don't need to add this one.

KuuCi:
bet

PR #1384, opened 2024-07-22
Do dtype conversion in torch hook to save memory

irenedea:
## Unit tests
Existing unit tests should cover this change. 

## Manual tests
Before
image

After

image

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -435,7 +435,7 @@ class HuggingFaceCheckpointer(Callback):
 
         cpu_offload = True
 
-        # Add a dtensor->cpu tensor hook to avoid CUDA OOM
+        # Add hook to move tensors to cpu to avoid CUDA OOM
         def tensor_hook(
             module: nn.Module,
             state_dict: Dict[str, Any],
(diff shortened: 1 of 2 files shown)


dakinggg (approved):
lgtm, can you get @bigning to take a look too please?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -436,7 +436,7 @@ def _save_checkpoint(self, state: State, logger: Logger):
         cpu_offload = True
 
         # Add a dtensor->cpu tensor hook to avoid CUDA OOM
```
snarayan21, line 438:
```suggestion
        # Add hook to move tensors to cpu to avoid CUDA OOM
```

snarayan21 (approved):
lgtm! minor comment

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -449,20 +449,23 @@ def dtensor_to_tensor_hook(
...
+                del tensor
             if dist.get_global_rank() != 0:
-                for fqn in dtensor_fqns:
-                    del state_dict[fqn]
+                state_dict = {}
             return state_dict
 
         hooks = []
         for _, module in state_dict_model.named_modules():
-            if isinstance(module, FSDP):
```
bigning, line 462: why do we need to remove this ? Does it make any difference ? without this condition, for modules wrapped with FSDP,  i think we run the `tensor_hook` for both the module and the fsdp wrapped module, which might be not necessary ?

bigning (commented):
thanks for the change!

bigning (approved):

PR #1385, opened 2024-07-22
Fix load and save planner config logic

irenedea:
Fixes the load and save planner logic. 'dict_items' object is not subscriptable, so it needs to be converted to a list.


dakinggg (approved):
Please add pr description, but lgtm

snarayan21 (approved):
fire

PR #1386, opened 2024-07-23
[kushalkodnad/tokenizer-registry] Introduce new registry for tokenizers

kushalkodn-db:
# What Does This PR Do?

The purpose of this PR is to introduce a new registry, specifically for tokenizers. This way, in the `llmfoudry/utils/builders.py` `build_tokenizer()` method, we can build a tokenizer from the registry, as long as the tokenizer inherits the `transformers.PreTrainedTokenizerBase` interface. For instance, the existing `TiktokenTokenizerWrapper` in `llmfoundry/tokenizers/tiktoken.py` can be built from this registry, as opposed to using if-else clauses in `build_tokenizer`, which can become tedious if we add support for more tokenizers.

### `llmfoundry/registry.py`

This is where I created the `tokenizers` registry. See the brief description of the llmfoundry tokenizers registry, defined in `_tokenizers_description`. Then, I followed the `callbacks` registry creation to do the same thing for `tokenizers` when calling `create_registry`.

### `llmfoundry/utils/builders.py`

The main idea is to replace the if-else clause that currently supports "tiktoken". As long as the tokenizer is a sub-class of `transformers.PreTrainedTokenizerBase`, then the tokenizer will be constructed.

### `llmfoundry/tokenizers/__init__.py`

Following the `callbacks` example, the addition to this file shows how to register a `transformers.PreTrainedTokenizerBase` sub-class to the tokenizers registry. Here, I registered the `tiktoken` tokenizer to the tokenizers registry, because it inherits the same interface.

### DIFF
--- a/llmfoundry/registry.py
+++ b/llmfoundry/registry.py
@@ -155,6 +155,19 @@ schedulers = create_registry(
     description=_schedulers_description,
 )
 
+_tokenizers_description = (
+    'The tokenizers registry is used to register tokenizers that implement the transformers.PreTrainedTokenizerBase interface. '
+    +
+    'The tokenizer will be passed to the build_dataloader() and build_composer_model() methods in train.py.'
+)
+tokenizers = create_registry(
+    'llmfoundry',
+    'tokenizers',
+    generic_type=Type[PreTrainedTokenizerBase],
+    entry_points=True,
+    description=_tokenizers_description,
+)
+
 _models_description = (
     """The models registry is used to register classes that implement the ComposerModel interface.
 
--- a/llmfoundry/registry.py
+++ b/llmfoundry/registry.py
@@ -396,6 +396,7 @@ __all__ = [
     'optimizers',
     'algorithms',
     'schedulers',
+    'tokenizers',
     'models',
     'dataset_replication_validators',
     'collators',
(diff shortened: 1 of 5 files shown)


tests/tokenizers/test_registry.py
```diff
@@ -0,0 +1,63 @@
...
+    def __call__(self, text: str) -> BatchEncoding:
+        raise NotImplementedError()
+
+
+def test_tokenizer_registry():
+    tokenizers.register('dummy', func=DummyTokenizer)
+    tokenizer = build_tokenizer(tokenizer_name='dummy', tokenizer_kwargs={})
+    assert type(tokenizer) == DummyTokenizer
+
+    text = 'Hello World'
```
dakinggg, line 50: you can delete everything below this for the test (and your dummy class doesn't need to implement anything). the test just needs to test that the object is constructed.

dakinggg (approved):

llmfoundry/registry.py
```diff
@@ -155,6 +155,19 @@
     description=_schedulers_description,
 )
 
+_tokenizers_description = (
+    'The tokenizers registry is used to register tokenizers that implement the transformers.PreTrainedTokenizerBase interface. '
+    +
+    'The tokenizer will be passed to the build_dataloader() and build_composer_model() methods in train.py.'
+)
+tokenizers = create_registry(
```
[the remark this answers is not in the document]
kushalkodn-db: done

tests/tokenizers/test_registry.py
kushalkodn-db: is this better? i removed the tokenize, encode, decode, and __call__ methods.

b-chu (approved):

PR #1387, opened 2024-07-23
Refactor loss function for ComposerMPTCausalLM

irenedea:
Pulls out part of the loss function into helpers
image


snarayan21 (approved):
lgtm, saw there was a test that failed but i think the `get_targets` change addresses that. would like @dakinggg to take a look before merging as well

dakinggg (commented):
can you manual test before and after a simple run and just make sure the loss is the same?

PR #1390, opened 2024-07-23
Avoid race condition in convert text to mds script

dakinggg:
Loading tokenizers from hf also has a race condition. This PR should avoid it by prepopulating the cache from the main process.

Fails before: `debug-race-9-CBEU3C`
Succeeds after: `debug-race-12-OaxlPc`

### DIFF
--- a/llmfoundry/command_utils/data_prep/convert_text_to_mds.py
+++ b/llmfoundry/command_utils/data_prep/convert_text_to_mds.py
@@ -22,7 +22,6 @@ from tqdm import tqdm
 from transformers import AutoTokenizer, PreTrainedTokenizerBase
 
 from llmfoundry.data.data import AbstractConcatTokensDataset
-from llmfoundry.utils.builders import build_tokenizer
 from llmfoundry.utils.data_prep_utils import (
     DownloadingIterable,
     download_file,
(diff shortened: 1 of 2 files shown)


llmfoundry/command_utils/data_prep/convert_text_to_mds.py
```diff
@@ -394,6 +394,13 @@ def convert_text_to_mds(
         reprocess (bool): Whether to always reprocess the given folder of text files
         trust_remote_code (bool): If true, allows custom code to be executed to load the tokenizer
     """
+    # Load the tokenizer once on the main process so that the files are cached to avoid race conditions
+    # in the Hugging Face load code
+    AutoTokenizer.from_pretrained(
```
milocress, line 399:
```suggestion
    _ = AutoTokenizer.from_pretrained(
```

maybe this to indicate that we're using it for the side effect and not the return value? Could take or leave this suggestion tbh

milocress (approved):

llmfoundry/command_utils/data_prep/convert_text_to_mds.py
dakinggg: `_` actually does store it (its just a variable name), so i will leave as is

PR #1391, opened 2024-07-23
Support rope scaling

milocress:
Adds support for llama3 Rope scaling.


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -45,6 +45,7 @@
 import logging
 
 from transformers import PreTrainedModel, PreTrainedTokenizerBase
+from transformers.models.llama.modeling_llama import LlamaRotaryEmbedding
```
ShashankMosaicML, line 48: I think this line is the same as line 57?

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -88,14 +90,59 @@
 log = logging.getLogger(__name__)
 
 
+class InvalidConfigAccessError(KeyError):
+    pass
+
+
+_ALLOWED_LLAMA_CONFIG_KEYS = {
+    # these are the only config keys that are set and are safe to read from
```
dakinggg, line 98:
```suggestion
    # These are the only config keys that are set and are safe to read from
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -88,14 +90,59 @@
...
+    'rope_scaling',
+    'rope_theta',
+    'max_position_embeddings',
+    'hidden_size',
+    'num_attention_heads',
+
+    # not set but llama modeling code tries to read this attribute
```
dakinggg, line 105:
```suggestion
    # Not set but llama modeling code tries to read this attribute
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -88,14 +90,59 @@
...
+    'partial_rotary_factor',
+
+    # benign transformers attributes needed for __init__
```
dakinggg, line 108:
```suggestion
    # Benign transformers attributes needed for __init__
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -88,14 +90,59 @@
...
+    '_get_generation_defaults',
+    'label2id',
+    'id2label',
+    'torch_dtype',
+    'problem_type',
+    '__class__',
+}
+
+
+class PartialLlamaConfig(LlamaConfig):
```
dakinggg, line 118: Can you add a comment explaining why this class exists?

tests/models/test_rope_scaling.py
```diff
@@ -0,0 +1,36 @@
...
+    'rope_hf_config': {
+        'factor': 8.0,
+        'low_freq_factor': 1.0,
+        'high_freq_factor': 4.0,
+        'original_max_position_embeddings': 8192,
+        'type': 'llama3',
+    },
+}
+
+rope_dail_config = {}
```
dakinggg, line 20: delete?

dakinggg (commented):
Note: when transformers 4.43.2 is released tomorrow, we should bump the min version to that in this PR

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -134,6 +181,18 @@ def gen_rotary_embedding(
                 device=
                 'cpu',  # FSDP does not materialize modules with meta buffers, hence device is set to cpu
             )
+        elif rope_hf_config['type'] == 'llama3':
+            llama_rope_config = {**rope_hf_config}
+            llama_rope_config['rope_type'] = rope_hf_config.pop('type')
```
dakinggg, line 186:
```suggestion
            llama_rope_config['rope_type'] = rope_hf_config.get('type')
```

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -134,6 +181,18 @@ def gen_rotary_embedding(
                 device=
                 'cpu',  # FSDP does not materialize modules with meta buffers, hence device is set to cpu
```
snarayan21, line 182: Can you also swap out the `dynamic`, `linear`, and `no_scaling` rope modifications to go through LlamaRotaryEmbedding and PartialLlamaConfig as well? transformers is deprecating the old code path

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -134,6 +181,18 @@ def gen_rotary_embedding(
...
             )
+        elif rope_hf_config['type'] == 'llama3':
+            llama_rope_config = {**rope_hf_config}
+            llama_rope_config['rope_type'] = rope_hf_config.get('type')
+            return LlamaRotaryEmbedding(
+                config=PartialLlamaConfig(
```
snarayan21, line 188: so i'm understanding -- the new class takes in just the complete llama config right

tests/models/test_rope_scaling.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+from transformers.models.llama.modeling_llama import LlamaRotaryEmbedding
```
snarayan21, line 4: can you add a test that checks that the new llama3 embedding type from HF outputs the same embeddings when `apply_scaling` is called as the rope modification code I shared with you?

snarayan21 (commented):
some comments

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -88,14 +90,59 @@
...
+
+        return super().__getattribute__(key)
+
+    def __getitem__(self, key: str):
+        if key not in _ALLOWED_LLAMA_CONFIG_KEYS:
+            raise InvalidConfigAccessError(key)
+
+        return super().__getitem__(key)
+
+    def _get_generation_defaults(self):
```
irenedea, line 132: why do you need this? It always feels a little sketch to override private methods

llmfoundry/models/mpt/modeling_mpt.py
irenedea: nit: docstring pls

llmfoundry/models/mpt/modeling_mpt.py
milocress: yessir

llmfoundry/models/mpt/modeling_mpt.py
milocress: I removed it the override it still works, so will delete.

tests/models/test_rope_scaling.py
milocress: It's needed because the codepath is jank.

tests/models/test_rope_scaling.py
dakinggg: i don't think that is necessary. we don't need to test _everything_ from hf

tests/models/test_rope_scaling.py
dakinggg: oh right, i see

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -108,32 +154,20 @@ def gen_rotary_embedding(
             'cpu',  # FSDP does not materialize modules with meta buffers, hence device is set to cpu
         )
     elif rope_impl == 'hf':
+        llama_rope_config = {**rope_hf_config}
+        llama_rope_config['rope_type'] = rope_hf_config.get('type')
```
dakinggg, line 158:
```suggestion
        llama_rope_config['rope_type'] = rope_hf_config.get('type')
        if 'type' in llama_rope_config:
            del llama_rope_config['type']
```

dakinggg: i think will fix the failing tests

dakinggg (approved):

PR #1395, opened 2024-07-25
Dtensor oom

dakinggg:
Previously we were accidentally keeping extra copies of dtensors in GPU memory, which would cause an OOM. This bug was introduced in https://github.com/mosaicml/llm-foundry/pull/1384. This PR fixes it by not creating extra copies of the full tensor on GPU.

Manual tests:
- [Done] Tested a previous run with dtensors that would OOM
- [Done] Tested a normal run with fp32 and bf16 and confirmed checkpoint sizes were correct to make sure the dtype conversion is still getting applied (there are also unit tests for this)


bigning (approved):
thanks for the fix!

PR #1396, opened 2024-07-26
Add pre register method for mlflow

dakinggg:
Adds a new method to allow subclasses to modify the saved files before registering to mlflow.


irenedea (approved):
lgtm

PR #1399, opened 2024-07-26
Enable passing epsilon when building norm layers

gupta-abhay:
Several public checkpoints use different values of epsilon for normalization layers (for e.g. 1e-6 instead of 1e-5 in gemma2). This enables passing in the right epsilon when needed.


dakinggg (approved):
lgtm

PR #1402, opened 2024-07-28
Propagate `name_or_path` through HF Checkpointer

snarayan21:
Makes sure to propagate model `name_or_path` through HF checkpointer (similar to how we do for generation config).

Note that since [PretrainedConfig](https://huggingface.co/docs/transformers/v4.43.3/en/main_classes/configuration#transformers.PretrainedConfig) always has the attr `name_or_path`, we don't need to check `hasattr`. All instances of PreTrainedModel will have `name_or_path` (which defaults to "").

Note that if the `pretrained_model_name_or_path` is a local path then `name_or_path` will also be a local path. And so when saving out LoRA adapters, for example, the adapter_config.json will have `base_model_name_or_path` as the local path. This PR does not solve this problem. IMO we should add a new optional arg to ComposerHFCausalLM that's the true model name and propagate that through model transforms & HF checkpointer instead of `name_or_path`.

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -517,6 +517,7 @@ class HuggingFaceCheckpointer(Callback):
                     new_model_instance.generation_config.update(
                         **original_model.generation_config.to_dict(),
                     )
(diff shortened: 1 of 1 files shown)


snarayan21:
not needed rn

PR #1406, opened 2024-07-29
Remove curriculum learning error when duration less than saved timestamp

b-chu:
A run may crash after a checkpoint successfully saves but before an iteration is incremented. This can occur when checkpoint_saver is defined in terms of ba, so the checkpoint event happens before the iteration increment check happens. In the scenario of a crash, the resumption would also fail because the timestamp saved is greater than the duration of the current datamix. This change emits a warning instead of always erroring.


snarayan21 (commented):
mind providing an example of when this would be necessary? i'm finding it hard to follow

milocress (approved):

PR #1408, opened 2024-07-29
Enable QuickGelu Function for CLIP models

gupta-abhay:
- CLIP Models need the quick_gelu activation function
- Sourcing from [HF](https://github.com/huggingface/transformers/blob/main/src/transformers/activations.py#L90), but implemented as Callable for resolve_fn semantics

### DIFF
--- a/tests/models/layers/test_ffn.py
+++ b/tests/models/layers/test_ffn.py
@@ -1,0 +1,62 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+import torch.nn as nn
+import torch.distributed as dist
+
+from llmfoundry.models.layers.layer_builders import build_ffn
+
+
+@pytest.mark.gpu
+def test_quickgelu_activation():
+    d_model = 32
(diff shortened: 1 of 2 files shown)


dakinggg (commented):
can you please add a simple unit test that exercises this code path?

gupta-abhay:
@dakinggg - added something, not sure if this is what you had in mind.

tests/models/layers/test_ffn.py
```diff
@@ -0,0 +1,62 @@
...
+    device: torch.device = torch.device(f'cuda:{rank}')
+
+    ffn1 = build_ffn(
+        name=ffn_config['ffn_type'],
+        d_model=d_model,
+        expansion_ratio=expansion_ratio,
+        device=device,
+        bias=not no_bias,
+        ffn_kwargs=ffn_config,
+    )
```
dakinggg, line 33: can you also assert that the ffn1.act is what you expect?

dakinggg:
@gupta-abhay thats much more than what i meant :D but no objection. left one comment, which is what I actually meant

dakinggg (approved):

PR #1409, opened 2024-07-29
Add better error handling for FSDP activation checkpointing and use_orig_params

j316chuck:
Add better error handling for FSDP activation checkpointing and use_orig_params

### DIFF
--- a/scripts/train/train.py
+++ b/scripts/train/train.py
@@ -1,6 +1,571 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
+import gc
+import logging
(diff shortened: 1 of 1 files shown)


snarayan21 (commented):
@j316chuck is this what you had meant to commit? modifications to all of train.py?

milocress:
> @j316chuck is this what you had meant to commit? modifications to all of train.py?

I think these changes should be in [`llmfoundry/command_utils/train.py`](https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/command_utils/train.py)

dakinggg (changes requested):
This error should be in Composer, not Foundry

PR #1410, opened 2024-07-29
Log original config

josejg:
Follow-up to #1351. This PR changes the logging behavior to log the originally submitted config. The current behavior applies transformations, then merges them into the original config. Because of the merging, the logged config might not be a valid config, making reproducible runs harder.

### DIFF
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -341,8 +341,6 @@ def make_dataclass_and_log_config(
         transforms,
     )
 
-    logged_cfg.update(unstructured_config, merge=True)
-
     arg_config_keys = set(unstructured_config.keys())
     extraneous_keys = set.difference(arg_config_keys, dataclass_fields)
 


llmfoundry/utils/config_utils.py
```diff
@@ -341,8 +341,6 @@ def make_dataclass_and_log_config(
         transforms,
     )
 
-    logged_cfg.update(unstructured_config, merge=True)
```
dakinggg, line 344: I think you want to move the `logged_cfg = copy.deepcopy(unstructured_config)` line down here after the transforms right?

josejg: I considered both options, and decided to err on the side of reproducibility. In particular, I was thinking that if there is a faulty transformation, then once the data is logged post-transformation there is no way to get back the original config and reproduce the issue. In that case it would be preferable to log the original config to be able to reproduce the same behavior .

dakinggg: hmm ok. i guess i can see it both ways, but i accept.

PR #1414, opened 2024-07-30
Additional registry entrypoint documentation

dakinggg:
Adds some additional documentation for using the entrypoints registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ dependencies = [
 ]
 
 # Note: Even though in python code, this would be llmfoundry.registry.loggers,
-# when specified in the entry_points, it should be "llmfoundry_loggers". That is,
+# when specified in the entry_points, it has to be "llmfoundry_loggers". That is,
 # the segments of the name should be joined by an _ in the entry_points section.
 [project.entry-points."llmfoundry_loggers"]
 my_logger = "foundry_registry.loggers:MyLogger"
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ dependencies = [
 my_logger = "foundry_registry.loggers:MyLogger"
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -309,10 +309,15 @@ dependencies = [
...
 ]
 
+# Note: Even though in python code, this would be llmfoundry.registry.loggers,
+# when specified in the entry_points, it should be "llmfoundry_loggers". That is,
+# the segments of the name should be joined by an _ in the entry_points section.
 [project.entry-points."llmfoundry_loggers"]
 my_logger = "foundry_registry.loggers:MyLogger"
 ```
 
+If developing new components via entrypoints, it is important to note that Python entrypoints are global to the Python environment. This means that if you have multiple packages that register components with the same key, the last one loaded will be the one used. This can be useful for overriding components in LLM Foundry, but can also lead to unexpected behavior if not careful. Additionally, if you change the pyproject.toml, you will need to reinstall the package for the changes to take effect. You can do this quickly by installing with `pip install -e . --no-deps` to avoid reinstalling dependencies.
```
dakinggg, line 319:
```suggestion
If developing new components via entrypoints, it is important to note that Python entrypoints are global to the Python environment. This means that if you have multiple packages that register components with the same key, the last one installed will be the one used. This can be useful for overriding components in LLM Foundry, but can also lead to unexpected behavior if not careful. Additionally, if you change the pyproject.toml, you will need to reinstall the package for the changes to take effect. You can do this quickly by installing with `pip install -e . --no-deps` to avoid reinstalling dependencies.
```

mvpatel2000 (approved):

README.md
```diff
@@ -309,10 +309,15 @@ dependencies = [
     "llm-foundry",
 ]
 
+# Note: Even though in python code, this would be llmfoundry.registry.loggers,
+# when specified in the entry_points, it should be "llmfoundry_loggers". That is,
```
josejg, line 313:
```suggestion
# when specified in the entry_points, it has to be "llmfoundry_loggers". That is,
```

PR #1415, opened 2024-07-31
Read Package Version Better

eitanturok:
In [composer](https://github.com/mosaicml/composer/blob/dev/composer/_version.py), [streaming](https://github.com/mosaicml/streaming/blob/main/streaming/_version.py), and [MegaBlocks](https://github.com/databricks/megablocks/pull/128), we read the package version in the `setup.py` file from the `_version.py` file. However, in llm-foundry we are missing a `_version.py` file and instead read the version from the `llmfoundry/__init__.py` file. This PR changes this.

To ensure consistency with our other repos, I put the package version in the `_version.py` file. Having a `_version.py` file better follows good design principles as it is better to separate the version out into its own file.


snarayan21 (commented):
minor change, other than that lgtm

setup.py
```diff
@@ -15,17 +15,15 @@
 _REPO_REAL_PATH = os.path.dirname(os.path.realpath(__file__))
 _PACKAGE_REAL_PATH = os.path.join(_REPO_REAL_PATH, _PACKAGE_DIR)
 
-# Read the repo version
+# Read the composer version
```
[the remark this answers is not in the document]
eitanturok: Fixed. Good catch.

snarayan21 (approved):
lgtm

llmfoundry/_version.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""The LLM Foundry Version."""
+
+__version__ = '0.10.0'
```
mvpatel2000, line 6:
```suggestion
__version__ = '0.11.0.dev'
```

llmfoundry/__init__.py
```diff
@@ -87,5 +87,3 @@
     'tokenizers',
     'utils',
 ]
-
-__version__ = '0.11.0.dev0'
```
dakinggg, line 91: You need to reexport this here so it is still available as `llmfoundry.__version__` (https://github.com/mosaicml/composer/blob/722888eeb577804a4f6de2fae47fc38f42d9bcbd/composer/__init__.py#L6)

eitanturok: Just exported it here.

eitanturok:
Weird, I initially imported `__version__` into the `__init__.py` file. Looking at the GA logs [here](https://github.com/mosaicml/llm-foundry/actions/runs/10185658717/job/28175679399?pr=1415#step:4:754), turns out the `pycln` hook was removing `__version__`, probably because it was an unused import. 

To fix this, I added `__version__` to the `__all__`.

dakinggg (approved):

PR #1421, opened 2024-08-02
Remove type ignore

dakinggg:
The typing issue has been fixed in composer, so the type ignore is no longer needed.


snarayan21 (approved):
nice

PR #1422, opened 2024-08-03
resize embeddings fix for hf checkpointer

jdchang1:
Added a check to call resize token embeddings in huggingface checkpointer.


dakinggg:
We are discussing this PR offline to see if it is necessary

jdchang1:
Found a fix without needing this PR. Thank you!

PR #1430, opened 2024-08-05
Raise error when not enough data when converting text to MDS 

KuuCi:
In response to https://databricks.atlassian.net/browse/MCLOUD-4727, where users were seeing that if they didn't have enough data, no MDS shards would be made, but no error is thrown until much later.

### DIFF
--- a/llmfoundry/command_utils/data_prep/convert_text_to_mds.py
+++ b/llmfoundry/command_utils/data_prep/convert_text_to_mds.py
@@ -29,6 +29,7 @@ from llmfoundry.utils.data_prep_utils import (
     merge_shard_groups,
 )
 from llmfoundry.utils.exceptions import (
+    DatasetTooSmallError,
     InputFolderMissingDataError,
     OutputFolderNotEmptyError,
 )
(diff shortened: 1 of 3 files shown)


llmfoundry/command_utils/data_prep/convert_text_to_mds.py
```diff
@@ -468,6 +469,11 @@ def convert_text_to_mds(
             trust_remote_code,
         )
 
+    index_path = os.path.join(local_output_folder, 'index.json')
+    with open(index_path, 'r') as index_file:
+        if not json.load(index_file)['shards']:
+            raise ValueError('Input data was too small, no shards written to output folder.')
```
dakinggg: @jjanezhang will want this to be a custom error like the other ones i think?

jjanezhang: Yes! we should add a custom error here telling the user they have too few dataset examples

KuuCi:
vincent-test-FBzlIg for manual test

jjanezhang (approved):
lgtm, if the input folder is not helpful to surface to ft users, then no need to pass it

llmfoundry/utils/exceptions.py
```diff
@@ -348,6 +349,14 @@ def __init__(self, dataset_name: str, split: str) -> None:
         super().__init__(message, dataset_name=dataset_name, split=split)
 
 
+class DatasetTooSmallError(UserError):
+    """Error thrown when the dataset is too small to be processed."""
+
+    def __init__(self) -> None:
+        message = f'Your dataset is too small to be converted to MDS.'
```
dakinggg, line 356:
```suggestion
        message = f'Your dataset is too small and produced no complete samples during preprocessing. Please provide more data.'
```

dakinggg (approved):

PR #1434, opened 2024-08-07
Automatically get the portion of the dataset config that is constructor args

dakinggg:
Incremental improvement to remove the hardcoding of some of the dataset config keys in the validation function.
(1) Makes the streaming constructor class of the dataset constructor a property so that it can be accessed on the class and/or overridden
(2) Removes the dataset constructor args from the hardcoded validation list and uses `inspect` to get them dynamically instead
(3) Makes the allowed keys an arg to the validation function to give the caller more control

TODO:
- [x] Run the regression tests against this PR

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -1,5 +1,6 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
+import inspect
 import logging
 import os
 from typing import Any, Dict, Optional, Tuple, Union
@@ -17,6 +18,8 @@ from llmfoundry.data.finetuning.collator import (
     validate_target_settings,
 )
 from llmfoundry.data.finetuning.tasks import (
+    DEFAULT_TARGET_PROMPTS,
+    DEFAULT_TARGET_RESPONSES,
     DOWNLOADED_FT_DATASETS_DIRPATH,
     SUPPORTED_EXTENSIONS,
     dataset_constructor,
@@ -38,10 +41,6 @@ __all__ = [
 
 # HuggingFace hardcodes the ignore index to -100
 _HF_IGNORE_INDEX = -100
-
-# Default settings to use for target responses and target prompts
-_DEFAULT_TARGET_RESPONSES = 'last'
-_DEFAULT_TARGET_PROMPTS = 'none'
 
 
 def build_finetuning_dataloader(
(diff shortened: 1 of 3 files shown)


llmfoundry/data/finetuning/tasks.py
```diff
@@ -851,6 +854,14 @@ def build_from_hf(
         Returns:
             Dataset: The tokenized dataset.
         """
+        if hf_kwargs is None:
+            hf_kwargs = {}
```
b-chu, line 858: Default this to empty dict

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -171,7 +180,25 @@ def build_finetuning_dataloader(
         given a starting workload YAML.
     """
     dataset_cfg = dataset
-    _validate_config(**dataset_cfg)
+    is_streaming = dataset_cfg.get('remote') is not None or dataset_cfg.get(
+        'streams',
+    ) is not None
+    if is_streaming:
+        dataset_constructor_keys = inspect.signature(
```
b-chu, line 187: What does inspect.signature do?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -229,30 +254,15 @@ def build_finetuning_dataloader(
...
-            batching_method=dataset_cfg.get('batching_method', 'random'),
-            max_seq_len=dataset_cfg['max_seq_len'],
-            allow_unsafe_types=dataset_cfg.get('allow_unsafe_types', False),
             replication=replication_factor,
             packing_ratio=dataloader_batch_size / dataset_batch_size,
+            **{
+                k: v
+                for k, v in dataset_cfg.items()
+                if k in dataset_constructor_keys and
+                k not in {'packing_ratio', 'streams'}
```
b-chu, line 264: Move this outside the function

b-chu: Though it seems like it'd be generally useful to have a util function that takes in a function and gets you the parameters and parses your dictionary to only contain valid arguments for that function

b-chu (commented):
Thanks for the PR!

llmfoundry/data/finetuning/tasks.py
dakinggg: you can't default mutable objects in python arguments because then each call will share the same object

dakinggg: did you mean something else?

llmfoundry/data/finetuning/dataloader.py
dakinggg: https://docs.python.org/3/library/inspect.html#inspect.signature, gets the function signature

llmfoundry/data/finetuning/dataloader.py
dakinggg: i don't think we really need a helper for a dictionary filter, but yeah, i can not inline it

llmfoundry/data/finetuning/tasks.py
b-chu: Nope, forgot about this, my bad

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -171,7 +180,25 @@ def build_finetuning_dataloader(
         given a starting workload YAML.
     """
     dataset_cfg = dataset
-    _validate_config(**dataset_cfg)
+    is_streaming = dataset_cfg.get('remote') is not None or dataset_cfg.get(
+        'streams',
+    ) is not None
```
b-chu, line 185:
```suggestion
    is_streaming = (dataset_cfg.get('remote') is not None or dataset_cfg.get('streams') is not None)
```
I think reformatting this will give you better wrapping?

b-chu (approved):
Thanks for the PR!

PR #1435, opened 2024-08-07
Make autopacking faster

b-chu:
Adds a profiling flag to collators to avoid unnecessary padding.

1. Seq2Seq collator pads to the length of the longest processed example instead of max_seq_len
2. The packing collator doesn't concatenate tensors when profiling
3. The packing collator returns early instead of repadding the freshly concatenated tensors to max_seq_len

Tested with reference runs:

`autopacking-oom-mpt-baseline-Pr7MgS`
- global_train_batch_size: 256
- device_train_microbatch_size: 16
- num_gpus: 8
- max_seq_len: 131072
- packing ratio: 1310.7
- PR packing ratio: 1310.7

`autopacking-oom-mpt-baseline-CPadi0`
- global_train_batch_size: 8
- device_train_microbatch_size: 1
- num_gpus: 8
- max_seq_len: 32768
- packing ratio: 190.1
- PR packing ratio: 190.1

- global_train_batch_size: 64
- device_train_microbatch_size: 1
- num_gpus: 64
- max_seq_len: 131072
- PR packing ratio: 897.1 `autopacking-oom-mpt-fix-mcV61Z`

### DIFF
--- a/llmfoundry/data/finetuning/collator.py
+++ b/llmfoundry/data/finetuning/collator.py
@@ -224,6 +224,10 @@ class Seq2SeqFinetuningCollator:
             sizes. Default: ``False`` ensures that all sequences are max_seq_len.
         batch_metadata (dict, optional): A dictionary of metadata which will be added
             to the batch.
+        pad_to_longest (bool, optional): Whether to pad to the longest sequence,
+            which may result in smaller but inconsistent batch sizes. This is
+            primarily used to profile packing.
+            Default: ``False`` ensures that all sequences are max_seq_len.
     """
 
     def __init__(
@@ -235,6 +239,7 @@ class Seq2SeqFinetuningCollator:
         target_prompts: str = 'none',
         allow_pad_trimming: bool = False,
         batch_metadata: Optional[Dict[str, Any]] = None,
+        pad_to_longest: bool = False,
     ):
         self.tokenizer = tokenizer
         self.max_seq_len = max_seq_len
@@ -246,6 +251,8 @@ class Seq2SeqFinetuningCollator:
         # Trimming will always be skipped on at least the first __call__
         self._allow_pad_trimming = allow_pad_trimming
         self._seen_first_batch = False
+
+        self._pad_to_longest = pad_to_longest
 
         illegal_keys = [
             'input_ids',
@@ -320,24 +327,34 @@ class Seq2SeqFinetuningCollator:
     ) -> Dict[str, torch.Tensor]:
         # Steps explained in comments
         processed_examples = []
-        for example in examples:
-            input_ids, labels = stitch_turns_decoder_only(
+        input_ids_and_labels = [
+            stitch_turns_decoder_only(
                 example_turns=example['turns'],
                 target_prompts=self.target_prompts,
                 target_responses=self.target_responses,
                 eos_token_id=self.tokenizer.eos_token_id,
-            )
+            ) for example in examples
+        ]
 
+        if self._pad_to_longest:
+            max_seq_len = max([
+                len(input_ids) for input_ids, _ in input_ids_and_labels
+            ])
+            max_seq_len = min(max_seq_len, self.max_seq_len)
+        else:
+            max_seq_len = self.max_seq_len
+
+        for input_ids, labels in input_ids_and_labels:
             orig_size = len(input_ids)
             # We may need to truncate the input_ids / labels in order to maintain max_seq_len
-            if orig_size > self.max_seq_len:
-                input_ids = input_ids[:self.max_seq_len]
-                labels = labels[:self.max_seq_len]
+            if orig_size > max_seq_len:
+                input_ids = input_ids[:max_seq_len]
+                labels = labels[:max_seq_len]
 
                 # Check to make sure there are still loss-generating tokens. Error if not.
                 if len([l for l in labels if l != _HF_IGNORE_INDEX]) == 0:
                     raise ValueError(
-                        f'Truncating to max_seq_len={self.max_seq_len} has removed all loss-generating tokens. ' +\
+                        f'Truncating to max_seq_len={max_seq_len} has removed all loss-generating tokens. ' +\
                         f'Pre-truncation sequence length was {orig_size}. ' +\
                         'This sample should have been filtered out before reaching the collator. If using ' +\
                         'pre-tokenized streaming data, this may have resulted from using different ' +\
@@ -348,7 +365,7 @@ class Seq2SeqFinetuningCollator:
                 # Still issue a warning when truncating
                 if not self._warned_truncated:
                     warnings.warn(
-                        f'Truncating sequence of length={orig_size} to fit max_seq_len={self.max_seq_len}. ' +\
+                        f'Truncating sequence of length={orig_size} to fit max_seq_len={max_seq_len}. ' +\
                         f'If truncation is a problem, consider increasing max_seq_len.',
                     )
                     self._warned_truncated = True
@@ -358,7 +375,7 @@ class Seq2SeqFinetuningCollator:
             # Annoyingly, we need to pad everything but input_ids
             # and attention_mask ourselves
             n_total = len(input_ids)
-            i_pad = [_HF_IGNORE_INDEX] * (self.max_seq_len - n_total)
+            i_pad = [_HF_IGNORE_INDEX] * (max_seq_len - n_total)
             if self.tokenizer.padding_side == 'left':
                 labels = i_pad + labels
             else:
@@ -376,7 +393,7 @@ class Seq2SeqFinetuningCollator:
         batch = self.tokenizer.pad(
             processed_examples,
             padding='max_length',
-            max_length=self.max_seq_len,
+            max_length=max_seq_len,
             return_tensors='pt',
         )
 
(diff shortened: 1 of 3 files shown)


llmfoundry/data/finetuning/collator.py
```diff
@@ -320,24 +323,29 @@ def _process_and_batch_decoder_only(
...
-            )
+        input_ids_and_labels = [stitch_turns_decoder_only(
+            example_turns=example['turns'],
+            target_prompts=self.target_prompts,
+            target_responses=self.target_responses,
+            eos_token_id=self.tokenizer.eos_token_id,
+        ) for example in examples]
+
+        if self._is_profiling:
+            max_seq_len = max([len(input_ids) for input_ids, _ in input_ids_and_labels])
```
irenedea, line 334: Is this doing almost the same thing as `allow_pad_trimming`? @alextrott16, do you have any thoughts here on how/if we can consolidate this functionality? We're avoiding padding to the max seq length because it's too slow at super long seq length and with packing, we trim the padding off anyway.

llmfoundry/data/finetuning/collator.py
```diff
@@ -235,6 +235,7 @@ def __init__(
         target_prompts: str = 'none',
         allow_pad_trimming: bool = False,
         batch_metadata: Optional[Dict[str, Any]] = None,
+        is_profiling: bool = False,
```
irenedea, line 238: let's rename this flag. Would prefer a flag that describes the functionality of what happens rather than an overall state so that the flag makes sense if you use the collator independently. 

Ideally, if the below comment makes sense, we can just use the allow_pad_trimming flag

llmfoundry/data/packing.py
```diff
@@ -277,43 +309,21 @@ def _first_fit_bin_packing(
...
-    keys = packed_examples[0].keys()
-    batch = {}
-    for key in keys:
-        batch[key] = torch.stack([
-            pad_tensor(example[key], pad_vals[key])
-            for example in packed_examples
-        ])
-    return batch
+def _pad_tensor(tensor: torch.Tensor, pad_value: int, max_seq_len: int, padding_side: str):
+    return tensor
```
irenedea, line 313: remove?

llmfoundry/data/packing.py
```diff
@@ -367,6 +379,8 @@ def auto_packing_ratio(
         num_packing_ratios=num_packing_ratios,
         device_batch_size=device_batch_size,
     )
+    dist.barrier()
```
irenedea, line 382: are you planning to keep this dist barrier?

llmfoundry/data/packing.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         pad_token_id: int,
         padding_side: Literal['left', 'right'],
         max_leftover_bins_to_keep: Optional[int] = None,
+        is_profiling: bool = False,
```
irenedea, line 33: can you add a doc string for this class and its args? ty!

llmfoundry/data/packing.py
```diff
@@ -183,6 +211,7 @@ def _first_fit_bin_packing(
     num_bins: int,
     max_bin_size: int,
     existing_bins: List[Tuple[int, Dict[str, torch.Tensor]]],
+    is_profiling: bool = False,
```
irenedea, line 214: I think it'll make more sense to have this as a class method and just use `self.is_profiling`

irenedea: Similar reason to wanting to rename the is_profiling param in Seq2SeqCollator. It doesn't make a lot of sense if you look at this function individually.

llmfoundry/data/finetuning/collator.py
irenedea: also since this is only supported for decoder_only_format, please add a checks that errors if this is not the case

llmfoundry/data/finetuning/collator.py
alextrott16: As discussed offline, yeah this has _nearly_ the same behavior as `allow_pad_trimming` but the latter doesn't trim the first batch. And that's a problem during profiling since it just grabs a single batch.

So, I recommend sticking with the way @b-chu has it implemented here :)

llmfoundry/data/packing.py
b-chu: Done

llmfoundry/data/packing.py
b-chu: Removed

llmfoundry/data/packing.py
b-chu: Removed

llmfoundry/data/packing.py
b-chu: Done

llmfoundry/data/finetuning/collator.py
b-chu: Renamed and added support for _process_and_batch_encoder_decoder

llmfoundry/data/finetuning/collator.py
```diff
@@ -320,24 +327,33 @@ def _process_and_batch_decoder_only(
...
                 example_turns=example['turns'],
                 target_prompts=self.target_prompts,
                 target_responses=self.target_responses,
                 eos_token_id=self.tokenizer.eos_token_id,
-            )
+            ) for example in examples
+        ]
+
+        if self._pad_to_longest:
+            max_seq_len = max([
```
irenedea, line 340: You want to also include the true max seq len here
```suggestion
            max_seq_len = max([self.max_seq_len] + [
```

llmfoundry/data/packing.py
```diff
@@ -492,6 +510,7 @@ def profile(raw_batch_size: int) -> Tuple[Optional[float], Optional[float]]:
             pad_token_id=0,  # <-- Doesn't need to be correct for profiling
             padding_side='left',  # <-- Doesn't need to be correct for profiling
             max_leftover_bins_to_keep=max_leftovers_to_keep,
+            is_profiling=False,
```
irenedea, line 513: shouldn't this be true?

b-chu: Done

llmfoundry/data/finetuning/collator.py
irenedea: oh my bad it should be 
```
max_seq_len = min(self.max_seq_len, max([...]))
```
right?

irenedea: I think the test was failing because the examples were not getting truncated

llmfoundry/data/packing.py
```diff
@@ -177,143 +317,22 @@ def _combine_in_place(
...
-    batch = {}
-    for key in keys:
-        batch[key] = torch.stack([
-            pad_tensor(example[key], pad_vals[key])
-            for example in packed_examples
-        ])
-    return batch
+def _pad_tensor(
+    tensor: torch.Tensor, pad_value: int, max_seq_len: int, padding_side: str
+):
```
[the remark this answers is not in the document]
b-chu: Done

llmfoundry/data/packing.py
```diff
@@ -20,7 +20,17 @@
...
+    """Utility collator for packing to reduce padding.
+
+    Args:
+        collator (Callable): The base collator to use.
+        target_batch_size (int): The number of bins.
+        max_seq_len(int): The maximum sequence length of a bin.
+        pad_token_id (int): The padding token id.
+        padding_side (Literal['left', 'right']): The side to pad on.
+        max_leftover_bins_to_keep (Optional[int]): The number of leftover bins to keep.
+        is_profiling (bool): Whether the collator is being used for profiling.
```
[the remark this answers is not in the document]
b-chu: Done

llmfoundry/data/packing.py
```diff
@@ -115,15 +130,140 @@ def _pack_trimmed_examples(
         self.n_packed_examples += self.out_size
         self._leftover_bins = leftover_bins[:self.max_leftover_bins_to_keep]
 
+        if self._is_profiling:
```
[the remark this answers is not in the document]
b-chu: Done

llmfoundry/data/finetuning/collator.py
b-chu: Yea fixed it

irenedea (approved):
lgtm! just remove print statements :)

llmfoundry/data/packing.py
```diff
@@ -115,15 +130,140 @@ def _pack_trimmed_examples(
...
+                bins.append((size, example))
+                continue
+
+            # Add it to the first bin it fits in
+            added = False
+            for bidx in range(len(bins)):
+                if bins[bidx][0] + size <= max_bin_size:
+                    bin_size, packed_example = bins.pop(bidx)
+                    bin_size = bin_size + size
+                    if not self._is_profiling:
```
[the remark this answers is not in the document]
irenedea: discussed offline, just add some comments describing that we aren't going to combine the tensor when profiling for efficiency so we just return the first example, which isn't "correct"

llmfoundry/data/packing.py
```diff
@@ -86,24 +102,28 @@ def pack(self, batch: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
             ]
         # Cut everything down to size
         sizes, trimmed_examples = _trim_batch(batch)
-        return self._pack_trimmed_examples(trimmed_examples, sizes)
+        packed_batch = self._pack_trimmed_examples(trimmed_examples, sizes)
+        assert packed_batch is not None
```
irenedea, line 106: use ValueError over assert
```suggestion
        if packed_batch is None:
            raise ValueError(...)
```

b-chu: This is purely for pyright, not actual error handling

irenedea: also can you add 
```
if self._is_profiling:
    raise ValueError('`pack` should not be called while profiling. Please use `_pack_trimmed_examples`.')
```

b-chu: `pack` is never called when profiling

irenedea: So that someone doesn't call `pack` when creating a BinPackCollator in profiling mode

PR #1436, opened 2024-08-07
Remove flash patching for HF

dakinggg:
HF has fixed the bugs that required the monkeypatching, so we can remove it now.

Manual test that loss, throughput, and memory are the same before and after
Screenshot 2024-08-07 at 12 45 38 PM
Screenshot 2024-08-07 at 12 45 47 PM
Screenshot 2024-08-07 at 12 45 57 PM

### DIFF
--- a/tests/models/hf/test_hf_config.py
+++ b/tests/models/hf/test_hf_config.py
@@ -277,4 +277,5 @@ def test_use_flash():
     )
     assert isinstance(attention_layer, flash_attn_class)
 
+    # Make sure that HF has not cast the parameters to bf16
     assert next(model.parameters()).dtype == torch.float32
--- a/tests/models/hf/test_hf_config.py
+++ b/tests/models/hf/test_hf_config.py
@@ -7,9 +7,11 @@ from typing import Any, Dict, Mapping
 from unittest.mock import Mock, patch
 
 import pytest
+import torch
 from omegaconf import OmegaConf as om
 from transformers import PretrainedConfig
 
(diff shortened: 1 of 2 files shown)


tests/models/hf/test_hf_config.py
```diff
@@ -235,3 +237,44 @@ def test_nested_override():
...
+@pytest.mark.gpu
+def test_use_flash():
+    model_cfg = {
+        'name': 'hf_causal_lm',
+        'pretrained_model_name_or_path': 'codellama/CodeLlama-7b-hf',
+        'config_overrides': {
+            'num_hidden_layers': 2,
+            'hidden_size': 32,
+            'intermediate_size': 64,
+            'torch_dtype': 'bfloat16',
```
snarayan21, line 251: is it necessary that the `torch_dtype` in the config is set to bf16?

I thought we would want to test that the model loaded in fp32 is able to use flash attn?

snarayan21 (approved):
clarification comment on test, probably my misunderstanding. other than that, lgtm!

always hype to remove patches

tests/models/hf/test_hf_config.py
dakinggg: ah yeah, i snuck into this test that hf doesn't silently put the master weights into the specified dtype :P

snarayan21: so that's why you have the check that the params are in fp32 then?

can you add a comment clarifying that :)

dakinggg: done

PR #1437, opened 2024-08-07
Add special errors for bad chat/ift types

milocress:
Adds the `InvalidExampleTypeError` and `InvalidMessageTypeError` for when `example` is not a `Mapping` and `message` is not an `Iterable`, respectively.

PR #1438, opened 2024-08-07
UnknownExampleType Error Message display only keys

ethanma-db:
Display only the keys, need to cast to list first:

https://databricks.atlassian.net/browse/MCLOUD-4621

Forgot to push this lol


dakinggg (commented):
Please include a manual test run that shows that this works

milocress:
This is addressed in #1437

PR #1450, opened 2024-08-13
Enabled generalizing build_inner_model in ComposerHFCausalLM

gupta-abhay:
- Will allow the static function to be called externally
    - Users can now use this function (instead of re-writing it for all transformer classes)
    - Can call hyper-specfic function classes as well.

### DIFF
--- a/llmfoundry/models/hf/hf_causal_lm.py
+++ b/llmfoundry/models/hf/hf_causal_lm.py
@@ -234,10 +234,13 @@ class ComposerHFCausalLM(HuggingFaceModelWithFSDP):
                 + 'Please `pip install llm-foundry[gpu]`.',
             )
 
-        assert hasattr(
-            model_cls,
-            'from_pretrained',
-        ), 'HF Model class is not supported, check arguments to function call!'
+        if not (
+            hasattr(model_cls, 'from_pretrained') and
+            hasattr(model_cls, 'from_config')
+        ):
+            raise AttributeError(
+                f'{model_cls=} has missing `from_pretrained` and `from_config` support.',
+            )
 
         # Hugging Face copies the modules into the
         # transformers modules cache. On particular systems, this operation seems to cause contention between
--- a/llmfoundry/models/hf/hf_causal_lm.py
+++ b/llmfoundry/models/hf/hf_causal_lm.py
@@ -9,6 +9,7 @@ import warnings
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     List,
     Optional,
@@ -193,6 +194,8 @@ class ComposerHFCausalLM(HuggingFaceModelWithFSDP):
         config_overrides: Dict[str, Any],
         load_in_8bit: bool,
         pretrained: bool,
+        config_fn: Optional[Callable] = AutoConfig,
+        model_fn: Optional[Callable] = AutoModelForCausalLM,
         prepare_for_fsdp: bool = False,
     ) -> Union[PreTrainedModel, 'PeftModel']:
         """Builds the inner model for the ComposerHFCausalLM.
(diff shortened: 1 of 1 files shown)


llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -193,6 +194,8 @@ def build_inner_model(
         config_overrides: Dict[str, Any],
         load_in_8bit: bool,
         pretrained: bool,
+        config_fn: Optional[Callable] = AutoConfig,
+        model_fn: Optional[Callable] = AutoModelForCausalLM,
```
dakinggg, line 198: please add these to the docstring

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -193,6 +194,8 @@ def build_inner_model(
         config_overrides: Dict[str, Any],
         load_in_8bit: bool,
         pretrained: bool,
+        config_fn: Optional[Callable] = AutoConfig,
```
dakinggg, line 197: do you really intend for these to be functions? They should be classes that support `.from_pretrained`/`.from_config`, which return model/config as appropriate, right?

gupta-abhay: true - but was not sure what "type" they are since these are classes?

gupta-abhay: ok - I think I've fixed this.

dakinggg: Yeah, I mean, the most complete way would be to define the interface that is require to be implemented. I don't think that is necessary. Instead, how about we just specify it as the type `Type` (its not `Optional`, because it can't be `None`). And then in code, we can check `hasattr(auto_config_class, 'from_pretrained')` etc and raise error if not defined. and can also raise errors if they return the wrong thing.

dakinggg: The correct type is actually this private class from transformers ithink https://github.com/huggingface/transformers/blob/9d2ab8824c2a820a0ac9f2cd884ef838ca77b481/src/transformers/models/auto/auto_factory.py#L403

dakinggg: and autoconfig has no base class (https://github.com/huggingface/transformers/blob/9d2ab8824c2a820a0ac9f2cd884ef838ca77b481/src/transformers/models/auto/configuration_auto.py#L859)

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -193,6 +194,7 @@ def build_inner_model(
         config_overrides: Dict[str, Any],
         load_in_8bit: bool,
         pretrained: bool,
+        model_cls: Union[AutoModel, PreTrainedModel] = AutoModelForCausalLM,
```
dakinggg, line 197: technically `issubclass(AutoModelForCausalLM, AutoModel) == False`. I think this line should be
```suggestion
        model_cls: Union[Type, Type[PreTrainedModel]] = AutoModelForCausalLM,
```

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -231,6 +234,14 @@ def build_inner_model(
                 + 'Please `pip install llm-foundry[gpu]`.',
             )
 
+        if not (
+            hasattr(model_cls, 'from_pretrained') and
+            hasattr(model_cls, 'from_config')
+        ):
+            raise AttributeError(
+                f'{model_cls=} has missing `from_pretrained` and `from_config` support.',
```
dakinggg, line 242:
```suggestion
                f'{model_cls=} is missing `from_pretrained` and `from_config` support.',
```

dakinggg (approved):
Please update the PR title to reflect the actual changes before merging

PR #1452, opened 2024-08-14
Add debug logs for model registration

irenedea:
Add debug logs for model registration so we can better track its progress.


dakinggg (commented):
There are already a bunch of logs in the composer function that is called. Does this actually add?

irenedea:
@dakinggg ah yeah you're right! I think this is probably duplicating.

PR #1453, opened 2024-08-14
Implement ruff rules enforcing PEP 585

snarayan21:
Since we don't support python 3.8 anymore, we can get in line with pep 585 and use native python type hints instead. https://peps.python.org/pep-0585/

Added ruff rules to automatically enforce.


dakinggg (approved):

b-chu (approved):
Thanks!

PR #1457, opened 2024-08-15
Add user error for UnicodeDeocdeError in convert text to mds

irenedea:
Addresses the following error:
```
for chunk in iter(partial(f.read, 1000000), ''):
  File "<frozen codecs>", line 322, in decode
[2024-08-15 15:09:09.939000+00:00][15086]: UnicodeDecodeError: 'utf-8' codec can't decode byte 0x95 in position 203337: 
```


dakinggg (approved):
Can you add a test?

irenedea:
@dakinggg added! 
8233d86

PR #1459, opened 2024-08-17
Correct error message for inference wrapper

josejg:
Correct error message for inference wrapper. Afaict openai uses `.message` not `._message` with the version we install (1.3.8) - https://github.com/openai/openai-python/releases/tag/v1.3.8. All future versions also use `.message`


dakinggg (approved):
Looks like it was ._message in pre 1.0

PR #1466, opened 2024-08-19
Update CI tests to v0.1.2

KuuCi:
We were calling `uses: mosaicml/ci-testing/.github/actions/pytest-gpu@v0.1.2` which depends on having scripts in ci-testing ref v0.1.2, which are not in v0.1.0, causing directory not found.

### DIFF
--- a/.github/workflows/pr-gpu.yaml
+++ b/.github/workflows/pr-gpu.yaml
@@ -4,7 +4,7 @@ on:
     branches:
     - main
     - release/*
-  pull_request_target:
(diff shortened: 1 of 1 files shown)


KuuCi:
For manual test, we set pull_request_target to pull_request: https://github.com/mosaicml/llm-foundry/actions/runs/10459273593/job/28962904068?pr=1466

KuuCi:
Failing because updating CI tests, refer to manual test above for proof of it working

PR #1467, opened 2024-08-19
Parent class hf models

jdchang1:
Parent class HuggingFace composer models for easier extendability.

### DIFF
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -193,15 +193,13 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         """
         from llmfoundry.utils.builders import build_metric
 
-        train_metric_names = list(cls.default_train_metrics) + (
-            additional_train_metrics or []
-        )
+        train_metric_names = list(cls.default_train_metrics
+                                 ) + (additional_train_metrics or [])
         train_metrics = [
             build_metric(metric, {}) for metric in train_metric_names
         ] if use_train_metrics else []
-        eval_metric_names = list(cls.default_eval_metrics) + (
-            additional_eval_metrics or []
-        )
+        eval_metric_names = list(cls.default_eval_metrics
+                                ) + (additional_eval_metrics or [])
         eval_metrics = [
             build_metric(metric, {}) for metric in eval_metric_names
         ]
@@ -220,8 +218,7 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         config_overrides: dict[str, Any],
         load_in_8bit: bool,
         pretrained: bool,
-        model_cls: Optional[Union[_BaseAutoModelClass,
-                         PreTrainedModel]] = None,
+        model_cls: Optional[Union[_BaseAutoModelClass, PreTrainedModel]] = None,
         prepare_for_fsdp: bool = False,
     ) -> Union[PreTrainedModel, 'PeftModel']:
         """Builds the inner model for the ComposerHFCausalLM.
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -53,8 +53,8 @@ class BaseHuggingFaceModel(HuggingFaceModel):
 
     model_cls: Union[_BaseAutoModelClass,
                      PreTrainedModel] = AutoModelForCausalLM
-    default_train_metrics: list = []
-    default_eval_metrics: list = []
+    default_train_metrics: tuple = ()
+    default_eval_metrics: tuple = ()
 
     def __init__(
         self,
@@ -193,13 +193,13 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         """
         from llmfoundry.utils.builders import build_metric
 
-        train_metric_names = cls.default_train_metrics + (
+        train_metric_names = list(cls.default_train_metrics) + (
             additional_train_metrics or []
         )
         train_metrics = [
             build_metric(metric, {}) for metric in train_metric_names
         ] if use_train_metrics else []
-        eval_metric_names = cls.default_eval_metrics + (
+        eval_metric_names = list(cls.default_eval_metrics) + (
             additional_eval_metrics or []
         )
         eval_metrics = [
@@ -220,8 +220,8 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         config_overrides: dict[str, Any],
         load_in_8bit: bool,
         pretrained: bool,
-        model_cls: Union[_BaseAutoModelClass,
-                         PreTrainedModel] = AutoModelForCausalLM,
+        model_cls: Optional[Union[_BaseAutoModelClass,
+                         PreTrainedModel]] = None,
         prepare_for_fsdp: bool = False,
     ) -> Union[PreTrainedModel, 'PeftModel']:
         """Builds the inner model for the ComposerHFCausalLM.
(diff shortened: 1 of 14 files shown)


llmfoundry/models/hf/model_wrapper.py
```diff
@@ -76,6 +131,285 @@ def loss(self, outputs: ModelOutput, batch: Mapping):
...
+        cls,
+        pretrained_model_name_or_path: str,
+        pretrained_lora_id_or_path: Optional[str],
+        trust_remote_code: bool,
+        init_device: str,
+        use_flash_attention_2: bool,
+        use_auth_token: bool,
+        config_overrides: dict[str, Any],
+        load_in_8bit: bool,
+        pretrained: bool,
```
dakinggg, line 209: you removed two args here, which is backwards breaking. can we put them back?

llmfoundry/models/hf/model_wrapper.py
```diff
@@ -24,29 +42,68 @@
 # HuggingFace hardcodes the ignore index to -100
 _HF_IGNORE_INDEX = -100
 
+log = logging.getLogger(__name__)
 
 class HuggingFaceModelWithFSDP(HuggingFaceModel):
```
dakinggg, line 47: to avoid this being a backwards compatibility breaking change, how about we make a new base class for everything to inherit from and deprecate HuggingFaceModelWithFSDP (which is not a very good name for the new class anyway)

dakinggg: here is the warning we raise for deprecations: https://github.com/mosaicml/llm-foundry/blob/d3a9ac9b874442657fd0df48afae5de6414e9a2f/llmfoundry/utils/warnings.py#L15

llmfoundry/models/hf/model_wrapper.py
dakinggg: can disregard this since making a new class

dakinggg: although might be nice to maintain it so we can just say "adjust your constructor, but otherwise the interface on the new class is the same as the interface on the causal lm class"

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -71,6 +56,11 @@ class ComposerHFCausalLM(HuggingFaceModelWithFSDP):
         tokenizer (PreTrainedTokenizer): The tokenizer that the model will use.
     """
 
+    model_cls: Union[_BaseAutoModelClass,
+                     PreTrainedModel] = AutoModelForCausalLM
+    default_train_metrics: list = DEFAULT_CAUSAL_LM_TRAIN_METRICS
```
dakinggg, line 61: same

llmfoundry/models/hf/hf_base.py
```diff
@@ -0,0 +1,446 @@
...
+            pretrained_model_name_or_path (str): The pretrained model name or path.
+            pretrained_lora_id_or_path (Optional[str]): The pretrained LORA ID or path.
+            trust_remote_code (bool): Whether to trust remote code.
+            init_device (str): The initialization device.
+            use_flash_attention_2 (bool): Whether to use flash attention 2.
+            use_auth_token (bool): Whether to use an authentication token.
+            config_overrides (Dict[str, Any]): The configuration overrides.
+            load_in_8bit (bool): Whether to load in 8-bit.
+            pretrained (bool): Whether the model is pretrained.
+            model_cls (Union[Type, Type[PreTrainedModel]]): Kept for backwards compatibility.
```
dakinggg, line 239: the current implementation doesn't really keep BC because it doesn't _use_ the `model_cls` arg. I would recommend defaulting it to `None`, and then, if it is not `None`, using it instead of the class attribute. That should preserve all functionality.

dakinggg (commented):
Will want to run the regression tests against this PR before merging. I can show you how to do that once the PR is otherwise done.

llmfoundry/models/hf/hf_t5.py
```diff
@@ -45,23 +45,45 @@ class ComposerHFT5(HuggingFaceModelWithFSDP):
...
     def __init__(
         self,
         tokenizer: PreTrainedTokenizerBase,
         pretrained_model_name_or_path: str,
-        pretrained: Optional[bool] = True,
+        pretrained: bool = True,
         trust_remote_code: bool = True,
         use_auth_token: bool = False,
-        config_overrides: Optional[Mapping] = None,
+        config_overrides: Optional[dict[str, Any]] = None,
```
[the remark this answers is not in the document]
dakinggg: was a comment to myself.

llmfoundry/models/hf/hf_base.py
jdchang1: added a line the sets `model_cls` the way you described. I did change the typing to be Optional for param. Hopefully that doesn't break BC

llmfoundry/models/hf/hf_causal_lm.py
jdchang1: changed to tuple

llmfoundry/models/hf/hf_base.py
```diff
@@ -0,0 +1,446 @@
...
+
+class BaseHuggingFaceModel(HuggingFaceModel):
+    """Wrapper around HuggingFaceModel.
+
+    Base class for HuggingFace based models.
+    """
+
+    model_cls: Union[_BaseAutoModelClass,
+                     PreTrainedModel] = AutoModelForCausalLM
+    default_train_metrics: list = []
```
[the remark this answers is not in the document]
jdchang1: changed to tuple

llmfoundry/models/hf/hf_t5.py
```diff
@@ -45,23 +45,45 @@ class ComposerHFT5(HuggingFaceModelWithFSDP):
         tokenizer (PreTrainedTokenizer): The tokenizer that the model will use.
     """
 
+    model_cls: Union[_BaseAutoModelClass,
+                     PreTrainedModel] = AutoModelForSeq2SeqLM
+    default_train_metrics: list = DEFAULT_ENC_DEC_METRICS
```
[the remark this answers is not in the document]
jdchang1: changed to tuple

llmfoundry/models/hf/model_wrapper.py
```diff
@@ -43,6 +45,12 @@ def __init__(
         peft_config: Optional['PeftConfig'] = None,
         should_save_peft_only: bool = True,
     ):
+        warnings.warn(
+            VersionedDeprecationWarning(
+                '`HuggingFaceModelWithFSDP` is deprecated. In the future please use `BaseHuggingFaceModel`.',
+                remove_version='0.12.0',
```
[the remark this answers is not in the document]
jdchang1: updated to 0.13.0

llmfoundry/models/hf/hf_base.py
```diff
@@ -0,0 +1,449 @@
...
+            use_logits=use_logits,
+            metrics=metrics,
+            eval_metrics=eval_metrics,
+            shift_labels=shift_labels,
+            allow_embedding_resizing=allow_embedding_resizing,
+            peft_config=peft_config_object,
+            should_save_peft_only=should_save_peft_only,
+        )
+
+    def forward(self, batch: Mapping):
```
dakinggg, line 125: This is essentially identical to the forward in the base class (https://github.com/mosaicml/composer/blob/72b2507b8dd63b49891d099c0540740571c100f1/composer/models/huggingface.py#L484-L493), so lets just remove while we're here.

dakinggg (approved):
Regression tests passed, LGTM!

llmfoundry/models/hf/hf_base.py
dakinggg: yes, that is fine, because the new version is _more_ permissive.

PR #1473, opened 2024-08-21
Track loss spikes and high losses during training

joyce-chen-uni:
KillLossSpike callback that 1) compares the train loss against the running average loss, and 2) checks the magnitude of running losses at the end of every batch/training step. This is intended to catch 1) rapid steep spikes and 2) divergent runs respectively.  

If we identify a spike or divergence, we log a message to the run metadata. This message is added to the TRAIN_UPDATED run event in MAPI, so it will be displayed to the user in the run events. The message recommends the user to stop and resubmit the run with a lower learning rate.

This change will make it easier to query spiky runs. We also log the loss window leading up to the identified spike for data analysis purposes.

Once we have collected enough data and feel confident about our cancellation threshold, we will switch to the hard user error `LossSpikeError` which cancels the run without retry and prompts the user to resubmit with a lower learning rate.


mvpatel2000:
@dakinggg to triage review

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,69 @@
...
+        self.outlier_counter = 0
+        self.loss_window = []
+
+    def batch_end(self, state: State, logger: Logger) -> None:
+
+        if not isinstance(state.loss, torch.Tensor):
+            raise NotImplementedError('Multiple losses not supported yet')
+        train_loss = state.loss.item()
+
+        # Only start early stopping once a full window of loss data
```
dakinggg, line 33: can you please factor the detection out into a function that you can then unit test?

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,69 @@
...
+            # Half of the running losses are greater than our "high loss" threshold, after the first window
+            elif (state.timestamp.batch >= self.window_size * 2) and (sum(1 for loss in self.loss_window if loss > self.loss_cap) >= self.window_size / 2):
+                log.info(f'High losses >{self.loss_cap} detected.')
+                for destination in logger.destinations:
+                    if isinstance(destination, MosaicMLLogger):
+                        destination.log_metadata({'high_loss': f'Persistently high (>{self.loss_cap}) training losses detected.'})
+                # raise LossSpikeError()
+
+        else:
+            log.info(f'Full loss window size not reached ({len(self.loss_window)} < {self.window_size}). Collecting loss data...')
```
dakinggg, line 65: I don't think you want to log this as it will log on every batch

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,69 @@
...
+
+        self.loss_window.append(train_loss)
+        if len(self.loss_window) > self.window_size:
+            self.loss_window.pop(0)
```
dakinggg, line 69: you can use a deque with a max length i believe

llmfoundry/callbacks/kill_loss_spike_callback.py
joyce-chen-uni: It should log on every batch up until 100, I'm fine with taking it out though

joyce-chen-uni: Removed

llmfoundry/callbacks/kill_loss_spike_callback.py
joyce-chen-uni: Done!

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,133 @@
...
+                else:
+                    raise LossSpikeError(
+                        outlier_multiplier=self.outlier_multiplier,
+                        running_loss_avg=round(running_loss_avg),
+                        outlier_counter=self.outlier_counter,
+                    )
+
+            elif self.detect_high_losses(int(state.timestamp.batch)):
+                if self.log_only:
+                    for destination in logger.destinations:
```
dakinggg, line 121: lets log it either way

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,101 @@
...
+        window_size (int): The size of the rolling window used to track recent losses. Default is 100.
+        loss_cap (int): The maximum allowable loss. If the training loss consistently exceeds this value, 
+                        it is considered a diverging or unstable run. Default is 10.
+
+    Raises:
+        LossSpikeError: If log_only is False and a loss spike or persistently high loss is detected, this error is
+                        raised to stop the run with an error message.
+    """
+	
+    def __init__(self, log_only:bool, patience:int=4, outlier_multiplier:int=2, window_size:int=100, loss_cap:int=10):
```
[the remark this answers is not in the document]
joyce-chen-uni: Didn't know to run this previously but formatting should be good now!

llmfoundry/utils/exceptions.py
```diff
@@ -387,3 +387,48 @@ class RunTimeoutError(InternalError):
...
+    def __init__(
+        self,
+        loss_cap: float,
+        window_size: int,
+    ) -> None:
+        if loss_cap and window_size:
+            message = f'Training stopped due to consistently high losses. The training loss exceeded the threshold of {loss_cap} \
+                        for more than half of the {window_size} most recent training steps. Please try submitting the run again with a lower learning rate.'
+
+        else:
```
dakinggg, line 428: why is the else needed? it looks to me like it would never get triggered

llmfoundry/utils/exceptions.py
```diff
@@ -387,3 +387,48 @@ class RunTimeoutError(InternalError):
...
+        outlier_multiplier: float,
+        running_loss_avg: int,
+        outlier_counter: int,
+    ) -> None:
+        if outlier_multiplier and running_loss_avg and outlier_counter:
+            message = f'Training stopped due to a loss spike. The training loss was more than {outlier_multiplier} times greater than \
+                        the running average loss (approx. {running_loss_avg}) over {outlier_counter} consecutive training steps. \
+                        Please try submitting the run again with a lower learning rate.'
+
+        else:
```
dakinggg, line 406: same, why is the else needed?

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,131 @@
...
+from composer.loggers import Logger, MosaicMLLogger
+
+from llmfoundry.utils.exceptions import HighLossError, LossSpikeError
+
+log = logging.getLogger(__name__)
+
+__all__ = ['KillLossSpike']
+
+
+class KillLossSpike(Callback):
```
dakinggg, line 22: can you tag this with `experimental_class` decorator please?

llmfoundry/utils/exceptions.py
joyce-chen-uni: yeah, it was more just in case the error would be raised elsewhere in future without the same info, but removed these now

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,137 @@
...
+        window_size (int): The size of the rolling window used to track recent losses. Default is 100.
+        loss_cap (int): The maximum allowable loss. If the training loss consistently exceeds this value,
+                        it is considered a diverging or unstable run. Default is 10.
+
+    Raises:
+        LossSpikeError: If log_only is False and a loss spike or persistently high loss is detected, this error is
+                        raised to stop the run with an error message.
+    """
+
+    def __init__(
```
dakinggg, line 49: one thing i forgot about, you should consider whether you want this to happen on all ranks, or just rank 0

tests/callbacks/test_kill_loss_spike_callback.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2024 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
+import unittest
+from collections import deque
+from unittest.mock import patch
+
+from llmfoundry.callbacks.kill_loss_spike_callback import KillLossSpike
+
+
+class TestKillLossSpike(unittest.TestCase):
```
dakinggg, line 10: can you also test that setting log_only works properly? i.e. the default settings won't kill a run

joyce-chen-uni: Yep I have manually tested this by submitting runs with `kill_loss_spike: {log_only: true}`. For example: [https://dbc-559ffd80-2bfc.cloud.databricks.com/ml/experiments/2436561715528145/runs/43cc2a5856e74d2985f554be2396ff8d/model-metrics?o=7395834863327820](url)

dakinggg: can we unit test it too though?

joyce-chen-uni: Decomposed the spike-handling and unit tested log_only setting

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,185 @@
...
+from composer.utils import dist
+
+from llmfoundry.utils.exceptions import HighLossError, LossSpikeError
+from llmfoundry.utils.warnings import experimental_class
+
+log = logging.getLogger(__name__)
+
+__all__ = ['KillLossSpike']
+
+MIN_WINDOW_SIZE = 100
```
dakinggg, line 23: suggest prefix these with `_` to indicate private

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,185 @@
...
+                        raised to stop the run with an error message.
+    """
+
+    def __init__(
+        self,
+        log_only: bool = True,
+        patience: int = 4,
+        outlier_multiplier: float = 2,
+    ):
+        self._enabled = True
```
dakinggg, line 60: assuming the `_enabled` flag is still wip, but it should be set to `dist.get_global_rank() == 0` here, and then just used to gate all the calls to only happen if enabled

llmfoundry/callbacks/kill_loss_spike_callback.py
dakinggg, line 23: might as well make all of these configurable in the callback with defaults so that if you want to try different values its a yaml change not a code change

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,185 @@
...
+                window_size=self.window_size,
+            )
+
+    def fit_start(self, state: State, logger: Logger) -> None:
+        # Set the window size to a fraction of the total number of training batches for the run, minimum 100 batches.
+        if state.max_duration is not None:
+            if state.max_duration.unit == TimeUnit.EPOCH and state.dataloader_len is not None:
+                total_training_steps = state.dataloader_len * state.max_duration.value
+            elif state.max_duration.unit == TimeUnit.BATCH or state.max_duration.unit == TimeUnit.TOKEN:
+                total_training_steps = state.max_duration.value
```
dakinggg, line 145: this doesn't seem right for tokens unless im misunderstanding. cause max duration value would not be the same as total training steps

llmfoundry/callbacks/kill_loss_spike_callback.py
joyce-chen-uni: Sure

llmfoundry/callbacks/kill_loss_spike_callback.py
joyce-chen-uni: Yeah, this is messy but that's why I added an additional check in batch_end so if the max_duration / 20 for tokens is not yet 100 batches, set to 100 batches there and continue appending losses

llmfoundry/callbacks/kill_loss_spike_callback.py
joyce-chen-uni: So I tried this initially, but that meant the callback was not turned on at all (no logs + didn't stop when expected). Setting it in fit_start fixed this -- could be because the callback seems to be initialized before distributed training is started?

dakinggg: hmm, that should not be the case.... it works fine for other things that we do it for: https://github.com/mosaicml/composer/blob/2fe9b10a0160466641aaf5890b5e865bc090fc78/composer/loggers/mlflow_logger.py#L151. The callback is created after dist has been initialized (and that function just reads from an env var anyway)

joyce-chen-uni: Tried again and looks good now, must have been some issue with my run previously

llmfoundry/callbacks/kill_loss_spike_callback.py
joyce-chen-uni: Made loss_cap and window_size configurable

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,198 @@
...
+            elif state.max_duration.unit == TimeUnit.TOKEN:
+                # This is an approximation of the total batches from the total tokens, assuming the ratio of tokens:batch is constant.
+                total_training_steps = current_step * (
+                    state.max_duration.value / current_token
+                )
+            self.window_size = max(
+                self.window_size,
+                round(float(total_training_steps * _WINDOW_FRACTION)),
+            )
+        self.loss_window = deque(self.loss_window, maxlen=self.window_size)
```
dakinggg, line 162: rather than this checking of whether window size has been initialized, I believe you can just use the `fit_start` event to set loss window appropriately before the first batch. you can initialized loss window to none in the constructor.

llmfoundry/callbacks/kill_loss_spike_callback.py
```diff
@@ -0,0 +1,198 @@
...
+                    self.loss_window.append(train_loss)
+                    return
+
+            # If user does not provide a loss cap, set loss cap to the maximum loss from the first loss window. Hard cap at loss=10.
+            if not self.loss_cap_set and current_step == self.window_size:
+                self.loss_cap = min(max(self.loss_window), self.loss_cap)
+                self.loss_cap_set = True
+
+            running_loss_avg = float(np.mean(self.loss_window))
+            log.info(f'Running loss average: {running_loss_avg}')
```
dakinggg, line 191: remove this log, as it will happen on every batch

dakinggg (commented):
looks mostly good to me! a couple small comments

llmfoundry/callbacks/kill_loss_spike_callback.py
joyce-chen-uni: I think we need some history of batches in order to set the correct loss window if max_duration is given in tokens? And then it was just a decision to keep all the window operations in the same place (I can break the epoch/batch ones off into fit_start if you think that's better)

dakinggg: oh yeah, you're right. all good

dakinggg (approved):

PR #1476, opened 2024-08-22
Remove extra bos for prompt/response data with llama3.1

dakinggg:
This is a modification to https://github.com/mosaicml/llm-foundry/pull/1003 and the explanation from there still stands. The difference is that llama3.1 does not have an `add_bos_token` attr because it is not a `LlamaTokenizerFast`, but is a `PreTrainedTokenizerFast`.

Existing unit test covers the existing usage, and manual test for llama3.1 is here:
```
In [1]: from llmfoundry.data.finetuning.tasks import _tokenize_with_bos_removal

In [2]: import transformers

In [3]: llama318t = transformers.AutoTokenizer.from_pretrained('meta-llama/Meta-Llama-3.1-8B-Instruct')

In [4]: llama318t(text='hello', text_target='goodbye')
Out[4]: {'input_ids': [128000, 15339], 'attention_mask': [1, 1], 'labels': [128000, 19045, 29474]}

In [5]: _tokenize_with_bos_removal(llama318t, text='hello', text_target='goodbye')
Out[5]: {'input_ids': [128000, 15339], 'attention_mask': [1, 1], 'labels': [19045, 29474]}
```

Before and after loss curve. Similarly to the previous PR, loss is lower after because we're not forcing the model into an unfamiliar use of bos.
Screenshot 2024-08-21 at 6 29 21 PM

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -323,7 +323,8 @@ def _tokenize_with_bos_removal(
     # an add_bos_token attr that we can check explicitly, so instead we rely on checking if both the
     # text and the text_target start with bos_token_id to determine whether to strip bos.
     has_bos_token = hasattr(
-        tokenizer, 'bos_token_id'
+        tokenizer,
+        'bos_token_id',
     ) and tokenizer.bos_token_id is not None
     input_ids_starts_with_bos = False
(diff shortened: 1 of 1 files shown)


snarayan21 (approved):
lgtm! makes sense that this impacts loss, still wild to see though

PR #1483, opened 2024-08-26
Bump to composer 0.24.0 and torch 2.4

snarayan21:
Updates to torch 2.4, as composer 0.24.0 now supports it. We pin to only torch 2.4 so cuda extension libraries are also usable.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -17,11 +17,11 @@ jobs:
     strategy:
       matrix:
         include:
-        - name: "2.3.1_cu121"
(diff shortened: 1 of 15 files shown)


dakinggg:
@snarayan21 I usually do this in multiple steps. (1) pr to make the docker images and move the upper pin, (2) pr to add the tests, (3) pr to move the lower pin

dakinggg:
And please bump to composer 0.24 separately from the torch bump

snarayan21:
closing in favor of other PRs

PR #1488, opened 2024-08-27
Set use_cache back to True for HF checkpointer

eldarkurtic:
Most HF models have `use_cache` set to `True` by default, which is manually changed to `False` in llm-foundry (most likely due to https://github.com/huggingface/transformers/issues/28056). This PR sets `use_cache` back to True before saving the model with the HF checkpointer.

This makes it a bit more convenient to use models trained with llm-foundry, without having to manually edit `config.json` and `generation_config.json` to set `use_cache`.


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -500,7 +500,12 @@ def tensor_hook(
 
         if dist.get_global_rank() == 0:
             log.debug('Saving Hugging Face checkpoint in global rank 0')
-
+            
+            if hasattr(original_model.config, 'use_cache'):
```
dakinggg, line 504: This should not be set on the original model/config because you might continue training after this function is run. I'd be fine setting it for the new model/config that are getting saved out though.

eldarkurtic: great point, thanks!

eldarkurtic:
@dakinggg seems like some tests are failing now, and if I am reading this correctly, it is because MPT models, by default, have `use_cache: False` (contrary to all other models on HF-hub) and before saving to HF checkpoints we are setting `use_cache: True`. 

What is the reason behind this choice for MPT models? I haven't been able to find another model in HF-hub that disables the usage of cache in `config.json`. 

As for the fix for the tests, I assume we can add a patch which checks if the model is MPT, and if yes, don't set `use_cache: True`. What do you think?

dakinggg:
@eldarkurtic Its fine to have it be True since we explicitly set `use_cache=False` for training now

eldarkurtic:
@dakinggg okay, got it. So it is safe to leave these tests red?

dakinggg:
@eldarkurtic well, no. We can't merge with failing tests, but you can update the test so that it passes

PR #1493, opened 2024-08-28
Enable passing in external position ids

gupta-abhay:
For several uses (such as multimodal), position_ids are determined by external functions which account for image + text positions. For such scenarios, we should enable passing in pre-computed `position_ids` to the modeling function.


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1121,6 +1126,7 @@ def forward(
         input_ids: Optional[torch.LongTensor] = None,
         past_key_values: Optional[list[tuple[torch.FloatTensor]]] = None,
         attention_mask: Optional[torch.ByteTensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
```
dakinggg, line 1129: any particular reason you put this in the middle of the args? it'd be better to put it at the end probably so it doesn't break any existing calls to this function

dakinggg (approved):
Can you add a simple test of the functionality (i.e. just forward with position ids and compare against passing it as None and make sure they are the same or something)?

llmfoundry/models/mpt/modeling_mpt.py
gupta-abhay: fixed - and added test. ptal!

dakinggg (approved):

PR #1494, opened 2024-08-28
Align logged attributes for errors and run metadata in kill_loss_spike_callback.py

joyce-chen-uni:
- Log loss spike/high loss metrics instead of message to run metadata (message will be reconstructed in mapi)
- Only log to run metadata if log_only to avoid duplication of message in train_updated event and failed_exception event
- Add loss_window to error attributes


jjanezhang (approved):
lgtm, thank you :)

dakinggg (approved):

PR #1495, opened 2024-08-29
Enable cross attention layers

gupta-abhay:
- enables encoder decoder models if needed
- enables cross attention for flamingo (needed for multimodal)
- enables different kv dim for kv projections (needed for multimodal)

### DIFF
--- a/llmfoundry/models/layers/attention.py
+++ b/llmfoundry/models/layers/attention.py
@@ -464,12 +464,7 @@ class GroupedQueryAttention(nn.Module):
         self.sliding_window_size = sliding_window_size
         self.reuse_kv_layer_idx = reuse_kv_layer_idx
 
-        if kv_dim is not None:
-            self.kv_dim = kv_dim
-            assert fused_qkv is False, 'Cannot use separate kv_dim from d_model for fused_qkv'
-        else:
-            self.kv_dim = self.d_model
-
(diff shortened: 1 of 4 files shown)


tests/models/layers/test_attention.py
```diff
@@ -271,3 +271,70 @@ def test_sliding_window(sliding_window_size: int, attn_impl: str):
...
+    attn_config_fused['fused_qkv'] = True
+
+    attn_config_unfused = generic_attn_kwargs.copy()
+    attn_config_unfused['fused_qkv'] = False
+
+    attn_fused = build_attention_layer(
+        name=attn_name,
+        attn_kwargs=attn_config_fused,
+    )
+    attn_unfused = build_attention_layer(
```
dakinggg, line 321: guessing you need to set the weights of the second layer equal to the weights of the first layer in order to get determinism

llmfoundry/models/layers/attention.py
```diff
@@ -572,7 +572,8 @@ def __init__(
 
     def forward(
         self,
-        x: torch.Tensor,
+        hidden_states: torch.Tensor,
```
dakinggg, line 575: to avoid breaking existing calls to these functions, can you leave the argument name and ordering (i.e. put the new arg at the end)?

dakinggg (approved):

PR #1496, opened 2024-08-29
tokenizer is never built when converting finetuning dataset

eldarkurtic:
```python
    tokenizer = None                                                      <--- because of this line
    tokenizer_kwargs = tokenizer_kwargs
    tokenizer_kwargs.update({'model_max_length': max_seq_len})
    if tokenizer:
        tokenizer = build_tokenizer(tokenizer, tokenizer_kwargs)          <--- tokenizer is never built
```


dakinggg (approved):
ah, good catch!

PR #1497, opened 2024-08-29
Removing error message for reusing kv cache with torch attn

ShashankMosaicML:
KV cache sharing is independent of the attention implementation, hence it should work for torch attention as well. Thus, this PR removes the error check which checks if KV cache sharing is being used with torch attention.


dakinggg (approved):
Add a PR description please

PR #1503, opened 2024-08-31
Update ci-testing to 0.2.2

dakinggg:
Updates ci-testing to 0.2.2, which includes switching from pip to uv. Also adds a numpy<2 pin to the databricks dependency group, as the databricks-sql-connector package requires numpy<2.

### DIFF
--- a/.github/workflows/pr-gpu.yaml
+++ b/.github/workflows/pr-gpu.yaml
@@ -27,10 +27,10 @@ jobs:
           markers: "gpu"
           pip_deps: "[all]"
           pytest_command: "coverage run -m pytest"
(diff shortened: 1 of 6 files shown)


.github/workflows/pr-gpu.yaml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
     - main
     - release/*
-  pull_request_target:
```
dakinggg, line 7: dont merge

mvpatel2000 (approved):

mvpatel2000 (commented):
you can remove passing in container to workflow 
also remove extra checkout for cpu

b-chu (approved):
Looks good!

PR #1510, opened 2024-09-03
Simplify CL API

b-chu:
Changes the API to

- specify the dataset and use the dataloader settings from train_loader
- not need to respecify the first dataset in the callback

Same loss curve as the old callback
image

### DIFF
--- a/llmfoundry/callbacks/curriculum_learning_callback.py
+++ b/llmfoundry/callbacks/curriculum_learning_callback.py
@@ -9,7 +9,8 @@ the future.
 
 import copy
 import logging
-from typing import Any
+import warnings
+from typing import Any, Optional, Union
 
 from composer import DataSpec
 from composer.core import State, Time, TimeUnit, ensure_time
@@ -23,6 +24,7 @@ from llmfoundry.utils.exceptions import (
     BaseContextualError,
     TrainDataLoaderLocation,
 )
+from llmfoundry.utils.warnings import VersionedDeprecationWarning
 
 log = logging.getLogger(__name__)
 
@@ -32,19 +34,21 @@ __all__ = ['CurriculumLearning']
 class CurriculumLearning(CallbackWithConfig):
     """Starts an epoch with a different dataset when resuming from a checkpoint.
 
+    Example duration:
+    <number>tok
     Example schedule:
     [
         {
             'duration': <number>tok,
-            'train_loader': <dataloader parameters>, # matches top level train_loader
+            'dataset': <dataset parameters>,
         },
         {
             'duration': <number>tok,
-            'train_loader': <dataloader parameters>,
+            'dataset': <dataset parameters>,
         },
         {
             'duration': <number>tok,
-            'train_loader': <dataloader parameters>,
+            'dataset': <dataset parameters>,
         ],
     ]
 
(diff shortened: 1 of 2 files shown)


mvpatel2000 (commented):
Deferring to Daniel

Should @codestar12 also review

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -53,48 +55,47 @@ class CurriculumLearning(CallbackWithConfig):
...
         schedule: list[dict[str, Any]],
     ):
         # Ensure all duration units are in epochs or tokens and values are positive
         self._schedule = schedule
         if len(self._schedule) == 0:
             raise ValueError('The schedule must have at least one datamix.')
-        for index, datamix in enumerate(self._schedule):
+        first_datamix = {
+            'duration': duration,
+            'dataset': train_config['train_loader']['dataset'],
```
snarayan21, line 82: question -- did we want to modify how this CL callback is being used or bring back the previous one as well @codestar12 ?

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -211,18 +213,20 @@ def _build_train_loader(
...
             f'Building train loader in CurriculumLearning callback for dataset {self._schedule_index}',
         )
-        assert self.tokenizer is not None
+        assert self._tokenizer is not None
         try:
             return build_dataloader(
                 train_loader_config,
-                self.tokenizer,
-                self.device_train_batch_size,
+                self._tokenizer,
```
snarayan21, line 228: just want to make sure, this should work with finetuning dataloader as well right?

snarayan21 (commented):
mind doing a before/after manual test that shows that the loss curves are the same with this change?

mvpatel2000:
We should ensure @codestar12 reviews

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -53,48 +55,47 @@ class CurriculumLearning(CallbackWithConfig):
...
-            ):
-                raise ValueError((
-                    'The first datamix in the schedule must match the '
-                    'train_loader in the train_config.'
-                ))
-
         self._schedule_index = 0
-        self.device_train_batch_size = train_config['device_train_batch_size']
-        self.tokenizer = None
+        self._train_loader_config: dict[str, Any] = train_config['train_loader']
```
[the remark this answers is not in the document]
b-chu: Yes

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Yes

llmfoundry/callbacks/curriculum_learning_callback.py
snarayan21: what was the decision lol

llmfoundry/callbacks/curriculum_learning_callback.py
```diff
@@ -53,48 +57,59 @@ class CurriculumLearning(CallbackWithConfig):
...
+        duration: Optional[Union[Time, str, int]] = None,
     ):
+        if duration is None:
+            warnings.warn(
+                VersionedDeprecationWarning(
+                    'Specifying the full schedule in the CurriculumLearning ' +
+                    'callback is deprecated. Please specify the duration of ' +
+                    'the first datamix separately and change the schedule ' +
+                    'use datasets instead of dataloaders.',
+                    remove_version='0.13.0'
```
dakinggg, line 85: should be 0.15.0, right? So the deprecation warning is included in two releases (13 and 14)

dakinggg (approved):

llmfoundry/callbacks/curriculum_learning_callback.py
b-chu: Done

PR #1515, opened 2024-09-09
Sort callbacks so that CheckpointSaver goes before HuggingFaceCheckpointer

irenedea:
Sort the callbacks so that CheckpointSaver goes first, then HuggingFaceCheckpointer, and finally all other callbacks. This saves time during final checkpoint saving and uploading.

A rough sketch on the reordering based on Llama 70B profiling:
Before
image

After
image

e2e run in staging: `ift-meta-llama-3-8b-aoprwt-reordered-q01HKT`

### DIFF
--- a/llmfoundry/command_utils/train.py
+++ b/llmfoundry/command_utils/train.py
@@ -548,6 +548,15 @@ def train(cfg: DictConfig) -> Trainer:
         spin_dataloaders=train_cfg.spin_dataloaders,
     )
 
+    from composer.callbacks.checkpoint_saver import CheckpointSaver
+
+    print('before', trainer.state.callbacks)
+
+
+    trainer.state.callbacks = sorted(trainer.state.callbacks, key=lambda c: 0 if isinstance(c, CheckpointSaver) else 1 if isinstance(c, HuggingFaceCheckpointer) else 2)
+
+    print('after', trainer.state.callbacks)
+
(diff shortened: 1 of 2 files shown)


llmfoundry/command_utils/train.py
```diff
@@ -187,6 +188,24 @@ def _initialize_dist_with_barrier(dist_timeout: Union[int, float]):
...
+        trainer (Trainer): Trainer object
+    """
+
+    def _sort_key(c: Callback) -> int:
+        # CheckpointSaver goes first because the blocking time is shortest while upload is async.
+        if isinstance(c, CheckpointSaver):
+            return 0
+        if isinstance(c, HuggingFaceCheckpointer):
+            return 1
+        return 2
```
mvpatel2000, line 204: should other callbacks be before or after ckpting? imo before as they might change model

mvpatel2000 (commented):
test?

llmfoundry/command_utils/train.py
irenedea: Ah good point, i'll switch to before

mvpatel2000:
> @mvpatel2000 What tests would you like to see? I can add a test for the helper function that checks to see that the callbacks are in the expected order. Anything else?

I think that's fine!

mvpatel2000 (approved):

PR #1516, opened 2024-09-10
Add HuggingFaceCheckpointer option for only registering final checkpoint

irenedea:
This would allow us to avoid saving an extra final checkpoint at the end of training.

Example Usage
```
  hf_checkpointer:
    precision: bfloat16
    save_folder: ./local_checkpoints
    save_interval: 1h
    mlflow_logging_config:
      task: llm/v1/chat
      metadata:
        task: llm/v1/chat
        pretrained_model_name: meta-llama/Meta-Llama-3-8B
        databricks_model_family: LlamaForCausalLM (llama-3)
        databricks_model_source: genai-fine-tuning
        databricks_model_size_parameters: 8b
    mlflow_registered_model_name: ift-meta-llama-3-8b-ohuja8
    final_register_only: true
  ```

Branch for testing: `final-register-only-fallback` This adds https://github.com/irenedea/llm-foundry/commit/f63b5e0eff0d0b052e28dcd8711af332da0e024e which raises an error.

fallback: `test-final-register-only-fallback-ATwQG5`
normal: `test-final-register-only-SKerij`

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -165,7 +165,8 @@ class HuggingFaceCheckpointer(Callback):
             be flattened when editing MPT files.
         final_register_only (bool): If true, only register the model in the MLFlow
             registry on the last batch and do not save the HuggingFace checkpoint. If
-            registration fails, then we will fallback to saving the HuggingFace checkpoint.
+            registration fails or mlflow_registered_model_name is not set, then we will
+            fallback to saving the HuggingFace checkpoint.
     """
 
     def __init__(
@@ -470,6 +471,14 @@ class HuggingFaceCheckpointer(Callback):
         upload_to_save_folder: bool,
         register_to_mlflow: bool,
     ):
+        """Save a HuggingFace formatted checkpoint.
+
+        Args:
+            state (State): The training state.
+            logger (Logger): The logger.
+            upload_to_save_folder (bool): Whether to upload the HF checkpoint to the save folder.
+            register_to_mlflow (bool): Whether to register the model to MLFlow
+        """
         del logger  # unused
 
         self.last_checkpoint_batch = state.timestamp.batch
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -161,6 +161,9 @@ class HuggingFaceCheckpointer(Callback):
             keys ``input_example`` and ``signature``.
         flatten_imports (Sequence[str]): A sequence of import prefixes that will
             be flattened when editing MPT files.
+        final_register_only (bool): If true, only register the model in the MLFlow
+            registry on the last batch and do not save the HuggingFace checkpoint. If
+            registration fails, then we will fallback to saving the HuggingFace checkpoint.
     """
 
     def __init__(
@@ -173,6 +176,7 @@ class HuggingFaceCheckpointer(Callback):
         mlflow_registered_model_name: Optional[str] = None,
         mlflow_logging_config: Optional[dict] = None,
         flatten_imports: Sequence[str] = ('llmfoundry',),
+        final_register_only: bool = False,
     ):
         _, _, self.save_dir_format_str = parse_uri(save_folder)
         self.overwrite = overwrite
@@ -184,6 +188,7 @@ class HuggingFaceCheckpointer(Callback):
         }[precision]
         self.flatten_imports = flatten_imports
         self.using_peft = False
+        self.final_register_only = final_register_only
 
         # mlflow config setup
         self.mlflow_registered_model_name = mlflow_registered_model_name
@@ -249,7 +254,7 @@ class HuggingFaceCheckpointer(Callback):
         self.last_checkpoint_batch: Optional[Time] = None
         self.mlflow_loggers = []
 
-        self.child_processes: list[SpawnProcess] = []
+        self.register_processes: list[SpawnProcess] = []
         # Temporary save directory used by child_processes.
         self.temp_save_dir = None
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: llmfoundry/callbacks/hf_checkpointer.py [hunk 1]
```diff
@@ -339,12 +367,23 @@ def _is_last_batch(self, state: State):
...
         return x.item() == 0
 
+    def _any_register_processes_error(self) -> bool:
+        has_errors = any(
+            process.exitcode is not None and process.exitcode != 0
+            for process in self.register_processes
+        )
+        x = torch.tensor(1 if has_errors else 0).to(device='cuda')
+        dist.all_reduce(x, reduce_operation='MAX')
+        return x.item() == 1
```
KuuCi, line 385: A bit unfamiliar with this part so might need a second opinion

KuuCi (approved):
lgtm aside from that one snipbit I don't quite understand.

llmfoundry/callbacks/hf_checkpointer.py
irenedea: cc @dakinggg or @snarayan21 ?

snarayan21: the logic seems fine to me...

snarayan21 (commented):
seems chill to me, would like @dakinggg to take a pass

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -185,8 +189,16 @@ def __init__(
         self.flatten_imports = flatten_imports
         self.using_peft = False
 
+        self.final_register_only = final_register_only
+
         # mlflow config setup
         self.mlflow_registered_model_name = mlflow_registered_model_name
+        if self.final_register_only and self.mlflow_registered_model_name is None:
```
dakinggg, line 196: i think this error is a bit overkill. If I'm not registering anything, its probably okay for the code to accept either value of `final_register_only`. Not super opinionated, so feel free to leave if you disagree.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -339,12 +374,23 @@ def _is_last_batch(self, state: State):
...
         x = torch.tensor(1 if not_done else 0).to(device='cuda')
         dist.all_reduce(x, reduce_operation='MAX')
         return x.item() == 0
 
+    def _any_register_processes_error(self) -> bool:
+        has_errors = any(
+            process.exitcode is not None and process.exitcode != 0
+            for process in self.register_processes
+        )
+        x = torch.tensor(1 if has_errors else 0).to(device='cuda')
```
dakinggg, line 390: mildly unfortunate this only works with gpu. would be better to get the device from state. up to you if you want to do or not since it was like this before your pr.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -412,7 +458,13 @@ def transform_model_pre_registration(
         """
         return model
 
-    def _save_checkpoint(self, state: State, logger: Logger):
+    def _save_checkpoint(
```
dakinggg, line 461: docstring please since there are now extra args

llmfoundry/command_utils/train.py
```diff
@@ -584,7 +584,12 @@ def train(cfg: DictConfig) -> Trainer:
             )
 
         hf_checkpointer_callback = hf_checkpointer_callbacks[0]
-        hf_checkpointer_callback._save_checkpoint(trainer.state, trainer.logger)
+        hf_checkpointer_callback._save_checkpoint(
+            trainer.state,
+            trainer.logger,
+            upload_to_save_folder=True,
+            register_to_mlflow=False,
```
dakinggg, line 591: i think it would be better for register to be true for this call

llmfoundry/callbacks/hf_checkpointer.py
irenedea: To confirm, you're suggesting that if final_register_only is True and mlflow_registered_model_name is None, we save the final model as normal?

I think that's fair since we fallback to normal save if register fails anyway. I'll raise a warning here instead of a error.

llmfoundry/command_utils/train.py
irenedea: yeah that's reasonable, can do that. fyi setting register=True would be a change in behavior from what was happening originally.

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: yeah that sounds good to me

llmfoundry/command_utils/train.py
dakinggg: yeah, when I added this I meant to check what would happen, and I intended for it to work with register in my head. and its not a breaking change, just does a bit more, so i think its ok.

llmfoundry/callbacks/hf_checkpointer.py
irenedea: done

llmfoundry/callbacks/hf_checkpointer.py
irenedea: done

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -300,14 +327,28 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
             # Wait for all child processes spawned by the callback to finish.
             timeout = 3600
             wait_start = time.time()
-            while not self._all_child_processes_done():
+            state.device
```
dakinggg, line 330: typo?

dakinggg (approved):
one typo i think, otherwise lgtm

llmfoundry/callbacks/hf_checkpointer.py
irenedea: oops yes

irenedea: i did that to check something in the IDE
```suggestion
```

PR #1517, opened 2024-09-10
Replace FSDP args

KuuCi:
To resolve this Jira:
https://databricks.atlassian.net/browse/GRT-2978

### DIFF
--- a/llmfoundry/command_utils/eval.py
+++ b/llmfoundry/command_utils/eval.py
@@ -100,7 +100,8 @@ def evaluate_model(
             mosaicml_logger._flush_metadata(force_flush=True)
 
     fsdp_config = parallelism_config.get(
-        'fsdp_config', None
+        'fsdp_config',
+        None,
     ) if parallelism_config else None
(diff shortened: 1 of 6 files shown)


snarayan21 (approved):
lgtm; thanks!!

llmfoundry/command_utils/eval.py
```diff
@@ -52,7 +52,7 @@ def evaluate_model(
     device_eval_batch_size: Union[int, float],
     eval_gauntlet_config: Optional[Union[str, dict[str, Any]]],
     eval_loader_config: Optional[Union[dict[str, Any], list[dict[str, Any]]]],
-    fsdp_config: Optional[dict[str, Any]],
+    parallelism_config: Optional[dict[str, Any]],
```
mvpatel2000, line 55: We should be deprecating and not hard deleting...

mvpatel2000: CC: @dakinggg  as well for input

PR #1520, opened 2024-09-12
Peft fsdp

dakinggg:
In the refactor in #1467, we accidentally moved the prepare for fsdp call before the super init. This results in peft models not being wrapped correctly, because the peft model change happens in the super init. This PR moves the prepare for fsdp call after the super init (same behavior as before #1467), and adds a unit test for the correct behavior.

Manual test (this oomd before this PR due to incorrect wrapping): `updated-image-6-9W2TKE`

### DIFF
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -121,7 +121,7 @@ class BaseHuggingFaceModel(HuggingFaceModel):
 
         # Prepare for FSDP needs to happen after the super init, so that any model
         # architecture changes are completed
-        self.prepare_inner_model(model, init_device)
+        self.prepare_inner_model(self.model, init_device)
 
     def loss(self, outputs: ModelOutput, batch: Mapping):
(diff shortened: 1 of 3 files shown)


llmfoundry/models/hf/hf_base.py
```diff
@@ -69,7 +69,7 @@ def __init__(
         config_overrides: Optional[dict[str, Any]] = None,
         use_logits: bool = True,
         shift_labels: bool = False,
-        peft_config: Optional['PeftConfig'] = None,
```
mvpatel2000, line 72: why no typing?

mvpatel2000 (approved):
Did you confirm test fails without change?

llmfoundry/models/hf/hf_base.py
dakinggg: this typing was just wrong. this function accepts a dict and then converts it to a peft config

dakinggg:
@mvpatel2000 yup, test fails before this PR.

PR #1523, opened 2024-09-13
Add defaults to HF checkpointer _save_checkpoint

dakinggg:
Just adds defaults so the new args don't break any usage of this function outside of foundry


irenedea (approved):
LGTM, generally prefer to not add defaults if we can fix the usage elsewhere especially for a private function, but not a strong preference.

PR #1525, opened 2024-09-16
Script to automatically split off eval set

mattyding:
## What changes are proposed in this pull request?
See go/sweeps-eval for more context. Corresponding MAPI changes are contained in https://github.com/databricks-mosaic/mcloud/pull/4562.

## How is this tested?
Added unit tests. 
Screenshot 2024-09-16 at 1 05 50 AM

The predecessor to the unit tests was this bash script I used to manually inspect the file outputs. Tests are the same as unit tests, but it invokes the `scripts/...` file instead of the `command_utils/...` one. Confirmed file outputs looked OK.
[_test.txt](https://github.com/user-attachments/files/17010423/_test.txt)

Everything passes:
Screenshot 2024-09-16 at 12 56 37 AM


llmfoundry/data/finetuning/tasks.py
```diff
@@ -702,7 +702,72 @@ def state_dict(self, num_samples: int,
...
+                    '*' + ext for ext in SUPPORTED_EXTENSIONS
+                ],
+                token=hf_kwargs.get('token', None),
+                revision=hf_kwargs.get('revision', None),
+                local_dir_use_symlinks=False,
+                local_dir=local_dataset_dir,
+            )
+            if _is_empty_or_nonexistent(dirpath=dataset_name):
+                log.error("Failed to safely load the dataset from HF Hub.")
+                raise InvalidFileExtensionError(
```
mattyding, line 745: Moved HF safe download logic into a separate function so I could reuse it. Code is basically unchanged.

Unrelated to meat of the PR, but per Daniel request, tried refactoring further to get rid of try-except block. I don't think it can be done without significant added complexity. You want to barrier/sync before throwing this InvalidFileExtensionError; however, this must be nested within this download logic so that only rank0 encounters it. So you can't both separate out this logic and have graceful exit without try-except. :'(

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,167 @@
...
+import logging
+import os
+import re
+import json
+import contextlib
+import datasets as hf_datasets
+import numpy as np
+from typing import Optional
+
+import composer.utils as utils
```
mattyding, line 13: import is formatted this way due to mocking difficulties - https://bhfsteve.blogspot.com/2012/06/patching-tip-using-mocks-in-python-unit.html

PR #1527, opened 2024-09-16
enable correct padding_idx for embedding layers

gupta-abhay:
- This functionality is primarily needed when you call the `resize_position_embeddings` call for a HF model (ie, adding new tokens to vocab). If the pad_token already exists for the config, you need to ensure that it gets set correctly post `_init_weights` call.
- By default, the behavior remains unchanged from today, since the `pad_token_id` - if not specified by the config, defaults to `None` from here: https://github.com/huggingface/transformers/blob/174890280b340b89c5bfa092f6b4fb0e2dc2d7fc/src/transformers/configuration_utils.py#L326 and the padding_idx is equivalently set to `None`


dakinggg (commented):
can you add a simple test for the new init function?

dakinggg (approved):

PR #1528, opened 2024-09-16
Replace in case of unicode errors

mattyding:
To improve user workflow, automatically replace any unicode errors in their code with a replacement character. With this change, we shouldn't encounter any `UnicodeDecodeErrors` at all, so I removed the `CannotUnicodeDecodeFile` that was added previously in https://github.com/mosaicml/llm-foundry/pull/1457

## How is this tested
Modified unit test

Screenshot 2024-09-16 at 1 20 44 PM


mattyding:
won't do

PR #1529, opened 2024-09-16
Throw error when tokenization fails due to mistyped dataset entries.

milocress:
Catches tokenization failures on custom HF datasets with missing/extra/mistyped columns.


irenedea (dismissed):
LGTM

irenedea:
unit tests?

PR #1530, opened 2024-09-16
Add deprecation warning to fsdp_config

KuuCi:
Deprecates fsdp_config and replaces it with parallelism_config. Replaces all instances of the test, as well as new unit tests to make sure that the proper warnings are thrown when users attempt to use fsdp_config


llmfoundry/command_utils/eval.py
```diff
@@ -65,6 +68,17 @@ def evaluate_model(
     should_log_config: bool = True,
     load_path: Optional[str] = None,
 ):
+    if fsdp_config:
+        warnings.warn(
+            VersionedDeprecationWarning(
+                'The argument fsdp_config is deprecated. Please use parallelism_config instead.',
+                remove_version='0.12.0',
```
snarayan21, line 75: what's our policy on this? should this be a later version?

dakinggg: yes, +2 versions

llmfoundry/command_utils/eval.py
```diff
@@ -62,9 +64,21 @@ def evaluate_model(
...
     load_path: Optional[str] = None,
 ):
+    if fsdp_config:
+        warnings.warn(
+            VersionedDeprecationWarning(
+                'The argument fsdp_config is deprecated. Please use parallelism_config instead.',
+                remove_version='0.13.0',
+            ),
+        )
+    if fsdp_config and parallelism_config:
```
dakinggg, line 78: for anyone using positional arguments, would be good to check parallelism_config and raise an error if it looks like fsdp config

KuuCi: I think I did it right.....

>>> from composer.utils import dist, get_device, reproducibility, parallelism
>>> print(list(parallelism.FSDPConfig.__annotations__.keys()))
['activation_checkpointing', 'activation_checkpointing_reentrant', 'activation_cpu_offload', 'auto_wrap', 'te_checkpoint_wrapper', 'te_shard_fp8_weight', 'backward_prefetch', 'backward_prefetch_limit', 'cpu_offload', 'data_parallel_shard_degree', 'data_parallel_replicate_degree', 'device_mesh', 'forward_prefetch', 'forward_prefetch_limit', 'ignored_modules', 'keep_low_precision_grads', 'limit_all_gathers', 'load_monolith_rank0_only', 'load_planner', 'mixed_precision', 'process_group', 'save_planner', 'sharded_ckpt_prefix_dir', 'sharding_strategy', 'state_dict_type', 'sync_module_states', 'use_orig_params', 'verbose']

dakinggg (commented):
can you add unit tests for calling this function with the different args and make sure the expected outcomes happen?

mvpatel2000 (commented):
PR descriptions are your friend

mvpatel2000 (approved):

PR #1532, opened 2024-09-17
Delete unneeded inner base model in PEFT HF Checkpointer

snarayan21:
We don't need the inner base model when HF checkpointing PEFT models. This can cause unnecessary memory usage later on.


dakinggg (commented):
@snarayan21 does the peft model store a copy or something...?

dakinggg (approved):
Don't totally understand where the extra memory comes into play, but no harm in deleting a reference we don't need anymore.

PR #1536, opened 2024-09-19
Add dataset swap callback

b-chu:
Adds a callback that swaps the dataset when switching indices


snarayan21 (commented):
seems chill to me, would like @dakinggg to give final approval

b-chu:
Need @codestar12 review

PR #1537, opened 2024-09-21
Add allowed keys to chat hf dataset format 

j316chuck:
# Description
Add `dataset` and `id` to allowed keys for chat formatted hf datasets. Enables us to finetune on datasets like [tulu-v2](https://huggingface.co/datasets/allenai/tulu-v2-sft-mixture/tree/main)

# Tests
Before - `tulu-ft-llama3-8b-instruct-trainer-2CHlea` 🔴 
```
[rank6]: UnknownExampleTypeError: Found keys {'dataset', 'id', 'messages'} in dataset.
[rank6]: Unknown example type. For prompt and response finetuning, the valid prompt keys
[rank6]: are {'prompt'} and the valid response keys are {'response', 'completion'}. For
[rank6]: chat finetuning, the allowed keys are {'messages'}
```
After - `tulu-ft-llama3-8b-instruct-trainer-V5i6u6` ✅

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -141,18 +141,13 @@ def _get_example_type(example: Example) -> ExampleType:
     if not isinstance(example, Mapping):
         raise InvalidExampleTypeError(str(type(example)))
     if (
-        len(example.keys()) == 1 and any(
-            allowed_message_key in example
-            for allowed_message_key in ALLOWED_MESSAGES_KEYS
-        )
-    ):
-        return 'chat'
-    elif (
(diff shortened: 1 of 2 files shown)


b-chu:
The dataset just looks like it's formatted incorrectly. Why not just drop the two additional columns from the dataset?

XiaohanZhangCMU:
@j316chuck sounds like the change is only meant for one particular dataset? I would suggest keep the code unchanged, but download the dataset then load it back with just the "messages" column.

j316chuck:
sure makes sense, reuploaded the dataset and got it by changing the dataset: `tulu-ft-llama3-8b-trainer-qeJHNN

PR #1540, opened 2024-09-22
Make ICL tasks not required for eval

snarayan21:
Currently icl tasks is required as a key to run eval. This doesn't make sense if all I'm trying to do is run an eval dataset via eval_loader(s). So we should set to false instead.


KuuCi (approved):
This makes sense to me, lgtm

snarayan21:
just want to make sure @dakinggg has a chance to review before I merge, in case there was some reason we chose to make icl tasks required...

dakinggg (approved):

PR #1543, opened 2024-09-23
Hfcheckpointer optional generation config

KuuCi:
Current hf_checkpointer assumes that generation config exists, which isn't necessarily the case for certain models. DM for testing!

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -588,7 +588,7 @@ class HuggingFaceCheckpointer(Callback):
                     del new_base_model_instance
                 else:
                     new_model_instance = type(original_model)(new_config)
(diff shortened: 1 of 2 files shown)


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -588,9 +588,10 @@ def tensor_hook(
                     del new_base_model_instance
                 else:
                     new_model_instance = type(original_model)(new_config)
-                    new_model_instance.generation_config.update(
-                        **original_model.generation_config.to_dict(),
-                    )
+                    if new_model_instance.generation_config:
```
dakinggg, line 591: this probably isn't the right check. I think if it is empty dict you still want to update it. also please add a test for this.

dakinggg (approved):

PR #1544, opened 2024-09-24
Add MLflow `log_model` option

nancyhung:
## Context
In order to support customers with sensitive storage network configurations, we have to use the `log_model` API. This will cause duplicate artifact uploads, which is not efficient, so we will only reserve rolling out to customers who require this.

This PR contains the first of 2 changes:
1. When saving the final HF checkpoint, use `log_model` instead of uploading to MLflow artifacts. 
- Functionally, a user can still find their HF checkpoint files in UC if they really wish to download the model weights and serve somewhere else.
- Instead of calling `save_model`, `register_model`, and uploading to UC directly via the remote uploader downloader object, this change simplifies the control logic with the `mlflow.log_model` function. This function is also critical to support secure training requirements, such as customer firewalls or private endpoints. Logging a model to MLflow will call the necessary steps to save and register a model for deployment.
- In this change, we only affect the logic while saving the final HF checkpoint. All other logic remains the same.

2. Next in a follow-up PR, we'll modify the intermediate checkpointing logic to also use `log_model` but not register the model. That way, a user can still manually register their intermediate checkpoints for evaluation.

## Testing
When incorporating this in MAPI, we should enable `final_register_only` to only upload using the `log_model` logic instead of uploading a duplicate copy to MLflow artifacts. All tests were done in AWS staging.

**Works for older models**
[Databricks staging] Llama3 8b
Run: `llama3-log-model-xusOti`
Llama3 8b was able to be successfully deployed here: https://e2-dogfood.staging.cloud.databricks.com/ml/endpoints/test-log-model?o=6051921418418893.

**Works for newest models with extra security**
[MCT] Llama3.2 1b
Run: `llama3-log-model-4eJUKo`
Experiment: https://dbc-559ffd80-2bfc.cloud.databricks.com/ml/experiments/2854093459220376?viewStateShareKey=55a332dc80d7200b6a6301d8f0163155ce9aac54d21436c9d292f0745e0bff05
Endpoint: https://dbc-559ffd80-2bfc.cloud.databricks.com/ml/endpoints/testfinetuning?o=7395834863327820
image

[MCT] Llama3.1 405b
Run: `405b-register-1-xB3dOx`

Tested that mlflow.log_model registers model in private link workspace
image
Registered model: https://adb-1622130341351604.4.azuredatabricks.net/explore/data/models/rkg-ft/default/llamatest?o=1622130341351604
image
Model stuck in pending example: 
image


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -135,6 +135,43 @@ def _register_model_with_run_id_multiprocess(
...
+        )
+        logging.getLogger('composer').setLevel(composer_logging_level)
+    mlflow_logger.log_model(
+        flavor='transformers',
+        artifact_path="model",
+        input_example=input_example,
+        task=task,
+        metadata={
+            "task": task,
+            "databricks_model_source": "genai-fine-tuning",
```
dakinggg, line 167: the metadata should not be hardcoded, it should be passed in like with the existing register call

dakinggg (commented):
What testing have you done? We need to make sure everything e2e shows up properly

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -135,6 +135,43 @@ def _register_model_with_run_id_multiprocess(
...
+    """
+    # Setup logging for child process. This ensures that any logs from composer are surfaced.
+    if composer_logging_level > 0:
+        # If logging_level is 0, then the composer logger was unset.
+        logging.basicConfig(
+            format=
+            f'%(asctime)s: rank{dist.get_global_rank()}[%(process)d][%(threadName)s]: %(levelname)s: %(name)s: %(message)s',
+        )
+        logging.getLogger('composer').setLevel(composer_logging_level)
+    mlflow_logger.log_model(
```
dakinggg, line 160: i don't think you're passing in the model? Note: because we edit the files before registering (https://github.com/mosaicml/llm-foundry/blob/d85c83b15d5b07a1b8cd00eaa7e400aaf7b22ea7/llmfoundry/callbacks/hf_checkpointer.py#L438-L447), this needs to be incorporated with the the local files somehow.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -729,6 +769,29 @@ def tensor_hook(
                         monitor_process = None
 
                     # Spawn a new process to register the model.
+                    # Slower method to register the model via log_model.
+                    if self.use_mlflow_log_model:
```
irenedea, line 773: If `self.use_mlflow_log_model` is true, do we need to both log and register the model? Can we only log to avoid duplicate artifact uploads?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -135,6 +135,43 @@ def _register_model_with_run_id_multiprocess(
...
+    if composer_logging_level > 0:
+        # If logging_level is 0, then the composer logger was unset.
+        logging.basicConfig(
+            format=
+            f'%(asctime)s: rank{dist.get_global_rank()}[%(process)d][%(threadName)s]: %(levelname)s: %(name)s: %(message)s',
+        )
+        logging.getLogger('composer').setLevel(composer_logging_level)
+    mlflow_logger.log_model(
+        flavor='transformers',
+        artifact_path="model",
```
irenedea, line 162: you'll need to use single quotes for strings for code quality ci

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: We do need to log and register because that's the only way to resolve the infinitely pending model registration issues we've been debugging for Azure PL support. [Debugging context is here](https://docs.google.com/document/d/1fh38Zw_-0qrVGuwUis0RgjJ8acli2_Lk2Q6bH0PvIDw/edit#heading=h.mpoax8wdsew0).

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -639,23 +724,29 @@ def tensor_hook(
                         self.flatten_imports,
                     )
 
-                if self.remote_ud is not None:
```
dakinggg, line 642: can't simply delete the remote uploader code. other things rely on this.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -76,6 +76,11 @@ def _maybe_get_license_filename(
 
     If the license file does not exist, returns None.
     """
+    # Early return if no local directory exists
```
dakinggg, line 79: this should never happen right?

dakinggg: assuming that is correct, please remove

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -107,32 +112,82 @@ def _maybe_get_license_filename(
...
 ):
-    """Call MLFlowLogger.register_model_with_run_id.
-
-    Used mainly to register from a child process.
+    """
+    Call MLFlowLogger.log_model.
+    
+    Used mainly to log from a child process.
+
+    Inputs:
```
dakinggg, line 131: Follow docstring format from other places (e.g. `Args:` and the indentation and such

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -107,32 +112,82 @@ def _maybe_get_license_filename(
...
             f'%(asctime)s: rank{dist.get_global_rank()}[%(process)d][%(threadName)s]: %(levelname)s: %(name)s: %(message)s',
         )
-        logging.getLogger('composer').setLevel(composer_logging_level)
-
-    # Register model.
-    mlflow_logger.register_model_with_run_id(
-        model_uri=model_uri,
-        name=name,
-        await_creation_for=await_creation_for,
+        logging.getLogger('llmfoundry').setLevel(python_logging_level)
```
dakinggg, line 148: it should still set composer log level too

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -107,32 +112,82 @@ def _maybe_get_license_filename(
...
+                    os.remove(os.path.join(kwargs['path'], 'model', tokenizer_file_name))
+        except Exception as e:
+            log.error(f"Exception when removing duplicate tokenizer files in the model directory", e)
+    mlflow.transformers.save_model = save_model_patch
+
+    if registered_model_name is not None:
+        mlflow_logger.log_model(
+            transformers_model=transformers_model_path,
+            flavor=flavor,
+            artifact_path='model', # TODO: where should we define this parent dir name?
```
dakinggg, line 174: fix?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -107,32 +112,82 @@ def _maybe_get_license_filename(
...
+            metadata=log_model_metadata,
+            task=task,
+            registered_model_name=registered_model_name, # not the full path? mlflow_logger.model_registry_prefix
+            await_creation_for=await_creation_for
+        )
+    else:
+         mlflow_logger.log_model(
+            transformers_model=transformers_model_path,
+            flavor=flavor,
+            artifact_path='model', # TODO: where should we define this parent dir name?
```
dakinggg, line 185: fix?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -107,32 +112,82 @@ def _maybe_get_license_filename(
...
+            for tokenizer_file_name in tokenizer_files:
+                dupe_file = os.path.isfile(os.path.join(kwargs['path'], 'model', tokenizer_file_name))
+                if dupe_file:
+                    log.debug(f"Removing duplicate tokenizer file: {tokenizer_file_name}")
+                    os.remove(os.path.join(kwargs['path'], 'model', tokenizer_file_name))
+        except Exception as e:
+            log.error(f"Exception when removing duplicate tokenizer files in the model directory", e)
+    mlflow.transformers.save_model = save_model_patch
+
+    if registered_model_name is not None:
```
dakinggg, line 170: you don't need a big if/else here. presumably log_model is fine with you passing `registered_model_name=None`

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -107,32 +112,82 @@ def _maybe_get_license_filename(
...
+        mlflow_logger.log_model(
+            transformers_model=transformers_model_path,
+            flavor=flavor,
+            artifact_path='model', # TODO: where should we define this parent dir name?
+            input_example=input_example,
+            metadata=log_model_metadata,
+            task=task,
+            registered_model_name=registered_model_name, # not the full path? mlflow_logger.model_registry_prefix
```
dakinggg, line 178: full path or no?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -171,7 +226,7 @@ class HuggingFaceCheckpointer(Callback):
 
     def __init__(
         self,
-        save_folder: str,
+        save_folder: Optional[str],
```
dakinggg, line 229: this is probably a change for part 2 not for this pr?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -618,22 +673,16 @@ def tensor_hook(
...
 
+            # This context manager casts the TE extra state in io.BytesIO format to tensor format
+            # Needed for proper hf ckpt saving.
+            context_manager = te.onnx_export(
+                True,
+            ) if is_te_imported and state.precision == Precision.AMP_FP8 else contextlib.nullcontext(
+            )
+            with context_manager:
+                new_model_instance.save_pretrained(temp_save_dir)
+            original_tokenizer.save_pretrained(temp_save_dir)
```
dakinggg, line 684: should move the next if statement out too (`if new_model_instance....`)

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -702,14 +751,6 @@ def tensor_hook(
                         True,
                     ) if is_te_imported and state.precision == Precision.AMP_FP8 else contextlib.nullcontext(
                     )
-                    with context_manager:
```
dakinggg, line 705: not sure if an equivalent to this is necessary or not, can you determine that?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -732,17 +773,25 @@ def tensor_hook(
...
                                 mlflow_logger,
-                            'composer_logging_level':
-                                logging.getLogger('composer').level,
-                            'model_uri':
-                                local_save_path,
-                            'name':
+                            'transformers_model_path':
+                                temp_save_dir,
+                            'flavor':
+                                'peft' if self.using_peft else 'transformers',
```
dakinggg, line 785: we haven't added peft support for `log_model` in Composer yet. Will need to think about how to handle this.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -732,17 +773,25 @@ def tensor_hook(
...
+                            'python_logging_level':
+                                logging.getLogger('llmfoundry').level,
+                            'task':
```
dakinggg, line 788: we should just be ** the `self.mlflow_logging_config`, rather than picking out pieces of it.

dakinggg (commented):
In the linked run, I don't see a registered model connected to your run. It should show up under registered models.
Screenshot 2024-10-25 at 7 46 28 PM

nancyhung:
@dakinggg That's weird I do see the registered model in MLflow for this run: 
image

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: Other functions were flattening the fields, so I thought that was best practice. I can pass in the full object, which I agree is cleaner.

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: I hit this error before when the `local_pretrained_dir` was empty, so I added this early return. This seems like syntactically expected behavior though -- why not keep this check?

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: no

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: deleted this part

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: I included this outside the if statement. Presumably if this was necessary, I don't want to modify the existing logic

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: What do we do for peft models then? Do I need to fork the old model registration logic for PEFT models until we find a solution?

dakinggg: I think you can just leave this flavor as transformers only...the peft logging API is marked as experimental in Composer.

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: I think my point is just that this function doesn't expect to be called if the dir is empty, so I'd be worried about it hiding an error upstream by returning silently.

dakinggg: medium conviction opinion, would remove the check (or error if the dir is empty)

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: Saw this in the logs 
```
2024/10/26 01:21:26 INFO mlflow.transformers: A local checkpoint path or PEFT model is given as the `transformers_model`. To avoid loading the full model into memory, we don't infer the pip requirement for the model. Instead, we will use the default requirements, but it may not capture all required pip libraries for the model. Consider providing the pip requirements explicitly.
```
so we don't need this code

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: Wouldn't I need the original register_model call for PEFT models though?

dakinggg: to preserve the existing behavior, yes. But it was marked as experimental and I'm pretty sure no one used it, and we will redo it now that mlflow supports peft model logging directly.

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: cool i'll error instead

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -732,21 +800,44 @@ def tensor_hook(
...
-                            'model_uri':
-                                local_save_path,
-                            'name':
-                                self.mlflow_registered_model_name,
-                            'await_creation_for':
-                                3600,
-                        },
-                    )
+                    # Slower method to register the model via log_model.
+                    if self.using_peft:
```
nancyhung, line 804: i added back original logic to respect peft ptal @dakinggg

dakinggg: seems right, will review thoroughly once CI passes

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -733,21 +803,44 @@ def tensor_hook(
...
-                                logging.getLogger('composer').level,
-                            'model_uri':
-                                local_save_path,
-                            'name':
-                                self.mlflow_registered_model_name,
-                            'await_creation_for':
-                                3600,
-                        },
-                    )
+                    # Slower method to register the model via log_model.
```
dakinggg, line 806: As discussed offline, the call to `self.pre_register_edit` a few lines up assumes that there is an mlflow model at `local_save_path`. This is no longer true for non peft. So `self.pre_register_edit` needs to be called on `temp_save_dir` (which will contain the hf checkpoint, that log_model will copy over and register)

nancyhung:
Closing this in favor of @irenedea 's PR https://github.com/mosaicml/llm-foundry/pull/1627

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -108,6 +127,93 @@ def _maybe_get_license_filename(
...
+                f'Exception when removing duplicate tokenizer files in the model directory',
+                e,
+            )
+
+    mlflow.transformers.save_model = save_model_patch
+
+    mlflow.set_tracking_uri(mlflow_logger.tracking_uri)
+    if mlflow_logger.model_registry_uri is not None:
+        mlflow.set_registry_uri(mlflow_logger.model_registry_uri)
+    mlflow.start_run(run_id=mlflow_logger._run_id,)
```
dakinggg, line 192: we shouldn't need start run, because we pass run id directly to log_model. did you test and it was still necessary?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -108,6 +127,93 @@ def _maybe_get_license_filename(
...
+    # Setup logging for child process. This ensures that any logs from composer are surfaced.
+    if python_logging_level > 0:
```
dakinggg, line 194: should set up logging at the start of the function

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -298,16 +404,7 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
...
-                    self.remote_ud.init(state, logger)
-                except PermissionError as e:
-                    if 'Client Error' in str(
-                        e,
-                    ):  # thrown from composer.utils._wrap_mlflow_exceptions
-                        raise StoragePermissionError(
-                            'Error when write to save_folder.',
-                        ) from e
-                    raise e
+                self.remote_ud.init(state, logger)
```
dakinggg, line 407: did you intend to delete this try/except? i assume not

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -675,101 +772,149 @@ def tensor_hook(
...
-                    # Restore the monitor process.
-                    if monitor_process is not None:
-                        mlflow_logger.monitor_process = monitor_process  # type: ignore
-                    self.register_processes.append(process)
 
-                    # Save the temporary directory to be cleaned up later.
-                    if use_temp_dir:
-                        self.temp_save_dir = temp_save_dir
+                    new_model_instance.save_pretrained(register_save_dir)
+                    original_tokenizer.save_pretrained(register_save_dir)
```
dakinggg, line 794: should do a none check here on tokenizer before calling save_pretrained on it

llmfoundry/callbacks/hf_checkpointer.py
nancyhung: i need to remove this

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -373,6 +376,14 @@ def _create_optimizer(original_model: torch.nn.Module) -> torch.optim.Optimizer:
...
     new=MockSpawnProcess,
 )
+@patch(
+    'llmfoundry.callbacks.hf_checkpointer._maybe_get_license_filename',
+    new=MagicMock(),
+)
+@patch(
+    'mlflow.start_run',
+    new=MagicMock(),
+)
```
nancyhung, line 386: if start run is unnecessary, then delete this

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -477,6 +491,26 @@ def test_final_register_only(
     'llmfoundry.callbacks.hf_checkpointer.SpawnProcess',
     new=MockSpawnProcess,
 )
+@patch(
+    'mlflow.start_run',
+    new=MagicMock(),
+)
```
nancyhung, line 497: and delete this if unnecessary

dakinggg (approved):

PR #1545, opened 2024-09-24
Refactor MLflow `register_model`

nancyhung:
Previously, we were using some hacky logic in Composer to add retry logic to MLflow model registration. Now that MLflow has baked this into their `register_model` API, we can call that instead.


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -128,7 +128,7 @@ def _register_model_with_run_id_multiprocess(
         logging.getLogger('composer').setLevel(composer_logging_level)
 
     # Register model.
-    mlflow_logger.register_model_with_run_id(
+    mlflow_logger.register_model(
```
dakinggg, line 131: we call this API because otherwise the run id doesn't get properly attached to the registered model, not because of retries.

PR #1547, opened 2024-09-24
Release docker img

KuuCi:
Updates the release process to also create a docker image upon hitting a tag with v*. Some parts are commented out for testing and will be reintroduced upon approval.

This was triggered off of this branch:
https://github.com/mosaicml/llm-foundry/actions/runs/11022339619/job/30611648283
image
As can be seen in the image building process, it properly clones the correct ref + repo along with all of the other installs.
image

image


.github/workflows/release.yaml
```diff
@@ -10,43 +10,111 @@ jobs:
...
-  pypi-packaging:
-    name: Build and Publish llm-foundry PyPI Package
+  # rm for testing
+  # pypi-packaging:
+  #   name: Build and Publish llm-foundry PyPI Package
+  #   needs:
+  #   - code-quality
+  #   runs-on: linux-ubuntu-latest
+  #   steps:
+  #   - name: Checkout source
```
mvpatel2000, line 20: Is this ready for review?

KuuCi: was going to uncomment it back in after review, but can uncomment it before hand

release.Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2022 MosaicML LLM Foundry authors
+# SPDX-License-Identifier: Apache-2.0
```
mvpatel2000, line 2: im a bit confused, how is this different?

KuuCi: Doesn't uninstall and rm foundry (mainly last two lines need to be removed from original docker file)

mvpatel2000: can we just gate that by an arg instead of copying? it's kind of painful to maintain two copies -- i imagine someone will forget to bump both

KuuCi:
image

mvpatel2000 (commented):
LGTM, @dakinggg  plz also review

.github/workflows/release.yaml
```diff
@@ -50,3 +50,67 @@ jobs:
...
+        password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+
+    - name: Define Docker tags
+      id: define-tags
+      run: |
+        BRANCH_NAME="${{ github.ref_name }}"
+        TAG_NAME=$(echo "${BRANCH_NAME}" | sed 's/\//_/g')
+        echo "BRANCH_NAME=${BRANCH_NAME}" >> $GITHUB_ENV
+
+        echo "DOCKER_TAG=mosaicml/llm-foundry:test_release_${TAG_NAME}" >> $GITHUB_ENV
```
dakinggg, line 80: is this supposed to have a `test` prefix?

KuuCi: No, shouldn't have test prefix

.github/workflows/release.yaml
```diff
@@ -50,3 +50,67 @@ jobs:
...
+        file: Dockerfile
+        push: true
+        tags: |
+          ${{ env.AWS_DOCKER_TAG }}
+          ${{ env.AWS_LATEST_TAG }}
+        build-args: |
+          BASE_IMAGE=mosaicml/pytorch:2.4.0_cu124-python3.11-ubuntu20.04-aws
+          BRANCH_NAME=${{ env.BRANCH_NAME }}
+          TE_COMMIT=901e5d2
+          DEPS_GROUPS="[all]"
```
dakinggg, line 99: this is `DEP_GROUPS` not `DEPS_GROUPS`. Could you update the manual testing to show this works please?

.github/workflows/release.yaml
dakinggg, line 80: change back?

.github/workflows/release.yaml
```diff
@@ -50,3 +50,67 @@ jobs:
...
+      id: define-tags
+      run: |
+        BRANCH_NAME="${{ github.ref_name }}"
+        TAG_NAME=$(echo "${BRANCH_NAME}" | sed 's/\//_/g')
+        echo "BRANCH_NAME=${BRANCH_NAME}" >> $GITHUB_ENV
+
+        echo "DOCKER_TAG=mosaicml/llm-foundry:release_${TAG_NAME}" >> $GITHUB_ENV
+        echo "AWS_DOCKER_TAG=mosaicml/llm-foundry:release_${TAG_NAME}_aws" >> $GITHUB_ENV
+        echo "LATEST_TAG=mosaicml/llm-foundry:release-latest" >> $GITHUB_ENV
+        echo "AWS_LATEST_TAG=mosaicml/llm-foundry:release_aws-latest" >> $GITHUB_ENV
```
dakinggg, line 83: can we be consistent and put the _aws at the end like the other case?

KuuCi: Leaving as is after dms

dakinggg (approved):

PR #1549, opened 2024-09-25
Bump version to 0.13.0.dev0

dakinggg:
Bump version to 0.13.0.dev0 and remove associated deprecations. In preparation for 0.12.0 release.

For the deprecations, realized that the wrong version had been used (would have only been deprecated for one version instead of two), so bumped those


llmfoundry/command_utils/eval.py
```diff
@@ -82,7 +82,7 @@ def evaluate_model(
         warnings.warn(
             VersionedDeprecationWarning(
                 'The argument fsdp_config is deprecated. Please use parallelism_config instead.',
-                remove_version='0.13.0',
+                remove_version='0.14.0',
```
snarayan21, line 85: we're punting this deprecation until later?

snarayan21 (approved):
lgtm, but we're pushing back fsdp_config deprecation another version?

llmfoundry/command_utils/eval.py
dakinggg: see pr description. its just for this function, and it was added with only version of deprecation when it should have been two

mvpatel2000 (approved):

PR #1551, opened 2024-09-25
Validate Cluster Access Mode

KuuCi:
To resolve:
https://databricks.atlassian.net/browse/ES-1264005
Basically this error happens if the cluster access mode is not either `Single User` or `Shared`

Testing:
image

image

### DIFF
--- a/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
+++ b/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
@@ -547,10 +547,8 @@ def validate_and_get_cluster_info(
         if res is None:
             raise ClusterDoesNotExistError(cluster_id)
 
-        policy_id = res.policy_id
-        assert policy_id is not None
-        print(res.policy_id)
-        print(w.cluster_policies.get(policy_id=policy_id))
+        print(res)
+        time.sleep(10)
 
         assert res.spark_version is not None
         stripped_runtime = re.sub(
--- a/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
+++ b/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
@@ -552,7 +552,7 @@ def validate_and_get_cluster_info(
         ).upper()[len('DATASECURITYMODE.'):]
 
         # NONE stands for No Isolation Shared
-        if data_security_mode in ('NONE'):
+        if data_security_mode == 'NONE':
             raise ClusterInvalidAccessMode(
                 cluster_id=cluster_id,
                 access_mode=data_security_mode,
--- a/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
+++ b/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
@@ -550,10 +550,9 @@ def validate_and_get_cluster_info(
         data_security_mode = str(res.data_security_mode).upper()[len('DATASECURITYMODE.'):]
         spark_conf = res.spark_conf
         if spark_conf:
-            compute_mode = spark_conf.get('spark.databricks.cluster.computeMode')
-            print('compute_mode: ', compute_mode)
+            print('spark_conf: ', spark_conf)
         else:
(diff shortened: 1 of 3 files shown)


llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -546,6 +547,14 @@ def validate_and_get_cluster_info(
         if res is None:
             raise ClusterDoesNotExistError(cluster_id)
 
+        data_security_mode = str(res.data_security_mode).upper()[len('DATASECURITYMODE.'):]
+
+        # USER_ISOLATION stands for Shared Access Mode
+        if data_security_mode not in ('USER_ISOLATION', 'SINGLE_USER'):
```
dakinggg, line 553: are you confident these are the only allowed options?

KuuCi: I'm confident that that's what the spark says ;-;
`[2024-09-18 13:43:22.187000+00:00][16308]: SingleClusterComputeMode(CLUSTER_ID) is not Shared or Single User `

dakinggg: What _is_ the data security mode that doesn't work? Would it be better to exclude that one rather than guess at the the ones to include?

KuuCi: I think `data_security_mode == None` is only one not allowed.... from what I've seen. But I also know they deprecated some of the things. Down to just exclude `data_security_mode == None` (No isolation Shared)

dakinggg: is there any documentation about this you could find?

dakinggg: found this

    """Data security mode decides what data governance model to use when accessing data from a cluster.
    
    * `NONE`: No security isolation for multiple users sharing the cluster. Data governance features
    are not available in this mode. * `SINGLE_USER`: A secure cluster that can only be exclusively
    used by a single user specified in `single_user_name`. Most programming languages, cluster
    features and data governance features are available in this mode. * `USER_ISOLATION`: A secure
    cluster that can be shared by multiple users. Cluster users are fully isolated so that they
    cannot see each other's data and credentials. Most data governance features are supported in
    this mode. But programming languages and cluster features might be limited.
    
    The following modes are deprecated starting with Databricks Runtime 15.0 and will be removed for
    future Databricks Runtime versions:
    
    * `LEGACY_TABLE_ACL`: This mode is for users migrating from legacy Table ACL clusters. *
    `LEGACY_PASSTHROUGH`: This mode is for users migrating from legacy Passthrough on high
    concurrency clusters. * `LEGACY_SINGLE_USER`: This mode is for users migrating from legacy
    Passthrough on standard clusters. * `LEGACY_SINGLE_USER_STANDARD`: This mode provides a way that
    doesn’t have UC nor passthrough enabled."""

dakinggg: lets just exclude the none one for now. can always revisit if we're wrong

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -546,6 +547,17 @@ def validate_and_get_cluster_info(
         if res is None:
             raise ClusterDoesNotExistError(cluster_id)
 
+        data_security_mode = str(
+            res.data_security_mode,
+        ).upper()[len('DATASECURITYMODE.'):]
+
+        # NONE stands for No Isolation Shared
+        if data_security_mode in ('NONE'):
```
dakinggg, line 555:
```suggestion
        if data_security_mode == 'NONE':
```

dakinggg (commented):
can you do a new manual test confirming the update still works?

dakinggg (approved):

dakinggg (approved):

PR #1553, opened 2024-09-26
Use `allenai/c4` instead of `c4` dataset

eitanturok:
HuggingFace has deprecated the [c4](https://huggingface.co/datasets/c4) dataset and recommends using [allenai/c4](https://huggingface.co/datasets/allenai/c4) dataset.

See the warning in a recent github workflow [here](https://github.com/mosaicml/llm-foundry/actions/runs/11056558091/job/30718370052#step:3:2047).


eitanturok:
Sorry, this had some other commits, closing and opening back up.

PR #1555, opened 2024-09-26
Insufficient Permissions Error when trying to access table

KuuCi:
To resolve this when permissions thrown by spark are not properly wrapped.

Instead of regex every variant of security perm errors, it's a lot easier to just rewarp the errors thrown by spark. Three scenerios were manually tested and passed:

User doesn't have access to catalog
User doesn't have access to schema (but has access for catalog)
User doesn't have SELECT for table (has access to catalog + schema)

This way, it's a lot easier to catch the permission spark errors.


dakinggg (approved):
Please write a more complete PR description, the ES ticket is internal

dakinggg (approved):
New version is much nicer :) Although I think you still need to update the tests

PR #1556, opened 2024-09-27
Cleanup TP

eitanturok:
This PR cleans up TP:
* Removes `@experimental_function('Tensor Parallelism')` from `build_tp_stratigies` because we already raise this warning in composer's TP [implementation](https://github.com/mosaicml/composer/blob/e5e2f744baa9f5bec8ff24b58f5a6b2a564919d0/composer/core/state.py#L600).
* Don't do a deep-copy of tp_config, instead create a new tp_config in `train.py`. This makes things a bit cleaner.
* Check that `tp_config` contains both 'strategy' and 'tensor_parallel_degree' fields.
* We previously only checked the tp edge cases but never had a simple example that tested tp actually working. So I added the ` test_tp_train` test to do this.
* I gave the tests better names

### DIFF
--- a/llmfoundry/command_utils/train.py
+++ b/llmfoundry/command_utils/train.py
@@ -5,6 +5,7 @@ import logging
 import os
 import time
 import warnings
+from copy import deepcopy
 from typing import Any, Optional, Union
 
 import torch
@@ -43,6 +44,7 @@ from llmfoundry.utils.builders import (
     build_save_planner,
     build_scheduler,
     build_tokenizer,
+    build_tp_strategies,
 )
 from llmfoundry.utils.config_utils import (
     TRAIN_CONFIG_KEYS,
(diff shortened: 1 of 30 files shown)


llmfoundry/utils/builders.py
```diff
@@ -701,7 +701,6 @@ def _validate_cfg(icl_cfg: dict[str, Any]):
     return evaluators, logger_keys
 
 
-@experimental_function('Tensor Parallelism')
```
dakinggg, line 704: I think we should leave this personally. We may want to remove the experimental designation in Composer and Foundry at different times.

llmfoundry/command_utils/train.py
```diff
@@ -518,8 +517,11 @@ def train(cfg: DictConfig) -> Trainer:
     # TP config
     if tp_config is not None:
         strategy = tp_config.pop('strategy', None)
-        assert isinstance(strategy, str), '`strategy` must be in `tp_config`.'
-        tp_config['layer_plan'] = build_tp_strategies(strategy, model)
+        layer_plan = build_tp_strategies(strategy, model)
+        tp_config = {
```
dakinggg, line 521: the compose tp config has other keys allowed. should we just be constructing that dataclass here?

tests/tp/test_tp_strategies.py
```diff
@@ -97,8 +99,34 @@ def test_ffn_tp_strategy():
 
 
 @pytest.mark.gpu
-def test_no_tp_with_one_gpu():
-    """Test that when we have one GPU, we use DDP and not FSDP-TP."""
+@pytest.mark.world_size(4)
+@pytest.mark.parametrize('tp_strategy', ['ffn'])
+def test_tp_train(tp_strategy: str):
```
dakinggg, line 104: tests like this are reasonably costly and also can be hard to debug. is there a simpler test that would achieve the same coverage?

llmfoundry/utils/builders.py
mvpatel2000: +1

llmfoundry/command_utils/train.py
mvpatel2000: +1

llmfoundry/utils/builders.py
eitanturok: added back

llmfoundry/command_utils/train.py
eitanturok: done

tests/tp/test_tp_strategies.py
eitanturok: I believe this test is okay because
* we only train for one step
* we want to make sure things work end-to-end
* 
I spoke to @mvpatel2000 about this and he said he thought it was okay.

llmfoundry/command_utils/train.py
```diff
@@ -517,9 +516,9 @@ def train(cfg: DictConfig) -> Trainer:
 
     # TP config
     if tp_config is not None:
-        strategy = tp_config.pop('strategy', None)
-        assert isinstance(strategy, str), '`strategy` must be in `tp_config`.'
-        tp_config['layer_plan'] = build_tp_strategies(strategy, model)
+        strategy = tp_config.pop('strategy')
+        layer_plan = build_tp_strategies(strategy, model)
```
snarayan21, line 520: is dataloading replication not needed in this PR? or is it already in

eitanturok: discussed offline.

PR #1562, opened 2024-09-30
Add error to catch more unknown example types

milocress:
Adds an error to catch more unknown example types


vanshcsingh (approved):

milocress:
> PR description?

same as title

snarayan21 (approved):

XiaohanZhangCMU:
care to add a test? in case people (me:)) mess it around in the future.

PR #1565, opened 2024-10-01
Add InvalidConversationError

b-chu:
Adds a new UserError for when the conversation can't be correctly sliced


milocress (approved):

mvpatel2000 (commented):
Should this require a unit test?

dakinggg (commented):
could you dig a little deeper? I don't think this error should ever be hit. how does it happen?

b-chu:
conversation_through_previous_turn ends in `<|eot_id|><|start_header_id|>assistant<|end_header_id|>`

the corresponding part in full_conversation ends in `<|eot_id|><|start_header_id|>user<|end_header_id|>`

b-chu:
The conversation goes system, user, assistant, user, assistant, assistant and I think we require alternating user and assistant

PR #1567, opened 2024-10-03
Allow parameters with requires_grad=False in meta init

sashaDoubov:
This PR allows the use of parameters with `requires_grad=False` within the meta init context, can be useful for `mixed init` edge cases.

### DIFF
--- a/llmfoundry/models/utils/meta_init_context.py
+++ b/llmfoundry/models/utils/meta_init_context.py
@@ -102,11 +102,9 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
             else:
                 param_cls = type(parameter)
                 kwargs = parameter.__dict__
-                # requires grad is separate from the kwargs
-                requires_grad = parameter.requires_grad
                 self._parameters[name] = param_cls(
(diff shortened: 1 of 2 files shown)


llmfoundry/models/utils/meta_init_context.py
```diff
@@ -102,8 +102,11 @@ def register_empty_parameter(
             else:
                 param_cls = type(parameter)
                 kwargs = parameter.__dict__
+                # requires grad is separate from the kwargs
+                requires_grad = parameter.requires_grad
                 self._parameters[name] = param_cls(
                     parameter.to(device),
+                    requires_grad=requires_grad,
                     **kwargs,
```
mvpatel2000, line 110:
```suggestion
                self._parameters[name] = param_cls(
                    parameter.to(device),
                    requires_grad=parameter.requires_grad,  # requires grad is separate from the kwargs
                    **kwargs,
```

mvpatel2000:
Test?

mvpatel2000 (approved):
LGTM. Please put a PR descriptin

PR #1570, opened 2024-10-05
Add load checkpoint callback

irenedea:
Depends on https://github.com/mosaicml/composer/pull/3641


dakinggg (commented):
request a test that the callback can be instantiated (will also force this PR to wait until the next composer release)

irenedea:
@dakinggg do we do this for other callbacks? Shouldn't instantiation be covered by whether or not I can import it (which it's failing until the composer release is out) and registry tests?

dakinggg:
@irenedea I don't think we've been super consistent (and you're right that the import covers what I said), but I tend try to write a simple instantiation test for things too (e.g. https://github.com/mosaicml/llm-foundry/blob/788c1f59ca5832842007fca620ec6bbf3abe9611/tests/callbacks/test_system_metrics_monitor.py#L9). Possibly should just write a single parametrized test that iterates over all the callbacks and tries to instantiate them.

irenedea:
@dakinggg added generic test for callback building here https://github.com/mosaicml/llm-foundry/pull/1577

dakinggg (approved):

PR #1571, opened 2024-10-05
Do not consider run_name when auto-detecting autoresume

irenedea:
Removes the consideration for run_name as there is already a default

### DIFF
--- a/llmfoundry/command_utils/train.py
+++ b/llmfoundry/command_utils/train.py
@@ -61,7 +61,6 @@ from llmfoundry.utils.exceptions import (
     TrainDataLoaderLocation,
 )
 from llmfoundry.utils.registry_utils import import_file
-from llmfoundry.utils.warnings import VersionedDeprecationWarning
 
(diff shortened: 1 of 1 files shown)


dakinggg (commented):
Why do we want to deprecate this? I thought we could just remove the run_name condition since run_name has a default.

irenedea:
ah, I may have understood what we discussed. I think it's odd to automatically switch to autoresume when run name is set, so we should deprecate this feature entirely.

irenedea:
I'm fine with keeping it as is (or making the change you suggested). I don't feel strongly and this would certainly eventually break some runs.

dakinggg:
The intention of this feature is that, generally speaking, you want autoresume to be on. The reason you would not want to do this is if you were sharing a save_folder between runs, so the condition was just trying to be a little safe about it. but we still want to set it to true in most cases.

dakinggg (approved):

PR #1577, opened 2024-10-07
Add build callback tests

irenedea:
Adds tests to check that we can build callbacks correctly.


tests/callbacks/test_callbacks.py
```diff
@@ -0,0 +1,34 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from composer.core import Callback
+
+from llmfoundry.registry import callbacks, callbacks_with_config
+from llmfoundry.utils.builders import build_callback
+
+
+def test_build_callbacks():
+    entry_points = callbacks.get_entry_points()
```
dakinggg, line 11: I think you want to use this https://github.com/mosaicml/llm-foundry/blob/23caea7532fb37f23fde2a1bfff798f565f9aeb2/llmfoundry/utils/registry_utils.py#L69?

irenedea: oops, yeah you're right. What are entry points?

dakinggg: entry points are just the part of the registry specified via pyproject.toml

tests/callbacks/test_callbacks.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def get_default_kwargs(callback_class: type):
+    type_hints = typing.get_type_hints(callback_class.__init__)
+    inspected_params = inspect.signature(callback_class.__init__).parameters
+
+    default_kwargs = {}
+
+    for param, tpe in type_hints.items():
+        print(param, tpe)
```
dakinggg, line 65: rm

dakinggg (approved):

tests/callbacks/test_callbacks.py
irenedea:
```suggestion
```

PR #1579, opened 2024-10-08
Handle long file names in convert text to mds

irenedea:
Sometimes we get `OSError: [Errno 36] File name too long` because the downloaded file names are too long.

### DIFF
--- a/llmfoundry/utils/data_prep_utils.py
+++ b/llmfoundry/utils/data_prep_utils.py
@@ -3,6 +3,7 @@
 
 import json
 import os
+import tempfile
 from glob import glob
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


llmfoundry/utils/data_prep_utils.py
```diff
@@ -131,7 +132,8 @@ def __iter__(self):
                 output_filename = os.path.join(
                     self.output_folder,
                     object_name.strip('/'),
-                )
```
snarayan21, line 134: just for my understanding, why don't we need to rename these files from the temp dir? I guess if someone hasn't specified the output_folder then it doesn't matter

snarayan21 (approved):
lgtm, just one comment for understanding

llmfoundry/utils/data_prep_utils.py
irenedea: ah like rename them with object_name? That was causing the name too long issue.

PR #1580, opened 2024-10-09
Make FinetuningStreamingDataset parameters more flexible

XiaohanZhangCMU:
Refactor build_text_dataloader and build_finetuning_dataloader to make them more flexible in terms of accepting extra keys used by StreamingDataset. The rest behavior should stay the same. For example, if the supplied extra keys to the dataloader are not defined in StreamingDataset, a normal TypeError is still expected to be thrown.

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -52,9 +52,7 @@ _ALLOWED_DATASET_KEYS = {
     'auto_packing_replication',
     'max_leftover_bins_to_keep',
     'pad_to_longest',
-    'warehouse_id',
-    'catalog',
-    'schema',
+    'delta_kwargs',
 }
 
 
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -246,6 +246,7 @@ def build_finetuning_dataloader(
         },
     )
 
+
     streaming_dataset = None  # for pyright
     sampler = None
     if is_streaming:
@@ -265,6 +266,7 @@ def build_finetuning_dataloader(
             if k in allowed_dataset_config_keys and
             k not in {'streams', 'packing_ratio'}
         }
+
         streaming_dataset = dataset_constructor.build_from_streaming(
             tokenizer=tokenizer,
             streams=streams,
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -258,18 +258,6 @@ def build_finetuning_dataloader(
             streams_cfg,
         ) if streams_cfg is not None else None
 
-        # Take the constructor args from above, minus args that have been created separately
-        # dataset_constructor_args = {
-        #     k: v
-        #     for k, v in dataset_cfg.items()
-        #     if k in allowed_dataset_config_keys and k not in {
-        #         'streams',
-        #         'packing_ratio',
-        #         'auto_packing_replication',
-        #         'seq_parallel_replication',
-        #         'pad_to_longest',
-        #     }
-        # }
         dataset_constructor_args = {
             k: v
             for k, v in dataset_cfg.items()
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -614,6 +614,7 @@ def _download_remote_hf_dataset(remote_path: str, split: str) -> str:
                     ]
                     raise FinetuningFileNotFoundError(
                         files_searched=files_searched,
+                        supported_extensions=SUPPORTED_EXTENSIONS,
                     ) from e
                 else:
                     log.debug(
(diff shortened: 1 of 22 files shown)


llmfoundry/data/finetuning/dataloader.py
```diff
@@ -257,9 +262,15 @@ def build_finetuning_dataloader(
         dataset_constructor_args = {
             k: v
             for k, v in dataset_cfg.items()
-            if k in dataset_constructor_keys and
-            k not in {'streams', 'packing_ratio'}
+            if k in allowed_dataset_config_keys and k not in {
```
milocress, line 265: Is there a way to do this without hardcoding the expected streaming keys? This requires that we manually keep track of any changes in the streaming repo and sync them to foundry whenever these args change on either side.

tests/data/test_dataloader.py
```diff
@@ -686,7 +686,7 @@ def test_finetuning_dataloader_streaming(
         'dataset': {
             'max_seq_len': 2048,
             'decoder_only_format': True,
-            'allow_pad_trimming': False,
+            # 'allow_pad_trimming': False, # This should not be allowed by Streaming
```
milocress, line 689: should we delete this line?

milocress (commented):
Deferring to @dakinggg , my intuition is there is some way of programmatically inspecting the streaming arguments and that this is preferable to hardcoding `streams`, `packing_ratio`, etc. as acceptable streaming arguments.

llmfoundry/data/finetuning/dataloader.py
XiaohanZhangCMU: ``streams`` needs to be hardcoded to get excluded. The others may be able to get programatically. But I am also inclined to keep the change minimal, in case it may affect any existing use cases unexpectedly.

I had to add these extra ones because this change [this](https://github.com/mosaicml/llm-foundry/blob/f034a8612708ca7b5c6a0523dbca9f03ffbf28e6/llmfoundry/data/finetuning/tasks.py#L655), breaks a unit test: tests/data/test_packing.py: test_auto_packing_with_streaming_dataloader.

llmfoundry/data/text_data.py
```diff
@@ -328,9 +324,11 @@ def build_text_dataloader(
         if 'streams' in dataset_cfg else None,
     )
 
-    valid_streaming_text_dataset_parameters = inspect.signature(
-        StreamingTextDataset,
-    ).parameters
+    valid_streaming_text_dataset_parameters = set(
```
dakinggg, line 327: this code turns this line and the next one into a noop right?

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -254,12 +259,23 @@ def build_finetuning_dataloader(
...
+        #         'seq_parallel_replication',
+        #         'pad_to_longest',
+        #     }
+        # }
         dataset_constructor_args = {
             k: v
             for k, v in dataset_cfg.items()
-            if k in dataset_constructor_keys and
-            k not in {'streams', 'packing_ratio'}
+            if k in set(dataset_constructor_keys).union(extraneous_keys)
```
dakinggg, line 276: pretty sure you still need to exclude streams and packing_ratio because they are passing in explicitly

XiaohanZhangCMU: That's true. didn't see that happen.

llmfoundry/data/text_data.py
XiaohanZhangCMU: 100%. logic is further cleaned up. plz tal.

llmfoundry/data/finetuning/dataloader.py
```diff
@@ -198,11 +198,16 @@ def build_finetuning_dataloader(
...
         dataset_constructor_keys,
     ).union(_ALLOWED_DATASET_KEYS)
-    _validate_config(
+
+    extraneous_keys = _validate_config(
         **dataset_cfg,
         allowed_dataset_keys=allowed_dataset_config_keys,
     )
 
+    allowed_dataset_config_keys = allowed_dataset_config_keys.union(
```
dakinggg, line 207: i don't think this gets used anywhere?

tests/data/test_dataloader.py
XiaohanZhangCMU: removed.

llmfoundry/data/finetuning/dataloader.py
XiaohanZhangCMU: Yes, u r right~~ I don't need to change the line.

tests/data/test_dataloader.py
```diff
@@ -686,7 +686,6 @@ def test_finetuning_dataloader_streaming(
         'dataset': {
             'max_seq_len': 2048,
             'decoder_only_format': True,
-            'allow_pad_trimming': False,
```
dakinggg, line 689: why did you need to remove this arg? this is a valid dataset config key

XiaohanZhangCMU: It was complaining this param, but not any more.

dakinggg (approved):
lgtm

PR #1585, opened 2024-10-11
Error if filtered dataset contains 0 examples

irenedea:
This presented as a StopIteration error during profiling because the dataset was completely empty.

### DIFF
--- a/tests/data/test_dataloader.py
+++ b/tests/data/test_dataloader.py
@@ -437,7 +437,7 @@ def test_finetuning_dataloader_safe_load(
         'dataset': {
             'hf_name': hf_name,
             'split': 'train',
-            'max_seq_len': 8,
+            'max_seq_len': 100,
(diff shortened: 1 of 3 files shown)


tests/data/test_dataloader.py
```diff
@@ -437,7 +437,7 @@ def test_finetuning_dataloader_safe_load(
         'dataset': {
             'hf_name': hf_name,
             'split': 'train',
-            'max_seq_len': 8,
+            'max_seq_len': 100,
```
snarayan21, line 440: jw, any particular reason to make this change?

snarayan21 (approved):
lgtm, just one question

tests/data/test_dataloader.py
irenedea: the dataset was empty bc the max_seq_len was too short and causing the new error to be raised :)

PR #1586, opened 2024-10-11
Change cluster errors from NetworkError to UserError

irenedea:
These should be considered UserErrors


b-chu (approved):
Makes sense!

jjanezhang (approved):
lgtm

PR #1587, opened 2024-10-11
Bump version to 0.14.0.dev0

irenedea:
Bump version to 0.14.0.dev0.


llmfoundry/command_utils/eval.py
```diff
@@ -78,18 +75,6 @@ def evaluate_model(
                     'parallelism_config cannot contain deprecated fsdp_config arguments.',
                 )
 
-    if fsdp_config:
```
snarayan21, line 81: could be remembering this wrong, were there some reg tests which had some issues with this? i think it should be ok but just want to make sure

milocress (approved):

llmfoundry/command_utils/eval.py
```diff
@@ -345,7 +328,6 @@ def evaluate(cfg: DictConfig) -> tuple[list[Trainer], pd.DataFrame]:
              device_eval_batch_size=eval_config.device_eval_batch_size,
              eval_gauntlet_config=eval_gauntlet_config,
              eval_loader_config=eval_loader_config,
-             fsdp_config=fsdp_config,
```
snarayan21, line 348: shouldn't we be passing in `parallelism_config` isntead?

tests/models/test_model.py
```diff
@@ -2560,7 +2562,7 @@ def test_hf_init(
         False,
     )
 
-    model = HuggingFaceModelWithFSDP(model, tokenizer)
+    model = HuggingFaceModel(model, tokenizer)
```
dakinggg, line 2565: would it be more correct to change this to BaseHuggingFaceModel (changing the args as appropriate)? glancing at what this was testing that might be better?

dakinggg (approved):
one comment, otherwise lgtm

llmfoundry/command_utils/eval.py
dakinggg: ah yeah, looks like that wasn't added before, but should have been?

irenedea: oops yes, added. good catch!

llmfoundry/command_utils/eval.py
```diff
@@ -127,7 +112,7 @@ def evaluate_model(
     fsdp_config = parallelism_config.get(
         'fsdp_config',
         None,
-    ) if parallelism_config else fsdp_config
```
dakinggg, line 130: below when making the trainer object, i think we should just pass along `parallelism_config` instead of making it again

dakinggg (approved):
looks right now, one other clean up comment

PR #1599, opened 2024-10-17
Add configurability for hf checkpointer register timeout

dakinggg:
Also increases the default from one hour two hours.


mvpatel2000 (approved):
intend to change defaults?

dakinggg:
@mvpatel2000 yup, see PR description

PR #1600, opened 2024-10-17
Finetuning dataloader validation tweaks

mvpatel2000:
Finetuning dataloader validation tweaks. Enables local streaming datasets. Also adds correct validation for decoder only. Also corrects replication parsing.

```
train_loader:
  name: finetuning
  dataset:
    local: ${variables.data_local}
    remote: ${variables.data_remote}  # None
    split: train
    shuffle: true
    max_seq_len: ${variables.max_seq_len}
    shuffle_seed: ${variables.global_seed}
    decoder_only_format: true  # now validated
    replication: 2  # now properly parsed
  drop_last: true
  num_workers: 8
```

### DIFF
--- a/llmfoundry/data/finetuning/dataloader.py
+++ b/llmfoundry/data/finetuning/dataloader.py
@@ -440,7 +440,9 @@ def _validate_config(
         set[str]: Return the extraneous keys.
     """
     if decoder_only_format is None:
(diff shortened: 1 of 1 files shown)


irenedea (approved):
LGTM ty!

PR #1601, opened 2024-10-18
Enforce compute cluster version

KuuCi:
To resolve:
https://databricks.slack.com/archives/C06RXC1NPQC/p1727371779924089

Essentially, we check if the compute runtime version is higher than our `databricks-connect` version to determine whether to use 'dbconnect' or not, but we don't check if the compute runtime version is _too high_. There is no formula to determine which version is too high, so we should just enforce the customer using a computer that has the same major runtime as the `databricks-connect` version

Test:
Error should be thrown:
image
with the following compute:
image

But should work:

### DIFF
--- a/llmfoundry/command_utils/data_prep/convert_dataset_hf.py
+++ b/llmfoundry/command_utils/data_prep/convert_dataset_hf.py
@@ -451,6 +451,7 @@ def convert_dataset_hf_from_args(
         ValueError: If the output directory already contains the requested splits
         ValueError: If `concat_tokens` is set but `tokenizer` is not
     """
+    os.environ['WORLD_SIZE'] = '1'
     if tokenizer_kwargs:
         parsed_tokenizer_kwargs = json.loads(tokenizer_kwargs)
     else:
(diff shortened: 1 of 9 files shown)


llmfoundry/command_utils/data_prep/convert_dataset_hf.py
```diff
@@ -451,6 +451,7 @@ def convert_dataset_hf_from_args(
         ValueError: If the output directory already contains the requested splits
         ValueError: If `concat_tokens` is set but `tokenizer` is not
     """
+    os.environ['WORLD_SIZE'] = '1'
```
dakinggg: are these changes supposed to be included?

dakinggg (commented):
Is this check potentially too restrictive? do we _actually_ need an exact match? or we just don't know?

llmfoundry/command_utils/data_prep/convert_dataset_hf.py
KuuCi: Actually going to close this pr, during testing seems like every version I tested worked (strangely), so perhaps old bug was actually transient, and was solved not by the version change but simply restarting the cluster. Will revisit if it comes up again.

KuuCi:
Not needed as all versions currently work... see above comment.

llmfoundry/command_utils/data_prep/convert_dataset_hf.py
KuuCi: Original motivation is we don't know what works and what gets deprecated

PR #1602, opened 2024-10-18
Updated error message for cluster check

nancyhung:
While digging into cluster access controls issues, it looks like the existing check does not check for cluster access controls. It only checks for Unity Catalog compatibility. Updating the error message accordingly.

Documentation: https://docs.databricks.com/api/workspace/clusters/get#data_security_mode


dakinggg:
@KuuCi could you take a look please?

dakinggg (approved):

PR #1606, opened 2024-10-22
Add env var for configuring the maximum number of processes to use for dataset processing

irenedea:
By default we have been doing `os.num_cpus() - 8` to get the number of processes to use for dataset processing. On notebooks, the number of processes is too high and causes the notebook to crash. 

This PR enables configuring the maximum number of processes via env var.


milocress (commented):
Can you add a manual test?

irenedea:
@milocress Added a unit test. Is there a particular manual test you'd like me to run? I don't expect the manual runs to have any visible change.

bigning (approved):
lgtm, thanks for the change!

PR #1607, opened 2024-10-22
Change Release Base Images to 

KuuCi:
This is to speed up release process by making us not need to reinstall all of transformer_engine. Docker caching doesn't work here because of https://github.com/mosaicml/llm-foundry/blob/8e78eb53f37a01ddb249da36438de0f86f7cf688/Dockerfile#L15, where if the branch name is different than main (and it will be since we expect it to be whatever branch is associated with the release), cache will be invaliadted.

Testing:
image

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,8 +7,54 @@ on:
   workflow_dispatch:
 
 jobs:
+  code-quality:
+    uses: ./.github/workflows/code-quality.yaml
+
+  pypi-packaging:
+    name: Build and Publish llm-foundry PyPI Package
+    needs:
+    - code-quality
+    runs-on: linux-ubuntu-latest
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.9"
+
+    - name: Build source and wheel distributions
+      run: |
+        if [[ "${{ github.ref }}" =~ refs\/tags\/v ]]; then
+          PYPI_PACKAGE_NAME="llm-foundry"
+        else
+          PYPI_PACKAGE_NAME="llm-foundry-test-$(date +%Y%m%d%H%M%S)"
+        fi
+
+        python -m pip install --upgrade build twine
+        python -m build
+        twine check --strict dist/*
+
+    - name: Publish 📦 to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      if: contains(github.ref, 'refs/tags/v')
+      with:
+        user: __token__
+        password: ${{ secrets.PROD_PYPI_API_TOKEN }}
+
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -93,7 +99,7 @@ jobs:
           ${{ env.AWS_DOCKER_TAG }}
           ${{ env.AWS_LATEST_TAG }}
         build-args: |
-          BASE_IMAGE=mosaicml/pytorch:2.4.0_cu124-python3.11-ubuntu20.04-aws
+          BASE_IMAGE=mosaicml/llm-foundry:2.4.0_cu124-latest
```
mvpatel2000, line 102: how is base image the same image ur building...?

KuuCi: They're different images, the images built here are 
`mosaicml/llm-foundry:release_${TAG_NAME}"`
to keep up with foundry releases

snarayan21: wait yeah, how are we building off the foundry images? that should be the result of this right

.github/workflows/release.yaml
```diff
@@ -108,7 +114,7 @@ jobs:
           ${{ env.DOCKER_TAG }}
           ${{ env.LATEST_TAG }}
         build-args: |
-          BASE_IMAGE=mosaicml/pytorch:2.4.0_cu124-python3.11-ubuntu20.04
+          BASE_IMAGE=mosaicml/llm-foundry:2.4.0_cu124-latest
```
snarayan21, line 117: same here

snarayan21: is this cached....?

.github/workflows/release.yaml
KuuCi: ohh. Huh, I guess the behavior is slightly weird.

so https://github.com/mosaicml/llm-foundry/blob/main/.github/workflows/docker.yaml is what builds `mosaicml/llm-foundry:2.4.0_cu124-latest` (which is largely cached), and then `https://github.com/mosaicml/llm-foundry/blob/main/.github/workflows/release.yaml` builds from that image except with the foundry directory in it...

but like isn't that fine?

.github/workflows/release.yaml
KuuCi: closing this to keep everything in one thread

.github/workflows/release.yaml
snarayan21: But currently the release workflow builds on the pytorch image, not the foundry image...i guess the foundry image does exist though...

mvpatel2000: I don't think its worth optimizing release workflow to add this complication. This feels a bit weird to me to build on top of the foundry build image...

snarayan21: ^ agree

snarayan21: This can have some unintended complications -- for example when doing foundry release, suppose I create the release branch but don't publish the release and kick off the release workflow yet. Then, in between, someone else merges some breaking PR. Then I launch this new release workflow which builds on the cached latest foundry image that contains the bad commit....right?

KuuCi: Ok, sounds fair, will close pr

PR #1609, opened 2024-10-22
Add split_eval_set command

jimmyxu-db:


llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,185 @@
...
+
+log = logging.getLogger(__name__)
+
+DELTA_JSONL_REGEX = re.compile(r"^tmp-t$")
+REMOTE_OBJECT_STORE_FILE_REGEX = re.compile(
+    r"^((s3|oci|gs):\/\/|dbfs:\/Volumes\/)[/a-zA-Z0-9 ()_\-.]+$"
+)
+HF_REGEX = re.compile(r"^[/a-zA-Z0-9 ()_\-.]+$")
+
+def get_dataset_format(data_path_folder: str) -> str:
```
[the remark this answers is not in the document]
dakinggg: no, this logic lives only in the API right now.

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,185 @@
...
+        return "delta"
+    if REMOTE_OBJECT_STORE_FILE_REGEX.match(data_path_folder):
+        return "remote_object_store"
+    if HF_REGEX.match(data_path_folder):
+        return "hugging_face"
+    return "unknown"
+
+TEMP_DIR = "tmp-split"
+
+def maybe_download_data_as_json(data_path_folder: str, data_path_split: str) -> str:
```
dakinggg, line 44:
```suggestion
def maybe_download_data_as_jsonl(data_path_folder: str, data_path_split: str) -> str:
```

tests/a_scripts/data_prep/test_split_eval_set.py
```diff
@@ -0,0 +1,217 @@
...
+def test_remote_object_store_file_regex():
+    """Test the regex pattern for remote object store file paths"""
+    assert REMOTE_OBJECT_STORE_FILE_REGEX.match("s3://bucket-name/path/to/file")
+    assert REMOTE_OBJECT_STORE_FILE_REGEX.match("oci://bucket-name/path/to/file")
+    assert REMOTE_OBJECT_STORE_FILE_REGEX.match("gs://bucket-name/path/to/file")
+    assert REMOTE_OBJECT_STORE_FILE_REGEX.match("dbfs:/Volumes/path/to/file")
+    assert not REMOTE_OBJECT_STORE_FILE_REGEX.match("https://bucket-name/path/to/file")
+    assert not REMOTE_OBJECT_STORE_FILE_REGEX.match("/local/path/to/file")
+    assert not REMOTE_OBJECT_STORE_FILE_REGEX.match("s3:/bucket-name/path/to/file")
+    assert not REMOTE_OBJECT_STORE_FILE_REGEX.match("s3://bucket-name/path/to/file with spaces")
```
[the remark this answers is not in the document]
dakinggg: yeah, i believe they can.

tests/a_scripts/data_prep/test_split_eval_set.py
```diff
@@ -0,0 +1,217 @@
...
+    assert not REMOTE_OBJECT_STORE_FILE_REGEX.match("s3://bucket-name/path/to/file?")
+
+def test_hf_regex():
+    """Test the regex pattern for Hugging Face dataset paths"""
+    assert HF_REGEX.match("dataset-name")
+    assert HF_REGEX.match("dataset_name")
+    assert HF_REGEX.match("dataset-name_with-mixed.characters-123")
+    assert not HF_REGEX.match("dataset/name")
```
dakinggg, line 44: er, shouldn't this pass? e.g. mosaicml/dolly_hhrlhf is our default testing dataset

tests/a_scripts/data_prep/test_split_eval_set.py
```diff
@@ -0,0 +1,217 @@
...
+
+# Default values
+OUTPUT_DIR = "tmp-split"
+TMPT_DIR = "tmp-t"
+DATA_PATH_SPLIT = "train"
+EVAL_SPLIT_RATIO = 0.1
+DEFAULT_FILE = TMPT_DIR + "/train-00000-of-00001.jsonl"
+
+
+def test_delta_jsonl_regex():
```
dakinggg, line 18: can you rewrite these tests to use `pytest.mark.parametrize`? e.g. this one would be like
```
@pytest.mark.parametrize("path,expectation", [("tmp-t", True), ("/tmp-t", False))
```

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,185 @@
...
+    # first pass: count total number of lines and determine sample size
+    total_lines = 0
+    with open(data_path, "r") as infile:
+        for _ in infile:
+            total_lines += 1
+    sample_size = int(eval_split_ratio * total_lines)
+    if max_eval_samples is not None:
+        sample_size = min(sample_size, max_eval_samples)
+
+    with temp_seed(seed) if seed is not None else contextlib.nullcontext():
```
dakinggg, line 146: rather than a temp seed context manager, its better to to explicitly create a new numpy rng instance with the seed you want and just use that for random operations in this scope.

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    Determine the format of the dataset from the provided data path
+
+    Args:
+        data_path_folder (str): Path to the training dataset folder
+
+    Returns:
+        str: The format of the dataset
+    """
+    if DELTA_JSONL_REGEX.match(data_path_folder):
```
dakinggg, line 34: I don't think you need to say anything about delta here do you? this is the same as the script just accepting a local path and can be treated generically as that.

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,185 @@
...
+import datasets as hf_datasets
+import numpy as np
+from typing import Optional
+
+import composer.utils as utils
+from llmfoundry.data.finetuning.tasks import maybe_safe_download_hf_data
+
+log = logging.getLogger(__name__)
+
+DELTA_JSONL_REGEX = re.compile(r"^tmp-t$")
```
dakinggg, line 18: probably better to not hardcode this and just have your functionality accept an arbitrary local path?

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,185 @@
...
+    with open(data_path, "r") as infile:
+        for _ in infile:
+            total_lines += 1
+    sample_size = int(eval_split_ratio * total_lines)
+    if max_eval_samples is not None:
+        sample_size = min(sample_size, max_eval_samples)
+
+    with temp_seed(seed) if seed is not None else contextlib.nullcontext():
+        random_numbers = np.random.rand(total_lines)
+        sample_indices = set(np.argsort(random_numbers)[:sample_size])
```
dakinggg, line 148: may not really matter at the number of rows we care about, but noting that this keeps a potentially large set in memory and sorts it. you could instead flip a coin for each row (np.random.uniform)

dakinggg (commented):
Looks reasonable, would want to also see a test run on MCT using this script

llmfoundry/command_utils/data_prep/split_eval_set.py
mattyding: Some notes:
1. These regexes are copied from MAPI and its logic
2. The issue is that the HF regex is quite generic so if we don't specify a Delta case here (by assuming a specific prefix) and check it before HF, it will mistakenly match on the HF case. This problem only occurs at this step because we convert Delta table to a local JSONL file and thus mess up its pathname formatting. The logic for processing HF datasets and local JSONL files differs because HF datasets have us look in path `foo/bar/split/file.parquet` or something and local JSONL file assumes `foo/bar/split.jsonl`.

llmfoundry/command_utils/data_prep/split_eval_set.py
mattyding: `random.uniform` seems fine assuming we're ok with slight variance in the number of samples, which I think is fine.

llmfoundry/data/finetuning/tasks.py
```diff
@@ -718,7 +718,72 @@ def state_dict(self, num_samples: int,
...
+
+    if not os.path.isdir(dataset_name):
+        local_dataset_dir = os.path.join(
+            DOWNLOADED_FT_DATASETS_DIRPATH,
+            dataset_name,
+        )
+
+        if _is_empty_or_nonexistent(dirpath=local_dataset_dir):
+            # Safely load the dataset from HF Hub with restricted file types.
+            hf_hub.snapshot_download(
```
mattyding, line 748: A note that I was previously running into cache errors when downloading HF data. The `safe_load` check would fail if the data was not present in cache. I couldn't diagnose the exact cause. 

This isn't blocking if we assume no one is using eval + a HF dataset (which I think is fair).

mattyding: Honestly should prob not support HF dataset splitting + throw an error. Just have all our e2e tests use some UC volume

dakinggg: yeah i think we had discussed just not supporting HF dataset for this to simplify things

mattyding: 👍 this also simplifies the HF regex issue above

llmfoundry/command_utils/data_prep/split_eval_set.py
mattyding: Actually, on second though, can use reservoir sampling to guarantee the same number of samples each time. In either case I think this is a low prio change but can add at some point. @jimmyxu-db

dakinggg:
Please also add a PR description (describing the functionality added, not any downstream API needs).

llmfoundry/command_utils/data_prep/split_eval_set.py
jimmyxu-db: maybe i'll rename this to local path vs remote object store?

llmfoundry/command_utils/data_prep/split_eval_set.py
jimmyxu-db: i changed this to be a string match on `"tmp-t"` instead of a regex, seemed simpler

llmfoundry/data/finetuning/tasks.py
jimmyxu-db: removed splitting support for hf datasets

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,181 @@
...
+import os
+import re
+from typing import Optional
+
+import composer.utils as utils
+import numpy as np
+
+log = logging.getLogger(__name__)
+
+LOCAL_PATH = 'tmp-t'
```
dakinggg, line 15: I don't think you want to hardcode this. Should this not just be a fallback to assume local path?

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,181 @@
...
+        str: The format of the dataset
+    """
+    if data_path_folder == LOCAL_PATH:
+        return 'local_file'
+    if REMOTE_OBJECT_STORE_FILE_REGEX.match(data_path_folder):
+        return 'remote_object_store'
+    return 'unknown'
+
+
+TEMP_DIR = 'tmp-split'
```
dakinggg, line 37: shouldn't we be using a real temp dir? or is there a reason we need to have this not be a real tempdir?

llmfoundry/command_utils/data_prep/split_eval_set.py
```diff
@@ -0,0 +1,181 @@
...
+    if not os.path.exists(data_path):
+        raise FileNotFoundError(
+            f'Expected dataset file at {data_path} for splitting, but it does not exist.',
+        )
+
+    return data_path
+
+
+@contextlib.contextmanager
+def temp_seed(seed: int):
```
dakinggg, line 95: remove?

llmfoundry/command_utils/data_prep/split_eval_set.py
jimmyxu-db: im not too familiar, is it acceptable to have
```
if remote object store:
  remote object store:
else:
  local path
```

i.e. have it be the defualt?

dakinggg: seems ok? just note that is what it defaults to

llmfoundry/command_utils/data_prep/split_eval_set.py
dakinggg: tempdirs typically get cleaned up by the system on some cadence, but even if that isn't in the case, it puts it in a unified place for temporary things that are no longer needed, and allows the user to configured where this file is written by configuring their temporary directory env vars (which we want to do in our services)

jimmyxu-db:
closing this PR as we are re-prioritizing the long term workflows solution over pushing out a sweep/eval PuPr asap, which this PR was meant to do

PR #1611, opened 2024-10-23
Use fun default composer run names

irenedea:
Our previous default "llm" is depressingly literal, and because it does not change, it is hard to distinguish runs from the same environment.

https://github.com/mosaicml/composer/blob/main/composer/trainer/trainer.py#L549 composer generates some really fun and random run names, we should take advantage of this.

### Testing
Tested on interactive by unsetting COMPOSER_RUN_NAME and RUN_NAME env vars.

created a run called `1729664919-daring-lori`
image

### DIFF
--- a/llmfoundry/command_utils/train.py
+++ b/llmfoundry/command_utils/train.py
@@ -312,9 +312,9 @@ def train(cfg: DictConfig) -> Trainer:
 
     # Optional parameters will be set to default values if not specified.
     env_run_name: Optional[str] = os.environ.get('RUN_NAME', None)
-    run_name: str = (
+    run_name: Optional[str] = (
         train_cfg.run_name if train_cfg.run_name else env_run_name
-    ) or 'llm'
+    )
     is_state_dict_sharded: bool = (
         fsdp_config.get('state_dict_type', 'full') == 'sharded'
     ) if fsdp_config else False
--- a/llmfoundry/command_utils/train.py
+++ b/llmfoundry/command_utils/train.py
@@ -323,9 +323,8 @@ def train(cfg: DictConfig) -> Trainer:
     save_filename: str = train_cfg.save_filename if train_cfg.save_filename else 'ep{epoch}-ba{batch}-rank{rank}.pt'
 
     # Enable autoresume from model checkpoints if possible
-    is_user_set_run_name: bool = train_cfg.run_name is not None or run_name is not None
     autoresume_default: bool = False
-    if is_user_set_run_name and \
+    if run_name is not None and \
         train_cfg.save_folder is not None \
         and not train_cfg.save_overwrite \
         and not train_cfg.save_weights_only:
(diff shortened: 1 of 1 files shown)


llmfoundry/command_utils/train.py
```diff
@@ -312,9 +312,9 @@ def train(cfg: DictConfig) -> Trainer:
 
     # Optional parameters will be set to default values if not specified.
     env_run_name: Optional[str] = os.environ.get('RUN_NAME', None)
-    run_name: str = (
+    run_name: Optional[str] = (
```
mvpatel2000, line 315: u can probably one line this and get rid of () now given line length

mvpatel2000 (commented):
https://github.com/mosaicml/llm-foundry/blob/c681a9c48002c5b7b1c847e264427ae6c9fbbd04/llmfoundry/command_utils/train.py#L325C68-L325C80

this variable can now be relaced with run_name is not None

env_run_name is also no longer needed and can be one-lined

llmfoundry/command_utils/train.py
```diff
@@ -311,18 +311,19 @@ def train(cfg: DictConfig) -> Trainer:
...
     ) if fsdp_config else False
     save_latest_filename: str = train_cfg.save_latest_filename if train_cfg.save_latest_filename else 'latest-sharded-rank{rank}' if is_state_dict_sharded else 'latest-rank{rank}.pt'
     save_filename: str = train_cfg.save_filename if train_cfg.save_filename else 'ep{epoch}-ba{batch}-rank{rank}.pt'
 
     # Enable autoresume from model checkpoints if possible
-    is_user_set_run_name: bool = train_cfg.run_name is not None or env_run_name is not None
+    is_user_set_run_name: bool = train_cfg.run_name is not None or run_name is not None
     autoresume_default: bool = False
     if is_user_set_run_name and \
         train_cfg.save_folder is not None \
```
mvpatel2000, line 329:
```suggestion
    autoresume_default: bool = False
    if run_name is not None and \
        train_cfg.save_folder is not None \
```
only used in one spot and also train_cfg is subset of run_name

llmfoundry/command_utils/train.py
```diff
@@ -311,20 +311,20 @@ def train(cfg: DictConfig) -> Trainer:
...
 
     # Optional parameters will be set to default values if not specified.
-    env_run_name: Optional[str] = os.environ.get('RUN_NAME', None)
-    run_name: str = (
-        train_cfg.run_name if train_cfg.run_name else env_run_name
-    ) or 'llm'
+    run_name: Optional[
+        str] = train_cfg.run_name if train_cfg.run_name else os.environ.get(
+            'RUN_NAME',
+            None,
```
mvpatel2000, line 317: this is really ugly lint lol.

mvpatel2000 (approved):
I love fun

PR #1613, opened 2024-10-24
Add UC not enabled error for delta to json conversion

irenedea:
Add UC not enabled error for delta to json conversion

https://databricks.atlassian.net/browse/ES-1290471

### DIFF
--- a/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
+++ b/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
@@ -714,11 +714,10 @@ def _check_imports():
         ) from e
 
     try:
+        from databricks import sql
         from databricks.sdk import WorkspaceClient
         from databricks.sql.client import Connection as Connection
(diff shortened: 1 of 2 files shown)


llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -711,10 +714,11 @@ def _check_imports():
         ) from e
 
     try:
-        from databricks import sql
         from databricks.sdk import WorkspaceClient
         from databricks.sql.client import Connection as Connection
         from databricks.sql.client import Cursor as Cursor
+
+        from databricks import sql
```
irenedea, line 721:
```suggestion
        from databricks import sql
        from databricks.sdk import WorkspaceClient
        from databricks.sql.client import Connection as Connection
        from databricks.sql.client import Cursor as Cursor
```

KuuCi (approved):
lgtm

PR #1616, opened 2024-10-25
Bump composer version to 0.26.0

irenedea:
Bump composer version to 0.26.0.

Depends on `llm-foundry-regression-tests-runner-7I6JZY` passing


dakinggg:
regression test run name?

mvpatel2000 (approved):

PR #1618, opened 2024-10-27
Change accumulate_train_batch_on_tokens default to True

dakinggg:
Following #1610, update the default value of `accumulate_train_batch_on_tokens` to `True`, for a more mathematically correct default. Note: this will slightly change loss curves for models trained with padding. The old behavior can be recovered if desired. by simply setting this to `False` explicitly. See #1610 and https://github.com/mosaicml/composer/pull/3677 for more discussion of this change.


snarayan21 (approved):
ah interesting. lgtm

Note: do we need a deprecation warning or no? This deviates from old behavior but i guess is a new feature...

dakinggg:
Open to other opinions, but given this is a correctness change, I think it is ok to change without deprecation (also deprecation warning for a default change is a bit unlikely to have any impact). Definitely will call this out in the next release notes.

PR #1623, opened 2024-10-28
Add Mlflow 403 PL UserError

mattyding:
## This PR

Adds networking UserError to Foundry.

Abandoning PFC change in favor of this. This try-catch should sufficiently cover the 403 networking errors we are seeing. Additionally, change is safer as it avoids duplicating logic in composer.

## Testing

Was gonna add a unit test that mocked a MLflow 500 exception but the test doesn't really fit anywhere in the testing suite and has strong assumptions, so not much benefit. Idk if others have thoughts

## Rollout (including migrations)

DLE release


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -296,7 +297,13 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
...
                 )
             if self.remote_ud is not None:
-                self.remote_ud.init(state, logger)
+                try:
+                    self.remote_ud.init(state, logger)
+                except PermissionError as e:
+                    log.error(
+                        f'Error initializing remote uploader/downloader. This is likely private link related. {e}',
+                    )
+                    raise ProbablyNetworkingError(is_contd_pretrain=state.run_name.startswith('contd-pretrain'))
```
dakinggg, line 306: the is_contd_pretrain check is pretty brittle. why do you need it? shouldn't this error be agnostic of task type since its just about whether you can upload to mlflow?

mattyding: Discussed offline. Original purpose was to throw CPT-specific error. Since this error doesn't occur on UC volume read (and we have no examples of actual runs that have errored on UC volume read), removing.

llmfoundry/utils/exceptions.py
```diff
@@ -524,3 +525,11 @@ def __reduce__(self):
 
     def __str__(self):
         return self.message
+
+
+class ProbablyNetworkingError(UserError):
```
dakinggg, line 530: should we just raise this as a permissions error? In the finetuning API, we can add the messaging around PL/SEG, since we know that is the cause there, but in general you could hit this for a variety of permissions error so we might want to keep the foundry message more general? wdyt

mattyding: This is a naming thing right? Rename this `ProbablyNetworkingError`->`StoragePermissionsError`.

That sounds fine to me.

dakinggg: makes sense to me, and yes, i just meant naming (and probably adjust the message a bit)

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -300,10 +300,7 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
                 try:
                     self.remote_ud.init(state, logger)
                 except PermissionError as e:
-                    log.error(
-                        f'Error initializing remote uploader/downloader. This is likely private link related. {e}',
-                    )
-                    raise ProbablyNetworkingError()
+                    raise StoragePermissionError('Error when write to save_folder.') from e
```
dakinggg, line 303:
```suggestion
                    raise StoragePermissionError('Error when writing to save_folder.') from e
```

dakinggg (approved):

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -297,7 +298,10 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
                     + f'Got {type(state.model)} instead.',
                 )
             if self.remote_ud is not None:
-                self.remote_ud.init(state, logger)
+                try:
+                    self.remote_ud.init(state, logger)
+                except PermissionError as e:
+                    raise StoragePermissionError('Error when writing to save_folder.') from e
```
nancyhung, line 304: can you check for the message I'm logging in the `PermissionError` I'm throwing to indicate it's specific to PL? If a user has a `PermissionError` that is not related to PL, we probably don't want to swallow this for every `PermissionError`

nancyhung:
```suggestion
                    if 'Client Error' e.message:
                        raise StoragePermissionError('Error when writing to save_folder, likely related to Private Link. Please contact support for next steps') from e
```

nancyhung (changes requested):
Can we also add a try/catch to the read operation when we download data via `uc.files.download(...)`? @mattyding

mattyding:
> Can we also add a try/catch to the read operation when we download data via `uc.files.download(...)`? @mattyding

I thought motivation for the PFC->wrapping change is that we have no examples of customer runs with data read errors and can't validate the PFC check. So any try/catch wrapping we do for read would be a best guess.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -297,7 +298,16 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
...
             if self.remote_ud is not None:
-                self.remote_ud.init(state, logger)
+                try:
+                    self.remote_ud.init(state, logger)
+                except PermissionError as e:
+                    if 'Client Error' in str(
+                        e,
+                    ):  # thrown from composer.utils._wrap_mlflow_exceptions
+                        raise StoragePermissionError(
+                            'Error when write to save_folder.',
```
nancyhung, line 308: do you want to propagate the error message up so that we can at least log the full error in mcloud?
```suggestion
                            f'Error when write to save_folder. Details: {e.message}',
```

nancyhung (approved):
lgtm

llmfoundry/callbacks/hf_checkpointer.py
mattyding: I think it's fine as-is since the earlier error will appear in the stacktrace and logs

PR #1625, opened 2024-10-29
Catch delta table not found error

milocress:
when the delta table isn't found, surface this error to the user.

### DIFF
--- a/llmfoundry/utils/exceptions.py
+++ b/llmfoundry/utils/exceptions.py
@@ -532,20 +532,23 @@ class UCNotEnabledError(UserError):
     """Error thrown when user does not have UC enabled on their cluster."""
 
     def __init__(self) -> None:
-        message = f'Unity Catalog is not enabled on your cluster.'
+        message = 'Unity Catalog is not enabled on your cluster.'
         super().__init__(message)
 
 
 class DeltaTableNotFoundError(UserError):
     """Error thrown when the delta table passed in training doesn't exist."""
 
-    def __init__(self, message: str) -> None:
-        self.message = message
-        super().__init__(message)
-
-    def __reduce__(self):
-        # Return a tuple of class, a tuple of arguments, and optionally state
-        return (DeltaTableNotFoundError, (self.message,))
-
-    def __str__(self):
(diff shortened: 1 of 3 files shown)


llmfoundry/utils/exceptions.py
```diff
@@ -532,3 +534,18 @@ class UCNotEnabledError(UserError):
...
         message = f'Unity Catalog is not enabled on your cluster.'
         super().__init__(message)
+
+
+class DeltaTableNotFoundError(UserError):
+    """Error thrown when the delta table passed in training does not exist or is not found."""
+
+    def __init__(self, message: str) -> None:
+        self.message = message
+        super().__init__(message)
```
jjanezhang, line 544: Can you pass me the table name so I can construct a good message for it in mcloud? or we could pass this message directly to the user if you feel it is perfect for the user.

milocress: The current message is:

```
Error processing `catalog`.`volume_name`.`table_name`: [DELTA_TABLE_NOT_FOUND] Delta table `volume_name`.`table_name` doesn't exist.
```

which I think is pretty good.

milocress:
The `docformatter` hook runs without error locally, it grates like sandpaper against my soul that this hook would run differently on the server.

llmfoundry/utils/exceptions.py
dakinggg: I think @jjanezhang wanted the table name as an attr on the error message to use downstream

milocress:
I think it's cool that docformatter won't tell me which file is incorrect, it's like a fun little scavenger hunt.

llmfoundry/utils/exceptions.py
jjanezhang: Yea I think I'm going to rewrite to be something like "Your {train/eval} data path {table name} does not exist. Please double check your delta table name". If you could pass the delta table name we check as an attr on the class that would be great!

dakinggg (approved):
approving from foundry side. please make sure you and jane agree on the unresolved conversation before merging

PR #1626, opened 2024-10-30
Only do `_initialize_dist_with_barrier` if world_size is greater than 1

irenedea:
we shouldn't need to do this for a single gpu


bigning:
can we test it with the notebook? Maybe try pip install your branch in notebook ?

irenedea:
discussed offline with @dakinggg, we shouldn't need to do this

PR #1631, opened 2024-11-01
Bump transformers version

dakinggg:
Only material change is to add a default chatml template in the training code, as transformers removed default chat templates in their library, but we don't want to break finetuning configurations of existing models that don't have a chat template specified.


llmfoundry/data/finetuning/tasks.py
```diff
@@ -243,17 +252,21 @@ def slice_out_last_turn(
         messages_through_current_turn: list[dict[str, str]],
         conversation_through_previous_turn: str,
     ) -> tuple[str, str]:
+        chat_template = None if tokenizer.chat_template is not None else _DEFAULT_CHAT_TEMPLATE
+
```
irenedea, line 256: nit: can you log that we're using the default chat template

irenedea (approved):
lgtm

llmfoundry/data/finetuning/tasks.py
dakinggg: I don't want to add a log here because it will run for every row in the dataset :)

PR #1632, opened 2024-11-02
Move loss generating token counting to the dataloader

dakinggg:
The previous PR to count loss generating tokens caused a throughput regression because the token counting is on the critical path in the Composer trainer. This PR moves the token counting into the collator so that it can be hidden in the dataloader. It does this by counting tokens per row in the batch, using the existing token counting function (not the most efficient, but doesn't matter because this is not dataloader bottlenecked, and allows reusing the existing code).

This PR should not change token counts or loss, and will barely affect throughput for larger models, but have a significant impact for smaller models.

125m model pretraining throughput, loss, token count, before and after this PR:
Screenshot 2024-11-02 at 3 19 20 PM
Screenshot 2024-11-02 at 3 19 26 PM
Screenshot 2024-11-02 at 3 19 34 PM

llama finetune throughput, loss, token count, before and after this PR:

### DIFF
--- a/llmfoundry/data/contrastive_pairs/dataloader.py
+++ b/llmfoundry/data/contrastive_pairs/dataloader.py
@@ -249,6 +249,11 @@ def build_pairs_dataloader(
         processed_batch: dict[str, torch.Tensor] = collate_fn(batch)
         if 'labels' in processed_batch:
             del processed_batch['labels']
+
+        if 'total_tokens' in processed_batch:
+            del processed_batch['total_tokens']
+        if 'loss_generating_tokens' in processed_batch:
+            del processed_batch['loss_generating_tokens']
         return processed_batch
 
     dl = DataLoader(
--- a/llmfoundry/data/utils.py
+++ b/llmfoundry/data/utils.py
@@ -117,9 +117,7 @@ def get_tokens_per_batch_func(
 
         loss_generating_tokens = None
         if 'labels' in batch:
-            loss_generating_tokens = int(
-                torch.sum(batch['labels'][...,1:] != CROSS_ENTROPY_IGNORE_INDEX).item(),
-            )
+            loss_generating_tokens = int((batch['labels'][...,1:] != CROSS_ENTROPY_IGNORE_INDEX).sum())
 
         # For encoder decoder models only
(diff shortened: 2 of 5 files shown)


llmfoundry/data/contrastive_pairs/dataloader.py
```diff
@@ -249,6 +249,11 @@ def collate_fn_without_labels(batch: list[Any]) -> dict[str, torch.Tensor]:
         processed_batch: dict[str, torch.Tensor] = collate_fn(batch)
         if 'labels' in processed_batch:
             del processed_batch['labels']
+
```
mvpatel2000, line 252:
```suggestion
```
nit remove extra \n

llmfoundry/data/utils.py
```diff
@@ -20,6 +20,55 @@
...
+                'input_ids': batch['input_ids'][row].unsqueeze(0),
+            }
+            if 'attention_mask' in batch:
+                row_batch['attention_mask'] = batch['attention_mask'][
+                    row].unsqueeze(0)
+            if 'labels' in batch:
+                row_batch['labels'] = batch['labels'][row].unsqueeze(0)
+            if 'decoder_attention_mask' in batch:
+                row_batch['decoder_attention_mask'] = batch[
+                    'decoder_attention_mask'][row].unsqueeze(0)
```
mvpatel2000, line 54: can we instead modify token_counting_func? It feels awkward to have a side-effect on other attributes here when it should just be appending new data

llmfoundry/data/utils.py
```diff
@@ -20,6 +20,55 @@
 log = logging.getLogger(__name__)
 
 
+class LossGeneratingTokensCollatorWrapper:
+    """Collator wrapper to add loss generating token counts to batch."""
+
+    def __init__(
+        self,
+        base_collator: Callable,
+        token_counting_func: Callable[[Batch], Union[int, dict[str, int]]],
```
mvpatel2000, line 29: Do we need to pass the counting fn as an arg? given there's only one option and no way to pipe it through? Why not just call fn directly

mvpatel2000 (approved):

PR #1639, opened 2024-11-05
Fix dataset download location

dakinggg:
Fixes an error in the previous PR by making sure all ranks use the same dataset download location to avoid contention between the ranks operating on the same hugging face dataset.

Example before: `ds-go-away-sl-1-7-egU4ZP`
Example after: `ds-go-away-sl-2-8-p1nLOh`

### DIFF
--- a/llmfoundry/data/finetuning/tasks.py
+++ b/llmfoundry/data/finetuning/tasks.py
@@ -93,6 +93,7 @@ from llmfoundry.utils.exceptions import (
     UnknownExampleTypeError,
 )
 #  yapf: enable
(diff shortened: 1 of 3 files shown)


j316chuck (approved):
LGTM, can you post a test run? 

Also is this syntax correct? 
https://github.com/mosaicml/llm-foundry/blob/main/llmfoundry/utils/file_utils.py#L21-L22 

Reading the docs it seems the API is a bit different: 
https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather_object

PR #1641, opened 2024-11-05
Put extra arg test back

XiaohanZhangCMU:
With streaming upgraded to 0.9.1, the unit test would lead to inf loop.


tests/data/test_dataloader.py
```diff
@@ -1549,15 +1550,23 @@ def test_text_dataloader_with_extra_keys():
 
     device_batch_size = 2
 
-    mock_stat = MagicMock()
-    mock_stat.st_size = 1024  # Mock st_size with a desired value
-    mock_stat.st_mode = 33188  # Regular file mode for Unix-based systems
+    def custom_stat_mock(path: Any):
```
dakinggg, line 1553: this is some really intense patching, are you sure this test can't be made simpler?

XiaohanZhangCMU: Hmm, should be, but I was not able to come up with one. Any suggestions? I basically need 
```os.path.exists(filelocks)``` to behave normally, and other ```os.stat``` to return a mock stat when testing with private-streaming.

PR #1643, opened 2024-11-06
Changes for removing unused terms in CE loss fn 

gupta-abhay:
We need to deprecate this in favor of the new changes to account for the correct loss calc. based on tokens in [this PR](https://github.com/mosaicml/llm-foundry/pull/1632)

### DIFF
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -79,6 +79,8 @@ from llmfoundry.models.utils.param_init_fns import generic_param_init_fn_  # typ
 from llmfoundry.models.layers.norm import LPLayerNorm  # type: ignore
 # isort: on
 
+from llmfoundry.utils.warnings import VersionedDeprecationWarning
+
 log = logging.getLogger(__name__)
 
 CROSS_ENTROPY_IGNORE_INDEX = -100
@@ -1346,6 +1348,7 @@ def compute_loss_from_logits(
     shift_labels: bool,
     labels: torch.Tensor,
     loss_fn: nn.Module,
+    sample_weighing_factor: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     targets = get_targets(labels) if shift_labels else labels
 
@@ -1358,6 +1361,13 @@ def compute_loss_from_logits(
         loss = losses.sum()
     else:
         loss = losses.sum() / (targets != loss_fn.ignore_index).sum()
+        if sample_weighing_factor is not None:
+            warnings.warn(
+                VersionedDeprecationWarning(
+                    message='sample_weighing_factor has been deprecated!',
+                    remove_version='0.15.0',
+                ),
+            )
 
     return loss
 
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -1346,7 +1346,6 @@ def compute_loss_from_logits(
     shift_labels: bool,
     labels: torch.Tensor,
     loss_fn: nn.Module,
-    sample_weighing_factor: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     targets = get_targets(labels) if shift_labels else labels
 
@@ -1359,12 +1358,6 @@ def compute_loss_from_logits(
         loss = losses.sum()
     else:
         loss = losses.sum() / (targets != loss_fn.ignore_index).sum()
-        if sample_weighing_factor is not None:
-            if sample_weighing_factor.shape[0] > 1:
-                raise ValueError(
-                    'Sample weighing factor is not supported when batch["sample_weighing_factor"].shape[0] > 1.',
-                )
-            loss = loss * sample_weighing_factor[0].item()
 
     return loss
 
@@ -1473,7 +1466,6 @@ class ComposerMPTCausalLM(HuggingFaceModel):
             self.shift_labels,
             batch['labels'],
             self.loss_fn,
-            batch.get('sample_weighing_factor', None),
         )
 
         if self.config.ffn_config['ffn_type'] in ffns_with_megablocks:
--- a/llmfoundry/models/mpt/modeling_mpt.py
+++ b/llmfoundry/models/mpt/modeling_mpt.py
@@ -1365,9 +1365,14 @@ def compute_loss_from_logits(
             warnings.warn(
                 VersionedDeprecationWarning(
                     message='sample_weighing_factor has been deprecated!',
-                    remove_version='0.15.0',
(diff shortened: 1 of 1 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1360,11 +1362,12 @@ def compute_loss_from_logits(
...
         loss = losses.sum() / (targets != loss_fn.ignore_index).sum()
         if sample_weighing_factor is not None:
-            if sample_weighing_factor.shape[0] > 1:
-                raise ValueError(
-                    'Sample weighing factor is not supported when batch["sample_weighing_factor"].shape[0] > 1.',
-                )
-            loss = loss * sample_weighing_factor[0].item()
+            warnings.warn(
+                VersionedDeprecationWarning(
+                    message='sample_weighing_factor has been deprecated!',
```
dakinggg, line 1367: this should keep the old functionality, and the version for removal should be current version + 2

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1360,11 +1362,12 @@ def compute_loss_from_logits(
     else:
         loss = losses.sum() / (targets != loss_fn.ignore_index).sum()
         if sample_weighing_factor is not None:
-            if sample_weighing_factor.shape[0] > 1:
-                raise ValueError(
-                    'Sample weighing factor is not supported when batch["sample_weighing_factor"].shape[0] > 1.',
-                )
-            loss = loss * sample_weighing_factor[0].item()
+            warnings.warn(
+                VersionedDeprecationWarning(
```
dakinggg, line 1366: should also store an attr so this warning only fires once

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1473,7 +1481,6 @@ def loss(self, outputs: CausalLMOutputWithPast,
             self.shift_labels,
             batch['labels'],
             self.loss_fn,
-            batch.get('sample_weighing_factor', None),
```
dakinggg, line 1476: should add this back too?

gupta-abhay: thanks for catching - done!

gupta-abhay:
Run logs:

```bash
WARNING:runtime_private_plugins.utils.config_utils:Sequence parallelism is only supported for accumulating the batch on tokens. Setting accumulate_train_batch_on_tokens to True.             
DEBUG: llmfoundry.command_utils.train: Initializing dist with device...                                                                    
DEBUG: llmfoundry.command_utils.train: Testing barrier with device...                                                                      
DEBUG: llmfoundry.command_utils.train: Barrier test passed with device.                                                                    
INFO: llmfoundry.command_utils.train: Building tokenizer...                                                                                
INFO: llmfoundry.command_utils.train: Building train loader...                                                                             
INFO: streaming.base.dataset: Because `predownload` was not specified, it will default to 8*batch_size if batch_size is not None, otherwise
 64.                                                                                                                                                                                          
INFO: llmfoundry.command_utils.train: Building eval loader...                                                                              
INFO: llmfoundry.command_utils.train: Initializing model...
DEBUG: llmfoundry.models.mpt.modeling_mpt: Using kaiming_normal_ initialization.                          
INFO: llmfoundry.command_utils.train: Building trainer...    
INFO: composer.utils.reproducibility: Setting seed to 24                                                                                   
INFO: composer.trainer.trainer: Run name: interactive-j1KMfR                                                                               
INFO: composer.core.state: Automatically setting data_parallel_shard to have parallelization degree 8.
/usr/lib/python3/dist-packages/composer/trainer/trainer.py:1630: UserWarning: Specifying `eval_interval=500ba` without an `eval_dataloader` has no effect. If trying to run an evaluator, make
 sure `eval_dataloader` is specified. Otherwise, set `eval_interval` to 0 or default value 1.                                                                                                 
  warnings.warn(                                                                               
INFO: composer.trainer.trainer: Stepping schedulers every batch. To step schedulers every epoch, set `step_schedulers_every_batch=False`.  
/usr/lib/python3/dist-packages/torch/storage.py:414: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module imp
licitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models 
for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects 
will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_on
ly=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.                           
  return torch.load(io.BytesIO(b))                                                                                                                                                            
INFO: composer.utils.reproducibility: Setting seed to 17                                                                                   
DEBUG: composer.utils.reproducibility: Restoring the RNG state                                                                             
DEBUG: composer.loggers.mosaicml_logger: Logging model initialized time to metadata                                                        
INFO: composer.trainer.trainer: Setting seed to 24                                                                                         
INFO: composer.utils.reproducibility: Setting seed to 24                                                                                   
INFO: llmfoundry.command_utils.train: Logging config                                                                                       
INFO: llmfoundry.command_utils.train: Starting training...                                                                                 
INFO: composer.trainer.trainer: Using precision Precision.AMP_BF16                                                                         
DEBUG: composer.trainer.trainer: Spinning the dataloaders                                                                                  
DEBUG: composer.trainer.trainer: Starting training loop                                                                                    
INFO: streaming.base.dataset: Because `num_canonical_nodes` was not specified, and `shuffle_algo` is py1e, it will default to be equal to p
hysical nodes.                                                                                                                                                                                
INFO: streaming.base.dataset: Because `shuffle_block_size` was not specified, it will default to max(4_000_000 // num_canonical_nodes, 1 <<
 18) if num_canonical_nodes is not None, otherwise 262144.
```

Do not see any logs for deprecation. cc: @dakinggg (Are we good on this?)

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1360,6 +1362,12 @@ def compute_loss_from_logits(
     else:
         loss = losses.sum() / (targets != loss_fn.ignore_index).sum()
         if sample_weighing_factor is not None:
+            warnings.warn(
+                VersionedDeprecationWarning(
+                    message='sample_weighing_factor has been deprecated!',
+                    remove_version='0.16.0',
```
dakinggg, line 1368:
```suggestion
                    remove_version='0.17.0',
```

dakinggg (approved):

PR #1644, opened 2024-11-06
Catch bad data prep

milocress:
When privatelink blocks a delta table from being downloaded, this fails silently in the dataset preparation script.

This PR adds a validation step to check that the downloaded file indeed exists and contains data to prevent this kind of silent error.

### DIFF
--- a/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
+++ b/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
@@ -1,6 +1,7 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
+import json
 import logging
 import os
 import re
@@ -26,6 +27,8 @@ from llmfoundry.utils.exceptions import (
     FailedToCreateSQLConnectionError,
     FaultyDataPrepCluster,
     InsufficientPermissionsError,
+    MisconfiguredHfDatasetError,
+    StoragePermissionError,
     UCNotEnabledError,
 )
 
@@ -681,7 +684,7 @@ def fetch_DT(
 
     log.info(f'Directory {json_output_folder} created.')
 
-    # validate_and_get_cluster_info allows cluster_id to be None if use_serverless is True
+    # Validate_and_get_cluster_info allows cluster_id to be None if use_serverless is True.
     method, dbsql, sparkSession = validate_and_get_cluster_info(
         cluster_id=cluster_id,
         databricks_host=DATABRICKS_HOST,
@@ -732,11 +735,40 @@ def fetch_DT(
     if dbsql is not None:
         dbsql.close()
 
-    # combine downloaded jsonl into one big jsonl for IFT
+    # Combine downloaded jsonl into one big jsonl for IFT.
     iterative_combine_jsons(
         json_output_folder,
         os.path.join(json_output_folder, json_output_filename),
     )
+
+    _validate_written_file(
+        json_output_folder,
+        json_output_filename,
+        delta_table_name,
+    )
+
+
+def _validate_written_file(
+    json_output_folder: str,
+    json_output_filename: str,
+    delta_table_name: str,
+):
+    # Validate downloaded dataset is actually downloaded.
+    with open(os.path.join(json_output_folder, json_output_filename)) as f:
+        is_empty = True
+        for line in f.readlines():
+            is_empty = False
+            try:
+                json.loads(line)
+            except:
+                raise MisconfiguredHfDatasetError(
+                    delta_table_name,
+                    split=json_output_folder,
+                ) from ValueError('line')
+        if is_empty:
+            raise StoragePermissionError(
+                f'Unable to download {delta_table_name}, check network permissions.',
+            )
 
 
 def _check_imports():
--- a/tests/a_scripts/data_prep/test_convert_delta_to_json.py
+++ b/tests/a_scripts/data_prep/test_convert_delta_to_json.py
@@ -1,9 +1,11 @@
 # Copyright 2022 MosaicML LLM Foundry authors
 # SPDX-License-Identifier: Apache-2.0
 
+import os
 import sys
 import unittest
 from argparse import Namespace
+from tempfile import NamedTemporaryFile
 from typing import Any
 from unittest.mock import MagicMock, mock_open, patch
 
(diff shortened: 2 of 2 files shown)


llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -732,12 +735,41 @@ def fetch_DT(
...
+        is_empty = True
+        for line in f.readlines():
+            is_empty = False
+            try:
+                json.loads(line)
+            except:
+                raise MisconfiguredHfDatasetError(
+                    delta_table_name,
+                    split=json_output_folder,
+                ) from ValueError('line')
```
dakinggg, line 767: whats the `from ValueError('line')` here?

milocress: oops fixed, meant to raise from `e` to give the trace more info in the case of a truly malformed dataset

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -732,12 +735,41 @@ def fetch_DT(
...
+    with open(os.path.join(json_output_folder, json_output_filename)) as f:
+        is_empty = True
+        for line in f.readlines():
+            is_empty = False
+            try:
+                json.loads(line)
+            except e:
+                raise MisconfiguredHfDatasetError(
+                    delta_table_name,
+                    split=json_output_folder,
```
dakinggg, line 766: this isn't really the split right? should we either make the split arg optional or put it as none here?

tests/a_scripts/data_prep/test_convert_delta_to_json.py
```diff
@@ -102,17 +108,20 @@ def test_stream_delta_to_json(
...
+                json_output_folder=json_output_folder,
+                http_path=http_path,
+                cluster_id=cluster_id,
+                DATABRICKS_HOST=DATABRICKS_HOST,
+                DATABRICKS_TOKEN=DATABRICKS_TOKEN,
+                use_serverless=use_serverless,
+                batch_size=batch_size,
+                json_output_filename=json_output_filename,
+            )
+        except FileNotFoundError:
```
irenedea, line 123: why do we not raise if there's a FileNotFoundError?

milocress: `fetch_DT` with the mocked functions in this test harness does not write a `json` file, but the check I added throws an error when no `json` file is written. I add this `try/except` block to ignore the `FileNotFoundError` in the mocked case.

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -732,12 +735,38 @@ def fetch_DT(
     if dbsql is not None:
         dbsql.close()
 
-    # combine downloaded jsonl into one big jsonl for IFT
+    # Combine downloaded jsonl into one big jsonl for IFT.
     iterative_combine_jsons(
```
nancyhung, line 739: wouldn't the output filename be empty before the iterative combine jsons command? Why do the validation after?

nancyhung (commented):
I wonder why `fetch` is writing empty rows. Do we why this fails silently? @milocress

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
milocress: I think `iterative_combine_jsons` writes to the output filename if the source dataset is across multiple files.

milocress: (Meaning we can't do the validation earlier because the file at the path may not exist)

milocress:
> I wonder why `fetch` is writing empty rows. Do we why this fails silently? @milocress

https://github.com/mosaicml/llm-foundry/blob/9b22b6de4f67c3e4129cf2dd6b1435f926bbbc7b/llmfoundry/command_utils/data_prep/convert_delta_to_json.py#L485C1-L499C18

I believe there is a situation where `cursor` is `None` and no exception is thrown. This PR adds a more general check to make sure the file is written, which catches a wider variety of silent errors that could happen in the stack.

tests/a_scripts/data_prep/test_convert_delta_to_json.py
nancyhung: nit: out of context - instead of catching the `FileNotFoundError`, shouldn't we update the mock to write to a temp path that we can then check?

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -732,12 +735,38 @@ def fetch_DT(
...
+    delta_table_name: str,
+):
+    # Validate downloaded dataset is actually downloaded.
+    with open(os.path.join(json_output_folder, json_output_filename)) as f:
+        is_empty = True
+        for line in f.readlines():
+            is_empty = False
+            try:
+                json.loads(line)
+            except Exception as e:
```
irenedea, line 763: let's log the json line that fails here

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -732,12 +735,38 @@ def fetch_DT(
...
+                raise MisconfiguredHfDatasetError(delta_table_name) from e
```
irenedea, line 764: I don't think we should actually raise a user error here right? Because we own the delta to jsonl conversion, if this fails it should be our fault.

irenedea: we can just let whatever exception happens go through

tests/a_scripts/data_prep/test_convert_delta_to_json.py
irenedea: ah yeah I agree with Nancy here, we should update the mock to write to a temp path

irenedea (commented):
some minor comments

tests/a_scripts/data_prep/test_convert_delta_to_json.py
```diff
@@ -25,9 +31,41 @@
...
+        with open(filename, 'w') as f:
+            f.write(json.dumps({'prompt': 'prompt', 'response': 'response'}))
+
+        assert os.path.exists(filename)
+
+    return _inner
+
+
+@contextmanager
+def UncreatedNamedTemporaryFile(suffix: str):
```
irenedea, line 53: wait why do you need this

irenedea (approved):

PR #1657, opened 2024-11-14
Retry catching BlockingIOError

KuuCi:
Transiently hitting this error (pretty rare):
`BlockingIOError: [Errno 11] Resource temporarily unavailable: '*****/embedding_data/train/shard.00000.mds'`

Adding a retry mechanism here because BlockingIOError means either a running out of resources issue or a linux I/O issue, so no risks for retrying


mvpatel2000 (approved):

jacobfulano (commented):
Thanks for adding this

PR #1679, opened 2024-11-30
Bugfix auto packing with streams + no remote path

mattyding:
## This PR

Edge case where auto packing fails for a streaming dataset specified through `streams` field if only a local path and no remote path is provided. This configuration is more broadly supported for Foundry.
```
  train_loader:
    name: finetuning
    dataset:
      streams:
        finetuning:
          local: /foo/bar/train/mds/
```
An error is thrown in StreamingDataset init because there is a [check](https://github.com/mosaicml/streaming/blob/f4832d41823c99fe9f80d6f0101fbc62041599e3/streaming/base/stream.py#L451-L454) that `local` contains an index file. The current auto packing script overwrites `local` with an empty tempdir. This fix temporarily has `remote` point to the user-provided `local`, which should contain an index file. 

## Testing
- MCLI run with streams + no remote path
- Added unit test

### DIFF
--- a/llmfoundry/data/packing.py
+++ b/llmfoundry/data/packing.py
@@ -485,6 +485,8 @@ def profile_packing(
             tmp_path_to_broadcast = tempfile.TemporaryDirectory().name
             gathered_paths = dist.all_gather_object(tmp_path_to_broadcast)
             tmp_path = gathered_paths[local_rank_zero]
+            if stream_config.get('remote') is None:
+                stream_config['remote'] = stream_config['local']
             stream_config['local'] = tmp_path
 
     # Determine the packing_ratio values we'll try
--- a/llmfoundry/data/packing.py
+++ b/llmfoundry/data/packing.py
@@ -474,7 +474,8 @@ def profile_packing(
 
     # If streaming dataset, use a temporary local folder for profiling
     local_rank_zero = dist.get_global_rank() - dist.get_local_rank()
-    if dataset_cfg.get('remote') is not None:
+    if dataset_cfg.get('remote'
+                      ) is not None and dataset_cfg.get('local') is None:
         tmp_path_to_broadcast = tempfile.TemporaryDirectory().name
         gathered_paths = dist.all_gather_object(tmp_path_to_broadcast)
         tmp_path = gathered_paths[local_rank_zero]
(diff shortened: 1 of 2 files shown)


tests/data/test_packing.py
```diff
@@ -194,28 +193,57 @@ def test_auto_packing_with_streaming_dataloader(tmp_path: Path):
...
+                'stream_without_remote': {
+                    'local': remote_dir,
+                },
+            },
+        },
+    })
 
-    state_dict = loader.dataset.state_dict(num_samples=2, from_beginning=False)
-    assert state_dict['sample_in_epoch'] == 2 * loader.dataset.packing_ratio
+    for cfg in [remote_cfg, streams_cfg]:
```
dakinggg, line 223: rather than adding a for loop to the test, please use `@pytest.mark.parametrize`. That way each test can fail individually instead of only all together

llmfoundry/data/packing.py
```diff
@@ -485,6 +485,8 @@ def profile_packing(
             tmp_path_to_broadcast = tempfile.TemporaryDirectory().name
             gathered_paths = dist.all_gather_object(tmp_path_to_broadcast)
             tmp_path = gathered_paths[local_rank_zero]
+            if stream_config.get('remote') is None:
```
dakinggg, line 488: I might be missing something, but I don't think this is the right fix. Shouldn't the fix be to just not overwrite the local dir if it is provided and already exists? (and in the above if statement too)

mattyding: That works too. This change also works because remote gets copied into local.

I can't repro this bug with remote unset and local set + autopacking (the above if statement)

dakinggg: oh thats odd that it only exists when using streams...well anyway, they should probably both do the same thing, and that should probably be to not overwrite local if its specified?

mattyding: Sure, my hesitation is because I don't know the original motivation for using tmp paths, but if you think this is a safe change then 👍

mattyding:
Re-tested that it trains E2E with the following cases:
- streams, local no remote
- no streams, local == remote
- no streams, remote only
- no streams, local only

dakinggg (approved):
CI failure looks unrelated, will have to figure that out separately

PR #1680, opened 2024-11-30
Add pre-tokenized Delta to MDS conversion script

mattyding:
## Testing
MCLI IFT and CPT runs trained successfully.


llmfoundry/command_utils/data_prep/convert_delta_to_mds.py
```diff
@@ -0,0 +1,167 @@
...
+        logging.info('Identified IFT data')
+        dtypes = {
+            'input_ids': 'ndarray',
+            'attention_mask': 'ndarray',
+            'labels': 'ndarray',
+        }
+        convert_x = lambda x: {
+            # join the turns into a single array
+            'input_ids':
+                np.concatenate([
```
mattyding, line 55: Q: Is this correct / how to convert multi-turn chat to MDS?

dakinggg: You should try to factor and reuse code from here if possible: https://github.com/mosaicml/llm-foundry/blob/41e02d778c22185dbcd3fd4fedd64ff264883ee9/llmfoundry/command_utils/data_prep/convert_finetuning_dataset.py#L173-L267

mattyding: The tokenized IFT/CHAT output is the output of the HF tokenizer (input_ids, attention_mask, labels). Just checked, multi-turn chat gets concatenated / formatted into a single turn. Assuming single-turn input for this function and removing multi-turn concat

mattyding: Took a look at the code you linked. IMO, not easy to re-factor + re-use because most of the logic for formatting examples, which is assumed to happen before data is written to Delta table. Like mentioned in the previous comment, it looks like multi-turn CHAT data gets concatenated into a single turns array (similar to the linked logic). The last remaining step is writing as MDS format, which is what this PR adds.

llmfoundry/command_utils/data_prep/convert_delta_to_mds.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+
+    if 'turns' in columns[0]:
+        logging.info('Identified IFT/CHAT data')
+        dtypes = {
+            'input_ids': 'ndarray',
+            'attention_mask': 'ndarray',
+            'labels': 'ndarray',
+        }
+        convert_x = lambda x: (
```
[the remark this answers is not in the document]
mattyding: Forge reduces/joins multi-turn CHAT into a single turn. I thought this was intended but after looking at the code, I don't think it is / PySpark might coerce the data into that format 🙃 . Will do some more debugging.

IFT regression tests might be necessary after all.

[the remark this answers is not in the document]
dakinggg: just following up that we confirmed multi turn is preserved, and so this script should support it too

mattyding:
Last few commits address comments, add support for multi-turn CHAT data. Provided testing plan + YAMLs in Slack thread. Looking for guidance on to test with PEFT enabled

llmfoundry/command_utils/data_prep/convert_delta_to_mds.py
```diff
@@ -0,0 +1,163 @@
...
+            columns=dtypes,
+            compression=compression,
+            hashes=hashes,
+            size_limit=limit,
+        ) as out:
+            try:
+                with open(json_full_filepath, 'r') as f:
+                    for line in f:
+                        turns = convert_x(json.loads(line))
+                        for turn in turns:
```
dakinggg, line 157: I don't think this is right? see https://github.com/mosaicml/llm-foundry/blob/1e997df6854de88a48252ee2872e48dfa6f55b40/llmfoundry/command_utils/data_prep/convert_finetuning_dataset.py#L244-L250 for another example of writing turns data to MDS

PR #1683, opened 2024-12-02
Add ability to log model instead of saving to folder

irenedea:


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -289,7 +291,11 @@ def __init__(
         final_register_only: bool = False,
         register_wait_seconds: int = 7200,
     ):
-        _, _, self.save_dir_format_str = parse_uri(save_folder)
+        if save_folder is None and mlflow_registered_model_name is None:
+            raise ValueError(
+                'No `save_folder` or `mlflow_register_model_name` set. Please set at least one of them.',
```
dakinggg, line 296:
```suggestion
                'No `save_folder` or `mlflow_registered_model_name` set. Please set at least one of them.',
```

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -278,8 +280,8 @@ class HuggingFaceCheckpointer(Callback):
 
     def __init__(
         self,
-        save_folder: str,
         save_interval: Union[str, int, Time],
+        save_folder: Optional[str] = None,
```
dakinggg, line 284: its not good to change arg order (technically backwards breaking change if positional args are allowed). would prefer to just make it `Optional` but without a default and keep the same order

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -389,12 +403,16 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
             self._save_checkpoint(
                 state,
                 logger,
-                register_to_mlflow=(
-                    self.mlflow_registered_model_name is not None and
-                    is_last_batch
+                log_to_mlflow=(
+                    self.save_folder is None or (
```
dakinggg, line 407: we probably need some check somewhere that if save_folder is None you actually have an mlflow logger to use?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -457,15 +475,24 @@ def run_event(self, event: Event, state: State, logger: Logger) -> None:
...
-                    register_to_mlflow=False,
-                )
+                if self.save_folder:
+                    log.error(
+                        'An error occurred in one or more registration processes. Fallback to saving the HuggingFace checkpoint.',
+                    )
+                    self._save_checkpoint(
+                        state,
+                        logger,
+                        upload_to_save_folder=True,
```
dakinggg, line 485: don't we want the fallback to be to log to mlflow without registering if save_folder is none?

llmfoundry/callbacks/hf_checkpointer.py
irenedea: We do a check on init that  errors if save_folder is None and mlflow_registered_model_name is None, so this should always be valid.

irenedea: actually I guess maybe we shouldn't require users to input a `mlflow_registered_model_name` for logging only

llmfoundry/callbacks/hf_checkpointer.py
irenedea: Hm we could do that. I was thinking to avoid the log_model path completely, but that could work

irenedea: Actually, assuming that logging happens prior to registration:

If the run errors during registration, then the model was probably already logged. If it errors during logging, we probably don't want to do it again.

irenedea: resolved offline, will keep as is bc the model will already be logged if there are registration failures

llmfoundry/callbacks/hf_checkpointer.py
irenedea: updated

dakinggg:
gpu tests failed, will wait to review until CI passes

PR #1687, opened 2024-12-05
Bump Composer to v0.28.0

snarayan21:
As title


dakinggg (approved):
If CI passes, good to go

PR #1688, opened 2024-12-05
Fix llama3 example yamls

j316chuck:
Added some flags to fix llama3 yaml examples

# Tests:
- Before nccl timeout: `llama3-1-70b-finetune-gjq2H5` 🔴 
- After: `llama3-1-70b-finetune-xS6h2K` ✅


snarayan21 (approved):
lgtm

PR #1690, opened 2024-12-06
Refactor HF checkpointer

milocress:
Refactors the HF checkpointer to be more general.

Manual tests:

- `refactor-hf-checkpointer-peft-4xfWFZ`
- `refactor-hf-checkpointer-no-peft-nMFI4o`


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -201,9 +204,18 @@ def save_model_patch(*args: Any, **kwargs: Any):
     if mlflow_logger.model_registry_uri is not None:
         mlflow.set_registry_uri(mlflow_logger.model_registry_uri)
 
+    if is_peft:
+        transformers_in_memory_model = pipeline(
```
dakinggg, line 208: won't this not be a peft model?

dakinggg: shouldn't you just pass the in memory model into this function directly for the peft case?

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -201,9 +204,18 @@ def save_model_patch(*args: Any, **kwargs: Any):
...
+                transformers_model,
+            )
+        )
+    else:
+        transformers_in_memory_model = None
+
     register_model_path = f'{mlflow_logger.model_registry_prefix}.{registered_model_name}' if mlflow_logger.model_registry_prefix and registered_model_name else registered_model_name
     mlflow_logger.log_model(
-        transformers_model=transformers_model,
+        transformers_model=transformers_in_memory_model or transformers_model,
```
dakinggg, line 218: pretty sure you should just be passing the model and tokenizer in here directly as
{'model': model, 'tokenizer': tokenizer}, since mlflow doesn't support local paths for peft models.

llmfoundry/callbacks/hf_checkpointer.py
milocress: For some reason mlflow requires either a Pipeline or a path to the model.

llmfoundry/callbacks/hf_checkpointer.py
milocress: Ah I'll try this.

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -201,9 +204,17 @@ def save_model_patch(*args: Any, **kwargs: Any):
     if mlflow_logger.model_registry_uri is not None:
         mlflow.set_registry_uri(mlflow_logger.model_registry_uri)
 
+    if is_peft:
+        transformers_in_memory_model = {
+            'model': AutoModel.from_pretrained(transformers_model),
```
dakinggg, line 209: why not just pass the model in directly to this function since its already in memory?

milocress: Good catch, fixed

dakinggg (commented):
The code looks right, but the test runs you linked I think don't use it? They appear to be an image with foundry main installed, without specifying a commit/branch to install over that. In particular, the peft checkpoint in the linked run doesn't have a "final_model_checkpoint" in the artifacts, as the non peft run does. I'd expect them to have the same file structure after this pr.

PR #1691, opened 2024-12-08
Catch bad split regex

milocress:
Datasets throws this error:

https://github.com/huggingface/datasets/blob/661d7bac29689e2d9eb74fba3d243939d6e9f25b/src/datasets/splits.py#L362

when a split doesn't match the regex. This we catch and throw to the user.


dakinggg (commented):
I think this error doesn't need to be try/catch, shouldn't we just replace bad characters in the split or something when we write the file to disk?

PR #1696, opened 2024-12-09
fix retry crash loop

milocress:
When a run is killed and automatically restarts after training is complete, it is thrown into a retry loop due to Composer erroring when no training will happen. This PR short-circuits to the checkpoint upload step when this error is detected during a retry. 

Manual tests: 
- `milo-gvMjEA` (initial)
- `milo-vsiPgi` (retry)


llmfoundry/command_utils/train.py
```diff
@@ -640,7 +640,37 @@ def train(cfg: DictConfig) -> Trainer:
...
-    trainer.fit()
+    try:
+        trainer.fit()
+    except ValueError as e:
+        msg = str(e)
+        if 'The max_duration' in msg and 'is less than or equal to the elapsed training duration' in msg and train_cfg.run_is_retry:
+            log.info(
+                'Training is already complete and detected retry. Skipping training and saving checkpoint.',
+            )
+            trainer.save_checkpoint_to_save_folder()
```
dakinggg, line 651: i believe you don't want to call the composer save, just the hf save. because if the you get this error that means the last composer checkpoint was already saved successfully.

dakinggg (commented):
manual test run?

PR #1697, opened 2024-12-12
Catch multiple slashes in source dataset into one slash

KuuCi:
As title suggests, trying to catch wrong input for UC where user accidentally adds multiple slashes

Test:
On main:
`test-uc-mlflow-IEv1zp`
image

On branch:
`test-uc-mlflow-8yeqFa`
image

Trains properly now with s broken dummy path: 
image

### DIFF
--- a/llmfoundry/utils/config_utils.py
+++ b/llmfoundry/utils/config_utils.py
@@ -6,9 +6,9 @@ import copy
 import logging
 import math
 import os
-import re
 import warnings
 from dataclasses import dataclass, fields
+from pathlib import Path
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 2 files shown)


jimmyxu-db (approved):

dakinggg (commented):
don't think you need to write your own regex

```
In [1]: from pathlib import Path

In [2]: str(Path('dbfs:/Volumes/blah/blah//blah.jsonl'))
Out[2]: 'dbfs:/Volumes/blah/blah/blah.jsonl'
```

KuuCi:
retested:
image

dakinggg (approved):

PR #1700, opened 2024-12-16
Adding preprocessors for QA and messages datasets

ShashankMosaicML:


gupta-abhay (approved):
lgtm

dakinggg (approved):
Could you add a quick unit test for each of the functions you added please? Approving preemptively because code lgtm, but please do that. don't need to test e2e, just unit test for the functions you're adding (input some dict, outputs the right dict)

ShashankMosaicML:
> Could you add a quick unit test for each of the functions you added please? Approving preemptively because code lgtm, but please do that. don't need to test e2e, just unit test for the functions you're adding (input some dict, outputs the right dict)

Done!

PR #1701, opened 2024-12-16
Make loaded peft adapters optionally trainable

snarayan21:
Previously, loaded peft adapters (from HF) were not trainable. This PR adds the option to continue training peft adapters that are initialized from some existing adapter HF checkpoint.

Added unit tests to make sure adapter parameters are trainable when needed. this requires setting `pretrained=True` since `pretrained_lora_id_or_path` is only used in that case.

### DIFF
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -72,7 +72,6 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         additional_train_metrics: Optional[list] = None,
         additional_eval_metrics: Optional[list] = None,
         should_save_peft_only: bool = True,
-        peft_is_trainable: bool = False,
     ):
         config_overrides = config_overrides or {}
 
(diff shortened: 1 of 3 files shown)


llmfoundry/models/hf/hf_base.py
```diff
@@ -72,6 +72,7 @@ def __init__(
         additional_train_metrics: Optional[list] = None,
         additional_eval_metrics: Optional[list] = None,
         should_save_peft_only: bool = True,
+        peft_is_trainable: bool = False,
```
dakinggg, line 75: do we need a config flag for this? shouldn't they just always be trainable?

snarayan21: hm...yeah that's true. I put this flag there because I didn't want to change the default behavior, but the default behavior is to have `is_trainable=False` which we don't want in foundry anyways.

Modifying to reflect this

dakinggg (commented):
do you have any sense why the default is `False`? mostly just wondering if i'm missing something, cause I'm not sure why they would have the default as fals.

snarayan21:
@dakinggg i think just because they assume that a saved adapter should be used directly for inference by default:
https://huggingface.co/docs/peft/en/package_reference/peft_model#peft.PeftModel.from_pretrained.is_trainable

PR #1710, opened 2025-01-13
Support for reusing the input to W_k and W_v

ShashankMosaicML:


llmfoundry/models/layers/attention.py
```diff
@@ -626,12 +637,15 @@ def forward(
...
                                              torch.Tensor]] = None,
         key_value_states: Optional[torch.Tensor] = None,
+        x_prev: Optional[torch.Tensor] = None,
         pos_id_within_seq: Optional[torch.Tensor] = None,
     ) -> tuple[torch.Tensor, Optional[torch.Tensor], Optional[tuple[
         torch.Tensor, torch.Tensor]]]:
         extra_kwargs = {}
         if prev_layer_key_value is not None:
             extra_kwargs['prev_layer_key_value'] = prev_layer_key_value
+        if x_prev is not None:
```
dakinggg, line 647: for this, and the line above, you don't need the extra kwargs dict right? you can just pass them in normally, if they are None, they will just be passed in as None, which is the default.

llmfoundry/models/layers/blocks.py
```diff
@@ -179,6 +180,8 @@ def forward(
             extra_kwargs['pos_id_within_seq'] = pos_id_within_seq
 
         if self.fuse_norm_attn_norm:
+            if x_prev is not None:
```
dakinggg, line 183: same comment on extra_kwargs

llmfoundry/models/layers/blocks.py
```diff
@@ -193,6 +196,8 @@ def forward(
             )
         else:
             a = self.norm_1(x)
+            if x_prev is not None:
```
dakinggg, line 199: same

llmfoundry/models/layers/blocks.py
```diff
@@ -342,13 +347,16 @@ def forward(
...
+        x_prev: Optional[torch.Tensor] = None,
         pos_id_within_seq: Optional[torch.Tensor] = None,
     ) -> tuple[torch.Tensor, torch.Tensor, Optional[torch.Tensor],
                Optional[tuple[torch.Tensor, torch.Tensor]]]:
         a = self.norm_1(x)
         extra_kwargs = {}
         if prev_layer_key_value is not None:
             extra_kwargs['prev_layer_key_value'] = prev_layer_key_value
+        if x_prev is not None:
+            extra_kwargs['x_prev'] = self.norm_1(x_prev)
```
dakinggg, line 359: same

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1024,6 +1049,8 @@ def forward(
             extra_kwargs = {}
             if prev_layer_key_value is not None:
                 extra_kwargs['prev_layer_key_value'] = prev_layer_key_value
+            if x_prev is not None:
```
dakinggg, line 1052: same

dakinggg (dismissed):

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: Fixed.

llmfoundry/models/layers/blocks.py
ShashankMosaicML: Fixed.

llmfoundry/models/layers/blocks.py
ShashankMosaicML: Fixed.

llmfoundry/models/layers/attention.py
ShashankMosaicML: Fixed.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1036,11 +1059,15 @@ def forward(
                 output_attentions=bool(output_attentions),
                 alibi_slopes=alibi_slopes,
                 flash_attn_padding_info=flash_attn_padding_info,
+                return_x_post_norm=True,
```
dakinggg, line 1062: pretty sure we wouldn't want this to be hardcoded True, and rather only be True if you actually need the tensor for saving

ShashankMosaicML: Fixed.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -633,46 +652,39 @@ def _get_override_block_args_list(
...
         b_idx: int,
         override_config: dict[str, Any],
-        reuse_kv_layer_idx_dict: dict[int, int],
+        reuse_state_layer_idx_dict: dict[int, int],
+        reuse_type: str,
     ) -> int:
         override_attn_config = override_config['attn_config']
-        if override_attn_config['reuse_kv_layer_idx'] >= 0:
-            reuse_kv_layer_idx = override_attn_config['reuse_kv_layer_idx']
+        if override_attn_config[reuse_type] >= 0:
```
dakinggg, line 664: would it make things simpler in the yaml if you allowed positive (but not 0) values for this in addition to negative? that way for a pivot layer you can just specify the conf one time, rather than a separate negative value for each layer after the pivot.

dakinggg (approved):
small comment that might make yaml spec cleaner, otherwise lgtm. would be nice to have wai check as well

llmfoundry/models/layers/attention.py
```diff
@@ -494,7 +495,17 @@ def __init__(
         self.n_heads = n_heads
         self.kv_n_heads = kv_n_heads
         self.sliding_window_size = sliding_window_size
+        if reuse_kv_x_layer_idx is not None:
+            if reuse_kv_layer_idx is not None:
+                raise ValueError(
+                    'Only one of reuse_kv_layer_idx and reuse_kv_x_layer_idx can be set.',
+                )
+            if self.fused_qkv:
+                raise ValueError(
```
waiwuc, line 504: Would be good to add a comment on why so. Is `reuse_kv_layer_idx` supported with fused_qkv?

llmfoundry/models/layers/blocks.py
```diff
@@ -192,10 +193,13 @@ def forward(
                 output_attentions=output_attentions,
                 alibi_slopes=alibi_slopes,
                 flash_attn_padding_info=flash_attn_padding_info,
+                x_prev=x_prev,
                 **extra_kwargs,
             )
         else:
             a = self.norm_1(x)
+            if x_prev is not None:
+                x_prev = self.norm_1(x_prev)
```
waiwuc, line 202: ideally we have test coverage on the norm as well, for both `fuse_norm_attn_norm` enabled and disabled. But not a big issue if it's hard to add.

tests/models/test_model.py
```diff
@@ -2831,8 +2922,8 @@ def test_get_modules_order_expanded():
     assert expected_list == MPTModel._get_modules_order_expanded(order)
 
 
-@pytest.mark.parametrize('reuse_kv_layer_idx', [-2, -1, 0])
-def test_resolve_reuse_kv_layer_idx(reuse_kv_layer_idx: int):
+@pytest.mark.parametrize('reuse_kv_layer_idx', [-2, 0])
```
waiwuc, line 2925: Why is -1 removed if we still check for these cases in line 2940 and 2954?

tests/models/test_model.py
```diff
@@ -2873,35 +2964,25 @@ def test_resolve_reuse_kv_layer_idx(reuse_kv_layer_idx: int):
...
-
-    elif reuse_kv_layer_idx == -2:
+    if reuse_kv_layer_idx == -2:
         assert _validate_helper(b_idx=2) == 0
     else:
         with pytest.raises(
             expected_exception=ValueError,
             match=
-            'The relative index of kv layer to reuse, override_attn_config\\[\'reuse_kv_layer_idx\'\\]=0, should be negative.',  # type: ignore
+            'The relative index of kv layer to reuse should be negative.',  # type: ignore
```
waiwuc, line 2985: Tests the 0 case - but what about -1?

waiwuc (approved):
Code looks good, just left some suggestions on documenting the logic a bit more clearly.

llmfoundry/models/mpt/modeling_mpt.py
ShashankMosaicML: We don't need a separate negative value for each subsequent layer. The following two are the same:

```
   converted_config_overrides:
      block_overrides:
        order:
        - name: default
        - name: reuse_1
        - name: reuse_2
        - name: reuse_3
      overrides:
          reuse_1:
            attn_config:
              reuse_kv_layer_idx: -1
          reuse_2:
            attn_config:
              reuse_kv_layer_idx: -2
          reuse_3:
            attn_config:
              reuse_kv_layer_idx: -3 
```

the above and the below are the same
```
   converted_config_overrides:
      block_overrides:
        order:
        - name: default
        - name: reuse_1
        - name: reuse_1
        - name: reuse_1
      overrides:
          reuse_1:
            attn_config:
              reuse_kv_layer_idx: -1
```

This is because if layer 3 reuses the kv cache of layer 2, then it is essentially just using the kv cache of layer 1. We take care of this recursive logic.

tests/models/layers/test_attention.py
```diff
@@ -448,3 +448,42 @@ def test_apply_temperature_tuning():
...
+        'attn_pdrop': 0.0,
+        'attn_impl': 'torch',
+        'qk_ln': False,
+        'qk_gn': False,
+        'clip_qkv': None,
+        'softmax_scale': None,
+        'sliding_window_size': -1,
+        'kv_n_heads': 2,
+        'fused_qkv': False,
+        'reuse_kv_x_layer_idx': 0,
```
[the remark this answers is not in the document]
ShashankMosaicML: Sorry my earlier comment was incorrect. We need a dummy `reuse_kv_x_layer_idx` so that the attention layer's `self.reuse_kv_layer_idx` gets set to something. Otherwise `get_qkv` will not use the `x_prev`

tests/models/test_model.py
ShashankMosaicML: The -1 case used to test that the config of the current and the previous layer matched so that the kv cache or kv input could be reused. However, now we also allow for a sliding window layer to reuse the kv cache or kv input of a full attention layer, hence this test and the corresponding check and error have been removed.

PR #1717, opened 2025-01-30
Bump version to 0.18.0.dev

milocress:
Removes `sample_weighing_factor` argument.


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -1362,17 +1360,7 @@ def compute_loss_from_logits(
     else:
         loss = losses.sum() / (targets != loss_fn.ignore_index).sum()
         if sample_weighing_factor is not None:
```
dakinggg, line 1362: since arg was deprecated and is now being removed, you should just remove the arg.

milocress: Sounds good, done

dakinggg (approved):

PR #1728, opened 2025-02-24
Bump FA2 to 2.7.4.post1

KuuCi:
FA2 2.6.3 doesn't support wheels with torch >2.4

FA updated the unpadding return values to return 5 outputs instead of 4:
https://github.com/Dao-AILab/flash-attention/blob/08f4c802c450708a86a92b226cba5663be81aead/flash_attn/bert_padding.py#L98

MPT doesn't actually support this FA change so we will gracefully fail MPT

Testing:
Regression test: https://databricks.slack.com/archives/C05T1A4UMT8/p1741210602072179


setup.py
```diff
@@ -52,7 +52,7 @@
 ]
 
 install_requires = [
-    'mosaicml[libcloud,wandb,oci,gcs,mlflow]>=0.28.0,<0.29',
```
dakinggg, line 55: composer upgrade should be done in a separate pr

setup.py
```diff
@@ -91,26 +91,26 @@
...
 
 extra_deps['tensorboard'] = [
-    'mosaicml[tensorboard]>=0.28.0,<0.29',
+    'mosaicml[tensorboard]>=0.29.0,<0.30',
 ]
 
 # Flash 2 group kept for backwards compatibility
 extra_deps['gpu-flash2'] = [
-    'flash-attn==2.6.3',
+    'flash-attn==2.7.4.post1',
```
dakinggg, line 107: we also need to upgrade flash in the composer images so that the base images have the right flash

setup.py
```diff
@@ -124,7 +124,7 @@
 ]
 
 extra_deps['te'] = [
-    'transformer-engine[pytorch]>=1.11.0,<1.12',
+    'transformer-engine[pytorch]>=1.13.0,<2.0',
```
dakinggg, line 127: transformer engine upgrade should be done in a separate pr

tests/models/hf/test_hf_mpt_gen.py
```diff
@@ -24,6 +24,12 @@ def test_init_hfhub_mpt(
         pytest.skip(f'{attn_impl=} not implemented for {device=}.')
     composer_device = get_device(device)
 
+    if device == 'gpu' and attn_impl == 'flash':
```
dakinggg, line 27: error should be in the code, not in the test. test can confirm that the error is raise

setup.py
KuuCi: TE 1.11.0 has dependency of flash-attn>2.1.0,<=2.6.3, which is incompatible with 'flash-attn==2.7.4.post1' so we need to bump

setup.py
KuuCi: https://github.com/mosaicml/composer/pull/3772, ready to merge when this pr is ready

setup.py
dakinggg: i understand you need to bump it to bump flash attention, but can you bump it in a separate PR?

KuuCi: ah gotcha

tests/models/hf/test_hf_mpt_gen.py
```diff
@@ -24,6 +24,19 @@ def test_init_hfhub_mpt(
         pytest.skip(f'{attn_impl=} not implemented for {device=}.')
     composer_device = get_device(device)
 
+    if attn_impl == 'flash':
```
dakinggg, line 27: is this really the only test that fails?

KuuCi: Well in the sense that only gpu no-flash runs... this will probably break with transformers update

KuuCi: Test was fixed via transformer bump

llmfoundry/models/hf/hf_causal_lm.py
```diff
@@ -80,6 +80,27 @@ def __init__(
         additional_eval_metrics: Optional[list] = None,
         should_save_peft_only: bool = True,
     ):
+        is_mpt_model = pretrained_model_name_or_path.startswith('mosaicml/mpt')
```
dakinggg, line 83: you don't need this code section any more right?

dakinggg (approved):
LGTM

PR #1729, opened 2025-03-01
Add user error to data prep script

milocress:
as title

### DIFF
--- a/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
+++ b/llmfoundry/command_utils/data_prep/convert_delta_to_json.py
@@ -738,6 +738,14 @@ def fetch_DT(
                 message=
                 f'You do not have permission to attach to the data preparation cluster you provided. {e}',
             ) from e
+        if isinstance(
+            e,
+            spark_errors.SparkConnectGrpcException,
+        ) and 'is not Shared or Single User' in str(e):
+            raise FaultyDataPrepCluster(
+                message=
+                f'The cluster you are using is not a shared or single-user cluster. {e}',
+            ) from e
         if isinstance(e, grpc.RpcError) and e.code(
         ) == grpc.StatusCode.INTERNAL and 'Job aborted due to stage failure' in e.details(
         ):


llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -738,6 +738,14 @@ def fetch_DT(
                 message=
                 f'You do not have permission to attach to the data preparation cluster you provided. {e}',
             ) from e
+        if isinstance(
```
dakinggg, line 741: is this different from the one at the top of this except? `and 'is not Shared or Single User Cluster' in str(e):`

mattyding: ES-1369903 is the relevant ticket with stacktrace. The issue is that `SparkConnectGrpcException` isn't propagated upward in `fetch` and gets cast to a `RuntimeException` on L537. Fix is to copy lines [533-534 ](https://github.com/mosaicml/llm-foundry/blob/5bf29cf8a3ca72ccdf1a231018a6c5e60f5877e9/llmfoundry/command_utils/data_prep/convert_delta_to_json.py#L533-L534) for `grpc.RpcError` and `spark_errors.SparkConnectGrpcException`[ types](https://github.com/mosaicml/llm-foundry/blob/5bf29cf8a3ca72ccdf1a231018a6c5e60f5877e9/llmfoundry/command_utils/data_prep/convert_delta_to_json.py#L708C13-L708C26).

milocress: ah this seems reasonable

mattyding:
Can you make the same change for [`grpc.RpcError`](https://github.com/mosaicml/llm-foundry/blob/870ca96f7e422357cb6e668b18090f67c42e5b8f/llmfoundry/command_utils/data_prep/convert_delta_to_json.py#L712) type since it looks like we don't propagate that up either?

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
dakinggg: can you also determine why the unit tests didn't catch the issue and fix it?

PR #1730, opened 2025-03-03
Bump TE for FA 2.7.1.post1 bump

KuuCi:
Needed for https://github.com/mosaicml/llm-foundry/pull/1728

Testing is here:
https://databricks.slack.com/archives/C05T1A4UMT8/p1741039448734289


dakinggg (commented):
what tests do we have for this? can you see what chuck might have left behind that we could run?

PR #1734, opened 2025-03-05
Fix dtype issue in transformers

dakinggg:
In transformers v4.49 (not supported here yet), `torch_dtype` takes effect, which would result in the master weights being bf16. This PR fixes that by hardcoding to `torch32` when we're loading, as we do not support master weights in lower precision.


llmfoundry/models/hf/hf_base.py
```diff
@@ -151,6 +151,7 @@ def build_config(
             trust_remote_code=trust_remote_code,
             use_auth_token=use_auth_token,
             attn_implementation=attn_implementation,
+            torch_dtype='float32',
```
irenedea, line 154: nit: could we make a global variable that is master weight precision?

irenedea (approved):
lgtm, one nit

llmfoundry/models/hf/hf_base.py
dakinggg: sure

PR #1735, opened 2025-03-05
Bump Transformer v4.49.0

KuuCi:
🚨🚨🚨 **MPT models on the huggingface hub will no longer be usable with trust_remote_code=True and older foundry versions will need to be used for that.** 🚨🚨🚨 

A couple of interesting notes:

Transformer tokenizers have added an [extra_special_tokens](https://github.com/huggingface/transformers/blob/159445d044623a4eba23ceb96dc7bd5bda51aa1a/src/transformers/tokenization_utils_base.py#L1451) attribute + made a new object [`AddedTokens`](https://github.com/huggingface/transformers/blob/159445d044623a4eba23ceb96dc7bd5bda51aa1a/src/transformers/tokenization_utils_base.py#L96) for special tokens.

Seems we have to explicitly set torch_dtype due to [this](https://github.com/huggingface/transformers/blob/51ed61e2f05176f81fa7c9decba10cc28e138f61/src/transformers/configuration_utils.py#L1000-L1013), where transformer will explicitly converts the torch_dtype to a string for JSON serialization. 

Looks like they've gotten rid of LlamaFlashAttention2 and refactored Attention:
https://github.com/huggingface/transformers/pull/35235

Seems like our test was actually hitting this strange interaction:
https://github.com/huggingface/transformers/issues/30305


tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -587,6 +645,7 @@ def test_huggingface_conversion_callback_interval(
             f'ba{batches_per_epoch}',
         ),
         trust_remote_code=True,
+        torch_dtype=precision,
```
dakinggg, line 648: so, i thought the issue was that transformers now uses `torch_dtype` on load, so the `precision` that we save in should already be used here when we reload. Is that not true?

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -873,6 +932,7 @@ def _assert_checkpoint_equivalence(
             loaded_model = transformers.AutoModelForCausalLM.from_pretrained(
                 checkpoint_path,
                 trust_remote_code=True,
+                torch_dtype=precision,
```
dakinggg, line 935: same q as above

tests/a_scripts/inference/test_convert_composer_to_hf.py
KuuCi: I think I saw this GH issue:
https://github.com/huggingface/transformers/issues/30305

llmfoundry/data/utils.py
```diff
@@ -211,7 +211,7 @@ def get_text_collator(
     collate_fn = transformers.DataCollatorForLanguageModeling(
         tokenizer=tokenizer,
         mlm=mlm_probability is not None,
-        mlm_probability=mlm_probability,
```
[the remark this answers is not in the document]
KuuCi: There was some sort of type checking to float as transformer does now, so is `mlm_probability` was `None`, `None` would still be passed in and then fail type checking.

llmfoundry/data/text_data.py
```diff
@@ -197,7 +197,7 @@ def __init__(
 
     # How to tokenize a text sample to a token sample
     def _tokenize(self, text_sample: Mapping) -> dict[str, list[int]]:
-        if self.tokenizer._pad_token is None:
```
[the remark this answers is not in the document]
KuuCi: I think they might have removed _pad_token from [here](https://github.com/huggingface/transformers/blob/e9756cdbc7dcec91ea0dde55c165d6276bd08252/src/transformers/tokenization_utils_fast.py#L82) since I was hitting:
```AttributeError: GPT2TokenizerFast has no attribute _pad_token
```

setup.py
```diff
@@ -85,6 +85,7 @@
     'pytest_codeblocks>=0.16.1,<0.18',
     'pytest-cov>=4,<7',
     'pyright==1.1.256',
+    'pydantic==1.10.10',
```
dakinggg, line 88: pinning pydantic is not the right answer here. we shouldn't have any dependency on pydantic or deepspeed and need to figure out what is causing that

tests/a_scripts/inference/test_convert_composer_to_hf.py
dakinggg: It is different when using different construction APIs

```
In [1]: import transformers

In [2]: llama1bcfg = transformers.AutoConfig.from_pretrained('meta-llama/Llama-3.2-1B-Instruct')

In [3]: llama1bcfg.torch_dtype
Out[3]: torch.bfloat16

In [4]: llama1b_frompt = transformers.AutoModelForCausalLM.from_pretrained('meta-llama/Llama-3.2-1B-Instruct')

In [5]: next(llama1b_frompt.parameters()).dtype
Out[5]: torch.float32

In [6]: llama1b_fromcfg = transformers.AutoModelForCausalLM.from_config(llama1bcfg)

In [7]: next(llama1b_fromcfg.parameters()).dtype
Out[7]: torch.bfloat16
```

🤦

dakinggg: ok their api here doesn't really make sense to me. but i guess this is right, given that. But we should then add to this test to test that the actual safetensors files written out contain the right dtype. can do like n_params*2 bytes, and the file should be about that size (and not n_params*4 bytes)

KuuCi: oh LOL

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -220,11 +220,74 @@ def check_hf_tokenizer_equivalence(
...
+            str,
+        ) else attr2.content if hasattr(attr2, 'content') else str(attr2)
         assert attr_value1 == attr_value2
 
-    assert tokenizer1.__dict__ == tokenizer2.__dict__
+    if 'extra_special_tokens' in tokenizer2.init_kwargs and 'extra_special_tokens' not in tokenizer1.init_kwargs:
+        tokenizer2.init_kwargs.pop('extra_special_tokens')
+
+    # Convert special tokens maps to strings for comparison
+    if '_special_tokens_map' in tokenizer1.__dict__ and '_special_tokens_map' in tokenizer2.__dict__:
```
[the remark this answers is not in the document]
KuuCi: yeah lgtm

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -220,11 +221,67 @@ def check_hf_tokenizer_equivalence(
...
+            for key in list(token_map1.keys()):
+                if hasattr(token_map1[key], 'content'):
+                    token_map1[key] = token_map1[key].content
+
+            # Process values in the second tokenizer's map
+            for key in list(token_map2.keys()):
+                if hasattr(token_map2[key], 'content'):
+                    token_map2[key] = token_map2[key].content
+
+            # Additional special processing for _added_tokens_decoder
```
dakinggg, line 256: what is this extra code doing?

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -274,6 +331,51 @@ def check_hf_model_equivalence(
...
+        bool: True if the safetensors files have the expected size, False otherwise
+    """
+    total_params = sum(p.numel() for p in model.parameters())
+    # Determine expected bytes per parameter based on precision
+    bytes_per_param = {
+        torch.float32: 4,
+        torch.float16: 2,
+        torch.bfloat16: 2,
+        torch.int8: 1,
+    }.get(expected_precision, 4)
```
dakinggg, line 358: prefer no default, any precision we're using should be added here

tests/models/hf/test_hf_config.py
```diff
@@ -310,11 +296,13 @@ def test_use_flash():
         attention_attr,
     )
     assert isinstance(attention_layer, flash_attn_class)
-
-    # Make sure that HF has not cast the parameters to bf16
-    assert next(model.parameters()).dtype == torch.float32
+    assert next(model.parameters()).dtype == torch.bfloat16
 
 
+@pytest.mark.skip(
```
dakinggg, line 302: still don't want this skipped

tests/a_scripts/inference/test_convert_composer_to_hf.py
KuuCi: thats fine

tests/models/hf/test_hf_config.py
KuuCi: forgot to rm skip here

tests/a_scripts/inference/test_convert_composer_to_hf.py
KuuCi: First for loop should be removed, it's no longer used. Basically the rest is to get the `AddedTokens` objects and convert them into strings for comparisons.

dakinggg (commented):
Basically LGTM! Please update the PR description to note that after this PR, the MPT models on the huggingface hub will no longer be usable with `trust_remote_code=True` and older foundry versions will need to be used for that.

Please also adjust the MPT error to say that it is no longer supported by foundry.

llmfoundry/models/hf/hf_base.py
```diff
@@ -228,13 +228,11 @@ def build_inner_model(
...
         """
-        if not trust_remote_code and pretrained_model_name_or_path.startswith(
+        if pretrained_model_name_or_path.startswith(
             'mosaicml/mpt',
         ):
             raise ValueError(
-                'trust_remote_code must be set to True for MPT models. Without this, the MPT model code will come from the transformers library, '
-                +
-                'which is significantly slower and not compatible with the LLM foundry training code, rather than the code release by MosaicML.',
+                'MPT models are no longer supported by LLM Foundry due to transformer version v4.49.0 incompatibilities'
```
dakinggg, line 235:
```suggestion
                'The MPT series of models on the Hugging Face Hub is no longer supported by LLM Foundry. Please use an older version of LLM Foundry (<0.18) or use a different model. Please open a GitHub issue and we can help you downgrade or work around the issue.'
```

llmfoundry/models/hf/hf_base.py
```diff
@@ -228,13 +228,13 @@ def build_inner_model(
...
-        ):
+        if pretrained_model_name_or_path.startswith('mosaicml/mpt',):
             raise ValueError(
-                'trust_remote_code must be set to True for MPT models. Without this, the MPT model code will come from the transformers library, '
+                'The MPT series of models on the Hugging Face Hub is no longer supported by LLM Foundry. '
+                +
+                'Please use an older version of LLM Foundry (<0.18) or use a different model. '
                 +
-                'which is significantly slower and not compatible with the LLM foundry training code, rather than the code release by MosaicML.',
+                'Please open a GitHub issue and we can help you downgrade or work around the issue.',
```
dakinggg, line 237:
```suggestion
                'Please open a GitHub issue if this is a problem for you and we can help you downgrade or work around the issue.',
```

PR #1738, opened 2025-03-10
Update layer_builders.py

Soumit0702:


llmfoundry/models/layers/layer_builders.py
```diff
@@ -100,12 +100,19 @@ def build_attention_layer(
     )
 
 
+from custom_layers import BandMatrix  # Adjust the import path based on the actual location of custom_layers.py
```
milocress, line 103: where is `BandMatrix` implemented?

PR #1740, opened 2025-03-12
Bump torch to 2.6

rithwik-db:
Upgrading pytorch to 2.6 on llmfoundry


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -566,14 +570,33 @@ def _get_hf_model(self, state: State):
...
             original_tokenizer = state.model.module.tokenizer
         elif isinstance(state.model.model, FSDP):
+            # Type assertions for FSDP case
+            assert isinstance(state.model, HuggingFaceModel)
+            assert isinstance(state.model.model.module, PreTrainedModel)
+            assert isinstance(state.model.tokenizer, PreTrainedTokenizerBase)
+
             original_model: PreTrainedModel = state.model.model.module
-            state_dict_model = state.model.model
+            state_dict_model = state.model.model.module
```
dakinggg, line 592: are you sure this is right? did this change?

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
```diff
@@ -236,7 +236,7 @@ def run_query(
             raise ValueError(f'cursor cannot be None if using method dbsql')
         cursor.execute(query)
         if collect:
-            return cursor.fetchall()
+            return cursor.fetchall()  # type: ignore
```
dakinggg, line 239: are you sure you're ignoring things that are needed to be ignored? This line of code has nothing to do with torch

llmfoundry/callbacks/hf_checkpointer.py
rithwik-db: oops that might be an autocorrect issue, thanks for catching that!

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
rithwik-db: This still shows up on the `precommit run -a` change so I had to add it, but yeah you're right this is not a `torch` related change.

llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -366,6 +366,7 @@ def _fsdp_wrap_fn(
 ) -> bool:
     # FSDP Wrap function for MPT Models
     if hasattr(module, '_fsdp_kwargs_dict'):
+        assert isinstance(module._fsdp_kwargs_dict, bool)
```
dakinggg, line 369: this is not a bool, its a dict (the return type is wrong)

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
dakinggg: im just wondering if whatever you're running locally isn't really matching CI, since this type ignore does not seem to be needed on main

llmfoundry/models/mpt/modeling_mpt.py
rithwik-db: Just going to `type: ignore` this unless you want me to convert this and the dependent functions to all return a dict...

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
rithwik-db: Let me try without these changes and see if anything comes up on CI/CD

.github/workflows/pr-gpu.yaml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
     - main
     - release/*
-  pull_request_target:
+  pull_request:
```
dakinggg, line 7: leaving comment to revert before merge, once testing is done

tests/tp/test_tp_strategies.py
```diff
@@ -165,13 +165,22 @@ def test_tp_train(tp_degree: int, tp_strategy: str, tmp_path: Path):
     # Compare loss and expected loss for TP
     import numpy as np
     expected_tp_loss = np.array([
```
dakinggg, line 167: I think this is fine btw. TP doesn't fully work until we incorporate fsdp2 anyway, and this isn't a very good test comparing against a fixed set of loss values. It would be a better test if it compared against a non TP model, with some allowed tolerance.

rithwik-db: Sound good, updated the value for now and we will revisit this when we add fsdp2 support soon enough

llmfoundry/command_utils/data_prep/convert_delta_to_json.py
rithwik-db: Okay I removed the unnecessary changes and it looks like it still works fine. I cross-checked the files changed with the original pyright failures [here](https://github.com/mosaicml/llm-foundry/actions/runs/13822012571) and looks like we don't have any extraneous files.

dakinggg (commented):
changes LGTM. kind of a sad amount of type ignores, but not blocking.

tests/a_scripts/inference/test_convert_composer_to_hf.py
```diff
@@ -1396,7 +1410,10 @@ def test_convert_and_generate(
         'scripts/train/yamls/pretrain/testing.yaml',
         pytest.param(
             'scripts/train/yamls/pretrain/testing-moe.yaml',
-            marks=pytest.mark.gpu,
+            marks=[pytest.mark.gpu, pytest.mark.skipif(
```
dakinggg, line 1413: instead of skipping all the sparse moe tests can we just switch them to use the grouped impl?

setup.py
```diff
@@ -119,7 +119,7 @@
 ]
 
 extra_deps['megablocks'] = [
-    'megablocks<1.0',
+    'megablocks[gg]<1.0',
```
dakinggg, line 122: should either have the gg extra _or_ the grouped_gemm dep here, not both.

llmfoundry/models/layers/ffn.py
```diff
@@ -305,6 +305,8 @@ def build_torch_dmoe(
 
     fc_type = kwargs.pop('fc_type', None)
     del fc_type  # Unused
+    mlp_impl = kwargs.pop('mlp_impl', None)
```
dakinggg, line 308: why is this needed? mlp_impl just isn't a valid arg for torch_dmoe

setup.py
rithwik-db: I'm not sure I follow...

dakinggg (commented):
changes lgtm

dakinggg (approved):

PR #1745, opened 2025-03-14
Fix meta initialization for FSDP training with HF models and TE Layers

jjuvonen-amd:
# What does this PR do?

Fixes weight initialization in meta device-initialized FSDP training of Hugging Face models. Specifically, it addresses weight initialization of Transformer Engine layers and modules that are skipped in model's `_init_weights` method.

## Changes

- Implements comprehensive initialization handling in `prepare_inner_model` to detect and initialize norm modules missed by `model._init_weights`
  - i.e. RMSNorm modules in [llama](https://github.com/huggingface/transformers/blob/052e652d6d53c2b26ffde87e039b723949a53493/src/transformers/models/llama/modeling_llama.py#L739)
- Fixes meta device initialization for Transformer Engine modules
- Adds a unit test for checking nan weights after FSDP HF model init with meta device
- Adds an example script demonstrating Transformer Engine FP8 integration with Llama models
  - Edit: example script left out from PR but kept in [llama-te-fp8-example branch fork](https://github.com/AMD-AIG-AIMA/llm-foundry/tree/llama-te-fp8-example)

## Why is this change required?

Some modules, particularly RMSNorm layers in Llama or Qwen, are skipped during model weight initialization, resulting in NaN values that propagate to the loss and break training. Also, all Transformer Engine layers are skipped in weight initialization. This is only an issue with FSDP and meta device init as this relies on model weight initialization after sharding empty weights to devices.

## Testing
Changes verified on AMD MI300X (8-GPU) with [llama3.1-8b](https://huggingface.co/meta-llama/Llama-3.1-8B) under FSDP training, with and without Transformer Engine optimizations, using the rocm/pytorch-training:v25.2 container.

### Unit tests

Unit test added:
- `tests/models/hf/test_hf_fsdp_meta.py`: checks if hf model weights have nan values with meta device initialization and FSDP

Run test with:
```bash
make test-dist-gpu WORLD_SIZE=2 EXTRA_ARGS="tests/models/hf/test_hf_fsdp_meta.py -v"
```

### End2end testing

Run Llama3.1-8b FSDP pretraining with meta initialization:
```bash
cd scripts/train
composer train.py yamls/pretrain/hf-llama-3.1-8B.yaml
```

Without the proposed changes, the outputs show nan loss values:
```bash
Train loss/train/total: nan
```

With the proposed changes, outputs show normal loss values:
```bash
Train loss/train/total: 7.5547
```

### TE-optimized Llama testing:
Edit: TE example script left out from PR but kept in [llama-te-fp8-example branch fork](https://github.com/AMD-AIG-AIMA/llm-foundry/tree/llama-te-fp8-example)

Run TE FP8 Llama3.1-8b FSDP pretraining with meta initialization:
```bash
cd scripts/train
composer train.py yamls/pretrain/te-hf-llama-3.1-8B.yaml
```

Without the proposed changes, the outputs show nan loss values:
```bash
Train loss/train/total: nan
```

With the proposed changes, outputs show normal loss values:
```bash
Train loss/train/total: 10.7082
```


llmfoundry/models/hf/hf_base.py
```diff
@@ -436,5 +443,37 @@ def prepare_inner_model(
...
+                    (te.LayerNormMLP, te.LayerNormLinear, te.Linear),
+                ):
+                    module.reset_parameters(defer_init=False)
+                    return
+
+            # Use the model's default initialization method
+            model._init_weights(module)
+
+            # Initialize any modules that were skipped in model._init_weights
+            if hasattr(module, 'weight') and module.weight is not None:
```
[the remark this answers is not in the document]
dakinggg: ah because to_empty does not initialize anything, just allocates the memory. If im understanding right, that makes this code here (hope there is a nan somewhere in the weight tensor) pretty fragile. Maybe better just be more explicit with the norm layers you are expecting not to be initialized? Even though that won't handle all cases.

scripts/train/te_hf_llama_example.py
```diff
@@ -0,0 +1,241 @@
+# Copyright 2025 MosaicML LLM Foundry authors
```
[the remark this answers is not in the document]
jjuvonen-amd: Sure, I understand. I'll move the example to [llama-te-fp8-example branch on our fork](https://github.com/AMD-AIG-AIMA/llm-foundry/tree/llama-te-fp8-example) and leave it out of this PR. This way it's still accessible for reference.

scripts/train/yamls/pretrain/te-hf-llama-3.1-8B.yaml
```diff
@@ -0,0 +1,108 @@
+variables:
```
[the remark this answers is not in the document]
jjuvonen-amd: got it, I'll leave this out too.

scripts/train/yamls/pretrain/hf-llama-3.1-8B.yaml
```diff
@@ -0,0 +1,103 @@
...
+model:
+  name: hf_causal_lm
+  init_device: meta
+  pretrained_model_name_or_path: ${variables.model_name}
+  pretrained: false
+  # Note: you must have set the HF_TOKEN environment variable and have access to the llama2 models
+  use_auth_token: true
+  use_flash_attention_2: true
+  config_overrides:
+    torch_dtype: bfloat16
```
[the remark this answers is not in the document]
jjuvonen-amd: Thank you for pointing this out, I wasn't aware of the recent change. I intended to keep master weights in float32 and have AMP autocasting to choose precision according to `precision: amp_bf16`. torch_dtype config override was there to suppress this warning:

```
Flash Attention 2.0 only supports torch.float16 and torch.bfloat16 dtypes, but the current dype in LlamaForCausalLM is torch.float32.
```

I'll remove the override.

llmfoundry/models/hf/hf_base.py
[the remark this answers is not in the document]
jjuvonen-amd: Good catch, and thanks for sharing the testing scrip. It appears that whether unitialized weights have zeros, nondeterministic data or NaNs depends on deterministic mode and torch.utils.deterministic.fill_uninitialized_memory bool (default: True). Found this explanation in [torch.empty() docs](https://pytorch.org/docs/2.4/generated/torch.empty.html#torch-empty). To test this, I modified your script:

```python
import torch
from composer.utils import dist, get_device

def test_nans(deterministic_mode):
    print(f"setting torch deterministic mode: {deterministic_mode}")
    torch.use_deterministic_algorithms(deterministic_mode)

    dist.initialize_dist(get_device('gpu'))
    from transformers.models.llama.modeling_llama import LlamaRMSNorm
    with init_empty_weights(include_buffers=False):
        rms = LlamaRMSNorm(32)
    print(rms.weight)
    rms.to_empty(device=f'cuda:{torch.cuda.current_device()}', recurse=False)
    print(rms.weight)
    print("done")

if __name__ == "__main__":
    test_nans(False)
    test_nans(True)
```

I changed unit test to set deterministic mode and fill_uninitialized_memory to ensure unitialized modules are filled with NaNs. 
Also, I changed intitialization for specifically norm modules such as LlamaRMSNorm, Qwen2RMSNorm, etc.

scripts/train/yamls/pretrain/hf-llama-3.1-8B.yaml
```diff
@@ -0,0 +1,103 @@
+variables:
+  data_local: /data/streaming/c4_2048
```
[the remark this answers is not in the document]
jjuvonen-amd: changed

tests/models/hf/test_hf_fsdp_meta.py
```diff
@@ -0,0 +1,64 @@
...
+
+from llmfoundry.models.hf.hf_causal_lm import ComposerHFCausalLM
+from llmfoundry.utils.builders import build_tokenizer
+
+
+@pytest.mark.gpu
+@pytest.mark.world_size(2)
+def test_hf_meta_init_fsdp():
+    # Use deterministic mode to ensure uninitialized weights are filled with NaNs
+    torch.use_deterministic_algorithms(False)
```
dakinggg, line 16: I thought we wanted to set this to true?

dakinggg, line 16: could you also please put these in a context manager so they don't affect other tests?

llmfoundry/models/hf/hf_base.py
```diff
@@ -436,5 +443,34 @@ def prepare_inner_model(
...
+                    (te.LayerNormMLP, te.LayerNormLinear, te.Linear),
+                ):
+                    module.reset_parameters(defer_init=False)
+                    return
+
+            # Use the model's default initialization method
+            model._init_weights(module)
+
+            # Initialize normalization modules that are skipped in model._init_weights
+            if _is_normalization_layer(module) and hasattr(
```
dakinggg, line 469: I wish there was a better way here...it does seem like there is some risk that a random other layer that was already initialized and happens to have the string "norm" in its name will end up reinitialize. What do you think about just hardcoding the fix for the llama named layer, and raising an error if any other layer has "norm" it and weight contains a nan? that should fix the issue you care about for llama, and be a small improvement for other edge cases. unless you have a better way to reliably tell if something was not initialized.

tests/models/hf/test_hf_fsdp_meta.py
jjuvonen-amd: ah true, fixed this now

jjuvonen-amd: yes, good point. I added a context manager for enabling deterministic settings and setting them back.

llmfoundry/models/hf/hf_base.py
jjuvonen-amd: Hardcoding layer names seems like the best option. I added the two RMSNorms I've encountered to a `EXPLICIT_INIT_ONES_NAMES` list. Weights of modules that are on the list are initialized by ones. Other weights are just checked for NaNs and warned if NaNs are found, and it wouldn't matter if they are norm or some other layer type. NaNs would break the training anyway.

I tried to see if there's some other robust way to detect uninitialized layers, but couldn't find a working way without other drawbacks. I explored:
- Detect initialization from Tensor metadata such as `tensor.storage.size()` - Didn't work. This value doesn't change.
- Compare weights before and after init (mean, std) - would add complexity and potentially problematic (i.e. if weights are zero-initialized, this may not work)

llmfoundry/models/hf/hf_base.py
```diff
@@ -436,5 +443,36 @@ def prepare_inner_model(
         # so that the (possible) embedding resizing doesn't destroy them
         prepare_hf_model_for_fsdp(model, init_device)
 
+        # Define explicitly which layers should be initialized with ones (skipped by model._init_weights)
```
dakinggg, line 446:
```suggestion
        # Define explicitly which layers should be initialized with ones (these are known to be skipped by the underlying Hugging Face model._init_weights)
```

tests/models/hf/test_hf_fsdp_meta.py
```diff
@@ -0,0 +1,83 @@
...
+
+        trainer = Trainer(
+            model=model,
+            device='gpu',
+            parallelism_config={'fsdp': fsdp_config},
+        )
+        assert trainer.state.fsdp_enabled
+
+        # Check for NaN weights
+        for name, param in trainer.state.model.model.named_parameters():
```
dakinggg, line 80: might as well assert that the layer you expect to be filled with all ones is indeed

dakinggg (approved):
LGTM! Couple minor comments

llmfoundry/models/hf/hf_base.py
dakinggg: I agree, thanks for trying to find a better way!

tests/models/hf/test_hf_fsdp_meta.py
jjuvonen-amd: Yes, I added assert for LlamaRMSNorm weights being ones

dakinggg:
Could you merge main? I think that should fix the failing tests

jjuvonen-amd:
> Could you merge main? I think that should fix the failing tests, we did a torch version upgrade on main.

Thanks, that fixed GPU tests.

dakinggg:
Sorry, there are some new typing issues coming from torch 2.6. You can add type ignores to any that aren't obvious how to fix.

jjuvonen-amd:
> Sorry, there are some new typing issues coming from torch 2.6. You can add type ignores to any that aren't obvious how to fix.

I pushed type checks to resolve those failed Pyright checks. I had to use type ignores for `torch.utils.deterministic.fill_uninitialized_memory` lines.

Thanks for your patience and feedback!

PR #1752, opened 2025-03-18
Use llmfoundry image instead of pytorch image for gpu tests

rithwik-db:
We replaced the original llmfoundry images with the pytorch images since the original llmfoundry images weren't built yet; so I am replacing this just before we start working on the llmfoundry release

### DIFF
--- a/.github/workflows/pr-gpu.yaml
+++ b/.github/workflows/pr-gpu.yaml
@@ -4,7 +4,7 @@ on:
     branches:
     - main
     - release/*
-  pull_request:
+  pull_request_target:
     branches:
     - main
     - release/**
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-gpu.yaml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
     - main
     - release/*
-  pull_request_target:
+  pull_request:
```
rithwik-db, line 7: Will replace before merging

dakinggg (commented):
just fyi this doesn't block release at all. change lgtm, will approve once pr target is reverted

dakinggg (approved):

PR #1753, opened 2025-03-18
bump dev version to 0.19.0.dev0

rithwik-db:
As title (also couldn't find anything to deprecate)


rithwik-db:
> https://pypi.org/project/llm-foundry/

Yup I couldn't find anything to deprecate

PR #1762, opened 2025-03-26
modified the packing slightly to enable inheritance

abaheti95:


llmfoundry/data/packing.py
```diff
@@ -106,7 +106,7 @@ def pack(self, batch: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
                 'sequence_id',
             ]
         # Cut everything down to size
-        sizes, trimmed_examples = _trim_batch(batch)
+        sizes, trimmed_examples = BinPackCollator._trim_batch(batch)
```
irenedea, line 109: Can you do `self._trim_batch`?

irenedea: Maybe we need classmethod for this to work properly with inheritance?

I should be able to create a class that only replaces the trim batch function:
```
class MyBinPackCollator(BinPackCollator):
     def _trim_batch(...):
           # my own logic

```

llmfoundry/data/packing.py
```diff
@@ -358,6 +364,9 @@ def auto_packing_ratio(
     tokenizer: PreTrainedTokenizerBase,
     device_batch_size: int,
     num_packing_ratios: int = 20,
+    max_ratio: float = 0,
```
irenedea, line 367: We shouldn't use actual numbers like 0 as a placeholder for something else. 
```suggestion
    max_ratio: Optional[float] = None,
```

irenedea (changes requested):

llmfoundry/data/packing.py
```diff
@@ -435,6 +455,8 @@ def profile_packing(
     max_ratio: float,
     num_packing_ratios: int,
     device_batch_size: int,
+    packing_collator: type[BinPackCollator] = BinPackCollator,
```
irenedea, line 458: nit: 
```suggestion
    packing_collator_cls: type[BinPackCollator] = BinPackCollator,
```

llmfoundry/data/packing.py
```diff
@@ -445,6 +467,8 @@ def profile_packing(
         max_ratio (float): Largest packing_ratio to test. Must be larger than `min_ratio`.
         num_packing_ratios (int): Number of packing_ratio values (spaced between `min_ratio` and `max_ratio`) to try.
         device_batch_size (int): The size of the batches (number of examples) per device.
+        packing_collator (BinPackCollator): The collator to use for packing.
```
irenedea, line 470:
```suggestion
        packing_collator_cls (BinPackCollator): The collator class to use for packing.
```

llmfoundry/data/packing.py
```diff
@@ -375,6 +386,9 @@ def auto_packing_ratio(
         tokenizer (PreTrainedTokenizerBase): The tokenizer for profiling.
         device_batch_size (int): The size of the batches (number of examples) per device.
         num_packing_ratios (int): The number of packing ratios to try.
+        max_ratio (Optional[float]): The largest packing ratio to try. Must be larger than `min_ratio`.
```
irenedea, line 389:
```suggestion
        max_ratio (Optional[float]): The largest packing ratio to try. Must be larger than `min_ratio`. Defaults to max_seq_len / 100.
```

irenedea (approved):
LGTM! pls addres nits before merging

llmfoundry/data/packing.py
abaheti95: Added the default

llmfoundry/data/packing.py
abaheti95: Updated the cls parameter name

PR #1764, opened 2025-03-29
Remove registration fallback

dakinggg:
Since we use `log_model` now for registration, even if registration itself fails, the model will still be logged to mlflow artifacts, so we no longer need the fallback to upload to save folder.


KuuCi (approved):
lgtm

irenedea (approved):
ty!

PR #1768, opened 2025-04-01
Remove a bunch of repeated calls to HF in the tests

dakinggg:
As title. Replaces repeated calls to HF in the tests with fixtures to avoid throttling.


irenedea (approved):
thank you!

PR #1769, opened 2025-04-01
Move save/load planner creation to after config logging

dakinggg:
Save/load planner creation puts python objects into the config, which means the config can no longer be logged (at least not without converting objects to primitive type representations. This just moves the object creation to after logging the config.

Manual test: `lp-test-3-tcrJbO`


ethantang-db (approved):
Looks good to me

PR #1770, opened 2025-04-01
Fix GPU Tests 3.10

KuuCi:
Yamls will parse 3.10 -> 3.1, which is invalid python file


dakinggg (approved):
LGTM, will force merge once you have manually tested

PR #1781, opened 2025-04-08
Making tokenizers optional in the building of LLMs

ethantang-db:
This PR makes it optional to have tokenizers when constructing LLMs, however respective parts that still require tokenizers will error out if it is not provided

### DIFF
--- a/llmfoundry/command_utils/train.py
+++ b/llmfoundry/command_utils/train.py
@@ -361,6 +361,7 @@ def train(cfg: DictConfig) -> Trainer:
 
     # Build tokenizer
     tokenizer = None
+    tokenizer_name - None
     if train_cfg.tokenizer:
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -3,12 +3,12 @@ default_language_version:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.2.2
+  rev: v0.11.4
```
dakinggg, line 6: pre-commit version upgrades should not be mixed with code changes. Any particular reason you made these changes?

ethantang-db: there were weird issues locally that made me have to upgrade, but I figured out how to revert and get it running wihtout upgrades

PR #1790, opened 2025-04-16
Update transformers to 4.51

dakinggg:
In the latest transformers version, they added a py.typed file, which means that the type checker starts using types from transformers. There are a number of legitimitate type issues surfaced that I tried to correct, but the vast majority result from transformers not actually being a type checked library and a couple specific difficulties:
(1) BatchEncoding (which is what tokenizers return) has a __getitem__ whose return type depends on the type of the index passed in (string vs int). The type checker can't handle this (correct answer would be to have overloads I believe).
(2) PeftModel forwards attr access to the underlying PreTrainedModel by overriding __getattr__. The type checker can't know this. The correct answer would be to do lots of unpacking of the model before accessing attrs, but I don't think its worth the energy.

Generally, if it wasn't clear to me what to do, I default to type ignore, since that is no worse of a state than before. For reviewers, probably best to skip reviewing all the type ignores, and focus on any places where I actually changed code.

Manual test run: `tr-types-2-9qAwxu`


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -640,25 +649,33 @@ def tensor_hook(
             # initialization cost.
             with init_empty_weights():
                 if self.using_peft:
-                    active_adapter = original_model.active_adapter
-                    base_model = original_model.get_base_model()
+                    from peft import PeftModel
+                    assert isinstance(original_model, PeftModel)
+                    active_adapter = original_model.active_adapter  # type: ignore
+                    base_model: PreTrainedModel = original_model.get_base_model(  # type: ignore
+                    )
```
ethantang-db, line 656: what happened here lol?

llmfoundry/data/data.py
```diff
@@ -72,24 +72,29 @@ def __init__(
         self.eos_text = eos_text
         self.should_wrap = not no_wrap
 
-        self.bos_tokens = self.tokenizer(
+        bos_iids = self.tokenizer(
```
ethantang-db, line 75: I'm assuming this isn't a typo... but what exactly do iids stand for?

llmfoundry/callbacks/hf_checkpointer.py
dakinggg: sometimes the formatter moves the type ignore around and then the type ignore ignores the wrong line

llmfoundry/data/data.py
dakinggg: input ids

llmfoundry/data/finetuning/tasks.py
```diff
@@ -343,16 +345,20 @@ def _tokenize_with_bos_removal(
     input_ids_starts_with_bos = False
     labels_starts_with_bos = False
     if has_bos_token and len(
-        tokenized_sample['input_ids'],
-    ) > 0 and len(tokenized_sample['labels']) > 0:
-        input_ids_starts_with_bos = tokenized_sample['input_ids'][
-            0] == tokenizer.bos_token_id
-        labels_starts_with_bos = tokenized_sample['labels'][
-            0] == tokenizer.bos_token_id
+        cast(Any, tokenized_sample['input_ids']),
```
ethantang-db, line 348: this is just for type checking?

dakinggg: yes, this is the "BatchEncoding" part of the PR description

dakinggg: I could have done type ignores, but casting to Any is actually the intention of the transformers code, because in the case where the index is a string (which is what we always have), its return type is Any.

ethantang-db (approved):

tests/eval/test_nlp_metrics.py
```diff
@@ -84,8 +92,15 @@ def test_in_context_learning_qa_cot_accuracy():
     assert metric.compute() == (2 / 4)
 
 
+from transformers import BatchEncoding
+
+
+def typed_batch_encoding_access(b: BatchEncoding, k: str) -> Any:
```
irenedea, line 98: nit: comment explaining why and when we introduced this

llmfoundry/data/finetuning/collator.py
```diff
@@ -398,18 +398,18 @@ def _process_and_batch_decoder_only(
             return_tensors='pt',
         )
 
-        batch['sequence_id'] = batch['attention_mask'] - 1
+        batch['sequence_id'] = batch['attention_mask'] - 1  # type: ignore
```
irenedea, line 401: nit: to get rid of these type ignores, you could do 
```
batch: Dict[str, torch.Tensor] = self.tokenizer.pad(
            processed_examples,
            padding='max_length',
            max_length=max_seq_len,
            return_tensors='pt',
        )
```

above

llmfoundry/data/finetuning/collator.py
```diff
@@ -487,38 +487,42 @@ def _process_and_batch_encoder_decoder(
             return_tensors='pt',
         )
         # We're still missing decoder_input_ids and decoder_attention_mask
-        batch['decoder_input_ids'] = torch.cat([
+        pad_id = self.tokenizer.pad_token_id
```
irenedea, line 490: nit: same here, you could probably just do some type casting/hint on batch?

llmfoundry/models/hf/hf_fsdp.py
```diff
@@ -229,10 +233,10 @@ def prepare_hf_enc_dec_model_for_fsdp(
     model.shared == model.encoder.embed_tokens and model.shared ==
     model.decoder.embed_tokens
     """
-    tied_embeddings = model.get_input_embeddings()
-    encoder = model.get_encoder()
-    decoder = model.get_decoder()
-    lm_head = model.get_output_embeddings()
+    tied_embeddings = model.get_input_embeddings()  # type: ignore
```
irenedea, line 236: lol that's annoying that PeftModel doesn't inherit from PretrainedModel

irenedea (approved):
fun PR, thank you! just had a few nits

tests/eval/test_nlp_metrics.py
dakinggg: oh shoot, this is unused, was just testing things, will delete

llmfoundry/data/finetuning/collator.py
dakinggg: thats technically not the type though...its a `BatchEncoding` which inherits `UserDict`, so behaves as a dict. im just gonna leave it. complying with the tokenizer output types is pretty messy.

llmfoundry/models/hf/hf_fsdp.py
dakinggg: It just mirrors all the attrs via overriding --getattr--

llmfoundry/data/finetuning/collator.py
dakinggg: same, gonna leave this one

llmfoundry/data/data.py
dakinggg: sure, done

PR #1792, opened 2025-04-16
extended hf_checkpointer for any additional content saving

ethantang-db:
Added a new method to override for the purpose of additional content saving

### DIFF
--- a/llmfoundry/callbacks/hf_checkpointer.py
+++ b/llmfoundry/callbacks/hf_checkpointer.py
@@ -848,6 +848,12 @@ class HuggingFaceCheckpointer(Callback):
                         self.flatten_imports,
                     )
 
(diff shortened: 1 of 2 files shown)


llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -852,3 +871,13 @@ def _save_checkpoint(
                 if use_temp_dir:
                     shutil.rmtree(temp_save_dir)
         dist.barrier()
+
+        if dist.get_global_rank() == 0:
+            self.save_additional_contents(
```
dakinggg, line 876: why does this need all these args? shouldn't it just need `state` and `temp_save_dir`? Also shouldn't this happen before upload and register?

dakinggg (approved):

PR #1794, opened 2025-04-17
Adding support for nope positional encoding in block overrides.

ShashankMosaicML:
Allows for certain layers to have no positional embedding, similar to Llama 4. Example usage:

```  
block_overrides:
  order:
  - name: default
  - name: nope_layer
  overrides:
    nope_layer:
      attn_config:
        nope: True
```


llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -182,6 +182,29 @@ def __init__(
...
+                            'nope position encoding block override cannot be used with learned_pos_emb.',
+                        )
+                    if self.attn_config['attn_impl'
+                                       ] == 'torch' and self.attn_config.get(
+                                           'alibi',
+                                           False,
+                                       ):
+                        raise ValueError(
+                            'nope position encoding block override cannot be used with alibi when using torch attention.',
+                        )
```
vchiley, line 206: NoPE can be used with RoPE?
but not other pos emb?

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -182,6 +182,29 @@ def __init__(
...
+            for override_def in block_overrides['overrides'].values():
+                if 'attn_config' in override_def and override_def[
+                    'attn_config'].get('nope', False):
+                    if override_def['attn_config'].get(
+                        'reuse_kv_layer_idx',
+                        None,
+                    ) is not None:
+                        raise ValueError(
+                            'nope position encoding block override cannot be used when reusing the kv cache of a previous layer.',
+                        )
```
vchiley, line 194: if we have reuse kv then shouldn't we invalidate pos emb is the same between layers

ie if we're reusing from a layer that has rope, then this layer should also have rope (but the same is true for nope or alibi etc)

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -195,6 +218,10 @@ def __init__(
             del kwargs['name']
         if 'loss_fn' in kwargs:
             del kwargs['loss_fn']
+        if self.attn_config.get('nope', False):
+            raise ValueError(
+                'nope cannot be specified as the default position encoding, it can only be specified as an override using block_overrides. Please use alibi or rope instead.',
+            )
```
vchiley, line 224: wait
why?

llmfoundry/models/mpt/configuration_mpt.py
ShashankMosaicML: Nope can't be used with alibi when using torch attention. It can still be used with alibi when using flash attention. The reason is that when using torch attention, we implement alibi through an attention mask, which is then passed to all the layers. If some layers have alibi, some have nope, then we will need to maintain multiple attention masks. Since we don't use alibi with torch, I think it is okay to not support nope with alibi on torch attention.

llmfoundry/models/mpt/configuration_mpt.py
ShashankMosaicML: Currently we compute rope rotation matrices for the entire model and then pass them to each of the layers. This was done for efficiency reasons. If we set nope as the default and rope for a few layers, that would require significant code changes and would add complexity to the code. Instead, I suggest that we set rope as default and nope for specific layers.

llmfoundry/models/mpt/configuration_mpt.py
ShashankMosaicML: Done.

tests/models/layers/test_flash_torch.py
```diff
@@ -760,3 +760,245 @@ def gen_bias(attn_impl: str):
...
+    })
+
+    n, s, f = 2, 4, cfg.d_model
+    assert cfg.d_model % cfg.n_heads == 0
+    if attn_type == 'grouped_query_attention':
+        cfg.kv_n_heads = 2
+
+    sequence_id = None
+
+    cfg.attn_impl = attn_impl
```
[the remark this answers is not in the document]
ShashankMosaicML: Sure, i'll try to make it shorter.

What this test tests is whether calling the attention module with nope (https://github.com/mosaicml/llm-foundry/blob/326b19cae8e07dd80caa5fb82bce8b8159508a34/tests/models/layers/test_flash_torch.py#L943C9-L952C10 ) produces the same result as if you compute attention explicitly, but don't apply any rotations or alibi (https://github.com/mosaicml/llm-foundry/blob/326b19cae8e07dd80caa5fb82bce8b8159508a34/tests/models/layers/test_flash_torch.py#L953-L981 )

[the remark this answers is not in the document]
ShashankMosaicML: Made the test shorter.

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -182,6 +182,22 @@ def __init__(
...
+                        )
+                    if self.attn_config['attn_impl'
+                                       ] == 'torch' and self.attn_config.get(
+                                           'alibi',
+                                           False,
+                                       ):
+                        raise ValueError(
+                            'nope position encoding block override cannot be used with alibi when using torch attention.',
+                        )
+
```
vchiley, line 200: Should this added code just be a part of `_validate_block_overrides`

llmfoundry/models/mpt/configuration_mpt.py
```diff
@@ -195,6 +211,10 @@ def __init__(
             del kwargs['name']
         if 'loss_fn' in kwargs:
             del kwargs['loss_fn']
+        if self.attn_config.get('nope', False):
```
vchiley, line 214:
```suggestion
        if self.attn_config.get('nope', False):
            # TODO: enable `nope` as a valid option to default position encoding. 
```

llmfoundry/models/mpt/configuration_mpt.py
ShashankMosaicML: Done.

PR #1795, opened 2025-04-19
Load model only on global rank 0 for mixed init

dakinggg:
Since `sync_module_states` syncs from global rank 0, we only actually need to download the model from HF on global rank 0, even though we were doing it on all local rank 0 just cause the code is a bit simpler that way. This adds a thread based synchronization so that we can download only on global rank 0 without needing to increase the dist timeout dramatically.

Manual testing:
- Big model loads: `405b-after-23-bp8HYb`
- Smaller model with multinode before and after:
before: `small-before-4-hxnlMv`, after: `small-after-12-UOTZcG`
Screenshot 2025-04-21 at 1 54 45 PM

### DIFF
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -428,7 +428,7 @@ class BaseHuggingFaceModel(HuggingFaceModel):
             download_thread.join(timeout=3600)
             log.debug('Download thread joined')
 
-        # Broadcast any
+        # Gather information about which ranks errored
         gathered_errors = dist.all_gather_object(error)
         ranks_with_error = [
             rank for rank, err in enumerate(gathered_errors) if err is not None
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -406,9 +406,11 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         download_thread.join(timeout=3600)
         log.debug('Download thread joined')
 
-        error = dist.broadcast_object_list([error], src=0)[0]
-        if error:
-            raise error
+        error_for_broadcast = [error]
+        dist.broadcast_object_list(error_for_broadcast, src=0)
+        error_to_maybe_raise = error_for_broadcast[0]
+        if error_to_maybe_raise:
+            raise error_to_maybe_raise
 
         # Use the pretrained generation config for the model if it exists.
         try:
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -357,7 +357,9 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         # Create a thread to busy wait for download to be complete
         download_thread = threading.Thread(
             target=download_thread_target,
+            daemon=True,
         )
+        log.debug('Starting download thread')
         download_thread.start()
 
         # initialize the model on the correct device
@@ -394,7 +396,9 @@ class BaseHuggingFaceModel(HuggingFaceModel):
             )
 
         rank0_done_global = True
+        log.debug('Joining download thread')
         download_thread.join(timeout=3600)
+        log.debug('Download thread joined')
 
         # Use the pretrained generation config for the model if it exists.
         try:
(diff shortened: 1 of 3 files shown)


llmfoundry/models/hf/hf_base.py
```diff
@@ -331,53 +334,111 @@ def build_inner_model(
...
+            download_status_queue.put(1)
+            log.debug('Download complete')
+
+        # Join and wait for the thread to complete on all ranks
+        if download_thread is not None:
+            log.debug('Joining download thread')
+            download_thread.join(timeout=3600)
+            log.debug('Download thread joined')
+
+        # Broadcast any
```
dakinggg, line 431:
```suggestion
        # Gather information about which ranks errored
```

dakinggg:
@irenedea @bowenyang008 this is ready for review now

llmfoundry/models/hf/hf_base.py
```diff
@@ -331,53 +334,111 @@ def build_inner_model(
...
+        # This approach is used because NCCL process group does not support updating the timeout
+        # and we don't want to require a super long distributed timeout for all operations
+        download_thread = None
+        download_status_queue = None
+        if dist.get_world_size() > 1:
+            download_status_queue = queue.Queue() if dist.get_global_rank(
+            ) == 0 else None
+
+            # Create a thread to busy wait for download to be complete
+            download_thread = threading.Thread(
```
irenedea, line 373: nit: download_thread makes it sound like we're downloading something on the thread itself, but we're just waiting. maybe `wait_for_done` or `wait_for_download`

llmfoundry/models/hf/hf_base.py
```diff
@@ -331,53 +334,111 @@ def build_inner_model(
...
+            download_status_queue.put(1)
+            log.debug('Download complete')
+
+        # Join and wait for the thread to complete on all ranks
+        if download_thread is not None:
+            log.debug('Joining download thread')
+            download_thread.join(timeout=3600)
+            log.debug('Download thread joined')
+
+        # Gather information about which ranks errored
```
irenedea, line 431: So if init_device=cpu or meta (also true for mixed but with mixed we have the assumption that rank 0 takes a long time compared to other ranks), we would first wait for rank 0 to finish and then be subject to the dist timeout here.

irenedea: seems fine, just thinking aloud

irenedea (approved):
lgtm, minor comments

llmfoundry/models/hf/hf_base.py
dakinggg: yup. seems like there could be an edge case there with cpu init, but the ranks took wildly different amounts of time to download and rank 0 finished more than a `dist_timeout` ahead, and then you would still timeout. I think thats ok though, can handle later if it comes up for some reason, but should really always be using mixed init here.

llmfoundry/models/hf/hf_base.py
dakinggg: done

llmfoundry/models/hf/hf_base.py
```diff
@@ -331,53 +334,111 @@ def build_inner_model(
...
-            if pretrained:
-                model = auto_model_cls.from_pretrained(
-                    pretrained_model_name_or_path,
-                    trust_remote_code=trust_remote_code,
-                    use_auth_token=use_auth_token,
-                    load_in_8bit=load_in_8bit,
-                    attn_implementation=requested_attention_implementation,
-                    config=config,
-                )
+        def download_thread_target(queue: Optional[queue.Queue]):
```
bowenyang008, line 337: sorry for the late review, I am just wondering if we can turn this to a convenient util, e.g.,

with foundry.barrier():
     do_something()

and it only exists upon thread joining. Though I am not sure if it has other use cases than HF download

dakinggg: we have some other places we have file based barriers and have utils for those in composer. could be useful to pull this into composer as util with the threading approach tho, since file based barrier only works with a shared file system. can put on the backlog as a follow up/refactor the next time we need it.

PR #1801, opened 2025-04-22
added attn_implementation for hf_base.py

ethantang-db:
Added attn_implementation as our new way of specifying the attention implementation we want to use for the huggingface model.


llmfoundry/models/hf/hf_base.py
```diff
@@ -80,19 +80,32 @@ def __init__(
         additional_train_metrics: Optional[list] = None,
         additional_eval_metrics: Optional[list] = None,
         should_save_peft_only: bool = True,
+        attn_implementation: str = 'eager',
     ):
+        if use_flash_attention_2:
+            warnings.warn(
```
dakinggg, line 86: I'd rather not deprecate this actually since its a yaml change and we try very hard to not change those.

llmfoundry/models/hf/hf_base.py
```diff
@@ -80,19 +80,32 @@ def __init__(
...
+                'use_flash_attention_2 is deprecated. Please use attn_implementation instead.' +
+                'Setting attn_implementation to flash_attention_2.',
+            )
+            attn_implementation = 'flash_attention_2'
+
+        if attn_implementation not in ['eager', 'flash_attention_2', 'sdpa']:
```
dakinggg, line 92: shouldn't we let transformers validate this?

llmfoundry/models/hf/hf_base.py
```diff
@@ -212,14 +225,14 @@ def build_inner_model(
...
         init_device: str,
-        use_flash_attention_2: bool,
         use_auth_token: bool,
         config_overrides: dict[str, Any],
         load_in_8bit: bool,
         pretrained: bool,
         model_cls: Optional[Union[type[_BaseAutoModelClass],
                                   type[PreTrainedModel]]] = None,
         prepare_for_fsdp: bool = False,
+        attn_implementation: str = 'eager',
```
dakinggg, line 235: removing an arg here is backwards breaking on this API

ethantang-db: ah, I added it back then

llmfoundry/models/hf/hf_base.py
```diff
@@ -80,7 +80,14 @@ def __init__(
         additional_train_metrics: Optional[list] = None,
         additional_eval_metrics: Optional[list] = None,
         should_save_peft_only: bool = True,
+        attn_implementation: str = 'eager',
     ):
+        if use_flash_attention_2 and attn_implementation != 'flash_attention_2':
+            warnings.warn(
+                'use_flash_attention_2 is set, this will override attn_implementation',
```
dakinggg, line 87:
```suggestion
                'use_flash_attention_2 is set, this will override attn_implementation. Set attn_implementation to flash_attention_2 directly to remove this warning.',
```

llmfoundry/models/hf/hf_base.py
```diff
@@ -80,7 +80,14 @@ def __init__(
         additional_train_metrics: Optional[list] = None,
         additional_eval_metrics: Optional[list] = None,
         should_save_peft_only: bool = True,
+        attn_implementation: str = 'eager',
```
dakinggg, line 83: this doesn't seem like quite the right behavior. I think attn_implementation should default to None, and then you warn if the user set both attn_implementation and use_flash_attention_2. Cause as is, you would immediately start getting warnings about attn_implementation being set incorrectly.

llmfoundry/models/hf/hf_base.py
```diff
@@ -80,7 +80,17 @@ def __init__(
         additional_train_metrics: Optional[list] = None,
         additional_eval_metrics: Optional[list] = None,
         should_save_peft_only: bool = True,
+        attn_implementation: Optional[str] = None,
     ):
+        if use_flash_attention_2:
+            if attn_implementation is not None and attn_implementation != 'flash_attention_2':
+                warnings.warn(
+                    'use_flash_attention_2 is set, this will override attn_implementation'
```
dakinggg, line 88:
```suggestion
                    'use_flash_attention_2 is set, this will override attn_implementation. '
```

scripts/train/README.md
```diff
@@ -363,7 +374,7 @@ For any HuggingFace model that supports Flash Attention (e.g. Llama and Mistral)
 ```yaml
 model:
     name: hf_causal_lm
-    use_flash_attention_2: True # Will be automatically set to True if Flash Attention V2 is installed and the model supports it
+    attn_implementation: "flash_attention_2"  # Will be automatically set to "flash_attention_2" if Flash Attention V2 is installed and the model supports it
```
dakinggg, line 377: that automatic comment is just not true. probably my mistake from a while ago, but please delete it

tests/models/hf/test_hf_base.py
```diff
@@ -22,6 +22,7 @@ def test_build_inner_model_fsdp():
         load_in_8bit=False,
         pretrained=False,
         prepare_for_fsdp=True,
+        attn_implementation='eager',
```
dakinggg, line 25: you shouldn't need the changes in this file cause the default is eager, right?

tests/models/hf/test_hf_config.py
```diff
@@ -335,3 +335,86 @@ def test_generation_config(
...
+
+@pytest.mark.gpu
+@pytest.mark.parametrize(
+    'attn_implementation',
+    ['eager', 'flash_attention_2', 'sdpa'],
+)
+def test_attn_implementation(attn_implementation: str):
+    model_cfg = {
+        'name': 'hf_causal_lm',
+        'pretrained_model_name_or_path': 'codellama/CodeLlama-7b-hf',
```
dakinggg, line 348: this will result in network calls to HF. Please follow patterns else where to use an already downloaded fixture for this (`tiny_codellama_model`)

tests/models/hf/test_hf_fsdp_meta.py
```diff
@@ -67,6 +67,7 @@ def test_hf_meta_init_fsdp(tiny_codellama_tokenizer: PreTrainedTokenizerBase):
             trust_remote_code=False,
             init_device='meta',
             use_flash_attention_2=False,
+            attn_implementation='eager',
```
dakinggg, line 70: same, shouldn't need this change?

tests/models/llm_embed/test_llm_embedding.py
```diff
@@ -88,6 +88,7 @@ def model(
             pretrained_model_name_or_path='bert-base-uncased',
             loss_fn='torch_crossentropy',
             use_flash_attention_2=False,
+            attn_implementation='eager',
```
dakinggg, line 91: same

dakinggg (approved):

llmfoundry/models/hf/hf_base.py
ethantang-db: done

tests/models/hf/test_hf_config.py
ethantang-db: done, some other tests in the file also use network calls?

tests/models/hf/test_hf_fsdp_meta.py
ethantang-db: done

tests/models/llm_embed/test_llm_embedding.py
ethantang-db: done

dakinggg (approved):

PR #1815, opened 2025-05-03
Allow subselecting the appropriate config for llama4

dakinggg:
Multimodal models may return a config with subconfigs from `AutoConfig`. This PR allows subclasses to automatically select a subconfig (and sets up the causal LM training class to select text config). With this, llama4 scout finetuning works (`llama4-scout-hf-real-7-a3JTUJ`). llama4 also works with flex attention (`llama4-scout-hf-flex-1-EUH7iy`)

Along for the ride, install the new HF extra that supposedly speeds up download `hf_xet`, and bumps the minimum hf hub version that supports it.

### DIFF
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -185,7 +185,7 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         transformers_version = transformers.__version__
         from packaging.version import parse
         transformers_version_is_dev = parse(transformers_version).is_devrelease
-        if not transformers_version_is_dev and cls.use_text_config and hasattr(config, 'text_config') and transformers_version_is_dev:
+        if cls.use_text_config and hasattr(config, 'text_config') and transformers_version_is_dev:
             config = config.text_config
 
         config.text_config.use_cache = False
--- a/llmfoundry/models/hf/hf_base.py
+++ b/llmfoundry/models/hf/hf_base.py
@@ -56,6 +56,16 @@ class BaseHuggingFaceModel(HuggingFaceModel):
     """Wrapper around HuggingFaceModel.
 
     Base class for HuggingFace based models.
+
+    Attributes:
+        model_cls (type): The model class to use. Default: ``AutoModelForCausalLM``.
+        subselect_config_attr (str): The attribute to use to subselect the config.
+            This is used if you want to select only using the text_config or vision_config
+            for a multimodal model. For example, AutoConfig.from_pretrained on Llama4 produces
+            a Llama4Config, and to use as a causal LM, we need to get the Llama4TextConfig.
+            Default: ``None``, which will use whatever AutoConfig produces.
+        default_train_metrics (tuple): The default training metrics to use.
+        default_eval_metrics (tuple): The default evaluation metrics to use.
     """
 
     model_cls: Union[type[_BaseAutoModelClass],
@@ -339,7 +349,6 @@ class BaseHuggingFaceModel(HuggingFaceModel):
             attn_implementation=attn_implementation,
             config_overrides=config_overrides,
         )
-        print(config)
 
         # We need to have all non-zero local ranks be not-pretrained
         # Rank 0 will still be pretrained, and distribute the weights appropriately
@@ -508,7 +517,6 @@ class BaseHuggingFaceModel(HuggingFaceModel):
         if prepare_for_fsdp:
             cls.prepare_inner_model(model, init_device)
 
-        print(model)
         return model
 
     def get_peft_config(self, peft_config_dict: dict[str, Any]) -> 'PeftConfig':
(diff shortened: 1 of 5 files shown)


llmfoundry/models/hf/hf_base.py
```diff
@@ -327,6 +339,7 @@ def build_inner_model(
             attn_implementation=attn_implementation,
             config_overrides=config_overrides,
         )
+        print(config)
```
bowenyang008, line 342: is this intended or should be used with a logger

llmfoundry/models/hf/hf_base.py
```diff
@@ -60,6 +60,7 @@ class BaseHuggingFaceModel(HuggingFaceModel):
 
     model_cls: Union[type[_BaseAutoModelClass],
                      type[PreTrainedModel]] = AutoModelForCausalLM
+    subselect_config_attr: Optional[str] = None
```
bowenyang008, line 63: nit: can you leave an inline comment what this is for?

bowenyang008 (approved):
LGTM

llmfoundry/models/hf/hf_base.py
dakinggg: nope, just left behind debugging lines, will remove

llmfoundry/models/hf/hf_base.py
```diff
@@ -181,6 +192,17 @@ def build_config(
...
         )
 
+        if cls.subselect_config_attr is not None and hasattr(
+            config,
+            cls.subselect_config_attr,
+        ):
+            config = getattr(config, cls.subselect_config_attr)
+
+            # Forward the above overrides to the subselected config too
+            config.use_cache = False
```
ethantang-db, line 202: should these overrides be hard coded? or should it be `config.use_cache = orig_config.use_cache` for example?

dakinggg: its the same overrides applied a few lines above this - cache should always be False for training. I suppose I could get them from the top level config, but seems a bit cleaner to just reapply them here? no strong opinion

irenedea: nit: make it clear that it's the same override

```
use_cache = False # Necessary due to https://github.com/huggingface/transformers/issues/28056

config = AutoConfig.from_pretrained(
            pretrained_model_name_or_path,
            use_auth_token=use_auth_token,
            attn_implementation=attn_implementation,
            torch_dtype=_MASTER_WEIGHTS_PRECISION,
            use_cache=use_cache
        )

...

config.use_cache = use_cache
```

llmfoundry/models/hf/hf_base.py
```diff
@@ -181,6 +192,17 @@ def build_config(
             False,  # Necessary due to https://github.com/huggingface/transformers/issues/28056
         )
 
+        if cls.subselect_config_attr is not None and hasattr(
+            config,
+            cls.subselect_config_attr,
+        ):
+            config = getattr(config, cls.subselect_config_attr)
```
irenedea, line 199: Could you add a nice error for when the user inputs a subselect_config_attr that is not found?

llmfoundry/models/hf/hf_base.py
```diff
@@ -56,10 +56,21 @@ class BaseHuggingFaceModel(HuggingFaceModel):
     """Wrapper around HuggingFaceModel.
 
     Base class for HuggingFace based models.
+
+    Attributes:
+        model_cls (type): The model class to use. Default: ``AutoModelForCausalLM``.
+        subselect_config_attr (str): The attribute to use to subselect the config.
```
irenedea, line 62:
```suggestion
        subselect_config_attr (optional, str): The attribute to use to subselect the config.
```

irenedea (approved):
couple of nits! lgtm

llmfoundry/models/hf/hf_base.py
```diff
@@ -171,16 +182,33 @@ def build_config(
...
+
+            # Forward the above overrides to the subselected config too
+            config.use_cache = use_cache
+            config.attn_implementation = attn_implementation
+            config.torch_dtype = _MASTER_WEIGHTS_PRECISION
+        elif cls.subselect_config_attr is not None:
+            raise ValueError(
+                f'Config {cls.subselect_config_attr} not found in {config}.',
+            )
+
```
irenedea, line 211:
```suggestion
        if cls.subselect_config_attr is not None:
            if not hasattr(config, cls.subselect_config_attr):
                raise ValueError(
                    f'Config {cls.subselect_config_attr} not found in {config}.',
                )
            config = getattr(config, cls.subselect_config_attr)

            # Forward the above overrides to the subselected config too
            config.use_cache = use_cache
            config.attn_implementation = attn_implementation
            config.torch_dtype = _MASTER_WEIGHTS_PRECISION
        elif cls.subselect_config_attr is not None:
            raise ValueError(
                f'Config {cls.subselect_config_attr} not found in {config}.',
            )

```

irenedea: simpler conditionals and more readable to error early

ethantang-db (approved):

PR #1820, opened 2025-05-15
Fix Dtensor initialization

bowenyang008:
Existing initialization utils can not handle DTensor slice properly, as DTensor slice could be a new DTensor (with a different sharding layout/placement) instead of a view to the original Tensor.

Solution: Init on DTensor.full_tensor and then copy the shard of the full_tensor to original Dtensor.

Test:
1. Added unit test
2. Manually tested param norm are the same between FSDP1 and FSDP2
3. 7B Pretrain Loss:
green: FSDP1
violet: FSDP2 w/o this PR
red: FSDP2 w/ this PR
w/ this PR FSDP2 and FSDP1 are bits wise the same at first 2 steps
loss_train_total (4)
4. Init time:
FSDP1:  Trainer construction took 2.95 seconds
FSDP2 w/ this PR:  Trainer construction took 4.05 seconds, Prepare FSDP2 took 0.12 seconds, Meta Init Device took 1.61 seconds
FSDP2 w/o this PR:  Trainer construction took 3.99 seconds, Prepare FSDP2 took 0.12 seconds, Meta Init Device took 1.56 seconds
Per iteration time: 1.84 seconds

### DIFF
--- a/llmfoundry/models/utils/param_init_fns.py
+++ b/llmfoundry/models/utils/param_init_fns.py
@@ -97,7 +97,7 @@ def materialize_module(
         nn.Module: The module with materialized parameters that can be initialized.
     """
     param_placement_map: dict[tuple[nn.Module, str],
-                              tuple[DeviceMesh, list[Placement]]] = {}
+                              tuple[DeviceMesh, tuple[Placement, ...]]] = {}
     with torch.no_grad():
         for submodule in module.modules():
             for name, param in submodule.named_parameters(recurse=False):
@@ -127,11 +127,8 @@ def materialize_module(
                     )
 
 
-def summon_dtensor(
-    init_fn: Callable[[nn.Module | torch.Tensor, Any], bool],
-) -> Callable[[nn.Module | torch.Tensor, Any], bool]:
-    """Decorator that makes initialization functions compatible with DTensor
-    parameters.
+def summon_dtensor(init_fn: Callable) -> Callable:
+    """Unshard and Reshard DTensor parameters for initialization.
 
     This decorator wraps an initialization function to handle both regular tensors/modules
     and those containing DTensor parameters by temporarily materializing DTensors as full
--- a/llmfoundry/models/utils/param_init_fns.py
+++ b/llmfoundry/models/utils/param_init_fns.py
@@ -128,8 +128,8 @@ def materialize_module(
 
 
 def summon_dtensor(
-    init_fn: Callable[[nn.Module | torch.Tensor, Any], None],
-) -> Callable[[nn.Module | torch.Tensor, Any], None]:
+    init_fn: Callable[[nn.Module | torch.Tensor, Any], bool],
+) -> Callable[[nn.Module | torch.Tensor, Any], bool]:
     """Decorator that makes initialization functions compatible with DTensor
     parameters.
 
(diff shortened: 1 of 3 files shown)


dakinggg (commented):
code lgtm, but can you check the rest of this file for other things that will be an issue with dtensor? e.g. the `stacked_init_helper_` below has the same issue

llmfoundry/models/utils/param_init_fns.py
```diff
@@ -444,7 +576,7 @@ def generic_param_init_fn_(
     did_init = False
     for module_init_fn in all_module_init_fns:
         did_init = module_init_fn(
-            module=module,
+            module,
```
dakinggg, line 579: why?

tests/models/utils/test_param_init_fns.py
```diff
@@ -228,3 +241,437 @@ def test_emb_padding_init(
...
+
+    # Convert the weight and bias to DTensor
+    dtensor_linear.weight = torch.nn.Parameter(
+        distribute_tensor(dtensor_linear.weight, mesh, [Shard(0)]),
+    )
+    dtensor_linear.bias = torch.nn.Parameter(
+        distribute_tensor(dtensor_linear.bias, mesh, [Shard(0)]),
+    )
+
+    # Mark one of the layers as residual to test the residual path
```
dakinggg, line 327: comment doesn't make sense

dakinggg (approved):
LGTM, if you could just add a comparison of initialization time for reference to the PR that'd be great.

llmfoundry/models/utils/param_init_fns.py
bowenyang008: the returned function from summon_dtensor requires first arg to be positional, as the kwargs can have arbitrary orders, let me know if you have a better way of handle both positional and kw arguments

tests/models/utils/test_param_init_fns.py
bowenyang008: doesn't make sense to me either 🤦 I just checked the code not the comments of AI gen tests

bowenyang008:
For the failing FFN TP test, can't tell much diff from the param norm, so would just attribute it to randomness
w/o this PR:
param name: up_proj.weight, param norm: 31.983922958374023
param name: up_proj.bias, param norm: 0.0
param name: down_proj.weight, param norm: 7.9869465827941895
param name: down_proj.bias, param norm: 0.0
param name: up_proj.weight, param norm: 31.980695724487305
param name: up_proj.bias, param norm: 0.0
param name: down_proj.weight, param norm: 7.955490589141846
param name: down_proj.bias, param norm: 0.0
w/ this PR:
up_proj.weight, param norm: 32.03810501098633
up_proj.bias, param norm: 0.0
down_proj.weight, param norm: 7.992878437042236
down_proj.bias, param norm: 0.0
up_proj.weight, param norm: 31.91594696044922
up_proj.bias, param norm: 0.0
down_proj.weight, param norm: 8.026201248168945
down_proj.bias, param norm: 0.0

bowenyang008:
> LGTM, if you could just add a comparison of initialization time for reference to the PR that'd be great.

updated in PR description

llmfoundry/models/utils/param_init_fns.py
dakinggg: ah i think just remove *args from the init_fn_wrapper definition. put a * as the first arg, and have the function only accept kwargs.

bowenyang008: init_fn_wrapper accepts both Tensor/Module, if it is kwargs only, it is a bit loose on the interface, e.g., need to handle "tensor"/"param" and what if "tensor" and "module" both exist etc, I can make it init_fn_wrapper(obj, *, **kwargs) though if you prefer

dakinggg: ah i see what you mean. ok sure `init_fn_wrapper(obj, *, **kwargs)` seems reasonable

bowenyang008: actually init_fn_wrapper(obj, *, **kwargs) would force all decorated functions to be used with kwargs, which is a bit too verbose IMO and against the single purpose of just handling DTensor. Let me know if you still prefer init_fn_wrapper(obj, *, **kwargs)

dakinggg: im in favor of kwargs everywhere :P feel free to do whichever you prefer here tho

bowenyang008: LoL, I will go with what I have now first

PR #1832, opened 2025-05-28
init with cursor: I need to upgrade this repo to use torch 2.7 with c…

bowenyang008:
…uda version 12.6. please check the commit with a message "Bump torch to 2.6 (#1740)" that used to upgrade the repo to use 2.6 and help me similar changes. You can assume the  base image with correct torch and cuda version is already available

### DIFF
--- a/mcli/mcli-1b-max-seq-len-8k.yaml
+++ b/mcli/mcli-1b-max-seq-len-8k.yaml
@@ -17,7 +17,7 @@ command: |
     --out_root ./my-copy-c4 --splits train_small val_small \
     --concat_tokens 8192 --tokenizer EleutherAI/gpt-neox-20b --eos_text '<|endoftext|>'
   composer train/train.py /mnt/config/parameters.yaml
-image: mosaicml/llm-foundry:2.6.0_cu124-latest
+image: mosaicml/llm-foundry:2.7.0_cu126-latest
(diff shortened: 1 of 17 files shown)


mcli/mcli-1b-max-seq-len-8k.yaml
```diff
@@ -1,7 +1,7 @@
 integrations:
 - integration_type: git_repo
   git_repo: mosaicml/llm-foundry
-  git_branch: v0.20.0
+  git_branch: v0.21.0
```
bowenyang008, line 4: this is a forwarding tag as will be releasing the new version shortly

mcli/mcli-1b-max-seq-len-8k.yaml
```diff
@@ -17,7 +17,7 @@ command: |
     --out_root ./my-copy-c4 --splits train_small val_small \
     --concat_tokens 8192 --tokenizer EleutherAI/gpt-neox-20b --eos_text '<|endoftext|>'
   composer train/train.py /mnt/config/parameters.yaml
-image: mosaicml/llm-foundry:2.6.0_cu124-latest
+image: mosaicml/llm-foundry:2.7.0_cu128-latest
```
bowenyang008, line 20: also a forwarding tag but it will be available right after this PR gets merged

dakinggg (approved):

PR #1837, opened 2025-05-30
WIP: Qwen model support via MPT

gupta-abhay:


dakinggg (commented):
Can you add to the `test_mpt_creation` tests to (1) actually test that no bias worked (will validate that we haven't broken `no_bias` on `Linears`, and (2) add a parametrization that use `attention_bias` and check that does the right thing. overall lgtm

gupta-abhay:
> Can you add to the `test_mpt_creation` tests to (1) actually test that no bias worked (will validate that we haven't broken `no_bias` on `Linears`, and (2) add a parametrization that use `attention_bias` and check that does the right thing. overall lgtm

added tests, ptal!

dakinggg (approved):
couple nits, lgtm

PR #1842, opened 2025-06-10
Add Head Dim as a configurable parameter for Attention and Rope

jdchang1:
Added head dim as an optional configurable parameter in MPT for Rope implementations and Attention layers.

### DIFF
--- a/llmfoundry/models/layers/blocks.py
+++ b/llmfoundry/models/layers/blocks.py
@@ -46,7 +46,7 @@ class MPTBlock(nn.Module):
         fc_type: Optional[dict[str, Any]] = None,
         device: Optional[str] = None,
         no_bias: bool = False,
-        attention_bias: bool = False,
+        attention_bias: bool = True,
         use_pad_tok_in_ffn: bool = True,
         **kwargs: Any,
     ):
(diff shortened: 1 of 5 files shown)


llmfoundry/models/mpt/modeling_mpt.py
```diff
@@ -159,13 +160,23 @@ def gen_rotary_embedding(
...
         if llama_rope_config['rope_type'] == 'no_scaling':
             llama_rope_config['rope_type'] = 'default'
-        partial_llama_config = PartialLlamaConfig(
-            rope_scaling=llama_rope_config,
-            rope_theta=rope_theta,
-            max_position_embeddings=max_seq_len,
-            hidden_size=d_model,
-            num_attention_heads=n_heads,
-        )
+        if head_dim is not None:
```
dakinggg, line 163: I don't think you need an if/else here, passing `head_dim` when its `None` is the same as not passing it i believe.

llmfoundry/models/layers/blocks.py
```diff
@@ -45,7 +46,7 @@ def __init__(
         fc_type: Optional[dict[str, Any]] = None,
         device: Optional[str] = None,
         no_bias: bool = False,
-        attention_bias: bool = True,
```
dakinggg, line 48: Did you need to change this for some reason?

jdchang1: oh no i did not need to change it

dakinggg (commented):
LGTM, add a test for constructing a small MPT model with and without head dim specified and make sure the expected layers are the expected sizes? You can add to the `test_mpt_creation` test I think, or make a separate one just for testing head dim specification

PR #1857, opened 2025-07-18
updated composer and streaming versions

ethantang-db:


llmfoundry/callbacks/monolithic_ckpt_callback.py
```diff
@@ -116,14 +114,18 @@ def _save_checkpoint(self, state: State, logger: Logger) -> None:
...
                 optimizer = state.optimizers[0]
-                state_dict['state']['optimizers'] = {
-                    type(optimizer).__qualname__:
-                        fsdp_get_optim_state_dict(
-                            state.model,
-                            optimizer,
-                            state_dict_type='full',
-                        ),
-                }
+                with fsdp_state_dict_type_context(
```
rithwik-db, line 117: I am not 100% sure but I feel like this should break monolithic checkpointing for FSDP2 since `fsdp_state_dict_type_context` uses FSDP1 functionality. I'm not sure why we didn't just do the following:

```
state_dict['state']['model'] = state.get_model_state_dict()
state_dict['state']['optimizers'] = state.get_optim_state_dict()
```

We may or may not also need to wrap this in a contextmanager that sets the state.fsdp_config.state_dict_type to `full` and then change it back to what it originally was after the contextmanager exits

ethantang-db: yeah this was an oudated merge

llmfoundry/callbacks/hf_checkpointer.py
```diff
@@ -673,17 +673,32 @@ def tensor_hook(
...
                 if self.using_peft:
+                    import peft
                     from peft import PeftModel
                     assert isinstance(original_model, PeftModel)
                     active_adapter = original_model.active_adapter  # type: ignore
                     base_model: PreTrainedModel = original_model.get_base_model(  # type: ignore
                     )
                     new_base_model_instance = type(base_model)(new_config)
 
+                    if peft.__version__ == '0.16.0':
```
rithwik-db, line 685: Can you add a `TODO: Need to deprecate` or something just to make it easier for us to just `grep deprecate` next release and remove this conditional?

ethantang-db: sure, but this shouldn't be deprecated unless our dependency of peft no longer contains 0.16.0

rithwik-db: Yeah that's fine, I just don't want this to go forgotten in the future and our release doc mentions `git grep deprecate` explicitly

PR #1865, opened 2025-08-01
Gating megablocks based on CUDA availability

rithwik-db:
See title

### DIFF
--- a/llmfoundry/models/layers/ffn.py
+++ b/llmfoundry/models/layers/ffn.py
@@ -27,10 +27,13 @@ try:
 except ModuleNotFoundError:
     is_te_imported = False
 
-try:
-    import megablocks
-    is_megablocks_imported = True
-except ModuleNotFoundError:
+if torch.cuda.is_available():
+    try:
+        import megablocks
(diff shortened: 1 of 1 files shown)


llmfoundry/models/layers/ffn.py
```diff
@@ -27,10 +27,13 @@
 except ModuleNotFoundError:
     is_te_imported = False
 
-try:
-    import megablocks
-    is_megablocks_imported = True
-except ModuleNotFoundError:
+if torch.cuda.is_available():
```
rithwik-db, line 32: This should solve the issue on `compose-rl`, but another solution would be to simply
```
try:
    import megablocks
    is_megablocks_imported = True
except:
    is_megablocks_imported = False
```
to catch any straggling issues?

irenedea: I like it as is, it's more intentional and specific. Can you add a note explaining that we are gating this because megablocks depends on triton and triton will error upon import if you do not have gpus

irenedea (approved):

Issue #1848, opened 2025-06-19
signaling error on shared filesystem is back


Skylion007:
## To reproduce

Tried to run llm-foundry on a shared file system:
Easy to get a race condition:
```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/mnt/weka/home/aaron.gokaslan/git/llm-foundry/scripts/train/train.py", line 9, in <module>
[rank0]:     train_from_yaml(yaml_path, args_list)
[rank0]:   File "/mnt/weka/home/aaron.gokaslan/git/llm-foundry/.venv/lib/python3.11/site-packages/llmfoundry/command_utils/train.py", line 666, in train_from_yaml
[rank0]:     return train(yaml_cfg)
[rank0]:            ^^^^^^^^^^^^^^^
[rank0]:   File "/mnt/weka/home/aaron.gokaslan/git/llm-foundry/.venv/lib/python3.11/site-packages/llmfoundry/command_utils/train.py", line 370, in train
[rank0]:     tokenizer = build_tokenizer(tokenizer_name, tokenizer_kwargs)
[rank0]:                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/mnt/weka/home/aaron.gokaslan/git/llm-foundry/.venv/lib/python3.11/site-packages/llmfoundry/utils/builders.py", line 545, in build_tokenizer
[rank0]:     os.remove(signal_file_path)
[rank0]: FileNotFoundError: [Errno 2] No such file or directory: '._signal_file_node0_QVBHXW'
```

jxdn:
ah find the same issue

souvikDevloper:
is your issue already resolved?

souvikDevloper:
hey hope my pr helps, let me know!

rehan243:
Oh yeah, that `build_tokenizer` call throwing errors on shared filesystems — we've hit this before. It’s usually some kind of file lock contention when multiple ranks try to read/write tokenizer cache files at the same time. If you're on something like WEKA, it can amplify the race conditions since the latency for small writes isn't great.

[...]

PR #1881, opened 2026-01-18
Handle missing tokenizer signal file

souvikDevloper:


souvikDevloper:
GPU checks failed before tests: mcli returns 401 “API key has expired.” Looks like infra/secret issue; please refresh
  and rerun.

souvikDevloper:
Title
  Avoid FileNotFoundError when cleaning up tokenizer signal file

  

  Summary:
  This guards cleanup of the tokenizer signal file so shared filesystems don’t throw a race-condition
  `FileNotFoundError`.

  Changes:
  - Suppress `FileNotFoundError` when removing the tokenizer signal file
  - Not run (local environment not set up)

  Issue
  Fixes #1848

   Additional context
  On shared filesystems, multiple ranks can race when cleaning up the signal file, so a best‑effort delete avoids
  `FileNotFoundError` while keeping the same synchronization behavior.

PR #1884, opened 2026-02-21
fix(eval): include non-Accuracy metrics in markdown results

### LINKED ISSUE #1722: Make metric name handling more flexible and transparent
I spent a lot of time trying to get a custom metric working only to realize that it wasn't showing up in the output simply because https://github.com/LocalResearchGroup/llm-foundry/blob/a4bd9fc08f8dac482970d42a94ef2cdda2659a60/llmfoundry/command_utils/eval.py#L473-L474 looks for "Accuracy" in the metric name. Would it be possible to (1) develop a more robust way to recognize metrics, (2) treat everything in `logger_keys` as a metric and record its name and value rather than assuming it is an "accuracy", or at least (3) make it more obvious why a metric without "Accuracy" in its name isn't showing up in the output?

I would be interested in developing a solution if we can align on the approach.

ry2009:
Fixes #1722.

This removes the hardcoded `"Accuracy"` name filter in `calculate_markdown_results()` so all metrics present in `logger_keys` are included.

Changes:
- parse metric keys defensively (`metrics/{benchmark}/{fewshot}/{metric}` and subcategory form)
- include metric names in markdown output via a `Metric` column
- keep the existing `Accuracy` column name for compatibility (it now stores generic metric values)
- convert tensor-like metric outputs with `.item()` when available
- add regression coverage for non-accuracy metrics (`LanguageCrossEntropy`, `CustomF1`) and averaged subcategory rows

Validation:
- `ruff check llmfoundry/command_utils/eval.py tests/a_scripts/eval/test_eval.py`
- `python3.11 -m py_compile llmfoundry/command_utils/eval.py tests/a_scripts/eval/test_eval.py`
- direct function harness for `calculate_markdown_results()` (full local pytest is blocked in this environment by missing `composer` dependency)

### DIFF
--- a/llmfoundry/command_utils/eval.py
+++ b/llmfoundry/command_utils/eval.py
@@ -397,10 +397,16 @@ def calculate_markdown_results(
     results = {}
 
     for key in logger_keys:
-        # dl_name is either 2-tuple (benchmark_name, num_fewshot)
-        # or 3-tuple (benchmark_name, num_fewshot, subcategory)
-        dl_name, metric_name = key.split('/')[1:-1], key.split('/')[-1]
-        if 'Accuracy' not in metric_name:
+        parts = key.split('/')
+        # Metrics are expected to be:
+        # metrics/{benchmark}/{num_fewshot}/{metric_name}
+        # or metrics/{benchmark}/{num_fewshot}/{subcategory}/{metric_name}
+        if len(parts) < 4 or parts[0] != 'metrics':
+            continue
+
+        dl_name, metric_name = parts[1:-1], parts[-1]
+        # Skip malformed keys and keep processing other metrics.
+        if len(dl_name) not in (2, 3):
             continue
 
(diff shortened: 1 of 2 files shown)


ry2009:
Validation update after latest push (`d030c69`):

- Local checks passed:
  - `ruff check llmfoundry/command_utils/eval.py tests/a_scripts/eval/test_eval.py`
  - `python3.11 -m py_compile llmfoundry/command_utils/eval.py tests/a_scripts/eval/test_eval.py`
  - direct `calculate_markdown_results()` harness confirms non-Accuracy metrics are included and subcategory averages are correct
- Current CI failures are infrastructure/auth before project tests start:
  - run: https://github.com/mosaicml/llm-foundry/actions/runs/22253519036
  - error: `mcli.api.exceptions.MultiMAPIException: Error 401: Context creation failed: API key has expired.`

